PR #5, opened 2013-10-08
Bug 416927: Enhance director mojo for supporting additional Install Units

sseshan1:
Here is a branch containing the original patch.

### DIFF
--- a/tycho-p2/tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/p2/director/DirectorMojo.java
+++ b/tycho-p2/tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/p2/director/DirectorMojo.java
@@ -55,6 +55,21 @@ public final class DirectorMojo extends AbstractProductMojo {
     /** @parameter default-value="true" */
     private boolean installFeatures;
 
+
+    /**
+      * Additional install units that may be required to complete the product installation	  
+	  * @parameter 
+	  */
+    private List<String> additionalIUs;
+
+    /**
+     * Additional file based repositories to be considered when using an InstallationSource of
+     * <code>repository<code>.
+     * 
+     * @parameter
+     */
+    private List<String> additionalRepositories;
+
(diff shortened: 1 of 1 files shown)


laeubi:
@sseshan1 is this still relevant for you?

laeubi:
See also https://bugs.eclipse.org/bugs/show_bug.cgi?id=416927

tycho-p2/tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/p2/director/DirectorMojo.java
```diff
@@ -55,6 +55,21 @@
...
 
+
+    /**
+      * Additional install units that may be required to complete the product installation	  
+	  * @parameter 
+	  */
+    private List<String> additionalIUs;
+
+    /**
+     * Additional file based repositories to be considered when using an InstallationSource of
```
mickaelistria, line 66: Why file-based?

tycho-p2/tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/p2/director/DirectorMojo.java
```diff
@@ -55,6 +55,21 @@
     /** @parameter default-value="true" */
     private boolean installFeatures;
 
+
+    /**
+      * Additional install units that may be required to complete the product installation
```
mickaelistria, line 60: Not necessary required. So it could just say `Additional installable units to include with the product`.
Is it possible to enforce a specific version and if yes, what's the syntax to be used? Documentation should clarify that.

laeubi:
@mickaelistria do you think this is useful so we should keep/improve this idea or should we drop this as there is no response from the original author?

mickaelistria:
Let's close it by lack of feedback. @sseshan1 Please post here if this is still relevant.

PR #7, opened 2017-03-22
New goal to synchronize metadata from pom.

perelengo:
As I do releases of eclipse based projects with maven-release-plugin, I needed some utility to update tycho based projects metadata after version change. I think it's a interesting utility, for wider use of tycho.

I created a new Mojo to do such task:

Synchronizes the metadata version of the current project and child projects with their corresponding pom version.

Allows updating properties.

By default bounds of OSGI version ranges referencing the version of an element that changed version will be updated to match the newVersion.

Only executes in the root project of an aggregation project, due to it updates all child projects, it's not necessary to execute again in child projects.

### DIFF
--- a/tycho-release/tycho-versions-plugin/src/main/java/org/eclipse/tycho/versions/SynchronizeMetadataMojo.java
+++ b/tycho-release/tycho-versions-plugin/src/main/java/org/eclipse/tycho/versions/SynchronizeMetadataMojo.java
@@ -1,0 +1,131 @@
+/*******************************************************************************
+ * Copyright (c) 2008, 2015 Sonatype Inc. and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Sonatype Inc. - initial API and implementation
+ *    Sebastien Arod - update version ranges
+ *    Pere Joseph Rodríguez - synchronize metadata with pom version
+ *******************************************************************************/
+package org.eclipse.tycho.versions;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.StringTokenizer;
+
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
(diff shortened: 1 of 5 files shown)


jsievers:
is this the same as https://eclipse.org/tycho/sitedocs/tycho-release/tycho-versions-plugin/update-pom-mojo.html ?

perelengo:
No, I'ts the inverse, updates Metadata from pom info.

So if you release with maven, then you have to update theMANIFEST.MF, the features.xml, site.xml, category.xml ....


________________________________
De: Jan Sievers <notifications@github.com>
Enviado: miércoles, 22 de marzo de 2017 17:24
Para: eclipse/tycho
Cc: Pere Joseph Rodríguez; State change
Asunto: Re: [eclipse/tycho] New goal to synchronize metadata from pom. (#7)


is this the same as https://eclipse.org/tycho/sitedocs/tycho-release/tycho-versions-plugin/update-pom-mojo.html ?

Tycho Versions Plugin - tycho-versions:update-pom<https://eclipse.org/tycho/sitedocs/tycho-release/tycho-versions-plugin/update-pom-mojo.html>
eclipse.org
tycho-versions:update-pom. Full name: org.eclipse.tycho:tycho-versions-plugin:1.0.0:update-pom. Description:

ctubbsii:
This issue should be reopened. The above referenced commit does not achieve the same thing. It forces the POM to match the manifest. But, what is needed here is for the manifest to match the POM. This makes sense, since lots of other Maven plugins/tasks help you manage incrementing the POM version (for example, the maven-release-plugin). This issue should be reopened, and addressed, so that the manifest can be automatically updated from the POM version, after the POM is updated by some other previous plugin.

perelengo:
Check this out: is another proposed solution that went production: https://wiki.eclipse.org/Tycho/Release_Workflow

ctubbsii:
I ended up using two executions of `com.google.code.maven-replacer-plugin`, the first to replace the Bundle-Version in the manifest with the version in the POM, and then a second to replace `-SNAPSHOT` with `.qualifier`. It's a terrible hack. I tied the execution to the `pre-clean` phase of the `clean` lifecycle.

ctubbsii:
Ideally, the entire manifest should be generated dynamically from a well-configured Maven plugin, version included, and the manifest put in the target/classes/META-INF directory during the build. The requirement to essentially maintain two project profiles separately, is a bit un-Maven-like.

PR #8, opened 2017-03-22
Signed-off-by: Pere Joseph Rodríguez <perelengo@hotmail.com>

perelengo:
As I do releases of eclipse based projects with maven-release-plugin, I needed some utility to update tycho based projects metadata after version change. I think it's a interesting utility, for wider use of tycho.

I created a new Mojo to do such task:

Synchronizes the metadata version of the current project and child projects with their corresponding pom version.

Allows updating properties.

By default bounds of OSGI version ranges referencing the version of an element that changed version will be updated to match the newVersion.

Only executes in the root project of an aggregation project, due to it updates all child projects, it's not necessary to execute again in child projects.

### DIFF
--- a/tycho-release/tycho-versions-plugin/src/main/java/org/eclipse/tycho/versions/SynchronizeMetadataMojo.java
+++ b/tycho-release/tycho-versions-plugin/src/main/java/org/eclipse/tycho/versions/SynchronizeMetadataMojo.java
@@ -1,0 +1,131 @@
+/*******************************************************************************
+ * Copyright (c) 2008, 2015 Sonatype Inc. and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Sonatype Inc. - initial API and implementation
+ *    Sebastien Arod - update version ranges
+ *    Pere Joseph Rodríguez - synchronize metadata with pom version
+ *******************************************************************************/
+package org.eclipse.tycho.versions;
+
+import java.io.File;
+import java.io.IOException;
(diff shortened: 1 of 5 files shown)


perelengo:
After you do a release with maven, you need to update the MANIFEST.MF, feature.xml, site.xml, etc... because release plugin updates pom info but not metadata to the next development version. 
Releases with maven plugin can only be done deactivating the strictVersion.
                        <plugin>
				<artifactId>tycho-packaging-plugin</artifactId>
				<configuration>
					<strictVersions>false</strictVersions>
				</configuration>
			</plugin>
As release plugin handles SCM versioning, it's very useful, but after releasing with maven release plugin, metadata (MANIFEST, features.xml,...) are out of sync with pom versions.

In difference to the setVersion goal, that updates versions of pom and metadata to do a new release, but without handling the SCM repo, this new goal upgrades metadata versions info from pom versions.

jsievers:
we use github as a mirror only. Can you push your change to gerrit? See Tycho's [contributor guide](https://wiki.eclipse.org/Tycho/Contributor_Guide) for details

perelengo:
Is a MUST to develop the unit tests?

I don'thave done so, but the existing ones keep working.


________________________________
De: Jan Sievers <notifications@github.com>
Enviado: viernes, 24 de marzo de 2017 8:30
Para: eclipse/tycho
Cc: Pere Joseph Rodríguez; Author
Asunto: Re: [eclipse/tycho] Signed-off-by: Pere Joseph Rodríguez <perelengo@hotmail.com> (#8)


we use github as a mirror only. Can you push your change to gerrit? See Tycho's contributor guide<https://wiki.eclipse.org/Tycho/Contributor_Guide> for details

jsievers:
of course the existing tests should not break.
If you add new functionality, there should be at least an integration test which documents how it should work, otherwise we wouldn't notice if we break it.

PR #9, opened 2017-03-30
Update instructions in README.md

luvalach:
Update instructions in README.md. It seems that Tycho is now included in Eclipse Installer Project Catalog, so guide will reflect this fact. Also Eclipse and JRE version have been updated.

Alternatively we could put here only link to Wiki [1], then we wouldn't have to maintain the same guide on two places. 
[1] https://wiki.eclipse.org/Developing_Tycho

### DIFF
--- a/setup/README.md
+++ b/setup/README.md
@@ -14,22 +14,17 @@ Tycho.setup is an Eclipse Installer (Oomph) setup file that does
 Usage for contributors:
 -----------------------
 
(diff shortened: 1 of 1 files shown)


jsievers:
we use github as a mirror only

could you push your change to gerrit? See https://wiki.eclipse.org/Tycho/Contributor_Guide

luvalach:
Sure, I can, thank you.

PR: https://git.eclipse.org/r/#/c/94880/

PR #13, opened 2018-09-22
Tycho p2 metadata mojo classifier support

renelink:
Hi,

I created a patch that allows users to define a classifier for plugins etc.
I need this feature, because I obfuscate my plugins and the obfuscated version has a classifier.
Without that patch the product configuration will always use the "normal" jar. Thus the generated p2 repository or the created ZIPs did not contain the obfuscated jar. 

With my patch I can now configure the tycho-p2-plugin in this way

				<plugin>
					<artifactId>tycho-p2-plugin</artifactId>
					<configuration>
						<p2ArtifactClassifier>obfuscated</p2ArtifactClassifier>
					</configuration>
				</plugin>

It would be very nice if you would merge that commit.
Don't hesitate to contact me, if you have any questions.

Greetings
René Link

### DIFF
--- a/tycho-p2/tycho-p2-plugin/src/main/java/org/eclipse/tycho/plugins/p2/P2MetadataMojo.java
+++ b/tycho-p2/tycho-p2-plugin/src/main/java/org/eclipse/tycho/plugins/p2/P2MetadataMojo.java
@@ -65,6 +65,9 @@ public class P2MetadataMojo extends AbstractMojo {
     @Parameter(defaultValue = "true")
     protected boolean attachP2Metadata;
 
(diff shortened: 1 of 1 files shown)


jsievers:
please propose the change in gerrit and open a bugzilla ticket

https://wiki.eclipse.org/Tycho/Contributor_Guide

mickaelistria:
I don't think this is necessary. See for example how signing jar is implmented and how the plugin allows to modify the jar and then make tycho-p2-plugin refresh its metadata accordingly. Same strategy can work for anything that modify the jar.

Issue #23, opened 2021-04-08
additional.bundles dependencies are not properly resolved


cdietrich:
additional.bundles dependencies are not properly resolved

this issue is initially reported here: https://www.eclipse.org/lists/tycho-user/msg09020.html
besides the xtext-eclipse and xtext-xtend tests builds with tycho 2.4.0-SNAPSHOT
i can also see the problem with
https://github.com/cdietrich/tych24ddBundlesBug

```
[ERROR] Internal error: java.lang.RuntimeException: org.osgi.framework.BundleException: Bundle org.xtext.example.mydsl.ide cannot be resolved:org.xtext.example.mydsl.ide [109]
[ERROR]   Unresolved requirement: Require-Bundle: org.xtext.example.mydsl
[ERROR]     -> Bundle-SymbolicName: org.xtext.example.mydsl; bundle-version="1.0.0.qualifier"; singleton:="true"
[ERROR]        org.xtext.example.mydsl [108]
[ERROR]          Unresolved requirement: Require-Bundle: org.eclipse.xtext.xtext.generator
[ERROR] -> [Help 1]
org.apache.maven.InternalErrorException: Internal error: java.lang.RuntimeException: org.osgi.framework.BundleException: Bundle org.xtext.example.mydsl.ide cannot be resolved:org.xtext.example.mydsl.ide [109]
  Unresolved requirement: Require-Bundle: org.xtext.example.mydsl
    -> Bundle-SymbolicName: org.xtext.example.mydsl; bundle-version="1.0.0.qualifier"; singleton:="true"
       org.xtext.example.mydsl [108]
         Unresolved requirement: Require-Bundle: org.eclipse.xtext.xtext.generator
```

org.eclipse.xtext.xtext.generator is part of xtext.sdk here in the tp https://github.com/cdietrich/tych24ddBundlesBug/blob/7e0dbd51eed5802857b45b4db499041457c4eb5c/org.xtext.example.mydsl.target/org.xtext.example.mydsl.target.target#L18

interestingly i can also find this in the log
```
[INFO] Fetching org.eclipse.xtext.xbase_2.25.0.v20210301-0909.jar from https://ftp.snt.utwente.nl/pub/software/eclipse/releases/2021-03/202103171000/plugins/ (2,46MB)
[...]

PR #41, opened 2021-04-09
test case for #23

cdietrich:
test case for #23
Signed-off-by: Christian Dietrich <christian.dietrich@itemis.de>

### DIFF
--- a/tycho-its/projects/issue23/bundle/build.properties
+++ b/tycho-its/projects/issue23/bundle/build.properties
@@ -1,0 +1,5 @@
+source.. = src/
+output.. = bin/
+bin.includes = .,\
+               META-INF/
+additional.bundles = org.junit
--- a/tycho-its/projects/issue23/bundle/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/issue23/bundle/META-INF/MANIFEST.MF
@@ -1,0 +1,9 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: bundle Plug-in
+Bundle-SymbolicName: bundle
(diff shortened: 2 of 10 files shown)


tycho-its/projects/issue23/bundle/build.properties
```diff
@@ -0,0 +1,5 @@
+source.. = src/
+output.. = bin/
+bin.includes = .,\
+               META-INF/
+additional.bundles = org.eclipse.emf.codegen.ecore
```
cdietrich, line 5: unfortunately the problem is not reproducible with a dependency closer to platform

tycho-its/projects/issue23/bundle/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
```
laeubi, line 1: itest projects should not contain individual git ignores

tycho-its/projects/issue23/bundle/build.properties
laeubi: I found org.junit useful for this purpose as it is part of the sdk but I think it does not mater much as long as we do not need to include large dependency chains.

laeubi (changes requested):
Thanks for the test-case, just a minor adjustment (git ignore). BTW you don't need to force-push, just push to your branch and we can later squash it.

cdietrich:
reworked to have two bundles, and use junit as add. bundle in the first

laeubi (commented):
.project files and .classpath files are not necessary for this also m2e.core prefs can be omited

laeubi (approved):
Looks good to me, if build succeeds we can merge this

Issue #43, opened 2021-04-10
extraRequirements should be part of the tycho-surefire configuration instead of the target platform


laeubi:
As mentioned in https://www.eclipse.org/tycho/sitedocs/tycho-surefire-plugin/test-mojo.html extra requirements are only meaningful to the tycho-surefire-plugin but are instead configured in the target-platform-configuration.

We should do the following:

1. add support to configure this in the tycho-surefire-plugin instead
2. add a **big warning** that this has changed and will be removed in the next release
3. remove it from the target-platform-configuration

laeubi:
Strange enough the type in the code for an extraRequirement is org.apache.maven.model.Dependency but in fact this is not a maven-dep but only used as a container object to pass the necessary data...

mickaelistria:
`extraRequirements` in `target-platform-configuration` can be useful for some other cases not related to tycho-surefire-plugin. It can be useful basically for anything that does requires to load or execute the bundle as it influence the target-platform and resolved dependencies, some other mojos can leverage it. It can also help a bunch to work around some deps management odditites. So in any case, we shouldn't remove this extraRequirements from target-platform-configuration.
tycho-surefire-plugin already has `dependencies` (https://www.eclipse.org/tycho/sitedocs/tycho-surefire-plugin/test-mojo.html#dependencies ) that is aimed at extending the dependencies. But it has some limitations, eg are those dependencies compatible with the target-platform and so on; but if we can remove these limitations, we could totally stop recommending using target-platform-configuration/dependency-resolution/extraRequirements from tycho-surefire-plugin and stick to `dependencies`.

laeubi:
> it can be useful basically for anything that does requires to load or execute the bundle

I think we should provide explicit solutions for such use-cases, I think you are refereeing to something like eclipse-run mojo? 

> It can also help a bunch to work around some deps management odditites.

Are there any examples/bug reports? I think we should try to fix or provide better alternatives

> But it has some limitations, eg are those dependencies compatible with the target-platform and so on; but if we can remove these limitations,

I'm currently working on a more generic solution for plugins to express extra requirements so they get passed to the P2 resolver. tycho-surefire can the use this as well I think.

mickaelistria:
> I think we should provide explicit solutions for such use-cases, I think you are refereeing to something like eclipse-run mojo?

[...]

PR #44, opened 2021-04-10
Move IDependencyMetadata to a shared place

laeubi:
relates to #43


laeubi:
@mickaelistria any objections here?

mickaelistria (approved):
That's a good change. I particularly like the replacement of boolean `primary` by the more explicit SEED/RESOLVE enum and the other refactorings factorizing around IDependencyMetadata.

Issue #37, opened 2021-04-09
Bug 572602 - Tycho should support classpath-container


laeubi:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=572602

laeubi:
https://github.com/eclipse/tycho/issues/93 shows another use-case for supporting the JRE-Classpath-Container

laeubi:
General support for this is avaible, we can open specific tickets for more support in this area

PR #49, opened 2021-04-11
Issue #37 - support for junit classpath container

laeubi:
This adds support for the JDT Classpath Container "JUNIT"


laeubi:
That's strange the test fails with 'Missing: org.eclipse.tycho:tycho-bundles-external:zip:2.4.0-SNAPSHOT' but if I run tycho with example projects I don't get this error :-(

laeubi:
Test are now working again, I'll try to cut this into smaller chunks.

laeubi:
Will require a different solution once 

- Tycho should respect "only visible for test sources" #66
- Support to supply additonal requirements through P2MetadataProvider #53

are resolved

PR #57, opened 2021-04-12
Bug 380169 - Make createEclipseInstallation in EclipseRunMojo thread safe

lara0905:
First step to make EclipseRunMojo thread safe. In further step use of the same workspace by multiple executions must be prevented.


mickaelistria (commented):
As we're progressively making efforts to enable more parallel work in Tycho; I think we should try to refine opportunities for parallelism.
Concretely, the only case where 2 eclipse-run are not allowed to run in parallel is if they share the same "work" directory; all other cases can allow parallel execution.
The current proposal doesn't implement that: it will allow 2 eclipse executions that share the same `work` area to execute in parallel. We should avoid that.
Basically, the lock should probably be implemented with a file marker in the work area.

mickaelistria (changes requested):
See previous comment.

mickaelistria:
Thanks a lot!

PR #58, opened 2021-04-12
Enable reuse of tycho classes in other Mojos

lara0905:


laeubi:
Maybe it would make more sense to extract what EclipseRunMojo does into a separate class that is then called from EclipseRunMojo and the ones that like to reuse it. Also the massive amount of constructor parameters might be better handled by some kind of EclipseRunBuilder.

mickaelistria:
In any case, I'd like to remind that Tycho at the moment does *not* maintain anything as API and will break things whenever it makes sense to Tycho without caring much about extensions.
Tycho guarantess some "usage" compatibility (mojos, parameters and so on) that may be better to rely on.
API compatibility is very expensive on the long run and won't be possible unless people who consume those APIs become major regular contributors to Tycho to balance the required effort.

lara0905:
We wouldn't mind adapting to API changes - it's still cleaner to change only our code after an API change than copying and changing Tycho code after an API change.

mickaelistria:
I think the proposal is good enough. Let's just merge it.

laeubi:
> I think the proposal is good enough. Let's just merge it.

@mickaelistria I'm not so certain about that. the constructor is obviously never used and will most probably be delete on the next refactoring. Also if we add new parameters its likely that they are not added to the constructor.

mickaelistria:
Sure, but as mentioned, we -as current maintainers of Tycho- are not really to care about this constructor nor to build only API for that. People who use this feature will have to take care of updating and so on. I'd perceive refactorings for a marginal unsupported use-cases being over-engineering, and the effort spent on other topics that increase without doubt Tycho quality and "value".

Issue #33, opened 2021-04-09
Bug 572570 - Tycho surefire must respect the requirements of the implicit test dependencies


laeubi:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=572570

PR #60, opened 2021-04-12
Fix #53 Support to supply additonal requirements through P2MetadataProvider

### LINKED ISSUE #53: Support to supply additonal requirements through P2MetadataProvider
This is a split from https://github.com/eclipse/tycho/pull/49 

Currently P2MetadataProvider can only supply either seed or resolve units. But JDT/PDE/P2 sometimes express a requirements on a more abstract level (e.g. additional bundles, test-container, ...). These requirements are not matching into the current categories and are best expressed as e.g. a compile-time dependency rather than a unit (e.g. concrete bundle or feature).

laeubi:

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/DependencyMetadataScope.java
+++ b/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/DependencyMetadataScope.java
@@ -13,15 +13,13 @@
 package org.eclipse.tycho;
 
 /**
- * {@link RequiredCapability} is the abstraction
- *
+ * {@link DependencyMetadataScope} refers to the scope where a dependency and/or requirement is
+ * relevant. For example some requirements are relevant for the initial resolving of a project (this
+ * can be seen s the minimal set of dependencies required to actually work with, e.g. to validate
+ * that all prerequisites are meet), others might be relevant for compiling (e.g. classpath entries,
+ * compiler levels) and others only for test execution (e.g. additional mock-services, fragments and
+ * so on).
  */
-public interface RequiredCapability {
-
-    String getNamespace();
-
-    String getId();
-
-    String getVersionRange();
-
+public enum DependencyMetadataScope {
+    SEED, RESOLVE, COMPILE, TEST;
 }
--- a/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/IDependencyMetadata.java
+++ b/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/IDependencyMetadata.java
@@ -12,13 +12,15 @@
  *******************************************************************************/
 package org.eclipse.tycho;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Set;
+import java.util.stream.Stream;
 
 public interface IDependencyMetadata {
 
     enum DependencyMetadataType {
-        SEED, RESOLVE;
+        SEED, RESOLVE, COMPILE;
     }
 
     Set<? /* IInstallableUnit */> getDependencyMetadata(DependencyMetadataType type);
@@ -26,5 +28,10 @@ public interface IDependencyMetadata {
     Set<? /* IInstallableUnit */> getDependencyMetadata();
 
     void setDependencyMetadata(DependencyMetadataType type, Collection<? /* IInstallableUnit */> units);
+
+    default <T> Stream<T> getDependencyMetadata(Class<T> type, DependencyMetadataType... types) {
+        return Arrays.stream(types).flatMap(t -> getDependencyMetadata(t).stream()).filter(type::isInstance)
+                .map(type::cast);
+    }
 
 }
--- a/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/IDependencyMetadata.java
+++ b/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/IDependencyMetadata.java
@@ -19,17 +19,13 @@ import java.util.stream.Stream;
 
 public interface IDependencyMetadata {
 
-    enum DependencyMetadataType {
-        SEED, RESOLVE, COMPILE;
-    }
-
-    Set<? /* IInstallableUnit */> getDependencyMetadata(DependencyMetadataType type);
+    Set<? /* IInstallableUnit */> getDependencyMetadata(DependencyMetadataScope type);
 
     Set<? /* IInstallableUnit */> getDependencyMetadata();
 
-    void setDependencyMetadata(DependencyMetadataType type, Collection<? /* IInstallableUnit */> units);
+    void setDependencyMetadata(DependencyMetadataScope type, Collection<? /* IInstallableUnit */> units);
 
-    default <T> Stream<T> getDependencyMetadata(Class<T> type, DependencyMetadataType... types) {
+    default <T> Stream<T> getDependencyMetadata(Class<T> type, DependencyMetadataScope... types) {
         return Arrays.stream(types).flatMap(t -> getDependencyMetadata(t).stream()).filter(type::isInstance)
                 .map(type::cast);
     }
--- a/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/ReactorProject.java
+++ b/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/ReactorProject.java
@@ -25,6 +25,8 @@ public interface ReactorProject extends IDependencyMetadata {
      */
     public static final String SOURCE_ARTIFACT_CLASSIFIER = "sources";
 
+    public static final String CTX_REQUIREMENTS = "tycho.requirements";
+
     public File getBasedir();
 
     public String getPackaging();
@@ -64,6 +66,8 @@ public interface ReactorProject extends IDependencyMetadata {
     public String getBuildQualifier();
 
     public String getExpandedVersion();
+
+    public Requirements getRequirements();
 
     // misc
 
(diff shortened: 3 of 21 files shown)


mickaelistria:
I don't get what this does better than the current capabilities of target-platform-configuration/dependency-resolution/extraRequirements ; as it seems to me the requirements here are immutable for the project, so those requirements cannot be set differently by different mojo and hence, they still affect the resolution results that's shared across multiple mojos.
Please add documentation in code and in commit message to better explain what's the goal of the various changes.

laeubi:
> I don't get what this does better than the current capabilities of target-platform-configuration/dependency-resolution/extraRequirements ; 

I have tried to explain it here in a bit more detail https://github.com/eclipse/tycho/issues/53 maybe you can add there anything that's unclear and I can try to enhance the description?

>as it seems to me the requirements here are immutable for the project, so those requirements cannot be set differently by different mojo and hence, they still affect the resolution results that's shared across multiple mojos.

Each mojo/plugin can contribute as many requirements as he likes. This looks a bit strange because of the split between tycho-maven/tycho-osgi ... Maybe it helps to take a look at https://github.com/eclipse/tycho/pull/49/files#diff-61ada7cb0bee2a7485a8cfe949369d96447be49526d02d4114d7192b878a22c8R42 what is using this feature?

mickaelistria:
> I have tried to explain it here in a bit more detail #53 maybe you can add there anything that's unclear and I can try to enhance the description?

Such description, comment and explanations should be in the commit message and in the code. Otherwise, it's too easy to loose track. But even in #53, I really miss to understand what the actual use-case that's missing.

> Each mojo/plugin can contribute as many requirements as he likes.

OK, so although it's not fully clear from me when looking at the code, that's what I understand as the goal. And it's a good goal.
However, what I don't understand is that "requirements" come from `reactorProject.getRequirements()`, I don't see how the mojos can add requirements that are specific to themselves without leaking to other mojos? Ie I don't see what's different if the "extraRequirements" are still applying to project and not to a mojo-specific p2 resolution.
Concretely, for surefire, I'd just expect that the resolution happens with the extra deps from dependencies and from the container, so all that could be code inside tycho-surefire-plugin.

laeubi:
The goal is not to have mojo specific requirements, they are still global but more fine grained e.g. surefire would for example use requirements from the category 'compile'+'test', the compile-mojo only those for 'compile', a update-site mojo only those of 'resolve' and so on.

That's implicit the case currently, e.g the compile mojo currently ignores resolved units it don't determines to be relevant. I'll try to enhance the javadoc here, its just a bit hard to make this perfectly clear without all parts in place.

mickaelistria:
Co basically it's more about reproducing Maven scopes? Maybe rename DependencyMetadataType to DependencyMetadataScope.

If the initial issue is about tycho-surefire-plugin able to read the test container and other test deps; couldn't tycho-surefire-plugin invoke some `OsgiBundleProject.getDependencyMetadata(TEST)` method and OsgiBundleProject taking care of statically reading the .classpath, processing the relevant containers and so on to populate the metadata?
I think it would be simpler and straighter to the point compared to the proposal here of mojo extending the requirements dynamically. This case seems standard enough to be part of the "core" resolution without introducing some indirections.

laeubi:
The problem is that I want to avoid to add a new "way" for each possible thing to integrate, also OsgiBundleProject do not has the necessary scope information to handle all cases.

mickaelistria:
> a new "way" for each possible thing to integrate

We're not trying to integrate each possible thing, we/you just want a way to have the JUnit container populating a set of test dependencies that are to be consumed by tycho-surefire-plugin.
By the introduction of DependencyMetadataType/DependencyMetadataType, you already introduce the new idea that the OsgiBundleProject (implements IDependencyMetadata) does deal with scopes; so just 1 extra TEST scope for test specific deps and reading .classpath as an extra way to populate them is way less "new way" compared to current state and regular Maven lifecycle than the addition of IRequirement, IRequiredCapability and so on as you're suggesting here.

laeubi:
ReactorProject implements IDependencyMetadata, and sadly this is just a container to exchange data between tycho-maven and tycho-osgi :-\ IRequiredCapability is neccesary because the tycho-maven part do not has access to the P2 IRequirement of tycho-osgi. that's a bit painfull here but I don't see a solution without massive refactoring (not only in tycho) to solve this.

Also this is required as IDependencyMetadata currently only can handle IUs but not requirements... my hope is to move more and more code to the requirements section over time so IDependencyMetadata becomes obsolete then and could be removed.

laeubi:
Just forget to mention: Even though the Requirements are "global" like the extraRequirements they are more specific as for example only projects that have a JUNIT container on its classpath will get additional (test/compile) requirements.

laeubi:
I have adjusted naming and javadoc now.

> we/you just want a way to have the JUnit container populating a set of test dependencies that are to be consumed by tycho-surefire-plugin.

This will hopefully supersede current additional.bundles special handling and add support for surefire-plugin to get its `<dependecies>` resolved reliable, so this is (for me) the entry-key into a new more flexible way in many areas. Maybe Tycho is near to feature complete but I still found things to improve at the moment :-)

laeubi:
BTW this is also related to #33 currently tycho-surefire only claims requirements to the junit ***bundles*** but Java 11 do not contain the javax.annotation package a test bundle that do not manually imports them fails with a real cryptic error message.

But including any ***bundle*** that provides this package is not desirable without the danger of messing things up if the test-setup contains a conflicting dependency.

So actually, surefire needs to expose a requirement to the ***package*** javax.annotation instead and that's where this feature should extend beyond the current concept of explicit required IUs. Aslo we currently embeds several junit.xxx fragment providers. Also these are actually additional requirements on bundles and/or packages that should be resolved through the target platform to prevent collisions with the ones provided there (#63 is related here).

mickaelistria:
Good you relate it to #33 ;) IMO #33 is a simpler expression of the same issue and a solution for #33 would probably allow to fix this current one in a simpler way or at least by isolating one critical item: the project needs to compute a set of "test dependencies" (an alternative resolution result) based on requirements of "compile" + extra requirements. When this is isolated, added 1 extra source for "test dependencies" (eg a container from .classpath) will become trivial.

laeubi:
@mickaelistria I would be happy to see this "trivial" solution so please provide a PR for it as it seems I wasted a massive amount of going in the wrong direction then :-\

This whole topic is not trivial at least for me and this was the less intrusive but still most flexible solution I could came up with.

mickaelistria:
> I would be happy to see this "trivial" solution so please provide a PR for it

Sorry, I don't have need for it at the moment so I'll prioritize other stuff

>  I wasted a massive amount of going in the wrong direction then :-\

It's not waste. "Simple solutions are usually the result of a hard work" ;)

> the less intrusive but still most flexible solution I could came up with.

That may be the issue: the story of properly resolving test dependencies is fairly standard workflow and important usage, it seems more interesting to go for an intrusive and "rigid" solution that roots in the basics of project lifecycle; just like Maven itself does.

laeubi:
Refactoring Tycho "from the roots" is currently a bit out of scope, so I seek for a solution that works right now.  So I would suggest we start with this and see how things evolve. If it proves to not fulfill the requirements on the long run we can adjust/refactor it.

mickaelistria:
Sorry, I think your proposal is currently a bit over-engineered and adds a lot of complexity to work around an actual issue (eg #33 and family). IMO, in the current form, the change isn't sufficiently profitable to be worth being merged and I'd rather avoid this solution being merged in and becoming a "long term workaround" we start building upon.
Let's just start by rephrasing the issue of #33 to something like "extra test dependencies are added without dependency resolution" and then from here place a `test`-scope dependency resolution and make tycho-surefire-plugin use it (basically `project.getTestDependencyArtifacts()` or `project.getDependencyArtifacts(IDependencyMetadataScope.TEST)`. I don't think having this is more difficult than what you achieved with this patch.

laeubi:
This is not to solve #33 that's your interpretation and its not about sure-fire and its not about only test-dependecies, I mentioned the issue to show that the current approach is not flexible enough.
If you think it could be solved better please suggest concrete changes with code, saying that its 'trivial' or 'not more difficult' but on the other hand refuse to show or even prove this is really frustrating given the amount of time for developing and debugging that stuff to a working solution.

I have already pointed out that maven-mojo code can't use the dependencies directly due to the P2 API is not available to maven and thus the "over-engineered" solution. If you think you can fix this with trivial changes a LOTS of code could be simplified! I just wonder why this has never happened in the past if its that trivial?

laeubi:
@akurtakov @jsievers  I fear we could need some mediation here, this is really blocking me in current development of my testing-storys.

I'm open to alternative approaches (proven in code), but refactoring the complete core of tycho dependency resolution is not really an option for me.

mickaelistria:
> I'm open to alternative approaches (proven in code), but refactoring the complete core of tycho dependency resolution is not really an option for me

What you're suggesting with this PR is a already refactoring of tycho dependency resolution. Except that your refactoring is more/too generic, not isolated and increasing complexity because it requires to implement flexibility/extensibility and scatters responsibilties across more places, while typical Maven usage is more on a centralized dependency definition and resolution.
Implementing https://github.com/eclipse/tycho/pull/60#issuecomment-818016831 would probably be less difficult to produce than the work if took you to produce this PR.

laeubi:
As mentioned before it dose not help to speculate about 'probably' add here or there, please just try to add the required API here that includes:

a) Reactor Project / P2MetadataProvider
b) P2 resolver /  Equinox Resolver
c) Compile Mojo (just as an example)

and I can try to adapt/add my implementation there....

(a) can't see classes from (b) that why we have 'DependencyMetadataGenerator' there, and (c) can't see classes from (b) because we have 'Dependecy' container that copies around strings that are actually IUs.

mickaelistria:
I'll start to work on it soon.

laeubi:
> I'll start to work on it soon.

Are there already any updates on this?

mickaelistria:
>  Are there already any updates on this?

I've done some progress, I plan to push a PR later today.

laeubi:
Obsolete

Issue #55, opened 2021-04-12
discard test-fragment packaging


laeubi:
Adding more stages to the test story has shown that the test-fragment packaging is too strict and does not scale well (e.g. has hard dependency to additional.bundles feature that should be generalized in #53) as this is not released/used in the wild we better discard this now before there are consumers of this feature.

The following should replace this feature:

- [ ] #53 replace the support for additional-bundles
- [ ] #37 replace the support for loading additional classes from test-classpath
- [ ] dev.properties already cover the case of test-classes to be loaded from test-source folders

PR #67, opened 2021-04-14
#55 - discard packaging of test-fragment

laeubi:

### DIFF
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/surefire/TestsInBundleTest.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/surefire/TestsInBundleTest.java
@@ -46,15 +46,6 @@ public class TestsInBundleTest extends AbstractTychoIntegrationTest {
     }
 
     @Test
-    public void testPackage() throws Exception {
-        Verifier verifier = getVerifier("surefire.combinedtests/bundle.test");
-        verifier.executeGoals(Arrays.asList("clean", "package"));
-        verifier.verifyErrorFreeLog();
-        assertTrue("test fragment not found",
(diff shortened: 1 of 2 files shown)


akurtakov:
What is the reason behind this? There are projects (like Platform, CDT, Linux Tools) that use test fragments and publish them to p2 repository so they can be installed later in a provisioned IDE and executed on different OS/WS/ARCH combo without rebuilding everything.

mickaelistria:
I also don't get why removing this now. Please add explanation in the commit message.

laeubi:
This was recently added by me to package classes in target/test-classes (what was previously empty) and generates some kind of generic Manifest, but I found that my initial idea was not ideal and its original intent becomes a bit obsolete now as PDE refuses to allow something similar in the IDE it does not make sense to go this path further and instead I'd https://github.com/eclipse/tycho/issues/53 will replace this.

So no worries this does not affect existing use-case.

mickaelistria:
OK, please just add those explanation to the commit message, also fix related tests and then feel free to merge as soon as CI is green.
Could it be a revert of https://github.com/eclipse/tycho/commit/d9a58240820776f11feb0a8dbaabfaaa00a4619d ?

laeubi:
I'll do as soon as the replacement for additional bundles is in place, just wanted to see if I break more than these three test cases :-)

The original commit was https://github.com/eclipse/tycho/commit/9485693f128bf3a634f657f0df6e27daab60fbdf but I think we could keep the extraction of manifest creation as it makes code more clear so I decided to not revert this completely.

laeubi:
Two test are failing but recent discussions have show that there is no way to support this, so I'll remove the failing test as they are invalid.

PR #71, opened 2021-04-15
Issue 70: Make org.eclipse.tycho.surefire.junit* libs compatible with Java 1.8

sewe:
See https://github.com/eclipse/tycho/issues/70

### DIFF
--- a/tycho-surefire/tycho-surefire-plugin/pom.xml
+++ b/tycho-surefire/tycho-surefire-plugin/pom.xml
@@ -44,6 +44,36 @@
 				<artifactId>maven-surefire-plugin</artifactId>
 				<version>${surefire-version}</version>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-enforcer-plugin</artifactId>
+				<version>3.0.0-M3</version>
+				<executions>
+					<execution>
+						<id>enforce-runtime-jdk-compatibility</id>
+						<goals>
+							<goal>enforce</goal>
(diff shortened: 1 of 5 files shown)


mickaelistria:
@akurtakov what do you think?

akurtakov:
Java 8 is dead for me and I would not move a finger for it but as long as there are people to keep doing the changes for it to work I welcome it. Some of them might fix other issues too someday using their gained knowledge and already existing environment.

tycho-surefire/tycho-surefire-plugin/pom.xml
```diff
@@ -44,6 +44,36 @@
...
+						<id>enforce-runtime-jdk-compatibility</id>
+						<goals>
+							<goal>enforce</goal>
+						</goals>
+						<phase>verify</phase>
+						<configuration>
+							<rules>
+								<enforceBytecodeVersion>
+									<!-- Surefire Booter Eclipse Application and JUnit/TestNG Runners -->
+									<includes>org.eclipse.tycho:org.eclipse.tycho.surefire.*</includes>
```
sewe, line 62: Relying on names is a bit awkward here. Ideally, one would place the `<enforceBytecodeVersion>` rule in the `org.eclipse.tycho.surefire.*` themselves, but they don’t have an _real_ `<dependencies>`, which the `<enforceBytecodeVersion>` rule can check but only `<artifactItems>` for the `dependency:unpack` goal, which it cannot see. So placing the sanity check in `tycho-surefire-plugin` is the best I could come up with.

Issue #72, opened 2021-04-16
Create a Java17 SE profile in Tycho


jarthana:
Add new JavaSE profile and associated code/tests in Tycho for JavaSE-17.

PR #73, opened 2021-04-16
Issue #72 - Create a Java17 SE profile in Tycho

jarthana:
Signed-off-by: Jay Arthanareeswaran <jarthana@in.ibm.com>


tycho-core/src/main/java/org/eclipse/tycho/core/ee/ExecutionEnvironmentUtils.java
```diff
@@ -49,8 +49,8 @@
         Properties listProps = readProperties(findInSystemBundle("profile.list"));
         List<String> profileFiles = new ArrayList<>(Arrays.asList(listProps.getProperty("java.profiles").split(",")));
         profileFiles.add("JavaSE-11.profile");
-        profileFiles.add("JavaSE-15.profile");
```
mickaelistria, line 52: Let's keep JavaSE-15 in place for the moment.

mickaelistria (changes requested):
See inline comment about keeping JavaSE-15

mickaelistria:
See CI build, Java15ResolutionTest needs to be fixed to not reference JavaSE-15 any more in product builds.

laeubi:
@mickaelistria @akurtakov I know we have discussed this before but I still wonder if it is beneficial to completely remove "old" profiles from tycho.

It was argued that they might impose maintenance cost or get outdated, but hand I never seen any complains or work in this area as the just seem to satisfy the user needs. On the other hand there where complains when users upgraded their tycho version and suddenly the build breaks.

Wouldn't it be an option to instead completely removing them add a flag to the profile (e.g. first line = `#deprecated this version of java is out of date since` .... ) and tycho simply omits a warning with this text if the profile is used + a hint that there is no further support/testing/bugfixes for this profile anymore?

mickaelistria:
> I still wonder if it is beneficial to completely remove "old" profiles from tycho.

Where do you see the profiles removed?

> On the other hand there where complains when users upgraded their tycho version and suddenly the build breaks.

Can you please point to specific complaints?

akurtakov:
Main reason behind not keeping old Java versions are/were:
* Static package list in profiles which are simply "wishes" with modular JVM. That's no longer an issue as package list is no longer part of the profile
* Publishing of the profile - we publish a profile which if not defined in toolchain.xml will simply be mapped to current running JVM aka "[WARNING] No system packages found in profile nor toolchain for JavaSE-15, using current JRE system packages.
This can cause faulty dependency resolution, consider adding a definition for a 'jdk' with id=JavaSE-15 in your toolchains.xml
[INFO] Published profile IUs: [config.a.jre.javase 15.0.0, a.jre.javase 15.0.0]" . Even if it has small performance hit thus it can be ignored from correctness POV it's a problem which we don't have a way to fix.
* Java version no longer being supported - my view on the topic is that when smth is officially not supported it's a duty of the tool to prevent usage of it. Now, it becomes questionable how long we keep profiles for unsupported JVMs but I don't agree we should have Tycho publishing many useless profiles just to print WARNING about publishing them when majority of users never need it nor actually required it.

I would we can discuss retention period but the question of keeping profiles for extended periods will need someone to find a way to fix my point 2.

mickaelistria:
`a.jre.javase` units in general are bad. Fixing point 2 needs work in p2 to get rid of the need to have this units at runtime: https://bugs.eclipse.org/bugs/show_bug.cgi?id=568150 . Then, when p2 can still execute without them, we can more easily consider not publishing them.

laeubi:
https://github.com/eclipse/tycho/pull/73/commits/5d137e4511083ac044e63a11bc8107bc76a117d4 has removed it and 
https://github.com/eclipse/tycho/pull/73/commits/d905485330cb0edb3b99c7cea23c458bcd8f2509 restored it. So atm the profile is there but it seems there where previously some demand to remove it.

Also you have suggested to 'Java15ResolutionTest needs to be fixed to not reference JavaSE-15 any more' that seems to suggest that a previously working example is broken now (and thus would need fixing for anyone using the same in any of his/her builds).

> "[WARNING] No system packages found in profile nor toolchain for JavaSE-15,

I think the solution should not be to remove profiles but to get rid of useless warnings. e.g. I have a build where everything is build with Java 11 and get always warnings about J13, 13... (whatever) even though I have never any need for those. My preffered solution would be that tycho collects the java versions of the build/updatsite/product/... and only publish those profiles that where relevant.

> Java version no longer being supported - my view on the topic is that when smth is officially not supported it's a duty of the tool to prevent usage of it. 

That's why I would find it more useful (from a tool-user-POV) to get a warning for old profiles (if used (!), see above) like

'Java 8 is EOL since ... its extended support period expires at ... Tycho will no longer support/test/enhance/fix/... this profile from version 2.X onwards'

or something similar to the user than simply make him feel never Tycho version do not support his previous working build anymore.

I found no good way to search the mailing list archives but remember that there was here and there complains about profiles that "magically" vanished and users suspect this to be a bug of tycho.

mickaelistria:
> Also you have suggested to 'Java15ResolutionTest needs to be fixed to not reference JavaSE-15 any more' that seems to suggest that a previously working example is broken now

This test is more meant to verify internal consistency of Tycho, not really testing some user-story. So it's not highlighting a broken use-case.

> I think the solution should not be to remove profiles but to get rid of useless warnings. e.g. I have a build where everything is build with Java 11 and get always warnings about J13, 13... (whatever) even though I have never any need for those

Than you should investigate where these warnings come from. If you're not using Java 13 nor referencing it at all, those warnings shouldn't show up. It means those warning are worth investigation and shouldn't be silenced.

> That's why I would find it more useful (from a tool-user-POV) to get a warning for old profiles (if used (!), see above) like 'Java 8 is EOL since ... its extended support period expires at ... Tycho will no longer support/test/enhance/fix/... this profile from version 2.X onwards'

Are you willing to implement that and make sure you properly maintain the dates on the profile files, and make sure you schedule updates to Tycho so that the support is dropped when announced, in a way that doesn't slow down Tycho development? If you see enough value to make it profitable to you, that could be possible to add such a strategy, but it's not a development and a maintenance I find profitable at all for my goals (and I'm pretty certain it's even worse for Alex) and not something I think really is high priority for Tycho to enable its users in being more successful. On the contrary, encouraging users to move to newer Java release is actually helping them more than letting them stick on unsupported releases for a long time and face a technical debt they won't be able to deal with; while also reducing (or keeping constant) maintenance and support cost of Tycho.
And in the worst case, people can still use older version of Tycho if they're stuck with older Java. After all, they're used to be stuck to old releases enough apparently. The slower ones to migrate are the ones who slow down innovation, so unless they really support backward compatibility with a lot of manpower or $$$, I don't think they're worth caring about; and without such support, project should sticks to versions that are technically relevant (eg Java releases with upstream support) and drop everything else it's not interested in.

But if you're really seeing a personal ROI here and are willing to spend effort on that, I think it would be fine to merge patches that do as you suggest.

mickaelistria:
Thanks @jarthana !

laeubi:
> Than you should investigate where these warnings come from. If you're not using Java 13 nor referencing it at all, those warnings shouldn't show up.

They 'magical' appear once I upgraded tycho :-)

And the even occur with tycho build itself, I'm not aware that tycho depends on J15 or J16 features somewhere:

```
[INFO] --------------< org.eclipse.tycho:tycho-bundles-external >--------------
[INFO] Building Tycho OSGi Runtime External Bundles 2.4.0-SNAPSHOT       [5/72]
[INFO] -------------------------[ eclipse-repository ]-------------------------
[INFO] 
[INFO] --- maven-clean-plugin:3.1.0:clean (default-clean) @ tycho-bundles-external ---
[INFO] Deleting /home/christoph/git/laeubi/tycho/tycho-bundles/tycho-bundles-external/target
[INFO] 
[INFO] --- tycho-packaging-plugin:2.3.0:build-qualifier-aggregator (default-build-qualifier-aggregator) @ tycho-bundles-external ---
[INFO] The project's OSGi version is 2.4.0.202104190814
[INFO] 
[INFO] --- maven-clean-plugin:3.1.0:clean (default-clean-1) @ tycho-bundles-external ---
[INFO] 
[INFO] --- maven-resources-plugin:3.2.0:resources (default-resources) @ tycho-bundles-external ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] Using 'UTF-8' encoding to copy filtered properties files.
[INFO] skip non existing resourceDirectory /home/christoph/git/laeubi/tycho/tycho-bundles/tycho-bundles-external/src/main/resources
[INFO] 
[INFO] --- target-platform-configuration:2.3.0:target-platform (default-target-platform) @ tycho-bundles-external ---
[INFO] 
[INFO] --- tycho-p2-publisher-plugin:2.3.0:publish-osgi-ee (default-publish-osgi-ee) @ tycho-bundles-external ---
[INFO] Published profile IUs: [config.a.jre.javase 11.0.0, a.jre.javase 11.0.0]
[WARNING] No system packages found in profile nor toolchain for JavaSE-15, using current JRE system packages.
This can cause faulty dependency resolution, consider adding a definition for a 'jdk' with id=JavaSE-15 in your toolchains.xml
[INFO] Published profile IUs: [config.a.jre.javase 15.0.0, a.jre.javase 15.0.0]
[WARNING] No system packages found in profile nor toolchain for JavaSE-16, using current JRE system packages.
This can cause faulty dependency resolution, consider adding a definition for a 'jdk' with id=JavaSE-16 in your toolchains.xml
[INFO] Published profile IUs: [config.a.jre.javase 16.0.0, a.jre.javase 16.0.0]
```

> Are you willing to implement that and make sure you properly maintain the dates on the profile files

probably yes, but thats not on my top-list. I just mention this as I see we can enhance the user-experience here and give good/better advice. Maybe (even if uncertein) you or Alex when next time touching the code see a smart/cheap opprtunity to add support for such a feature and reminds back the discussion...

> On the contrary, encouraging users to move to newer Java release is actually helping them more than letting them stick on unsupported releases for a long time and face a technical debt

But thats exactly the point! Getting a warning each build reminding me of outdated java (some companys require warning free builds and take actions on it) will encourage (at laest me) users much more to upgrade while a breaking build people more encourage users (including me) to use old tools as the cost might be to high to upgrade now.

Issue #74, opened 2021-04-19
Project compiler settings cannot be overridden when running Maven


sithmein:
It is not possible to override the Eclipse project compiler settings on the command line using `<compilerArgs>` in the _tycho-compiler-plugin_ configuration. According to the Eclipse batch compiler documentation this is  possible, e.g. by passing

```
<configuration>
  <arg>-err:-nullAnnot</arg>
</configuration>
```

However, the complete command line generated by Tycho looks as follows:

> [DEBUG] JDT compiler args: [-err:-nullAnnot, -warn:-nullAnnot, -properties,  /.../.settings/org.eclipse.jdt.core.prefs, -s, ...] 

The manually provided compiler arguments are always put at the front of the argument list. However, according to the batch compiler documentation the *last* argument determines the effective settings which would be the project configuration. This means it's impossible to override it on the command line.

mickaelistria:
Thanks!

PR #75, opened 2021-04-19
Bug #74: Fixed order of custom compiler arguments

sithmein:
Not a super-nice solution but it works. I couldn't find another way to change the argument order.


sithmein:
GitHub seems to associate this PR with the wrong email address. My other address has signed the ECA.

mickaelistria:
Please make sure the commit "author" is using the address you used to sign the ECA, or sign ECA with other address as well.

Issue #63, opened 2021-04-12
Bug 567913 - 'junit.runner' missing in tycho surefire environment for JUnit 5 tests


ernstblechaJKU:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=567913

> Stephan Wahlbrink 2020-10-15 13:43:27 EDT
>
>The package 'junit.runner' is missing in tycho surefire environment for JUnit 5 (Jupiter) tests since Tycho 1.7:
>
>> java.version=11.0.8
>> java.vendor=AdoptOpenJDK
>> BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=de_DE
>> Framework arguments:  -application org.eclipse.tycho.surefire.osgibooter.headlesstest -testproperties C:\eclipse.statet\_build\org.eclipse.statet-commons\jcommons\org.eclipse.statet.jcommons.util-tests\target\surefire.properties
>> Command-line arguments:  -data C:\eclipse.statet\_build\org.eclipse.statet-commons\jcommons\org.eclipse.statet.jcommons.util-tests\target\work\data -application org.eclipse.tycho.surefire.osgibooter.headlesstest -testproperties C:\eclipse.statet\_build\org.eclipse.statet-commons\jcommons\org.eclipse.statet.jcommons.util-tests\target\surefire.properties
>> 
>> !ENTRY org.eclipse.tycho.surefire.junit56 2 0 2020-10-15 19:30:30.591
>> !MESSAGE Could not resolve module: org.eclipse.tycho.surefire.junit56 [109]
>>   Unresolved requirement: Import-Package: org.junit.runner; version="[4.12.0,5.0.0)"; resolution:="optional"
>>   Unresolved requirement: Import-Package: org.junit.runner.manipulation; version="[4.12.0,5.0.0)"; resolution:="optional"
>>   Unresolved requirement: Import-Package: org.junit.runner.notification; version="[4.12.0,5.0.0)"; resolution:="optional"
>>   Unresolved requirement: Import-Package: org.junit.runners.model; version="[4.12.0,5.0.0)"; resolution:="optional"
>>   Unresolved requirement: Import-Package: org.junit.experimental.categories; version="[4.12.0,5.0.0)"; resolution:="optional"
>>   Unresolved requirement: Import-Package: org.junit.internal.builders; version="[4.12.0,5.0.0)"; resolution:="optional"
[...]

PR #77, opened 2021-04-19
Bug 567913: Add test case for JUnit 5 only bundle

ernstblechaJKU:
This is the propsed integration test referred to in issue #63 

This test will fail if junit.runner cannot be resolved

> Change-Id: Iea22f522013057077acb0ff9e2852147ffe37a4c
> Signed-off-by: Stephan Wahlbrink <sw@wahlbrink.eu>


laeubi:
@ernstblechaJKU thanks for this test-case, I'm currently working on improving test-support in tycho so this is really useful and I think is related to #33

@mickaelistria @akurtakov can you see any "wrong" usage in the supplied example? for me they look as valid seut a user might choose

laeubi:
Any idea why this includes a rename of an existing test? Is this intentional?

laeubi:
Even though the test is named 'JUnit54Test' instead JUnit5.7 is used but I suspect that this does not really matter to you:

> !MESSAGE Could not resolve module: org.eclipse.tycho.surefire.junit57

ernstblechaJKU:
> 
> 
> Any idea why this includes a rename of an existing test? Is this intentional?

It's not my intention - i only moved the test over from gerrit; i'll take a look as soon as possible

ernstblechaJKU:
> 
> 
> > Any idea why this includes a rename of an existing test? Is this intentional?
> 
> It's not my intention - i only moved the test over from gerrit; i'll take a look as soon as possible

Please excuse that it took me this long to take another look, it appears that the intention behind this was to move a test for JUnit4 to a more fitting folder (Testing JUnit4 and JUnit5 in combination).

So yes, although i _was_ not aware why this was done it seems to be intentional

laeubi:
@mickaelistria @akurtakov any chance to take a look at this?

akurtakov:
Would you please squash the merge commit in the previous one?

mickaelistria:
Thanks a lot for this contribution which really helped finding a solution. It got merged with https://github.com/eclipse/tycho/pull/333

PR #78, opened 2021-04-19
[WIP] Support for Test dependencies/classpath

mickaelistria:


mickaelistria:
cc @laeubi ; this drafts how test requirements/resolved dependencies/classpath can hook into API explicitly and how compiler/surefire can leverage them. For the moment the main methods about computing extensions are empty, they are placeholder where to hook things like parsing .classpath or whatever strategy makes sense to explicit test specific deps.

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
```diff
@@ -242,6 +247,11 @@ public void resolveClassPath(MavenSession session, MavenProject project) {
         addPDESourceRoots(project);
     }
 
+    private Collection<ClasspathEntry> computeExtraTestClasspath(ReactorProject reactorProject) {
+        // TODO read .classpath and so on to resolve to extra test classpath entries
```
laeubi, line 251: Actually that would be the interesting part for me, how do I hook into this without 

a) coding all possible classpath container right here (a container might be test related or not by configuration)
b) how could i contribute to the resolving phases so the required deps are actually there?

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
```diff
@@ -593,4 +603,26 @@ private void applyBestOfCurrentOrConfiguredProfile(String configuredProfileName,
...
+    }
+
+    public DependencyArtifacts getTestDependencyArtifacts(ReactorProject project) {
+        return TychoProjectUtils.getTestDependencyArtifacts(project);
+    }
+
+    @Override
+    public List<Dependency> getExtraTestRequirements() {
+        // TODO parse .classpath to resolve "test" scope deps in it and feed those extra dependencies
+        return Collections.emptyList();
```
laeubi, line 625: Actually "Dependency" might be a package with version range or a capability, how would this be expressed?

tycho-core/src/main/java/org/eclipse/tycho/core/resolver/DefaultTychoResolver.java
```diff
@@ -138,10 +139,28 @@ public void resolveProject(MavenSession session, MavenProject project, List<Reac
...
+            /*
+             * TODO At the moment, we don't resolve test-specific deps, so we set it to common deps.
+             * This will be refined later with addition of support of test-specific deps from
+             * .classpath.
+             */
+            List<Dependency> testDependencies = ((BundleProject) tychoProjectType).getExtraTestRequirements();
+            if (!testDependencies.isEmpty()) {
+                logger.info("Resolving test dependencies of " + project);
+                resolverConfiguration.getExtraRequirements().addAll(testDependencies);
+                testDependencyArtifacts = resolver.resolveDependencies(session, project, preliminaryTargetPlatform,
```
laeubi, line 154: How does P2 resolver knows about those things? at least in my previous testing the P2 resolver needed them to build the set of required Units or does it means it will resolve the deps twice?
Also as mentioned above one actually don't want to resolve a given unit but a given requirement that boils down to 

- namespace
- properties (e.g. versionrange, id and so on...)

laeubi (commented):
If I understand right you added a second set of (test) DependencyArtifacts (whats good I think to distinguish) but I don't understand how I would contribute a requirement e.g.

osgi.wiring.package;filter:="(&(version>=1.1)(version<2))"

mickaelistria:
> I don't understand how I would contribute a requirement e.g. osgi.wiring.package;filter:="(&(version>=1.1)(version<2))"

This approach doesn't allow that at the moment, and nothing in Platform/PDE does allow that as far as I know (beyond simply creating a new OSGi artifact). So I think this story deserve more complete proposal, starting from the IDE: how would you imagine such requirement being defined in the IDE (without creating an extra bundle) in a simpler way than the current "create a test fragment" approach?

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
mickaelistria: a. What existing containers do you have in mind in practice? I think in any case, there needs to be code that "expands" a containers to extra requirements/classpath entries somewhere, and Tycho could support the typical ones.

b. See later: this steps happens *after* the p2&OSGi resolution, which do take care of resolving deps.

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
mickaelistria: There is currently nothing that allows to express that in Tycho "extraRequirements" as far as I know. The issue seems relatively independent technically (ie we could already start supporting
```xml
  <extraRequirement>
    <type>osgi.wiring.package</type> <!-- or java.package to use p2 wording -->
    <id>com.acme.pack</id>
    <version>0.0.0</version>
  <extraRequirement>
```
and use the same thing for extra test requirements generated from containers or whatever.

I think this story of expliciting requirement on packages deserves its own issue and PR, and then we can easily join both.

laeubi:
JUnit Container currently works this way but you can't influence this as a user, and actually #33 suffers from that. PDE has some nasty hard coded "extra classpath entries" that accounts for that but I really would avoid that for tycho ...

tycho-core/src/main/java/org/eclipse/tycho/core/resolver/DefaultTychoResolver.java
mickaelistria: Not sure exactly what you mean by "those things".
p2 resolver does translate extra requirements (as org.apache.maven.model.Dependency) into p2 requirements and run against them.

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
laeubi: If we can extend this to generic namespace that would be useful. For example to pull in DS one should describe the requirement in the following forms:

```
Require-Capability: osgi.extender;
  filter:="(&(osgi.extender=osgi.component)(version>=1.4)(!(version>=2.0)))"
```
http://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-capabilities

In P2 this is expressed by the org.eclipse.equinox.internal.p2.metadata.IRequiredCapability requirement and you can require other "P2" properties to (e.g. ws, os, ...)

mickaelistria:
> JUnit Container currently works this way

I don't get it; currently the JUnit Container has nothing related to OSGi in its definition. Or are you talking about how PDE interprets and expands the JUnit container?
If so

> PDE has some nasty hard coded "extra classpath entries" that accounts for that but I really would avoid that for tycho ...

That seems to be a good pragmatic 1st iteration to just add requirement on `org.junit` bundle to test deps when the container is present; it's been successfully enough for PDE so far.
It seems to me like the discussion above Tycho currently having no way to add requirements on packages so far. I would not expect we can easily improve the classpath container -> package requirement -> dependency resolution story as long as the "package requirement" part is already missing in other similar existing parts (`extraRequirements`). Both seem connected but not inter-independent stories and progress can happen on both fronts from now on.

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
laeubi: But how P2&OSGi knows that there are required? Would that not mean we need to put classpath parsing here and there?

For the containers, I think tycho should support JUNIT, UserLib  and maybe JRE as an alternative to BREE (but thats more for the future) ClasspathVariables could also be useful on the long term, but as containers are extensible one would need some way to provide a tycho(-extras) plugin to support specific ones. For example I'm currently develop a plugin that will contain other classpath types so I'd like to supply users with a small plugin to support this special one too.

In essence it should be possible to put 
https://github.com/eclipse/tycho/blob/400c244e564cf2f52744a032f284e271ca43ffe5/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/container/JunitClassPathContainer.java
into an own plugin, it adds the necessary requirements and everything else gets aware of them.

mickaelistria: > Would that not mean we need to put classpath parsing here and there?

Yes, here and there exactly, as suggested by the TODO

> For the containers, I think tycho should support JUNIT, UserLib

OK, so let's just fill this TODO.

> ClasspathVariables could also be useful on the long term

OK, but let's make it another iteration later.

>  For example I'm currently develop a plugin that will contain other classpath types so I'd like to supply users with a small plugin to support this special one too.

That will require people who use your container and need it at build time to also add an extra tycho extension for there build to resolve the classpath container to a set of dependencies/classpath entries.
So basically enforcing new extensibility and so on. Why not, but better to postpone it in a further step, when we have support for standard containers working well enough to get clear ideas of how to factorize better.

> into an own plugin, it adds the necessary requirements and everything else gets aware of them.

It's too early to go that way IMO. We first need to have support for standard things (JUnit) working decently before making this extensible.

laeubi:
> I don't get it; currently the JUnit Container has nothing related to OSGi in its definition.

it might seem so, but looking at the code of JDT you will see that it resolves the junit bundles against the Eclipse-Host OSGi environment. You can take a look at my [JunitClassPathContainer](https://github.com/eclipse/tycho/blob/400c244e564cf2f52744a032f284e271ca43ffe5/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/container/JunitClassPathContainer.java) that resembles what JDT does internally.

> That seems to be a good pragmatic 1st iteration to just add requirement on `org.junit`

what works so far for juni4 gets trickier for junit5. PDE looks at the manifest of the bundle and check if it imports any "juni5" and then adds a fixed list of other deps to the classpath. It also adds javax.annotations to ALL bundles in the workspace leading to hard to find resoloution error as it "works" in the IDE but not in tycho.

> t seems to me like the discussion above Tycho currently having no way to add requirements on packages so far

And that's why I tried to add support for this in https://github.com/eclipse/tycho/pull/60 ;-)

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
laeubi: I don't think so. If JUnit would work generic others would work also. That's the whole purpose of #60 to **not** add special handling on different places but allow **one** point of extension.

mickaelistria:
> And that's why I tried to add support for this in #60 ;-)

This is doing much more that just adding support for package requirements. As mentioned, support for package requirements can start with support for them in `extraRequirements`, then it would be easier to reuse them in other places. Let's try to resolve smaller problems and then connect them instead of trying to fix everything at the same time. That will lead to better quality, and better productivity.

laeubi:
> This is doing much more that just adding support for package requirements

This does much more because you requested me to rename a class take a look at https://github.com/eclipse/tycho/pull/60/commits/fb80088beb7babf168062b11ea385e2fef7d7022 for a more focused commit ...

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
mickaelistria: Sorry, but I really don't want to start by adding extensibility at the same time as we're adding other complexity on the lifecycle + compiler/surefire mojos. Just take smaller steps, with more focused changes each providing a valuable piece of the puzzle which has functional value on its own, and everything will work better for everyone.

mickaelistria:
> take a look at fb80088

OK, please isolate this commit in its own PR and with some tests showing how it can be leveraged in `extraRequirements` and we can consider it on its own more easily.

laeubi:
@mickaelistria please don't get me wrong but I somehow get the felling we probably talk past each other. So please forget for a moment about junit containers, surefire extensions and so on.

The change you propose here is also a topic that I have (not yet) addressed yet but that is not specific to my original intent but more relates to what I have described here https://github.com/eclipse/tycho/issues/66 and could be summarized to:

> Tycho should be able to distinguish between "compile" scope (normal source folders and classpath containers in eclipse, 'complie' scoped dependencies in maven) and "test" scope (source folders and classpath container marked as test in eclipse, 'test' scoped dependencies in maven)

My concern is mainly how could a mojo/plugin/.... (in a generic way) express that it needs a certain (p2) Requirement/ (OSGi)Required-Capability, that ultimately results in one or more bundles resolved from the target platform without the nedd to modify the whole Resolve>compile>test>package chain!

This could be best expressed by the current description of surefire-that states:

> <dependencies> - Additional dependencies to be added to the test runtime. I[...] This parameter has only limited support for dependencies to artifacts within the reactor. Therefore it is recommended to specify extraRequirements on the target-platform-configuration plugin instead

That illustrated the problem: extraRequirements is required so that P2 resolves the dependencies and then later tycho-surefire can use them.

This currently requires:

1. the p2 resolver has special support for 'extraRequirements' to resolve the units
2. the compiler-mojo has special support for  'extraRequirements' to get the resolved units on the classpath.
3. tycho-surefire has special support for  'extraRequirements' to fetch the bundles from the target
4. probably eclipse run mojos and other should have support for this too, each of them requires to understand 'extraRequirements'

The same applies to 'additional-bundles' support, it also requires these three (and probably more) to be adjust (one more code path), and the same will be true for each supported classpath container with the current architecture.

And that's why I have tried to introduce **one generic** way to express a requirement that is once implemented in all these stages and can the reused for whatever purpose (extraRequirements, surefire dependencies, classpath containers, additional bundles, BREEs, custom plugins...)

mickaelistria:
> So please forget for a moment about junit containers, surefire extensions and so on.

Sorry, I won't. I really think all development in Tycho requires pragmatic stories to do development properly.
You seem to be driven by a vision and other pure design things with, but that's not how development of Tycho has worked best and not how I feel comfortable reviewing or maintaining things. Please build user-stories with concrete examples to express what's needed, put them interests and then when we agree this case is to be supported, we pragmatically implement the simplest solution, and then we improve it later, maybe for the sake of extensibility.
The thing is that you started from a relatively simple story of adding test deps to classpath and surefire, and now you're extrapolating it to something bigger that requires more effort to develop and maintain. Please work in smaller steps and better communicate about the actual case you're willing to add support for; that will make collaboration much easier.

laeubi:
> I really think all development in Tycho requires pragmatic stories to do development properly.

And whenever I do this you asked me to cut it down into smaller pieces unless the user-story is no longer visible or applies here and the feature is doomed :-\

from my POV there is no simple and brilliant small code change here to make that happens, anyways I will just forget about 'visions' for now and simply try to implement something that 'just works' here once your change is merged.

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
laeubi: > OK, so let's just fill this TODO.

So what's next, do you mean I should try to add something at the "todo points" or should the PR be merged and we later fill the TODOs?

laeubi:
@mickaelistria as menioned above should I fill the TODOs so we can proceed or can we merge this and resolve that later?

The build complains for two tests about:
```
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.tycho.p2.maven.repository [55]
  Unresolved requirement: Import-Package: org.eclipse.tycho
```

mickaelistria:
I'll look at the tests today. Then I think we can merge with the TODO and implement smarter & more useful implementations in further patches and then refactor/extract what makes sense as the requirements scale up.

laeubi:
@mickaelistria I noticed this before it seems PRs are rebuild even if there are no changes... any idea how we can prevent this?

mickaelistria:
Whenever a change happens on master, PRs are rebuilt by merging with the new master.

PR #86, opened 2021-04-20
Fix #82 Support resolving of non-project IUs in P2Resolver

### LINKED ISSUE #82: Support resolving of non-project IUs in P2Resolver
Currently the P2Resolver API do read the root units from the project, in some circumstances it is desirable to perform a resolution of units based on a different set.

Use-case is for example I have a mojo and like to fetch a unit from a target-platform including its dependencies.

Currently one can archive this indirectly by creating a dummy project, and use the deprecated P2Resolver.collectProjectDependencies method.

laeubi:
Currently the P2Resolver API do read the root units from the project, in some circumstances it is desirable to perform a resolution of units based on a different set.

Use-case is for example I have a mojo and like to fetch a unit from a target-platform including its dependencies.

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/resolver/P2ResolverImpl.java
+++ b/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/resolver/P2ResolverImpl.java
@@ -11,6 +11,7 @@
  *    Sonatype Inc. - initial API and implementation
  *    Rapicorp, Inc. - add support for IU type (428310)
  *    Christoph Läubrich - Bug 572481 - Tycho does not understand "additional.bundles" directive in build.properties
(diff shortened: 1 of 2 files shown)


laeubi:
@mickaelistria do you think we can merge this new API? It is not intrusive and if we later decide to remove it easy to remove (or move to different place) as well.

mickaelistria:
I'd rather avoiding a merge of an API without a client code in Tycho that showcases how it can/should be used. And I'm not speaking about a test, but more about an actual feature.
This patch seems OK, but I'd rather merge it as part of a PR that shows the customer code as well.

laeubi:
Please see https://github.com/eclipse/tycho/pull/88

laeubi:
Is there any example/itest where <dependecy> in SureFireplugin does not work but <extraRequirements> ? Maybe this is a better showcase for this?

laeubi:
I have added https://github.com/eclipse/tycho/pull/108 as another example.

PR #88, opened 2021-04-21
Use P2 resolver to extend the test-probe-classpath for additional-bundles

laeubi:
This show how the new P" resolver API could be used to resolve 'additional-bundles' for Tycho Surefire in the test-probe.

### DIFF
--- a/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java
+++ b/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.net.MalformedURLException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
@@ -55,9 +56,11 @@ import org.apache.maven.surefire.util.DefaultScanResult;
 import org.apache.maven.surefire.util.ScanResult;
 import org.apache.maven.toolchain.Toolchain;
 import org.apache.maven.toolchain.ToolchainManager;
+import org.codehaus.plexus.logging.Logger;
 import org.codehaus.plexus.util.FileUtils;
 import org.codehaus.plexus.util.cli.CommandLineUtils;
 import org.eclipse.osgi.internal.framework.EquinoxConfiguration;
+import org.eclipse.sisu.equinox.EquinoxServiceFactory;
 import org.eclipse.sisu.equinox.launching.BundleStartLevel;
 import org.eclipse.sisu.equinox.launching.DefaultEquinoxInstallationDescription;
 import org.eclipse.sisu.equinox.launching.EquinoxInstallation;
@@ -72,6 +75,7 @@ import org.eclipse.tycho.BuildDirectory;
 import org.eclipse.tycho.DefaultArtifactKey;
 import org.eclipse.tycho.ReactorProject;
 import org.eclipse.tycho.artifacts.DependencyArtifacts;
+import org.eclipse.tycho.artifacts.TargetPlatform;
 import org.eclipse.tycho.core.BundleProject;
 import org.eclipse.tycho.core.DependencyResolver;
 import org.eclipse.tycho.core.DependencyResolverConfiguration;
@@ -87,11 +91,17 @@ import org.eclipse.tycho.core.osgitools.project.BuildOutputJar;
 import org.eclipse.tycho.core.resolver.DefaultDependencyResolverFactory;
 import org.eclipse.tycho.core.resolver.shared.OptionalResolutionAction;
 import org.eclipse.tycho.core.resolver.shared.PlatformPropertiesUtils;
+import org.eclipse.tycho.core.shared.BuildProperties;
 import org.eclipse.tycho.core.utils.TychoProjectUtils;
 import org.eclipse.tycho.dev.DevBundleInfo;
 import org.eclipse.tycho.dev.DevWorkspaceResolver;
 import org.eclipse.tycho.launching.LaunchConfiguration;
+import org.eclipse.tycho.osgi.adapters.MavenLoggerAdapter;
 import org.eclipse.tycho.p2.facade.RepositoryReferenceTool;
+import org.eclipse.tycho.p2.resolver.facade.P2ResolutionResult;
+import org.eclipse.tycho.p2.resolver.facade.P2ResolutionResult.Entry;
+import org.eclipse.tycho.p2.resolver.facade.P2Resolver;
+import org.eclipse.tycho.p2.resolver.facade.P2ResolverFactory;
 import org.eclipse.tycho.p2.tools.RepositoryReferences;
 import org.eclipse.tycho.surefire.provider.impl.ProviderHelper;
 import org.eclipse.tycho.surefire.provider.spi.TestFrameworkProvider;
@@ -445,6 +455,11 @@ public abstract class AbstractTestMojo extends AbstractMojo {
 
     @Component(role = TychoProject.class, hint = "eclipse-plugin")
     private OsgiBundleProject osgiBundle;
+    @Component
+    private EquinoxServiceFactory equinox;
+
+    @Component
+    private Logger logger;
 
     /**
      * Normally tycho will automatically determine the test framework provider based on the test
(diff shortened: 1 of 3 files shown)


tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java
```diff
@@ -1252,6 +1267,24 @@ private String getBuildOutputDirectories() {
             }
             appendCommaSeparated(sb, outputJar.getOutputDirectory().getAbsolutePath());
         }
+        BuildProperties buildProperties = osgiBundle.getEclipsePluginProject(getReactorProject()).getBuildProperties();
```
mickaelistria, line 1270: Should this instead be included in `getExtraDependencies` & `getIUsToInstall` ?

mickaelistria (commented):
See comment inline. I'm not certain we need a specific resolution step for this since we already to a resolution of the test installation, includng extra dependencies, and `additional.bundles` just seem like 1 form of extra dependencies.

tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java
laeubi: getExtraDependencies + getIUsToInstall are the bundle to be installed in the OSGi runtime, here we need them on the classpath of the test-probe itself.

mickaelistria: That's not what `additional.bundles` is supposed to deal with; and I'd rather avoid avoid the test probes be extensible that way because it adds a lot of complexity on all sides.
What happened to the idea of generating a test fragment for the test code and extra depedencies? This would seem to fit much better into tycho-surefire-plugin as we could just rely on more "regular" dependency resolution without having to inject into some early classpath references to artifacts that may require to be running inside the OSGi container to work well.

laeubi: It's just as it works when I run it from eclipse ... in general I won't recommend using additional.bundles but some seem to use this as its currently the only way PDE support and (sadly) there is no much interest to change this.

Once JUNIT Classpath containers work I would recommend to use that for tycho but as long as this is not widely used both should be supported.

The test-fragment approach was a nice idea but was to limited and leads to classpath inconsistencies (especially with JUnit5) so I have dropped the idea evne though I found it more "OSGish", and again PDE has no interest to support something similar so thats a dead-end for me.

laeubi: And of course everything ads complexity but if we don't want to put tycho into "maintenance mode" we need to enhance from time to time. And on my last proposal you voted for having more explicit handling than a "global generic" so here we go ...

Anyways I think this will on the long run integrate nice with your aproach of "getTestdepednecies" and probabaly would make it possible to resolve https://github.com/eclipse/tycho/issues/63 and https://github.com/eclipse/tycho/issues/33 as it enable sure-fire to resolve additional (implicit) dependencies of the test-probe also.

mickaelistria: So PDE doesn't adds additional.bundles to the OSGi runtime, but adds them to the "root" classloader? If so, please add comments around this piece of code pointing to class where PDE does that. This code here doesn't smell very good and we need to be reminded about why we're doing it that way and not better.

I think the the fragment approach was the best one, and would also have been easier to implement than trying to bend things into the single existing fragment.

> to limited and leads to classpath inconsistencies (especially with JUnit5) 

Is there some other comment where you detailed those issues? I think the test-fragment approach was better, from OSGi and Maven POV.

laeubi: > So PDE doesn't adds additional.bundles to the OSGi runtime, but adds them to the "root" classloader?

not only the additional.bundle, actually PDE adds a "requiredbundles" classpath container to JDT and puts all bundles imported on the project (!) classpath.

anyways, this does not put anything on the "root" classpath but extends the Bundle-Classpath as if the jar was embedded in the bundle, and that how it is supposed to work. If you really like the test-fragment aproach, tycho surefire needs to be rewritten completely as it currently is not working that way and I found putting support for it adds more complexity as value as neither PDE nor Tycho-Surefire currently works this way (e.g. pax-exam implements it that way but then you can't reuse maven-surefire runner but need to create a custom junit runner).

laeubi: If you are interested in the technical details you can take a look at 

org.eclipse.pde.internal.core.RequiredPluginsClasspathContainer.addSecondaryDependencies(BundleDescription, HashSet<BundleDescription>, ArrayList<IClasspathEntry>)

(Secondary Dependencies are the synonym for additional.bundles)

laeubi: org.eclipse.pde.internal.core.RequiredPluginsClasspathContainer.addExtraClasspathEntries(HashSet<BundleDescription>, ArrayList<IClasspathEntry>) is the part that references to jars.extra.classpath that is handled here just a few lines above in the TychoTestModjo here... so from just a technical POV this behaves here very similar to how PDE behaves.

mickaelistria: Do you mean with Run As > JUnit Plugin Test, or Run As > JUnit Test ?

laeubi: I mean with whats available on the bundle-classpath or not. If you do 'Run As > JUnit Plugin Test' PDE automatically adds all used entries from additional.bundles to the required-bundles (or import packages depending on choice in manifest editor). 'Run As > JUnit Test' works as expected without any side-effects.

mickaelistria: OK, so I still don't get why "additional.bundles" have to be part of the test-probe then.
Does PDE generate a different MANIFEST.MF file augmented with additional.bundles; and overall create a different bundle augmented also with test classes? It seems quite similar to the idea of generating a test-fragment to me, except that PDE is actually generated a bundle-with-test artifact for the sake of test execution. Tycho-Surefire should mimic that by creating such a specific test artifact instead of hooking into lower-level classpath/classloading stuff.

mickaelistria: So if additional.bundles does modify the original MANIFEST.MF, why just using it instead of just modifying the original MANIFEST ?
If that helps, all that is also obscure to me. I still don't get how these patches about additional.bundles move towards the current goal, and even lost track of the current goal. Is it still that the JUnit container brings extra requirements to compilation and surefire execution without leaking into regular compilation & packaging?

laeubi: >  why just using it instead of just modifying the original MANIFEST

Its PDE's way of 'manifest-first' but your point is valid that this *might* be an abuse and we should not use it. I just don't know how to demonstrate the API enhancement more simple.

For the goal 'additional.bundles' is just an intermediate step as it allows to demonstrate concepts that are needed for more advanced (like JUnit containers) and I found some complains additional.bundles is not supported by Tycho so I found it a good opportunity to add first support for the simple case (just a bundle name, no versioning ...).

> Is it still that the JUnit container brings extra requirements to compilation and surefire execution without leaking into regular compilation & packaging?

I think this describes it well, just that I'm not want to limit this to JUnit container specifically, e.g. I like to support [osgi-test](https://github.com/osgi/osgi-test) (what is very interesting for Tycho Surefire or 'Plugin Tests' and [Apache sling OSGi Mocks](https://sling.apache.org/documentation/development/osgi-mock.html) that  is usefull for 'plain' Junittest.
Both have more or less large dependency chains and its a pain to use those with current PDE/Tycho tools.
Instead i'd like to have a Classpath Container added to my project, maybe add an additional plugin to my pom.xml and are ready to set, either "Plugin Test" as well as "Unittest" ... but that's just a vision.

laeubi:
I'll close this for now as the use-case it not well suited (maybe even invalid) it still can serve as an example on how to use the extended P2Resolver in other places.

Issue #90, opened 2021-04-22
Compiler doesn't copy resource files when build.properties defines nested jars


sravanlakkimsetti:
When we tried building platform with tycho 2.4.0 we are getting the comparator errors indicating those files are not getting packaged. Log is attached [buildtimeComparatorUnanticipated.log.txt](https://github.com/eclipse/tycho/files/6355746/buildtimeComparatorUnanticipated.log.txt)

I have modified [platform aggregator job](https://ci.eclipse.org/platform/job/eclipse-aggregator-master/) to use tycho 2.4.0-SNAPSHOT. This can be used for testing. 

Please look out for similar errors as given below in the console log. These warnings should not appear. 
[INFO] --- tycho-p2-plugin:2.4.0-SNAPSHOT:p2-metadata-default (default-p2-metadata-default) @ org.eclipse.ant.core ---
[WARNING] MavenProject: org.eclipse.ant:org.eclipse.ant.core:3.6.0-SNAPSHOT @ /home/jenkins/agent/workspace/eclipse-aggregator-master/eclipse.platform/ant/org.eclipse.ant.core/pom.xml: baseline and build artifacts have same version but different contents
   no-classifier: different
      org/eclipse/ant/internal/core/InternalCoreAntMessages.properties: present in baseline only

[INFO] MavenProject: org.eclipse.ant:org.eclipse.ant.core:3.6.0-SNAPSHOT @ /home/jenkins/agent/workspace/eclipse-aggregator-master/eclipse.platform/ant/org.eclipse.ant.core/pom.xml
    The main artifact has been replaced with the baseline version.
    The following attached artifacts have been replaced with the baseline version: [sources]

mickaelistria:
This regression happens for files that have nested jars in their build.properties. I managed to write a test that reproduces it with 2.4.0-SNAPSHOT, but not with 2.3.0 (as expected); I'm going to merge it because this is a blocker issue, so it's fine to get all builds failing until it's resolved.
I'll then investigate to find out which commit does cause this issue.

sravanlakkimsetti:
> This regression happens for files that have nested jars in their build.properties. I managed to write a test that reproduces it with 2.4.0-SNAPSHOT, but not with 2.3.0 (as expected); I'm going to merge it because this is a blocker issue, so it's fine to get all builds failing until it's resolved.
> I'll then investigate to find out which commit does cause this issue.

Please let me know when a new snapshot has been published. I would like to try again.

[...]

PR #91, opened 2021-04-22
Issue #90 - Test resource inclusion

mickaelistria:


tycho-compiler-plugin/src/test/java/org/eclipse/tycho/osgicompiler/test/OsgiCompilerTest.java
```diff
@@ -210,6 +210,15 @@ private void assertSameFile(File expected, File actual) throws IOException {
         assertEquals(expected.getCanonicalFile(), actual.getCanonicalFile());
     }
 
+    public void testCopyResourcesNoConf() throws Exception {
+        File basedir = getBasedir("projects/resources/p004");
+        List<MavenProject> projects = getSortedProjects(basedir, null);
+        MavenProject project = projects.get(0);
+        getMojo(projects, project).execute();
+        assertTrue(new File(project.getBasedir(), "target/classes/testresources/Test.class").canRead());
+        assertTrue(new File(project.getBasedir(), "target/classes/testresources/test.properties").canRead());
```
laeubi, line 219: as this file is not mentioned in bin.includes should it really be copied?

mickaelistria: Yes, the expected behavior is that the resources that are part of source folders are copied into the same folder as .class files.

laeubi: You right I was confused by the path :-\

mickaelistria:
As this is an important regression, I'm merging this test. It's fine to get all builds failing until this is resolved, as fixing this is highest priority and worth blocking everything else.

PR #99, opened 2021-04-28
PGP Sign p2 artifacts

mickaelistria:


tycho-gpg-plugin/src/main/java/org/apache/maven/plugin/gpg/AbstractGPGMojoExtension.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.apache.maven.plugin.gpg;
+
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.project.MavenProject;
+
+public abstract class AbstractGPGMojoExtension extends AbstractGpgMojo {
```
laeubi, line 16: Should it be really be GPG?

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
```diff
@@ -0,0 +1,153 @@
...
+import org.codehaus.plexus.util.xml.Xpp3DomWriter;
+import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
+import org.eclipse.equinox.p2.repository.artifact.IArtifactDescriptor;
+
+/**
+ * Modifies the p2 metadata (<code>artifacts.xml</code>) to add PGP signatures for each included
+ * artifact. Signatures are added as <code>pgp.signatures</code> property, in armored form.
+ */
+@Mojo(name = "sign-p2-artifacts", requiresProject = true, defaultPhase = LifecyclePhase.PREPARE_PACKAGE)
+public class SignRepositoryArtifactsMojo extends AbstractGPGMojoExtension {
```
laeubi, line 47: Instead of SignRepositoryArtifactsMojo > AbstractGPGMojoExtension>AbstractGpgMojo maybe AbstractGPGMojoExtension could be left out here or is it likeley that there are more signers required?

If yes instead of AbstractGPGMojoExtension maybe "AbstractPublicKeyGPGMojo" would be a better name

mickaelistria:
[Example project showing how this can be used](https://github.com/eclipse/tycho/files/6419458/example.zip)
cc @akurtakov

laeubi:
@mickaelistria you should add a repository-verify step to make sure signing does not break signatures (just to be sure...)

tycho-gpg-plugin/src/main/java/org/apache/maven/plugin/gpg/ProxySignerWithPublicKeyAccess.java
```diff
@@ -0,0 +1,87 @@
...
+        delegate.generateSignatureForFile(file, signature);
+    }
+
+    public String getPublicKeys() throws MojoExecutionException {
+        Commandline cmd = new Commandline();
+
+//        if ( StringUtils.isNotEmpty( executable ) ) {
+//            cmd.setExecutable( executable );
+//        } else {
+        cmd.setExecutable("gpg" + (Os.isFamily(Os.FAMILY_WINDOWS) ? ".exe" : ""));
```
laeubi, line 41: This would require pgp installed right? Wouldn't it be better to use an of the java implementations of PGP (like jgit for git) instead if rely on the OS? https://www.openpgp.org/software/developer/

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
```diff
@@ -0,0 +1,153 @@
...
+            }
+            File file = new File(repository,
+                    "plugins/" + artifact.getAttribute("id") + '_' + artifact.getAttribute("version") + ".jar");
+            if (!file.canRead()) {
+                continue;
+            }
+            File signatureFile = signer.generateSignatureForArtifact(file);
+            String signature;
+            try {
+                signature = Files.readString(signatureFile.toPath());
```
laeubi, line 112: Is there any documentation about the exact format the signature must be? Could it be an ASCII encoded one like

-----BEGIN PGP SIGNATURE-----
Version: BCPG v1.61

iQIcBAABAgAGBQJc+WhMAAoJEIWRH0JexhtR6rIP/1mQpWKyZuYMnrcer+hmufGu
Ok/FTBR3gIcytJMIPqPyDwYCaqWJsYFvgQ0Ks6j8ch9BQWUaCyN9SE4rZfufRf7f
GWL35BiMjz0ksO+dSbLFhNGRx/e/Db1SmdaW2hgh0nVG8RakAImNha2a1SDJQh/O
7dWSvt6lnzzR85uWdnqicBh+dlpHbtVhzLIS+Zksn7yxHqdLnp/hheS1UMuIouzN
atDRBTaJ1NgZ8wciiDNrEShoEmpiIlAX1vx2a2ha/5/m1cccw6LvazeYIMKxqeM0
IP7G3HoIZ404/p47FCExhbzB1FrDi82G9f4O0jCExvKLj5XWEC+XgRCvihEqF6+F
WsohZLiySvUoWUgQgqMUhtSkJU+3dEgo/syL2NXKkUjSWL16PhB9DzCD/Q3TxE85
pbjbb4waJ3K0gF/c82FiEhgU2uhNW0ijMgIXwADQEBg+PBKr7PRh1GcV9DvpKZwz
QiQM73WYb9Mr593SYvDWYJwPNiuKd7xugvu20exa3V0WP2a2PeJJ3I3OcZr6EPjy
UI0LxabJYYZRMMJ9bNDbg6C0YSKTYn75335PE6YxEzgiWJatAQnVfFLao59ZDpyo
FO+nii+Lq/z1bHjCafyclY6dgSZREqvyLQMYMZR8e3bX9XJTwHsR+UW7nWsvq3AI
7Dv3t7xx8MlFs4imHIgp
=rjO2
-----END PGP SIGNATURE-----

laeubi:
Can someone redirect me to a place where it is described how exactly P2 expects to place what properties in what format to make PGP signed artifacts work?

tycho-gpg-plugin/src/main/java/org/apache/maven/plugin/gpg/ProxySignerWithPublicKeyAccess.java
mickaelistria: This is using the same startegy as the regular maven-gpg-plugin This brings consistency, although it would be nice to be able to leverage more of the maven-pgp-plugin as API.
One benefit of using the CLI is that it also reuse other gpg settings user might have tailored on their OS. If using boucycastle, it means we have to deal with building the keyring and so on. I have spent some time with bouncycastle on p2 earlier, and it's something I'd like to avoid as much as possible when we can: it's difficult to write and maintain, and it requires to be extra careful to not leak some keys and things like that. Security is a can of worm, so when it comes built-in as a binary we can expect users who need it have it, it's better to use it IMO.

tycho-gpg-plugin/src/main/java/org/apache/maven/plugin/gpg/AbstractGPGMojoExtension.java
mickaelistria: No, I'll make it Gpg.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
mickaelistria: I'll look at that.

mickaelistria:
> Can someone redirect me to a place where it is described how exactly P2 expects to place what properties in what format to make PGP signed artifacts work?

See https://www.eclipse.org/eclipse/news/4.20/platform_isv.php#pgp-signature-verification
And with https://bugs.eclipse.org/bugs/show_bug.cgi?id=572816 and related patches, in 4.21, the signer public key will be shown to user when installing and user will be asked whether they trust it or not in order to complete installation.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
mickaelistria: So I looked at it, and the issue is that package: in order to reuse some pieces of the AbstractGpgMojo which are package protected, we have to use the same package name. I prefer keeping the intermediary class to "open" some package-protected content and keeping the actual business code of Tycho in a mojo of the tycho package.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
mickaelistria: It's actually already documented in the class javadoc/the mojo documentation

laeubi:
I have already read the news but there is only mentioned:

> and those artifacts have the pgp.signatures property set

but how must the  pgp.signatures property formated? Is it a has? a full ASCII PGP signature? a base64 encoded binary one....?

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
laeubi: Maybe it would be good to open a patch at maven then to allow more open reuse there also?

tycho-gpg-plugin/src/main/java/org/apache/maven/plugin/gpg/ProxySignerWithPublicKeyAccess.java
laeubi: > although it would be nice to be able to leverage more of the maven-pgp-plugin as API

What would be needed here? can we request this from the maven-pgp-plugin directly?

> One benefit of using the CLI is that it also reuse other gpg settings user might have tailored on their OS

Don't know if gpg has similar default locations like maven (.m2) or git (.git/config) that could be adopted? Especially as we are only using the signing part, and the "secret key" is already to be configured I wonder what would could be inherited here?

Calling some system-executable really is something I would prevent as it also opens a "can of worms", just think about I write some malicious code named gpg.exe especially at windows ...

mickaelistria: > What would be needed here? can we request this from the maven-pgp-plugin directly?

Basically everything that we have in the org.apache.maven.plugin.pgp package here would need to become more public. It would indeed be a good thing to request it to the maven-gpg-plugin.

> Calling some system-executable really is something I would prevent as it also opens a "can of worms", just think about I write some malicious code named gpg.exe especially at windows ...

Except that if this happens, then it's an issue of the environment, not of Tycho. So Tycho would still be correct invoking gpg.exe and the issue would happen with every other consumer. That reduce the responsibility load on Tycho.
In any case, we should just mimic what's done in the official maven-pgp-plugin instead of trying to do things differently, The requirement on the binary is minimal: anyone caring about pgp has a gpg binary installed.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
mickaelistria: yes, it would be good. However, it's not of high enough importance to me at the moment to deal wih it right now.

mickaelistria:
> but how must the pgp.signatures property formated?

This is also already part of the N&N description "well formatted (it must be armoured PGP blocks)". `armoured` is a strongly loaded term about the actual format that most people playing with PGP are expected to be familiar with. It says it all.

laeubi:
> `armoured` is a strongly loaded term about the actual format that most people playing with PGP are expected to be familiar with.

Thanks for the explanation. I think a PGPAdvice would be useful but i haven't found one yet, do I just missed it?

grafik

Issue #93, opened 2021-04-23
Tycho doesn't honor 'limit-modules' from '.classpath'


dhendriks:
I have a Plug-in Project with the following `.classpath`:

```
<?xml version="1.0" encoding="UTF-8"?>
<classpath>
	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11">
		<attributes>
			<attribute name="module" value="true"/>
			<attribute name="limit-modules" value="java.management,java.xml"/>
		</attributes>
	</classpathentry>
	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
	<classpathentry kind="src" path="src"/>
	<classpathentry kind="output" path="target/classes"/>
</classpath>
```

Note the `limit-modules`, which reduces the JDK modules available, to prevent a JDK module and a plug-in dependency from both providing the same package. This makes sure that I don't have any compile errors in Eclipse when JDT compiles the Java code.

However, for the Maven/Tycho build, it failed again due to a package being provided by both a JDK module and a plug-in dependency.

I had to add the following to the `pom.xml` to resolve this:

```
    <build>
        <plugins>
            <plugin>
                <groupId>org.eclipse.tycho</groupId>
                <artifactId>tycho-compiler-plugin</artifactId>
                <version>${tycho.version}</version>
                <configuration>
                    <compilerArgs>
                        <arg>--add-modules</arg>
                        <arg>java.base,java.management,java.xml</arg>
                        <arg>--limit-modules</arg>
                        <arg>java.base,java.management,java.xml</arg>
                    </compilerArgs>
                </configuration>
            </plugin>
        </plugins>
    </build>
```

Essentially this has the same information as in `.classpath` (`java.base` is omitted there as it is a dependency of one of the other modules).

I'm no expert on how Tycho works under the hood. I think it would be good for Tycho to honor the `limit-modules` information from the `.classpath` file.

laeubi:
@dhendriks thanks for you request. Tycho currently does not honors any settings in claspath but I'm currently working to support at least some basic things.

To speed up with integration, you could:
- Provide a small (executable) reproducer that shows your problem in form of a github repo, even better provide an itest showing your problem see [here](https://github.com/eclipse/tycho/tree/master/tycho-its) for some examples
- Provide a patch that fixes the problem, its all open source and everyone is encourage to participate!

[...]

PR #125, opened 2021-05-17
#37 add parsing support for JRE classpath container

laeubi:


laeubi:
This is a perquisite of #93 and #126

PR #138, opened 2021-05-26
Fix typos

masofcon:
Signed-off-by: George Suaridze <suag@1c.ru>


laeubi:
@masofcon can you check that you have signed the ECA with the same email address you used in the author address of the commit?

masofcon:
> @masofcon can you check that you have signed the ECA with the same email address you used in the author address of the commit?

yes. it's OK

image

But I cannot recheck it...

PR #143, opened 2021-06-03
exclude maven-antrun-plugin:run from lifecycle-mapping

LorenzoBettini:
Otherwise there's an error on the POM and in any case there's no
connector for maven-antrun-plugin

Signed-off-by: Lorenzo Bettini <lorenzo.bettini@gmail.com>

### DIFF
--- a/tycho-core/pom.xml
+++ b/tycho-core/pom.xml
@@ -155,6 +155,41 @@
 				</executions>
 			</plugin>
 		</plugins>
+		<pluginManagement>
+			<plugins>
+				<!--This plugin's configuration is used to store Eclipse m2e settings only. It has no influence on the Maven build itself.-->
+				<plugin>
+					<groupId>org.eclipse.m2e</groupId>
+					<artifactId>lifecycle-mapping</artifactId>
+					<version>1.0.0</version>
+					<configuration>
+						<lifecycleMappingMetadata>
+							<pluginExecutions>
+								<pluginExecution>
+									<pluginExecutionFilter>
+										<groupId>
+											org.apache.maven.plugins
+										</groupId>
+										<artifactId>
+											maven-antrun-plugin
+										</artifactId>
+										<versionRange>
+											[3.0.0,)
+										</versionRange>
+										<goals>
+											<goal>run</goal>
+										</goals>
+									</pluginExecutionFilter>
+									<action>
+										<ignore></ignore>
+									</action>
+								</pluginExecution>
+							</pluginExecutions>
+						</lifecycleMappingMetadata>
+					</configuration>
+				</plugin>
+			</plugins>
+		</pluginManagement>
 	</build>
 
 	<dependencies>


LorenzoBettini:
@laeubi This is just to avoid having an error on the POM in the IDE

mickaelistria:
I'd rather have such workarounds prevented in Tycho. Instead it would be great to make the maven-antrun-plugin add the necessary metadata and use the right API for m2e to not complain about it.

LorenzoBettini:
@mickaelistria as I said, I never found any connector for the maven-antrun-plugin...

mickaelistria:
It's putting lipstick on a pig ;)

LorenzoBettini:
:D anyway, please feel free to close it if you think it's not worthwhile (but I guess the error marker should then be documented in the contribution guide)

laeubi:
I always found the m2e way to suppress things like this in the pom rather "disruptive".. @LorenzoBettini have you checked if we can ignore this in the **Workspace** instead using a setup file?

>  I never found any connector for the maven-antrun-plugin...

I'm not sure how m2e handles this in details but if I understand Mickael correctly you could place some data inside the antrun-plugin itself?!

LorenzoBettini:
> I always found the m2e way to suppress things like this in the pom rather "disruptive".. @LorenzoBettini have you checked if we can ignore this in the **Workspace** instead using a setup file?
> 

Probably this can be excluded at the workspace level, with some kind of preference that is stored in the setup file. For sure it can be ignored at the workspace level (because m2e provides such a quickfix), but I don't know if that can be stored somehow in the setup

> > I never found any connector for the maven-antrun-plugin...
> 
> I'm not sure how m2e handles this in details but if I understand Mickael correctly you could place some data inside the antrun-plugin itself?!

not that I'm aware of... or maybe I do not understand what Mickael suggests

mickaelistria:
See https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html . Maven plugins can be made compatible with incremental builders (such as m2e) or at least declare whether they are compatible or not so that the tools can take the best decision without bothering user with that.

LorenzoBettini:
@mickaelistria what I've done in the past is, instead of ignoring a plugin goal, bind a goal to the build with a mapping similar to what suggested in the link you mentioned:

```xml
<lifecycleMappingMetadata>
  <pluginExecutions>
    <pluginExecution>
      <pluginExecutionFilter>
        <goals>
          <goal>some-goal</goal>
        </goals>
      </pluginExecutionFilter>
      <action>
        <execute>
          <runOnIncremental>see below</runOnIncremental>
          <runOnConfiguration>see below</runOnConfiguration>
        </execute>
      </action>
    </pluginExecution>
  </pluginExecutions>
</lifecycleMappingMetadata>
```

IIRC I bound the goal to a clean build, not incremental because that was making everything slow.

Were you suggesting something like that? It's still a m2e lifecycle mapping.

If you were suggesting something more programmatic (https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html#buildcontext) then I wouldn't know what would be the best option.

laeubi:
>  I wouldn't know what would be the best option.

The best option would be to convince the maven-antrun plugin add m2e metadata. I just can't tell if it should really be skipped, or instead should be executed as part of the build. As maven can do *anything* it might be good to at least execute it in the `runOnConfiguration` ...

mickaelistria:
The best option is actually to contribute the desired change to the Maven plugin.

LorenzoBettini:
I leave it up to you guys... since this has never been executed in Eclipse I still think it's best to exclude it... if it has to be executed by Eclipse as well I don't know how to contribute.
If you're happy with the current state, I think that at least this error marker should be documented in the contribution guide

laeubi:
> since this has never been executed in Eclipse I still think it's best to exclude it.

I think one can start with this, this would then be required a PR/Change in the https://github.com/apache/maven-antrun-plugin @LorenzoBettini would you like to create a PR there?

LorenzoBettini:
> > since this has never been executed in Eclipse I still think it's best to exclude it.
> 
> I think one can start with this, this would then be required a PR/Change in the https://github.com/apache/maven-antrun-plugin @LorenzoBettini would you like to create a PR there?

Sorry, I don't understand what you mean.. I said it should be excluded as I proposed in this PR you've just closed. What should I do in the maven plugin?

laeubi:
> What should I do in the maven plugin?

The maven-plugin should include the [META-INF/m2e/lifecycle-mapping-metadata.xml ](https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html#mapping) then the warning/error won'T apear anymor (not only for tycho)

It could be something like 
```
<lifecycleMappingMetadata>
  <pluginExecutions>
    <pluginExecution>
      <pluginExecutionFilter>
        <goals>
          <goal>run</goal>
        </goals>
      </pluginExecutionFilter>
      <action>
        <ignore/>
      </action>
    </pluginExecution>
  </pluginExecutions>
</lifecycleMappingMetadata>
```

LorenzoBettini:
@laeubi if you think that's the way to go, please open the PR yourself.
I still think that the solution I proposed in this closed PR is the way to go...

laeubi:
I'll take a look I'm just not sure if it would be a good idea to actually execute the ant-run plugin as it obviously can do something relevant to the build...

LorenzoBettini:
> I'll take a look I'm just not sure if it would be a good idea to actually execute the ant-run plugin as it obviously can do something relevant to the build...

to "execute" or to "exclude"?

PR #144, opened 2021-06-03
Task 142 update eclipse installer git hub

### LINKED ISSUE #142: Update Eclipse Installer for Tycho to use github repository
I followed the instructions in https://github.com/eclipse/tycho/blob/master/CONTRIBUTING.md to setup a development environment. The setup fails because the repository https://git.eclipse.org/r/tycho/org.eclipse.tycho doesn't exist:

Cloning Git repo https://git.eclipse.org/r/tycho/org.eclipse.tycho to C:\Eclipse\tycho-master\git\org.eclipse.tycho
java.lang.Exception: org.eclipse.jgit.api.errors.TransportException: https://git.eclipse.org/r/tycho/org.eclipse.tycho: tycho/org.eclipse.tycho unavailable
...

LorenzoBettini:
Closes #142 

I also updated the documentation to specify "HTTPS readonly" if not committer.

I tested the setup locally.

@laeubi if you could please review it

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,7 +29,7 @@ Step by step instructions:
 4. If you are behind a proxy, at this point you might want to double check your network settings by clicking in the _Network Proxy Settings_ at the bottom.
 5. Select _Eclipse IDE for Eclipse Committers_ . Click _Next_ .
 6. Under _Eclipse.org_ , double-click on _Tycho_ (single click is not enough!). Make sure that _Tycho_ is shown in the table on the bottom. Click _Next_.
-7. You can edit the _Installation Folder_ , but you do not have to select the _Target Platform_ here, this will be set later automatically. By choosing _Show all variables_ at the bottom of the page, you are able to change other values as well but you do not have to. Click _Next_ .
+7. You can edit the _Installation Folder_ , but you do not have to select the _Target Platform_ here, this will be set later automatically. By choosing _Show all variables_ at the bottom of the page, you are able to change other values as well but you do not have to. (Unless you have write access to the GitHub repository, make sure you select "HTTPS read-only" in the dropdown "Tycho Github repository"). Click _Next_.
 8. Press _Finished_ on the _Confirmation_ page will start the installation process. 
 9. The installer will download the selected Eclipse version, starts Eclipse and will perform all the additional steps (cloning the git repos, etc...). When the downloaded Eclipse started, the progress bar in the status bar shows the progress of the overall setup.
 10. Once the _Executing startup task_ job is finished you should have all the Tycho and Tycho Extras projects imported into 2 working sets called _Tycho_ and _Tycho Extras_ .
--- a/setup/README.md
+++ b/setup/README.md
@@ -26,6 +26,7 @@ The Tycho.setup file is part of the Eclipse Installer Project Catalog, so the Ty
   * Find and select "Tycho" under the "Eclipse Projects" catalog. Make sure that Tycho is shown in the table on the bottom.
   * Click "Next" to get the Variables page
   * You can edit the values used by the setup process (e.g. the "Installation Folder"). If you want to choose the Install Folder and "Root install folder", you have to select "Show all variables" at the bottom of the page. Then you are also able to specify the location where Eclipse will be installed. You do not have to select the Target Platform here, this will be set later automatically.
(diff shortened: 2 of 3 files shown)


laeubi:
@LorenzoBettini I think the "read only" should be the default as we are using forks to develop code for tycho also for committer.

@merks is it possible for a setup to use a **fork** of a github repository? The usual workflow for this is the following:

1. goto the repository
2. fork it into private account
3. checkout the fork
4. add the original one as 'upstream' in the config

While 1+2 might not be possible to archive, maybe one could have a (required) field `GitHub username/organisation` then use this variable in the git:GitCloneTask and also adding the original one as 'upstream'?

merks:
I'm not totally sure.  The remoteURI property is used and is generally specified as a relative path that is expanded based on choices that are automatically synthesized.  The user is presented these choices, but I think they can paste in a link to a clone which I think would then check out that fork.  It wouldn't configure the upstream thing automatically except maybe the fork itself already specifies that somehow?  But the Git Clone task also support Config Section and stuff nested under there to configure this type of thing...

If something else is needed GitCloneTaskImpl could be extended with new optional attribute to do additional things...

laeubi:
@LorenzoBettini can you check if this is possible with the hints of @merks ?

LorenzoBettini:
@laeubi concerning the fork one could simply paste the URI of his/her fork in the dialog, as @merks said.
Concerning the "read only" by default I don't know how to do that... @merks some hints please?

laeubi:
Yeah I just think we should make this an explicit, non default, required field something like 'forkName', and the checkout task should add the 'upstream' https://github.com/eclipse/tycho.git as an additional origin.

LorenzoBettini:
@laeubi currently I'm unable to do that but I also find it overkill... I'd merge the PR since at least it fixes the current state. Note that the behavior is the same as the old one when the Git repo was hosted on Gerrit. The fork thing is a general GitHub workflow so that's something that possible contributors are already aware of.

merks:
> 
> 
> @laeubi concerning the fork one could simply paste the URI of his/her fork in the dialog, as @merks said.
> Concerning the "read only" by default I don't know how to do that... @merks some hints please?

By this you mean the order in which the choices are listed (where the first choice is then the default).  Those choices are induced from this variable definition in the Project Catalog;
```
<setup:VariableTask
    xmi:version="2.0"
    xmlns:xmi="http://www.omg.org/XMI"
    xmlns:setup="http://www.eclipse.org/oomph/setup/1.0"
    name="github.remoteURIs"
    label="Github repository">
  <choice value="git@github.com:@{remoteURI}.git"
      label="Git (read-write)"/>
  <choice value="https://${github.user.id|username}@github.com/@{remoteURI}.git"
      label="HTTPS (read-write)">
    <annotation
        source="http://www.eclipse.org/oomph/setup/MatchChoice"/>
  </choice>
  <choice value="https://github.com/@{remoteURI}.git"
      label="HTTPS (read-only, anonymous)"/>
  <description>Choose from the available Github URIs</description>
</setup:VariableTask>
```
You could copy and paste that into your setup to change the order...

(Note that you can literally copy the selected XML above  and paste it into your Project in the Setup Editor.)

laeubi:
> Note that the behavior is the same as the old one when the Git repo was hosted on Gerrit. 

That's not really true. Gerrit does not use the concept of Forks/PRs but you have to checkout the code 'readonly' as a non-contributor and then can only submit 'changesets' by everyone with an gerrit-account. These changes never show up in the original git repo if not merged by a contributor. To make this workflow work, there was a special 'gerrit-configuration' in the checked out repository.

That's why I said that if the original URL is used, then it should be read-only to make this more 'gerrit-like'. 

As an alternative, we need to checkout the 'writeable' fork of the user and add a special 'github-workflow' config, otherwhise users using the setup need to pay special attention to reconfigure the push/pull URLs and origins to match their fork.

laeubi:
@LorenzoBettini any plans to further improve this or should we merge this 'as-is'

LorenzoBettini:
@laeubi I'd suggest merging this as it is (as you see, I've also updated the documentation); I currently don't have time to improve it further at moment, but, as I said, I think also in the current shape it's perfectly fine and it's similar to other Oomph setups based on GitHub...

laeubi:
Okay I'll trigger a build then and merge it afterwards (just to get all checks green).

PR #145, opened 2021-06-03
141: Bug 453708 - Support for site/repository-reference/@location in eclipse-repository

### LINKED ISSUE #141: Bug 453708 - Support for site/repository-reference/@location in eclipse-repository
https://bugs.eclipse.org/bugs/show_bug.cgi?id=453708

This is still open and it would be great to have this feature

LorenzoBettini:
Closes #141 

See also https://bugs.eclipse.org/bugs/show_bug.cgi?id=453708#c19

This PR also adds the type 1 (artifact) reference to the final content.xml (z and jar).

I've also tested locally that with this patch I can finally install features from a p2 repository relying on dependencies stored in referred p2 repositories

Task-Url: https://github.com/eclipse/tycho/issues/141
Signed-off-by: Lorenzo Bettini <lorenzo.bettini@gmail.com>

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.p2.tools.impl/src/main/java/org/eclipse/tycho/p2/tools/mirroring/MirrorApplication.java
+++ b/tycho-bundles/org.eclipse.tycho.p2.tools.impl/src/main/java/org/eclipse/tycho/p2/tools/mirroring/MirrorApplication.java
@@ -16,6 +16,7 @@ import java.net.URI;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.equinox.p2.core.IProvisioningAgent;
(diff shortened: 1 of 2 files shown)


LorenzoBettini:
@mickaelistria I was also planning to add a more involved IT verifying that a feature with a dependency (taken from a referred repository, e.g., JUnit from Orbit) can be effectively installed (using tycho-eclipserun and the p2 director application, something that I usually do on my projects), what do you think? Should I add this into another PR in case this is going to be merged?

LorenzoBettini:
The build failed, but it's not my fault: it timed out due to recent JIRO problems. Could anyone restart it please?

laeubi:
@LorenzoBettini the build is back to normal, if you could add an IT it would be great.

laeubi (approved):

mickaelistria:
IMO, it's good enough to merge. Some more tests would be welcome, but I don't think we should delay delivery of this feature in snapshots just for that. They can come in further commits.

LorenzoBettini:
@mickaelistria thanks for merging! I think this feature is really crucial :)

@laeubi the IT was already there and I updated. I was suggesting a more involved IT (maybe an end-to-end test). I'll propose something in another PR along the line I was suggesting

mickaelistria:
Please also add a note about it to https://github.com/eclipse/tycho/blob/master/RELEASE_NOTES.md

LorenzoBettini:
> Please also add a note about it to https://github.com/eclipse/tycho/blob/master/RELEASE_NOTES.md

here it is https://github.com/eclipse/tycho/pull/147

PR #150, opened 2021-06-12
Fix #149 Support using reactor projects as source for assemble-p2-maven-site

### LINKED ISSUE #149: Support using reactor projects as source for assemble-p2-maven-site
currently it is possible to use

- dependencies
- managedDependencies

it should also be possible to use the  reactor projects itself as a source for the `assemble-p2-maven-site`

laeubi:


laeubi:
An example is provided here: https://github.com/eclipse/jetty.project/pull/6404

Issue #159, opened 2021-06-23
DependencyComputer is computing several times reexported bundles dependencies


pdulth:
For Eclipse Capella project, the migration from Tycho 2.0 to 2.3 and 2.4 lead to a longer dependency computation . From ~30min of compile to 1h30 with 2.3.0 and ~45min with 2.4.0

From 2.3 to 2.4, there had been a fix around DependencyComputer leading to this improvement.
https://github.com/eclipse/tycho/commit/d8d4a0c694b0a6785e7c779c91ff8ae0bbedabd8  (Cache ModuleWiring wires to speed-up dependency computation )

Likewise, in DependencyComputer.populateFromWiring and its sub methods, for a moduleWiring, we compute several times getRequiredBundlePackages of the same "providerBundleWires"

pdulth:
I'm no expert of Tycho, WDYT of the patch?

pdulth:
On Capella
without the patch, dependency computation takes around ~11min
with the patch, ~3min

mickaelistria:
See recent change https://github.com/eclipse/tycho/pull/161 (and more importantly, the explanation). It could be that just by updating to newer Equinox/p2 fixes the issue?
About #160, please add more explanation in the commit message, as it's not trivial code, more explanation would make review and further analysis easier.

pdulth:
I will try with the latest snapshot, but i had also upgraded Capella to use Eclipse 2021-06 at same time, the osgi release seems to be the expected one in my previous tests.

pdulth:
Hi, 
On Eclipse Capella based on Eclipse 2021-06, (with mentioned org.eclipse.osgi-3.16.300)
Tycho 2.1.0 => 2min resolve dependencies
Tycho 2.3.0 => 30min resolve dependencies
Tycho 2.4.0-snapshot a8d7bb7 => 16min resolve dependencies
Tycho 2.4.0 with patch #160 => 3min resolve dependencies

pdulth:
The issue: 

[...]

PR #160, opened 2021-06-23
#159 Speed-up dependency computation of 'visibility:reexport' bundles

pdulth:


mickaelistria:
I don't understand in which particular case this change is supposed to improve the behavior and how it achieves it. Can you please provide this patch with some more info to understand it. This can be a commit message, more comments, an example, a test... all of them ;)

pdulth:
> I don't understand in which particular case this change is supposed to improve the behavior and how it achieves it. Can you please provide this patch with some more info to understand it. This can be a commit message, more comments, an example, a test... all of them ;)

I have updated the commit message and added example in the issue.
Thanks

PR #167, opened 2021-07-01
[DEMO] JustJ

mickaelistria:


laeubi:
@mickaelistria should we merge this? Or maybe move this to the tycho-demo part?

mickaelistria:
It shouldn't be a test because it's toolong to execute, but moving it to demo is a good idea.

laeubi:
@mickaelistria can you update the PR so it is located at the tycho demo then?

mickaelistria:
> can you update the PR so it is located at the tycho demo then?

I'm taken with many other topics to work on and which seem of higher value, so I don't think this is going to be high enough priority for me to work on it anytime soon.

akurtakov:
Please either finish this one or close. Keeping things open for months doesn't help in other ways but to use more infra so the PR is still verified after each commit.

mickaelistria:
https://github.com/eclipse/tycho/commit/498449a244addc614e95904caa59aea1596a936c is merged and shows how to include JustJ in a product and build it with Tycho.

Issue #166, opened 2021-07-01
Skip Tycho dependency resolution for clean-only builds by default


HannesWell:
When you want to clean Eclipse projects build with Maven/Tycho the dependency resolution is performed initially by default, even tough it is not necessary (please correct me if I'm wrong). For large projects the dependency resolution takes the vast majority of runtime in a `clean`-only build.

In order to speed-up `clean` builds, one can specify the property `-Dtycho.mode=maven` to skip the dependency resolution.

For convenience (on the command-line and in the IDE) and to speed up cleans for users that do not know about the mentioned property I suggest to skip the dependency resolution by default for clean only builds.
This could be accomplished in `TychoMavenLifecycleParticipant.disableLifecycleParticipation(Seesion)` by checking if the session only has the goal `"clean"`.

If nothing else speaks against it, I can provide a patch to change that default behavior for clean-only builds.

mickaelistria:
Tycho somehow modifies the Maven model by adding dependencies, but AFAIK,it does affect the typical "dirty" directories that clean is supposed to process. So ok, let's try it!

HannesWell:
> Tycho somehow modifies the Maven model by adding dependencies, but AFAIK,it does affect the typical "dirty" directories that clean is supposed to process. So ok, let's try it!

You meant `does not affect`, don't you?

Because yes, the modifications of the Maven-model performed by Tycho in `TychoMavenLifecycleParticipant.afterProjectsRead` do affect the model, but I think they are not necessary for running a clean build.
At least I got that impression when looking at `TychoMavenLifecycleParticipant.afterProjectsRead(MavenSession)` and what happens after the initial if statement that calls disableLifecycleParticipation(MavenSession).
The additional validation and configuration performed here seems unnecessary for cleaning and skipping the dependency resolution is the intention of this change.

I'm about to open a PR to address this.

PR #172, opened 2021-07-03
#166: consider all clean-lifecylce phases and add release-notes entry

HannesWell:
This PR adds an release-notes entry for issue #166 and extends the first solution for that issue by additionally considering the other two phases of the clean life-cycle.

What makes me wonder is that maven usually talks about phases when referring to the different 'steps' of a lifecycle.
But the session (and its request) only provides a List of goals. Are specified phases considered as goals in Maven?
The content of the goals list seems to be as expected when I attached a debugger and I did not find the word 'phase' in the `MavenSession` or `MavenExecutionRequest` class.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoMavenLifecycleParticipant.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoMavenLifecycleParticipant.java
@@ -234,11 +234,13 @@ public class TychoMavenLifecycleParticipant extends AbstractMavenLifecyclePartic
         }
     }
 
+    private static final Set<String> CLEAN_PHASES = Set.of("pre-clean", "clean", "post-clean");
+
     private boolean disableLifecycleParticipation(MavenSession session) {
         // command line property to disable Tycho lifecycle participant
         return "maven".equals(session.getUserProperties().get("tycho.mode"))
                 || session.getUserProperties().containsKey("m2e.version")
-                || session.getGoals().equals(List.of("clean")); // disable for clean-only session
+                || (session.getGoals().size() == 1 && CLEAN_PHASES.contains(session.getGoals().get(0))); // disable for clean-only session
     }
 
     private void configureComponents(MavenSession session) {
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoMavenLifecycleParticipant.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoMavenLifecycleParticipant.java
@@ -240,7 +240,7 @@ public class TychoMavenLifecycleParticipant extends AbstractMavenLifecyclePartic
         // command line property to disable Tycho lifecycle participant
         return "maven".equals(session.getUserProperties().get("tycho.mode"))
                 || session.getUserProperties().containsKey("m2e.version")
-                || (session.getGoals().size() == 1 && CLEAN_PHASES.contains(session.getGoals().get(0))); // disable for clean-only session
+                || CLEAN_PHASES.containsAll(session.getGoals());
     }
 
     private void configureComponents(MavenSession session) {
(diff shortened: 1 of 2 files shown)


laeubi:
Its confusing but maybe this helps a bit?
https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#a-build-phase-is-made-up-of-plugin-goals

> A plugin goal represents a specific task (finer than a build phase) which contributes to the building and managing of a project. It may be bound to zero or more build phases. A goal not bound to any build phase could be executed outside of the build lifecycle by direct invocation.

tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoMavenLifecycleParticipant.java
```diff
@@ -234,11 +234,13 @@ private void validateUniqueBaseDirs(List<MavenProject> projects) throws MavenExe
...
     }
 
+    private static final Set<String> CLEAN_PHASES = Set.of("pre-clean", "clean", "post-clean");
+
     private boolean disableLifecycleParticipation(MavenSession session) {
         // command line property to disable Tycho lifecycle participant
         return "maven".equals(session.getUserProperties().get("tycho.mode"))
                 || session.getUserProperties().containsKey("m2e.version")
-                || session.getGoals().equals(List.of("clean")); // disable for clean-only session
+                || (session.getGoals().size() == 1 && CLEAN_PHASES.contains(session.getGoals().get(0))); // disable for clean-only session
```
mickaelistria, line 243:
```suggestion
                || CLEAN_PHASES.containsAll(session.getGoals());
```

HannesWell: That makes it more general and simpler. Thanks.

HannesWell:
If this is fine, you can squash in merge the commits. I think there is no value in keeping the changes as single commits.

HannesWell:
> 
> 
> If this is fine, you can squash in merge the commits. I think there is no value in keeping the changes as single commits.
> 
> Its confusing but maybe this helps a bit?
> https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#a-build-phase-is-made-up-of-plugin-goals
> 
> > A plugin goal represents a specific task (finer than a build phase) which contributes to the building and managing of a project. It may be bound to zero or more build phases. A goal not bound to any build phase could be executed outside of the build lifecycle by direct invocation.

Thanks for the link, but this only confirms my understanding of goals and phases.
For example `clean`, `verify`, `package` etc. are phases (of the two life-cycles `clean` and `default`), while e.g. `maven-compiler-plugin:compile` is a goal (by default bound to the lifecycle phase `compile`).
But when I call `mvn clean verify` the code `MavenSession.getGoals()` returns a List composed of `"clean"` and `"verify"`.
From this I expect for a command like `mvn clean maven-compiler-plugin:compile verify` 
that `getGoals()` returns `["clean", "maven-compiler-plugin:compile", "verify"]`.

But I actually expected something like `MavenSession.getPhases()`, that would return `["clean", "verify"]` and getGoals() that returns `["maven-compiler-plugin:compile"]`.

From the example in your link I think the solution of this theoretical problem is likely that the order of goals and phases specified in a Maven command needs to be maintained, even if they are mixed. And this is only possible if you store them together in one list in the model. Therefore I assume that Maven treats phases specified in a command as goals too?

But this theoretical problem is actually not relevant for this PR and the code should work anyways.

laeubi:
clean is a "life-cycle" see https://maven.apache.org/ref/3.8.1/maven-core/lifecycles.html

In the context of tycho, [we define for a packaging-type a custom  mapping of Life-cycle Phases](https://github.com/eclipse/tycho/blob/master/tycho-maven-plugin/src/main/resources/META-INF/plexus/components.xml) depending on the packaging type.

Issue #169, opened 2021-07-03
Uses constraint violation while upgrading from tycho 2.2 -> 2.3


laeubi:
I try to upgrade a build from tycho 2.2 -> 2.3 (the same happens for the current SNAPSHOT) but the build fails now with:

> org.apache.felix.resolver.reason.ReasonException: Uses constraint violation. Unable to resolve resource [...a bundle from the build...] because it is exposed to package 'javax.xml.bind.annotation.adapters' from resources org.eclipse.osgi [osgi.identity; osgi.identity="org.eclipse.osgi"; type="osgi.bundle"; version:Version="3.16.300.v20210525-1715"; singleton:="true"] and jakarta.xml.bind-api [osgi.identity; osgi.identity="jakarta.xml.bind-api"; type="osgi.bundle"; version:Version="2.3.2"] via two dependency chains.

While this is something that might be resolved by tweaking the imports and use constraints, tycho seem to resolve / install conflicting bundles here, as in the runtime / product there is no issue with this as there is always only one provider for that package. The bundle itself is from an update-site so not part of the build and works with 2.3...

@mickaelistria any idea what can be the culprit here between 2.2 -> 2.3 ?

laeubi:
This currently prevents one using different versions of the same bundle in the same reactor build.

laeubi:
This currently makes it impossible to use `<resolveWithExecutionEnvironmentConstraints>false</resolveWithExecutionEnvironmentConstraints>` for me also. I'll try to create a minimal reproducer for this, this really is a major regression/blocker for me.

mickaelistria:
This is definitely a consequence of usage of the "good" resolver since 2.3: https://wiki.eclipse.org/Tycho/Release_Notes/2.3#Official_Equinox_Resolver_used_for_dependency_resolution_.28stricter_and_can_produce_errors_for_split_packages.29
and this, combined with https://bugs.eclipse.org/bugs/show_bug.cgi?id=525368 can lead to such issues; or it can also highlight that this bundle is incorrect from the very beginning and that there is no satisfactory solution for the constraints with the current TP and this was not detected earlier...

laeubi:
Nope this is defiantly **not** a good solution and a major degradation from 2.2>2.3.
Having multiple versions of the same bundle in one framework is one of the keypoints of OSGi and forcing all projects to always using the latest version is a no-go. Also Split package are not an issue here...

mickaelistria:
[...]

PR #173, opened 2021-07-03
itest demonstrate regression from 2.2 -> 2.3 as of #169

laeubi:
This demonstrate the issue #169 with a trivial use-case, just a single bundle with a plain requirement to an older version of a library and a newer one in the target.

Build with `mvn clean verify -Dtycho-version=2.2.0` succeeds while build with `mvn clean verify -Dtycho-version=2.3.0` fails with

```
[ERROR] Internal error: java.lang.RuntimeException: org.osgi.framework.BundleException: Bundle version.bundle1 cannot be resolved:version.bundle1 [21]
[ERROR]   Bundle was not resolved because of a uses constraint violation.
[ERROR]   org.apache.felix.resolver.reason.ReasonException: Uses constraint violation. Unable to resolve resource version.bundle1 [osgi.identity; osgi.identity="version.bundle1"; type="osgi.bundle"; version:Version="1.0.0"] because it is exposed to package 'org.apache.batik.ext.awt.g2d' from resources org.apache.batik.ext.awt [osgi.identity; osgi.identity="org.apache.batik.ext.awt"; type="osgi.bundle"; version:Version="1.9.1.v20190730-1743"] and org.apache.batik.awt.util [osgi.identity; osgi.identity="org.apache.batik.awt.util"; type="osgi.bundle"; version:Version="1.14.0.v20210324-0332"] via two dependency chains.
[ERROR] 
[ERROR] Chain 1:
[ERROR]   version.bundle1 [osgi.identity; osgi.identity="version.bundle1"; type="osgi.bundle"; version:Version="1.0.0"]
[ERROR]     require: (&(osgi.wiring.bundle=org.apache.batik.ext.awt)(bundle-version>=1.7.0))
[ERROR]      |
[ERROR]     provide: osgi.wiring.bundle: org.apache.batik.ext.awt
[ERROR]   org.apache.batik.ext.awt [osgi.identity; osgi.identity="org.apache.batik.ext.awt"; type="osgi.bundle"; version:Version="1.9.1.v20190730-1743"]
[ERROR] 
[ERROR] Chain 2:
[ERROR]   version.bundle1 [osgi.identity; osgi.identity="version.bundle1"; type="osgi.bundle"; version:Version="1.0.0"]
[ERROR]     require: (&(osgi.wiring.bundle=org.apache.batik.svggen)(bundle-version>=1.7.0))
[ERROR]      |
[ERROR]     provide: osgi.wiring.bundle; bundle-version:Version="1.14.0.v20210324-0332"; osgi.wiring.bundle="org.apache.batik.svggen"
[ERROR]   org.apache.batik.svggen [osgi.identity; osgi.identity="org.apache.batik.svggen"; type="osgi.bundle"; version:Version="1.14.0.v20210324-0332"]
[ERROR]     import: (&(osgi.wiring.package=org.apache.batik.ext.awt.g2d)(&(version>=1.14.0)(!(version>=2.0.0))))
[ERROR]      |
[ERROR]     export: osgi.wiring.package: org.apache.batik.ext.awt.g2d
[ERROR]   org.apache.batik.awt.util [osgi.identity; osgi.identity="org.apache.batik.awt.util"; type="osgi.bundle"; version:Version="1.14.0.v20210324-0332"]
[ERROR] -> [Help 1]
```

For me this is a blocker for the upcomming 2.4.x release ...

### DIFF
--- a/tycho-its/projects/resolver.differentVersions/bundle1/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/resolver.differentVersions/bundle1/META-INF/MANIFEST.MF
@@ -1,0 +1,11 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Bundle1
+Bundle-SymbolicName: version.bundle1
+Bundle-Version: 1.0.0
+Require-Bundle: org.apache.batik.svggen;bundle-version="1.7.0",
+ org.apache.batik.dom;bundle-version="1.7.0",
+ org.apache.batik.ext.awt;bundle-version="1.7.0"
+Bundle-ActivationPolicy: lazy
+Import-Package: org.osgi.framework;version="1.10.0"
+
--- a/tycho-its/projects/resolver.differentVersions/bundle1/build.properties
+++ b/tycho-its/projects/resolver.differentVersions/bundle1/build.properties
@@ -1,0 +1,4 @@
+jars.compile.order = .
+source.. = src/
+output.. = target/classes
+bin.includes = META-INF/, .
--- a/tycho-its/projects/resolver.differentVersions/bundle1/pom.xml
+++ b/tycho-its/projects/resolver.differentVersions/bundle1/pom.xml
@@ -1,0 +1,17 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>org.eclipse.tycho.its.resolver.versions</groupId>
+		<artifactId>versions-parent</artifactId>
+		<version>1.0.0</version>
+	</parent>
+
+
+	<artifactId>version.bundle1</artifactId>
+
+	<packaging>eclipse-plugin</packaging>
+
+</project>
--- a/tycho-its/projects/resolver.differentVersions/bundle1/src/test/Activator.java
+++ b/tycho-its/projects/resolver.differentVersions/bundle1/src/test/Activator.java
@@ -1,0 +1,24 @@
+package test;
+
+import org.osgi.framework.BundleActivator;
+import org.osgi.framework.BundleContext;
+
+public class Activator
+    implements BundleActivator
+{
+
+    public static BundleContext context;
+
+    public void start( BundleContext context )
+        throws Exception
+    {
+        Activator.context = context;
+    }
+
+    public void stop( BundleContext context )
+        throws Exception
+    {
+        Activator.context = null;
+    }
+
+}
--- a/tycho-its/projects/resolver.differentVersions/feature.new.batik/build.properties
+++ b/tycho-its/projects/resolver.differentVersions/feature.new.batik/build.properties
@@ -1,0 +1,1 @@
+bin.includes = feature.xml
--- a/tycho-its/projects/resolver.differentVersions/feature.new.batik/feature.xml
+++ b/tycho-its/projects/resolver.differentVersions/feature.new.batik/feature.xml
@@ -1,0 +1,131 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<feature
+      id="feature.new.batik"
+      label="Batik New"
+      version="1.0.0">
+
+   <description url="http://www.example.com/description">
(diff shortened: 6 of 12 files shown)


laeubi:
To make it clear 
```
org.apache.batik.svggen [
   osgi.identity; osgi.identity="org.apache.batik.svggen"; 
   type="osgi.bundle"; 
   version:Version="1.14.0.v20210324-0332"
]
```
should not be part of the resolution for this bundle because it is not a valid choice here (and that's what the OSGi resolver complains about here). Maybe this is even a bug of P2...

mickaelistria:
> Maybe this is even a bug of P2...

Probably. Have you tried generating with PDE a p2 repo for this bundle, and then tried to install it in some empty-ish product with latest Orbit repo available to see what ends up being installed? If a mix of 1.14.x and 1.9.x is installed by p2, then we can't fix it in Tycho.

laeubi:
Nope as I said this works fine in Tycho 2.2.0 the product using this gets included both versions of batik and everything works like expected. And as mentioned this is just a simple reproducer... so I expect that at least this simple case works, as there is no reference even to never versions anywhere... so there is really something going wrong.

mickaelistria:
The resolver in Tycho 2.2 was outdated and deprecated by Equinox, it was not supporting many OSGi constructs. So those bugs in the previous resolver may have compensated/hidden former bugs in p2. We need to 1st verify how p2 itself behaves for that case, independently of Tycho.

laeubi:
The resolver is only used to construct the access rules, while P2 is used to build the actual product so P2 is at least capable to build a runnable product. That's why I suspect that someting is wrong with the usage of the new resolver.

If the new way o resolving is breaking working builds it might be better to restore the old way for a while and have at least an option to use the old resolver. I just don't have an idea how to best debug this and can only supply this very minimal example.

mickaelistria:
The old resolver had a bunch of issues that caised migration to "official"
one, some of those issues were blockers for proper adoption by Platform.
IIRC some of them caused by adoption of Java 11and so on. See history of
the code for a reference to the bug and its relations.
So no, we basically cannot (and should't) revert to old resolver. We just
need to fix this one, but here I'm really under the impression that the
DependencyArtifacts, provided throigh p2 planner, are wrong in 1st place.

-- 
Mickael Istria
Eclipse IDE <https://www.eclipse.org/downloads/eclipse-packages/>
developer, for Red Hat Developers <https://developers.redhat.com/>

laeubi:
The new resolver causes heavy performance degradation in some cases see https://github.com/eclipse/tycho/issues/213
I really wonder if we actually need this level of verification (especially as it is only used for the compile phase) and should not completely rely on P2 instead? The only thing required would the that we derive the access-rules, but that should be possible (maybe using the equinox code in a more direct way).

mickaelistria:
> The new resolver causes heavy performance degradation in some cases see #213

That's something to investigate and fix on its own.

> I really wonder if we actually need this level of verification (especially as it is only used for the compile phase) and should not completely rely on P2 instead? The only thing required would the that we derive the access-rules, but that should be possible (maybe using the equinox code in a more direct way).

That's an interesting thing to discuss, with a value gain but with some risks as well. Can you please bring the topic to the mailing-list or into a dedicated issue for discussion to ensure we capture all the value points and risks before starting some work here?

laeubi:
> That's an interesting thing to discuss, with a value gain but with some risks as well.

Maybe we can enable [github-discussions](https://docs.github.com/en/discussions) for such purpose?

mickaelistria:
Using github-discussions is also something we need to discuss first on the mailing-list yo ensure whether and how it should be done in the best interest of the project.

Back to the initial issue, I think 1 reason why Equinox resolver is ran early is to deal with some cases of dependency resolution that aren't well captured by p2; those can be about split packages, services or other forms of capabilities.
So removing early EquinoxResolver would lead to p2-only potentially uncomplete resolution in some cases, that could then cause an erroneous build order or silent some important resolution errors, or fail later (eg during tests or worse, at install-time instead of build-time).

laeubi:
> or fail later [...] at install-time

One can't guess install-time anyways. And as mentioned before, if P2 does not support a given construct we should add support for it there (as otherwise installs might be incomplete...).

If someone really wan't that additional verification steps, it might be better to have this run with a dedicated mojo at the verify phase.

mickaelistria:
Currently, one mais advantage if Tycho compared to previous build systems is the early verification and failure. This shouldn-'t be an opt-in. Any change that removes a part of this verification would reduce the perceived value of Tycho.

akurtakov:
What is the status here?

laeubi:
Closing for now

PR #175, opened 2021-07-03
[Doc] improve ReadMe and Contributing (Debugging- and Snapshot-section)

HannesWell:
With this PR I suggest improvements for Tycho's documentation.
This includes the following changes:
- In README.md
  - fix the formatting of the doc list, on Github the listing was not rendered correctly.
  - add an item to the release notes
  - mention the snapshots as well

- In CONTRIBUTING
  - extend the snapshot section, inspired by the Release-Notes sections from [Tycho's Eclipse wiki](https://wiki.eclipse.org/Tycho/Release_Notes/2.4) so one can simply copy and paste the relevant snippets.
  - extend the build section to distinguish between builds for 'testing' if everything works and builds to install Tycho locally to test changes with a project-build.
  - Change the suggested way to debug Tycho to the (in my opinion) more convenient way of launching Maven from Eclipse in Debug-mode. Do you think the previous way should remain as an alternative?

Please don't hesitate to let me know, if you don't find any of those changes useful.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -152,11 +152,19 @@ In order to debug Tycho plugins inside Eclipse:
 
 1. Get the Tycho sources in Eclipse
 2. Create/get a project that highlights the bug
-3. Create a Maven Run-Configuration in your Tycho Eclipse-workspace to build the project, specify goals, profiles and properties as required.
+
+Inside the Eclipse IDE:
+
+3. Create a Maven Run-Configuration in your Tycho Eclipse-workspace to build the project and specify goals, profiles and properties as required
 4. Launch the Maven-configuration from your Eclipse in Debug-mode
 
-Before debugging a build, you should make sure that your local Tycho-sources correspond to the Tycho version used by the project being build.
-Otherwise the debugger may behaves unexpected.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -124,17 +136,27 @@ Tycho has Maven dependencies to Equinox and JDT, so these artifact are used from
...
+and use the corresponding Tycho (probably snapshot) version in the project being build.
+You can also debug that build with the steps below (from here you can jump to step 3 immediately).
 
 ## Debugging 
 
 In order to debug Tycho plugins inside Eclipse:
 
 1. Get the Tycho sources in Eclipse
 2. Create/get a project that highlights the bug
-3. Run the project with `mvnDebug clean install`
```
mickaelistria, line 135: I think we should keep a note about mvnDebug here as alternative, or maybe fond a link to maven doc to reference that mentions such debug strategies.

HannesWell: Agree on that. I added it as a second variant (IMO debugging the build directly from Eclipse is more convenient and therefore should be named first).

Unfortunately I did not found official doc about `mvnDebug` on the Apache Maven website.

PR #181, opened 2021-07-09
Merge Tycho-demo into Tycho

mickaelistria:


jmini:
*Tycho/How Tos/Dependency on pom-first artifacts* updated:
https://wiki.eclipse.org/index.php?title=Tycho%2FHow_Tos%2FDependency_on_pom-first_artifacts&type=revision&diff=443651&oldid=418903

PR #186, opened 2021-07-14
MacOS means Cocoa

### LINKED ISSUE #184: [Tycho SureFire 2.3.0] org.eclipse.swt.SWTError cannot be found by org.eclipse.ui.workbench on Mac aarch64
Sorry if this is the wrong place to report this.

I'm running some Eclipse unit tests on a Mac Mini M1 (aarch64) using Tycho Surefire 2.3.0 that is failing with the error:

`java.lang.ClassNotFoundException: org.eclipse.swt.SWTError cannot be found by org.eclipse.ui.workbench_3.122.200.v20210506-1640`

This is working fine on Windows, Linux and Mac x86 (Intel), it only happens when running on Mac aarch64.

Error log attached.

[error.log](https://github.com/eclipse/tycho/files/6811074/error.log)

akurtakov:
Remove old code mapping only x86_64 macOS to cocoa and everything else
to Carbon. Carbon is simply dead for years.
Fixes #184 .


laeubi (commented):
I really wonder if there is not something in equinox/platform we can reuse here...

akurtakov:
Not in such simple/unified way AFAICT - code there is spread in different places.

laeubi:
Really sad.. as Eclipse would have such logic obviously also somewhere implemented...

PR #211, opened 2021-08-02
test cache first then download (and don't fail when online doesn't work if it can be resolved to cache)

jcompagner:
also make sure that if the download fails that it doesn't fail the
system
This is offline first but it will go online if needed.

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/remote/RemoteRepositoryCacheManager.java
+++ b/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/remote/RemoteRepositoryCacheManager.java
@@ -77,14 +77,20 @@ class RemoteRepositoryCacheManager extends CacheManager {
 
     @Override
     public File createCacheFromFile(URI remoteFile, IProgressMonitor monitor) throws ProvisionException, IOException {
+        File cacheFile = getCacheFile(remoteFile, getCacheDirectory());
         if (offline) {
-            File cacheFile = getCacheFile(remoteFile, getCacheDirectory());
             if (cacheFile != null) {
                 return cacheFile;
             }
             throw new ProvisionException(getFailureStatus(remoteFile));
         }
-        return super.createCacheFromFile(remoteFile, monitor);
+        try {
+            return super.createCacheFromFile(remoteFile, monitor);
+        } catch (IOException e) {
+            return handleCreateCacheException(cacheFile, remoteFile, e);
+        } catch (ProvisionException e) {
+            return handleCreateCacheException(cacheFile, remoteFile, e);
+        }
     }
 
     private Status getFailureStatus(URI uri) throws ProvisionException {
@@ -100,8 +106,8 @@ class RemoteRepositoryCacheManager extends CacheManager {
         return new File(dataAreaFile, Integer.toString(hashCode));
     }
 
-    private <T extends Exception> File handleCreateCacheException(File cacheFile, URI repositoryLocation, T e)
-            throws T {
+    private File handleCreateCacheException(File cacheFile, URI repositoryLocation, Exception e)
+            throws ProvisionException {
         if (cacheFile != null) {
             String message = "Failed to access p2 repository " + repositoryLocation.toASCIIString()
                     + ", use local cache.";
@@ -114,7 +120,7 @@ class RemoteRepositoryCacheManager extends CacheManager {
             // original exception has been already logged
             return cacheFile;
         }
-        throw e;
+        throw new ProvisionException(getFailureStatus(repositoryLocation));
     }
 
     @Override


laeubi:
@jcompagner you should make sure that your ECA matches the mail address you used in your commit and have added your github user id to your eclipse profile.

mickaelistria:
Can you please separate the part about not failing in case of network issue and the part about using cache first into distinct commits/PRs?
While the first part can be straightforward to merge, the second one would require more work to behave as expected (eg support for -U and regular re-check like Maven snapshots)

jcompagner:
the above patch is just that, that it ignores better http errors and falls back to cached content.
I don't care to much about the offline first part (because that could be more problematic, the cache manager really should understand when to or not to cache based on a hash created from the target file, else updates could not be found if we never do that)
But i don't care to much about that, if tycho at least uses the cached stuff and just goes on and tries to build if there are network errors then thats fine for me (and thats the current pr)

It could be a bit nicer IF the p2.index file was also cached. but thats not something tycho can really do anything about.
For that we need to change p2 code itself.

mickaelistria:
Ok. Is the network issue logged as a warning in case of unaccessible remote repo, but file is found in cache?

jcompagner:
yes you see this:

https://github.com/eclipse/tycho/blob/master/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/remote/RemoteRepositoryCacheManager.java#L106

the only thing is if i turn of my network i see per repository in my target quite a few lines of those
that is because the p2.index is not cached by the p2 code and because of that the suffix list is always something like 6 (so it tries 6 things per "content.xxxx" and "artifacts.xxx" which doesn't happen if the index page was there because the index page right away says use only this suffix 
That did cost me quite some time to understand why the result was so different when i hit

mvn install
mvn install -o

and the both of them with or without a network...

so caching the p2.index would make it a lot faster when network is down and with a lot less logs (only the actual warning of i can't read content.xml.xy ) (which in online mode it would only do)

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/remote/RemoteRepositoryCacheManager.java
```diff
@@ -77,14 +77,19 @@ public File createCache(URI repositoryLocation, String prefix, IProgressMonitor
 
     @Override
     public File createCacheFromFile(URI remoteFile, IProgressMonitor monitor) throws ProvisionException, IOException {
+        File cacheFile = getCacheFile(remoteFile, getCacheDirectory());
         if (offline) {
-            File cacheFile = getCacheFile(remoteFile, getCacheDirectory());
             if (cacheFile != null) {
                 return cacheFile;
             }
-            throw new ProvisionException(getFailureStatus(remoteFile));
```
mickaelistria, line 85: Shouldn't the "throw" here be preserved in case of offline + cache miss? I have the impression modified code would try to download the file in that case.

mickaelistria (changes requested):

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/remote/RemoteRepositoryCacheManager.java
jcompagner: yes, not sure why i removed that, force pushed a new change

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/remote/RemoteRepositoryCacheManager.java
```diff
@@ -77,14 +77,20 @@ public File createCache(URI repositoryLocation, String prefix, IProgressMonitor
 
     @Override
     public File createCacheFromFile(URI remoteFile, IProgressMonitor monitor) throws ProvisionException, IOException {
```
mickaelistria, line 79: Shouldn't the "IOException" be removed as possible "throws" ?

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/remote/RemoteRepositoryCacheManager.java
```diff
@@ -77,14 +77,20 @@ public File createCache(URI repositoryLocation, String prefix, IProgressMonitor
...
-            File cacheFile = getCacheFile(remoteFile, getCacheDirectory());
             if (cacheFile != null) {
                 return cacheFile;
             }
             throw new ProvisionException(getFailureStatus(remoteFile));
         }
-        return super.createCacheFromFile(remoteFile, monitor);
+        try {
+            return super.createCacheFromFile(remoteFile, monitor);
+        } catch (IOException e) {
```
mickaelistria, line 89: Could this be merged as `IOException | ProvisionException e` ?

mickaelistria (approved):
I've added a few cosmetic comments just for the sake of the discussion.
But it's not a blocker for a merge, so let's merged.

Issue #80, opened 2021-04-20
Bug 571951 - Incorrect requirement version for configuration/plugins in publish-products


and-k:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=571951

and-k:
After taking a look at changes between the versions 0.22.0 and 0.23.0 I suspect that the implementation of https://bugs.eclipse.org/bugs/show_bug.cgi?id=456543 are causing this issue.
Unfortunately I cannot build these old versions anymore since they are referencing some old eclipse update sites which aren't available anymore.

and-k:
I'm currently trying to find a solution for this issue.

The main problem imho is that every plugin defined in the configuration of the product is being set to version 0.0.0 in [org.eclipse.equinox.internal.p2.publisher.eclipse.ProductFile.processPluginConfiguration()](https://github.com/eclipse/rt.equinox.p2/blob/c98a11a5c10db6ca7f16da8d4df45172790bd22d/bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/internal/p2/publisher/eclipse/ProductFile.java#L1060).
Later in [org.eclipse.equinox.p2.publisher.eclipse.ProductFileAdvice.getConfigData()](https://github.com/eclipse/rt.equinox.p2/blob/c98a11a5c10db6ca7f16da8d4df45172790bd22d/bundles/org.eclipse.equinox.p2.publisher.eclipse/src/org/eclipse/equinox/p2/publisher/eclipse/ProductFileAdvice.java#L182) only for explicitly defined plugins within the \<plugins> of a product the correct version are being set.
Plugins defined indirectly within the \<features> of a product are not being considered and are always only set to version 0.0.0

This code has been the same for years even with Tycho 0.22.0. The difference though is that with https://github.com/eclipse/tycho/commit/f59db5e1dd35f087325daa2c2451ffd69867f59c the dependency resolution in Tycho has been changed and therefore this issue occurred. The commit message suggests that the new dependency resolution was intended by the author.

I see two possible solutions now:
1. P2 would need to handle the plugins defined in \<configurations> not as BundleInfo since it sets the version to 0.0.0 if no explicit version is available which is always the case because in Eclipse the graphical "Product Configuration Editor" doesn't even allow to set a version for plugins in \<configurations>. Instead only a "BundleConfigurationInfo" should be introduced without a version.
[...]

PR #214, opened 2021-08-07
Enhanced Testcase for #80 to show problem with simpleconfigurator

laeubi:
Enhanced the testcase to include simpleconfigurator (and check it is working)

### DIFF
--- a/tycho-its/projects/product.differentVersions/feature/feature.xml
+++ b/tycho-its/projects/product.differentVersions/feature/feature.xml
@@ -14,4 +14,18 @@
          version="5.2.0"
          unpack="false"/>
 
+   <plugin
+         id="org.eclipse.equinox.simpleconfigurator"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
+   <plugin
+         id="org.eclipse.equinox.simpleconfigurator.manipulator"
+         download-size="0"
+         install-size="0"
+         version="0.0.0"
+         unpack="false"/>
+
 </feature>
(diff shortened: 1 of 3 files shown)


mickaelistria:
I don't think the 2 issues are strongly coupled (they're related but not equivalent), so it would IMO be better to write a separate test for the simpleconfigurator.

laeubi:
Checking for simple configurator is necessary  as it shows that everything works as expected.

mickaelistria:
You're putting a lot of content and checks into a single test. That is not how test scale and help best.
Having 2 different features and test cases would be better here to better identify which part of the functionality is working/failing.

laeubi:
These are not two features! The test check a specific problem and all parts are required to prove a working solution. For sure I can copy the whole test project but this won't save a single line not would it make things more clear. The initial test was simply incomplete.

mickaelistria:
Why is the activemq part necessary to test the simpleconfigurator part?
I'd really rather see that specific simpleconfigurator issue isolated from everything else to better track it specifically, as experiece just showed that the former test does cover 1 story well, and that this story didn't seem to cover the simpleconfigurator well; and the simpleconfigurator part does not require the activemq part to be failing. There are opportunities for 1 test to break independently of the other, so it's better to separate them.

laeubi:
The activemq part is required to actually check that an older version is used (what is not possible with simple-configurator alone), the simple-configurator is required to check, that even if the build succeeds, the product is created as a valid one.
Otherwhise there will be no builderror but the result is (maybe) a brocken product. thats why these both are required to be checked together.

mickaelistria:
So you agree that there are 2 distinct stories: 1 about verifying correct version is used (covered by activemq), and another about properly supporting simpleconfigurator? If so, it confirms to me 2 distinct feature/product/tests would be more profitable.

laeubi:
Obsolete for now.

PR #218, opened 2021-08-14
Create a github action as pre-validation check

laeubi:
Even though eclipse-infra is back online, I think it is good to get some experience with github actions.
Currently the itest are not run (due to special requirements for java8) but this might be still good as a first pre-validation, and the eclipse-ci can then run the full build+integration suite.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,32 @@
+# This workflow will build a Java project with Maven
+# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
+
+name: Build Tycho
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 11
+      uses: actions/setup-java@v2
+      with:
+        java-version: '11'
+        distribution: 'adopt'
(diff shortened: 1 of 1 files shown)


mickaelistria:
What's missing for integration tests to run?

laeubi:
This requires an additional JDK8 to be setup and configured. While this seems possible its not directly supported by gh-action right now and would require some manual setup.
To keep things simple and clean, I think we should start with a (fast) Action first and can later enhance this to also run the itests (maybe in a separate action).

akurtakov:
Why is JDTK 8 needed? IMHO changes should be done so everything works with Java 11+.

laeubi:
JdtBreeJava8Test.testCompilerJdtBreeJava8 requires a java8 toolchain to work.

mickaelistria:
Is there a way to add comment about that missing piece in the .github/workflows/maven.yml file directly?

laeubi:
You can add comments with the `#` symbol to the file.

mickaelistria:
OK. Can you please add a comment and missing integration-tests and why; and then I think we can merge.

mickaelistria (approved):
That's a good 1st iteration.

laeubi:
This will also send out notifications if a commit breaks the build.

laeubi:
Our first two actions have run https://github.com/eclipse/tycho/actions successfull :1st_place_medal: :+1:

laeubi:
And here how it looks like on a PR: https://github.com/eclipse/tycho/pull/219

Issue #216, opened 2021-08-13
Allow Tycho surefire debug settings to be customised


joeshannon:
The surefire test allows a [debugPort](https://www.eclipse.org/tycho/sitedocs/tycho-surefire-plugin/test-mojo.html#debugPort) to be set only. This results in the following to be added to the Java arguements: `cli.addVMArguments("-Xdebug", "-Xrunjdwp:transport=dt_socket,address=" + debugPort + ",server=y,suspend=y");`

The maven surefire plugin allows this to be customised with the [maven.surefire.debug](https://maven.apache.org/surefire/maven-surefire-plugin/examples/debugging.html) property. It would be useful if Tycho supported something like this too.

For context, my use case is that I want to allow a different host to connect to the process. This used to work but as of Java 11 only localhost connections are permitted by default (see e.g. [this issue](https://github.com/OpenLiberty/open-liberty/issues/7833)).

laeubi:
@joeshannon this sounds reasonable can you provide a PR for this?

joeshannon:
Yes I am happy to take a look. I think the parameter names might require some thought/agreement. 

What do you think would be a good parameter name? `tycho.surefire.debug` to correspond with Maven surefire or something different e.g. `debugArgs`?

For the exiting `debugPort`, we would need to decide whether to keep this as an alternative or perhaps deprecate it as would be configurable via the new debug args.

There is also the existing `debugOptions` but this relates to Equinox/OSGi debug options.

laeubi:
I think `tycho.surefire.debug` is fine. I would recommend to have some kind of fallback approach: if `tycho.surefire.debug` is not specified, use the `debugPort` as before.

laeubi:
Adding a hint to debugPort seems to suffice then! AFAIK you simply need to adjust the javadoc and it will appear in the site-doc of the next release.

mickaelistria:
Is patch #219 enough to cover this story?

joeshannon:
Yes I think so, thanks for reviewing/merging!

PR #219, opened 2021-08-16
Issue #216 Add hint to debugPort parameter for test mojo

joeshannon:
To help guide the user to argLine which can be used to achieve the same
customization as the maven.surefire.debug property in the Maven Surefire
plugin.

### DIFF
--- a/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java
+++ b/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java
@@ -137,7 +137,8 @@ public abstract class AbstractTestMojo extends AbstractMojo {
 
     /**
      * Set this parameter to suspend the test JVM waiting for a client to open a remote debug
-     * session on the specified port.
+     * session on the specified port. If further customization of JVM debug parameters is
+     * required then {@link argLine} can be used instead.
      */
     @Parameter(property = "debugPort")
     private int debugPort;


tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java
```diff
@@ -137,7 +137,8 @@
 
     /**
      * Set this parameter to suspend the test JVM waiting for a client to open a remote debug
-     * session on the specified port.
+     * session on the specified port. If further customization of JVM debug parameters is
```
joeshannon, line 140: I used the US spelling of customization (differing from my issue), I'm assuming that is preferred by Eclipse projects?

laeubi: Honestly, I have no idea :-D
@mickaelistria might can tel something about...

laeubi (approved):

tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java
mickaelistria: I don't think there is a rule that enforce the british or US variant of a word. So whatever works ;)

mickaelistria (approved):

mickaelistria:
Thank you!

PR #228, opened 2021-08-19
Fix #203 : assemble-maven-repository should include PGP signatures in the P2 site

### LINKED ISSUE #203: assemble-maven-repository should include PGP signatures in the P2 site
As P2 support PGP signatures as an alternative to code signed artifacts, and maven artifacts that are deployed to maven central are PGP signed already we should add this information to prevent any "resign" requirements later on.

laeubi:
Adds support for maven signatures in the  `assemble-maven-repository`

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,10 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## Next release...
 
+### [Support for PGP Signatures in maven-p2 sites](https://github.com/eclipse/tycho/issues/203)
+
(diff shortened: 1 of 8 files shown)


laeubi:
@mickaelistria any comments on this? This is now complete from my point of view and I would like to squash/merge it if the build succeeds to have this for the jetty-story so we get PGP signatures there as well.

mickaelistria:
I won't review into details as the work is isolated enough, not intrustive on the core parts and I don't plan to use this mojo (as mentioned earlier, this is IMO a step we don't need from Platform POV). So I trust you judgement and feel free to merge at your own convenience.

laeubi:
Alright, I'll need to fix the test then its ready to be merged.

PR #231, opened 2021-08-22
Clean up DependencyComputer

HannesWell:
With this PR I want to suggest to improve the the DependencyComputer in the following ways:
- remove PackageSource and use the ModuleCapability directly: The capability of a PackageSource is not used, so it seems it can be removed.
- use a LinkedHashSet instead of a ArrayList when `Collection.contains()` is called often.

I created the change with the intention of improving the runtime of DependencyComputer, especially from the second point, but unfortunately I could not measure a relevant improvement. Probably it was just very small.
But since this change also reduces the necessary code, I submit it nevertheless.


mickaelistria:
Thanks!

PR #245, opened 2021-08-27
Use latest bouncycastle 1.69

akurtakov:
* there is no longer bcpg-jdk16 rather use bcpg-jdk15on
* licensecheck tool no longer gives warnings for this version
* adjust code for API changes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -221,7 +221,7 @@
 			<dependency>
 				<groupId>org.mockito</groupId>
 				<artifactId>mockito-core</artifactId>
-				<version>3.11.2</version>
+				<version>3.12.4</version>
 				<scope>test</scope>
 			</dependency>
 			<dependency>


laeubi:
It seems that the versions from the bundle-product and the maven of bouncycaste collides here. I think to fix this we must wait for P2 to use latest version, then export the BC as system packages from the maven world and remove the bc bundles from the bundles product as we do with the maven API.

laeubi:
I think the root-cause for the error is the code in P2:

```
try {
  signature.init(new JcaPGPContentVerifierBuilderProvider().setProvider(new BouncyCastleProvider()),
		  publicKey);
} catch (PGPException ex) {
  setStatus(new Status(IStatus.ERROR, Activator.ID, ex.getMessage(), ex));
  return;
}
```

this creates a new BouncyCastleProvider that uses a very broad class-path scanning to find (possible alternative) providers, but I think this is not necessary here, from the other code I get the impression that actually `BcPGPContentVerifierBuilderProvider` should be used instead of the `JcaPGPContentVerifierBuilderProvider` @mickaelistria can you tell why here the Jca one is used?

laeubi:
I have [created a P2 build](https://git.eclipse.org/r/c/equinox/rt.equinox.p2/+/184567 ) to see if this causes any test-failures, but I'm still not able to [test this in an IDE](https://bugs.eclipse.org/bugs/show_bug.cgi?id=575489) due to lack of PGP update-site, is there any test-update-site that could be used that already has PGP signatures?
@mickaelistria  @akurtakov And even though I know its really late, if this works is there any chance to get this into P2 before the final 2021-09 release?

PR #250, opened 2021-08-31
Various cleanups

akurtakov:


tycho-extras/tycho-p2-extras-plugin/src/test/java/org/eclipse/tycho/plugins/p2/extras/PublishFeaturesAndBundlesMojoTest.java
```diff
@@ -94,10 +97,12 @@ private void initLegacySupport(List<MavenProject> projects, MavenProject current
         buildContext.setSession(session);
     }
 
-    private void generateContentToBePublished(File repositoryFolder) throws IOException {
+    private void generateContentToBePublished(String repositoryFolder) throws IOException {
```
mickaelistria, line 100: I don't think replacing a File by a String is a good idea; maybe by a Path?

mickaelistria (commented):
Would it be possible to avoid the bulk formatting that applied to many files? It makes it almost impossible to review specific changes.

tycho-extras/tycho-p2-extras-plugin/src/test/java/org/eclipse/tycho/plugins/p2/extras/PublishFeaturesAndBundlesMojoTest.java
akurtakov: Fixed with latest commit in the series.

mickaelistria (approved):
diisabling whtespace on the diff view helped at reviewing

PR #251, opened 2021-09-01
Upgrade BC version

laeubi:
This will require the latest P2 release to work.

@akurtakov I have tested this with a snapshot build of P2 and this works, so as soon as the P2 is upgraded this should succeed to build and could be merged.


akurtakov:
Thanks, I'll update to 4.21 content once it's released.

mickaelistria:
@laeubi Does your local testing involve updating the .target file used by Tycho to 4.21 stream? If so, could you please submit a PR for it as a prepartion step for adoption of 4.21?

laeubi:
I didn't changed the the target file, as mentioned in the last gerrit P2  currently includes an 'old' 1.65 (the same as tycho).
If I understand you correctly it will only be updated to 1.69 once the release is out as it consumes bouncy castle from the platform? Due to the package imports (1.65 or higher) it should work with any version (even though a range of (1.65, 2] would be more appropriate if BC uses semantic-versioning).

mickaelistria:
Can you please try rebasing on top of master (which includes newer p2 and newer BC in the Equinox container)

laeubi:
Done

PR #262, opened 2021-09-07
To tycho 2.5.0-SNAPSHOT

jmini:
Similar to 7a4bf731e2e95a100eb573504c492880e7108fee by @mickaelistria, I have the feeling the `target.usefile` project should also be updated to `2.5.0-SNAPSHOT`.


laeubi:
Just as a note: For the itest this property is always defined as the current tycho version (the property is only usefull if you like to run the project outside the itests)

mickaelistria:
Thanks!

Issue #266, opened 2021-09-14
Tycho does not work with mvnd


gnodet:
See https://github.com/mvndaemon/mvnd/issues/477

When building https://github.com/eclipse/eclipse.jdt.ls with `mvnd`, the tycho plugin fails the second time with the exception below.
This happens because mvnd caches the plugin realms, in this particular case, the tycho classloader is cached and reused across builds.  This cause the underlying equinox framework instance to be reused without being shutdown.

Not sure what's the best way to solve the problem: either having the tycho extension explicitely shut down the equinox framework when the build is finished or have only the P2 services being shutdown (no idea how atm)...

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-p2-repository-plugin:2.3.0:assemble-repository (default-assemble-repository) on project org.eclipse.jdt.ls.product: Could not assemble p2 repository: Mirroring failed: No repository found at file:/Users/gnodet/tmp/eclipse.jdt.ls/org.eclipse.jdt.ls.product/target/. -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho:tycho-p2-repository-plugin:2.3.0:assemble-repository (default-assemble-repository) on project org.eclipse.jdt.ls.product: Could not assemble p2 repository
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:215)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:156)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:148)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:117)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:81)
	at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:56)
	at org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:128)
	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:305)
	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:192)
	at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:105)
[...]

PR #267, opened 2021-09-14
Shutdown equinox when the maven session ends, #266

gnodet:


mickaelistria:
This change seems good per se. Should we merge it as a 1st operation? If so, please remove its "draft" status.

tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoMavenLifecycleParticipant.java
```diff
@@ -107,6 +111,20 @@ public void afterProjectsRead(MavenSession session) throws MavenExecutionExcepti
         }
     }
 
+    @Override
+    public void afterSessionEnd(MavenSession session) throws MavenExecutionException {
+        if (plexus.hasComponent(EquinoxServiceFactory.class)) {
+            try {
+                EquinoxServiceFactory factory = plexus.lookup(EquinoxServiceFactory.class);
+                if (factory != null) {
```
laeubi, line 119: Instead of having a new "refresh" method won't it be an option to check for the Disposable interface her instead?

gnodet: I don't think so because of the different semantics.  That's the problem and I had to change it. Plexus `dispose` method is called automatically by plexus when the bean is _unmanaged_, i.e. removed from the container or when the container is itself destroyed.  
I tried that in the first version but the problem is that once the bean is unmanaged, plexus won't dispose it again.  So the semantics we want is different because we still need the bean to be managed by plexus, but we need to shutdown the container.  I though it would be better to now mix the semantics and rather have a dedicated method (even if `dispose` calls `refresh` at the end.
Now, if you think you'd rather keep a single one, that's fine and it's a matter of adding a check as you indicate.

laeubi: I'm just not sure we should have a "dangerous" method in the public API, if yes we should better call it "shutdown" or something so its clear. But I think it would also be fine to check for disposeable its a very special situation here anyways I think.

laeubi: Or maybe we can check for "Closeable" and let the implementation implement this ?

gnodet: It could also be moved to `EmbeddedEquinox` which makes more sense, or even to a new specific interface.

laeubi: @mickaelistria any preference on how to handle this?

mickaelistria: And couldn't avoid a new method, and directly call `dispose` here with a comment describing why this needs to be invoked explicitly?
Or if a separate method is really necessary, let's call it `shutdownEquinoxContainer` or `disposeEquinoxContainer` to make it more explicit; and then dispose() would invoke this new method rather than the other way round.

laeubi: > And couldn't avoid a new method, and directly call `dispose` here with a comment describing why this needs to be invoked explicitly?

I think this is enough for now 

```
if (factory instanceof Disposable) {
  ((Dispose)factory).dispose();
}
```

laeubi:
@gnodet can you rebase and adjust the test then I think we are ready to go

laeubi (approved):
Looks good to me.

Issue #213, opened 2021-08-06
New Equinox Resolver in Tycho 2.3.x and above fails to resolve in aceptable time in certein situations


laeubi:
This is a summary ticket for a problem reported already in the following tickets:

- https://github.com/eclipse/tycho/issues/212
- https://github.com/eclipse/tycho/issues/197

The issue here is, that in certain situations tycho creates a resolver task that seems incredible hard to resolve. A very good example project could be found [here](https://github.com/eclipse/tycho/files/6922395/hanging_project.zip) from the issue #212. The project is very simple (only one bundle), does not contain any special configuration (plain updatesites, only two standard envs) and also the bundle itself has no special requirements (just a few simple bundle imports).

It still results in a massive resolve operation that takes up to 6 minutes! A profiling session reveals that there are millions of calls to ModuleContainerAdaptor, while the similar ipt01 tycho project just has a few hundreds of them. As even though the average time is low, the plain number of calls makes this very slow and at least in #197 results in crashes due to out-of-memory.

![profiling](https://user-images.githubusercontent.com/1331477/128478621-5906d77f-f6eb-4bf6-9449-a735f9844546.png)

Maybe @mickaelistria can tell a bit more if this is an issue of Tycho, it might even be an issue of equinox or Apache Felix Resolver used by equinox.

mickaelistria:
I've enabled extra logging for the resolver (this needs to be achieved through code by settings some debugOptions to the ModuleContainerAdapter in EquinoxResolver). It shows lots of lines of computation and, more interesting, warning reports. Here is a subset, but about all log is variation of that:

> 
> RESOLVER: Candidate permutation failed due to a conflict between imports; will try another if possible. (Uses constraint violation. Unable to resolve resource com.sample.hanging.bundle [osgi.identity; osgi.identity="com.sample.hanging.bundle"; type="osgi.bundle"; version:Version="1.0.0.qualifier"; singleton:="true"] because it is exposed to package 'com.sun.istack' from resources com.sun.xml.bind.jaxb-impl [osgi.identity; osgi.identity="com.sun.xml.bind.jaxb-impl"; type="osgi.bundle"; version:Version="2.3.3"] and com.sun.xml.bind [osgi.identity; osgi.identity="com.sun.xml.bind"; type="osgi.bundle"; version:Version="2.3.3.v20201118-1818"] via two dependency chains.
> 
> Chain 1:
[...]

PR #268, opened 2021-09-14
Issue #213 - EquinoxResolver slow to resolve

mickaelistria:


tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
```diff
@@ -202,6 +202,16 @@ protected ModuleContainer newState(DependencyArtifacts artifacts, Properties pro
             throws BundleException {
         ModuleContainer[] moduleContainerAccessor = new ModuleContainer[1];
         ModuleContainerAdaptor moduleContainerAdaptor = new ModuleContainerAdaptor() {
+
+            @Override
+            public String getProperty(String key) {
+                switch (key) {
+                case "equinox.resolver.revision.batch.size":
```
laeubi, line 209: Maybe we should add a comment /reference to bug to explain why this is necessary?

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
```diff
@@ -202,6 +202,16 @@ protected ModuleContainer newState(DependencyArtifacts artifacts, Properties pro
...
+                    return Integer.toString(1);
```
laeubi, line 210: Maybe we could replace this by System.getProperty("tycho.resolver.batch.size", "1") to allow configuration through system properties.

mickaelistria:
Merged as 47726f1745c7d8836a4ea02ba53ff4bb34b553c5

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
mickaelistria: I didn't add it. More options is more complexity and are only worth the effort when we know we need them.

PR #274, opened 2021-09-15
Update to 2021-09 release dependencies

akurtakov:


akurtakov:
@laeubi @mickaelistria  I see some test failures e.g. Tycho188P2EnabledRcpTest and my feeling is it's related to https://bugs.eclipse.org/bugs/show_bug.cgi?id=574952 . Could you please run the IT test on top of this PR and share what you think?

laeubi:
It seems jenkins is missing permissions to pick up the PR I have opened [Bug 576014](https://bugs.eclipse.org/bugs/show_bug.cgi?id=576014)

tycho-extras/pom.xml
```diff
@@ -28,7 +28,7 @@
 	<description>Additional tools for Tycho</description>
 
 	<properties>
-		<jgit-version>5.12.0.202106070339-r</jgit-version>
+		<jgit-version>5.13.0.202109080827-r</jgit-version>
```
laeubi, line 31: Just a general note: Should we move those to the "master" pom instead of having two locations where we define versions / managed dependencies?

akurtakov: It's fine for not widely used dependencies to be defined where they are used. E.g. jgit in extras, objectweb-asm in artifactcomparator. There is no need to define everything in single place as it makes the process of bigger changes like groupId change more expensive and there are leftovers in disconnected places.

laeubi:
I get 6 errors:
[ERROR] Errors: 
[ERROR]   JdtBreeJava8Test.testCompilerJdtBreeJava8:25 » Verification Exit code was non-...
[ERROR]   ProductMixedVersionsTest.testMixedPluginVersions:27 » Verification Exit code w...
[ERROR]   Tycho188P2EnabledRcpTest.buildProduct:69 » Verification Exit code was non-zero...
[ERROR]   P2InstalledTestRuntimeTest.testDifferentHarnessVersions:57 » Verification Exit...
[ERROR]   P2InstalledTestRuntimeTest.testProvisionAppAndRunTest:33 » Verification Exit c...
[ERROR]   P2InstalledTestRuntimeTest.testRunTestOnProvisionedApp:44 » Verification Exit ...

JdtBreeJava8Test is due to missing toolchain configuration (does anyone know if it is possible to configure toolchains by means of a maven-dependency?).

laeubi:
> my feeling is it's related to https://bugs.eclipse.org/bugs/show_bug.cgi?id=574952

ProductMixedVersionsTest should work in any case! Strange enough I run the test local and it fails, after that I changed the pom of the test to investigate different settings and now the test succeeds regardless if I revert the changes or not. 
After a while again fails for 2 or 3 runs then it succeeds with whatever configuration.

As I always run with 'clean' it shouldn't matter... I'm really confused

`tycho-its/projects/product.differentVersions$ mvn clean install -Dtycho-version=2.5.0-SNAPSHOT`

laeubi:
BTW the error the test fails with is the same I have seen while play around with P2  before the changes to Bug 574952 so maybe this was introduced by another change or this change changes the timing somewhere a little bit?

laeubi:
Tycho188P2EnabledRcpTest fails with different errors depending on how often it is run e.g. once with

```
[INFO] --- tycho-p2-director-plugin:2.5.0-SNAPSHOT:materialize-products (materialize-products) @ pi.eclipse-repository ---
[INFO] Installing product main.product.id for environment win32/win32/x86_64 to /home/christoph/git/laeubi/tycho/tycho-its/projects/product.installation/eclipse-repository/target/products/main.product.id/win32/win32/x86_64
Installing main.product.id 0.0.1.202109160617.
Installing pi.root-level-installed-feature.feature.group 1.0.0.202109160617.
Installation failed.
Cannot complete the install because one or more required items could not be found.
	Software being installed: Main Product 0.0.1.202109160617 (main.product.id 0.0.1.202109160617)
	Missing requirement: toolingwin32.win32.x86_64org.eclipse.equinox.p2.reconciler.dropins 0.0.1.202109160617 requires 'osgi.bundle; org.eclipse.equinox.p2.reconciler.dropins 0.0.0' but it could not be found
	Cannot satisfy dependency:
		From: Main Product 0.0.1.202109160617 (main.product.id 0.0.1.202109160617)
		To: org.eclipse.equinox.p2.iu; toolingmain.product.id.configuration [0.0.1.202109160617,0.0.1.202109160617]
	Cannot satisfy dependency:
		From: toolingmain.product.id.configuration 0.0.1.202109160617
		To: org.eclipse.equinox.p2.iu; toolingwin32.win32.x86_64org.eclipse.equinox.p2.reconciler.dropins [0.0.1.202109160617,0.0.1.202109160617]

```

and once with 

```
[INFO] Installing product main.product.id for environment win32/win32/x86_64 to /home/christoph/git/laeubi/tycho/tycho-its/projects/product.installation/eclipse-repository/target/products/main.product.id/win32/win32/x86_64
Installing main.product.id 0.0.1.202109160617.
Installing pi.root-level-installed-feature.feature.group 1.0.0.202109160617.
Installation failed.
Cannot complete the install because one or more required items could not be found.
	Software being installed: Main Product 0.0.1.202109160617 (main.product.id 0.0.1.202109160617)
	Missing requirement: Expression Language 3.7.0.v20200720-1126 (org.eclipse.core.expressions 3.7.0.v20200720-1126) requires 'java.package; org.w3c.dom 0.0.0' but it could not be found
	Cannot satisfy dependency:
		From: Main Product 0.0.1.202109160617 (main.product.id 0.0.1.202109160617)
		To: org.eclipse.equinox.p2.iu; org.eclipse.rcp.feature.group [4.17.0.v20200902-1800,4.17.0.v20200902-1800]
	Cannot satisfy dependency:
		From: Eclipse e4 core commands 0.12.900.v20200110-1732 (org.eclipse.e4.core.commands 0.12.900.v20200110-1732)
		To: java.package; org.eclipse.core.expressions 0.0.0
	Cannot satisfy dependency:
		From: Eclipse 4 Rich Client Platform 4.17.0.v20200831-1002 (org.eclipse.e4.rcp.feature.group 4.17.0.v20200831-1002)
		To: org.eclipse.equinox.p2.iu; org.eclipse.e4.core.commands [0.12.900.v20200110-1732,0.12.900.v20200110-1732]
	Cannot satisfy dependency:
		From: Eclipse RCP 4.17.0.v20200902-1800 (org.eclipse.rcp.feature.group 4.17.0.v20200902-1800)
		To: org.eclipse.equinox.p2.iu; org.eclipse.e4.rcp.feature.group [4.17.0.v20200831-1002,4.17.0.v20200831-1002]
```

laeubi:
Especially the second one is interesting I think as it does not contain any `tooling*****` item so something is real strange here... I'm currently try to build a local P2 without my patch to see if that changes anything (even though I don't understand how it could create random behavior with the same metadata generated in all cases if I look at the generated xmls ...)

akurtakov:
Yeah, I've noticed the same strange things.

laeubi:
I reverted my patch but the `Tycho188P2EnabledRcpTest` still fails. Even more confusing it also fails with Tycho 2.4.0 (would this also use a local build p2???)

laeubi:
@mickaelistria any idea here?

mickaelistria:
I didn't have time to look at it yet but plan to do it soon shortly.

akurtakov:
Another change that might be causing some issues is jdt.core now requiring Java-11 BREE. It still can produce old bytecode but needs to run on Java 11+ so maybe this affects resolving?

laeubi:
But why should this cause it to work/fail on different runs?

laeubi:
On the jenkins only these three fail:

- org.eclipse.tycho.test.iu.ProductWithIUTest.testRootFilesFromIUPackagingInstalledAndInRepo
- org.eclipse.tycho.test.product.Tycho188P2EnabledRcpTest.org.eclipse.tycho.test.product.Tycho188P2EnabledRcpTest
- org.eclipse.tycho.test.surefire.P2InstalledTestRuntimeTest.testProvisionAppAndRunTest

laeubi:
>  ProductWithIUTest.testRootFilesFromIUPackagingInstalledAndInRepo

Fails with 

> Missing requirement: Eclipse Content Mechanism 3.7.800.v20200724-0804 (org.eclipse.core.contenttype 3.7.800.v20200724-0804) requires 'java.package; javax.xml.parsers 0.0.0' but it could not be found

That could be related to the java 11 updates?

laeubi:
>     * org.eclipse.tycho.test.product.Tycho188P2EnabledRcpTest.org.eclipse.tycho.test.product.Tycho188P2EnabledRcpTest

Fail with

> Cannot complete the install because one or more required items could not be found.
> 	Software being installed: Main Product 0.0.1.202109161559 (main.product.id 0.0.1.202109161559)
> 	Missing requirement: Expression Language 3.7.0.v20200720-1126 (org.eclipse.core.expressions 3.7.0.v20200720-1126) requires 'java.package; org.w3c.dom 0.0.0' but it could not be found

seems also be related to a package no longer supplied by java 11

laeubi:
>     * org.eclipse.tycho.test.surefire.P2InstalledTestRuntimeTest.testProvisionAppAndRunTest

same here

> Missing requirement: Extension Registry Support 3.7.0.v20170222-1344 (org.eclipse.equinox.registry 3.7.0.v20170222-1344) requires 'java.package; javax.xml.parsers 0.0.0' but it could not

also missing package

laeubi:
From the description, [Bug 572099](https://bugs.eclipse.org/bugs/show_bug.cgi?id=572099) / commit fe498f234da05bcca171b34b06a61a61dfd97a0a could be related, it claims to:

> - Switched default profile used by JREAction to Java 11.
> - If no JRE profile given, JREAction generates a temporary profile based
> on default profile version and containing packages observed by the
> currently running JVM.
> - Updated JREActionTest to Java 11 expectations: 228 system packages and
> 23 capabilities

Maybe this is making a difference here?

laeubi:
> From the description, [Bug 572099](https://bugs.eclipse.org/bugs/show_bug.cgi?id=572099) / commit fe498f234da05bcca171b34b06a61a61dfd97a0a could be related

I can confirm that reverting that patch makes the test pass again. So the solution would be to adjust the test cases?

mickaelistria:
> So the solution would be to adjust the test cases?

I don't think so. It looks to me that Tycho was relying on some assumption of the JREAction that is now invalid. We need to clarifying what was the expectation from Tycho and either how we can make Tycho get rid of it, or fix Tycho to make this assumption still valid.
Looking at the p2 metadata in the `eclipse-repository/target/targetPlatformRepository` or `eclipse-repository/target/repository` that are then consumed to build p2 repo; I'm under the impression that some of the a.jre.javase units are missing system packages without an obvious good reason.

mickaelistria:
I think it's becoming necessary to have Tycho using its own JREAction that is capable of processing toolchains and other extra info about EE Tycho has and JREAction cannot leverage.

mickaelistria:
Merged as part of #280 , thanks to a commit that finally makes Tycho rely less on JREAction for locally known EEs: https://github.com/mickaelistria/tycho/pull/280/commits/e0f17213a686457f55252f019697e14c29ed722d

PR #275, opened 2021-09-16
Update maven.yml

laeubi:
testing in progress - do not merge

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -16,6 +16,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - name: Set up JDK 8
+      uses: actions/setup-java@v2
+      with:
+        java-version: '8'
+        distribution: 'adopt'
     - name: Set up JDK 11
       uses: actions/setup-java@v2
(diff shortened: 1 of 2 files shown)


laeubi:
@akurtakov @mickaelistria I have now configured two JDKs in the GH action and configured a toolchains.xml for the maven build but it seems this is not picked up by the integration tests. Is there a trick to define the toolchains.xml for the itests as well?

akurtakov:
Only reason for it to not work I can think of is forking IT tests and thus losing the --global-toolchain parameter. Can you put the file ~/.m2?

laeubi:
That helped for the test but two itests are still failing: https://github.com/eclipse/tycho/runs/3621612714 ... I'll try to investigate tomorrow.

laeubi:
Everything works now, @mickaelistria @akurtakov any concerns that I quash/merge this? This will then include the itests as well.

laeubi:
BTW this would allow us to use other jdks as well if desired.

mickaelistria:
No concern for a squash+merge as soon as you're ready.

Issue #271, opened 2021-09-15
Error when fetching repository from maven zip artifact


ghillairet:
I am using this feature introduced in tycho [2.2](https://wiki.eclipse.org/Tycho/Release_Notes/2.2#Support_for_consuming_maven_artifacts_made_of_zipped_P2_update_sites) but I am getting this exception when trying to load a repository, not sure if it is a problem with the repository or a bug in tycho.

Note that I can reproduce the issue with tycho 2.3, 2.4 and 2.5-SNAPSHOT, but I guess the issue could be in the p2 version.

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-packaging-plugin:2.5.0-SNAPSHOT:package-feature (default-package-feature) on project com.ataccama.ide.one.feature: Execution default-package-feature of goal org.eclipse.tycho:tycho-packaging-plugin:2.5.0-SNAPSHOT:package-feature failed: Load repository from url mvn:com.ataccama.extension.salesforce:com.ataccama.extension.salesforce.ide.repository:0.2.1-SNAPSHOT:zip failed (org.eclipse.equinox.p2.core.ProvisionException: No repository found at mvn:com.ataccama.extension.salesforce:com.ataccama.extension.salesforce.ide.repository:0.2.1-SNAPSHOT:zip.) -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho:tycho-packaging-plugin:2.5.0-SNAPSHOT:package-feature (default-package-feature) on project com.ataccama.ide.one.feature: Execution default-package-feature of goal org.eclipse.tycho:tycho-packaging-plugin:2.5.0-SNAPSHOT:package-feature failed: Load repository from url mvn:com.ataccama.extension.salesforce:com.ataccama.extension.salesforce.ide.repository:0.2.1-SNAPSHOT:zip failed (org.eclipse.equinox.p2.core.ProvisionException: No repository found at mvn:com.ataccama.extension.salesforce:com.ataccama.extension.salesforce.ide.repository:0.2.1-SNAPSHOT:zip.)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
[...]

PR #279, opened 2021-09-17
Add test case for #271

apeteri:


laeubi:
@apeteri thanks for this testcase, @mickaelistria do you have an idea why it is necessary to download the artifacts.xml as part of the build? Could this be related to the "lazy download" you have added a while ago? Maybe a workaround would be to always download artifacts+content xmls together...

mickaelistria:
> do you have an idea why it is necessary to download the artifacts.xml as part of the build?

I believe publish-product goal adds the product *and all its dependencies* into the repository. So since dependencies have to be downloaded at that stage (and weren't necessary earlier), the artifacts.xml is queried at that stage.
However, I don't know about the `mvn` protocol and why it would miss some file if they're not fetched earlier. It seems like a bug in local expansion of the `mvn:` based p2 repo.

laeubi:
I have replaced the maven site with a `file` reference by extracting it and then everything works. My guess is that if it is  downloaded before it is cached and the `publish-product` uses that cached location, while here it needs to fetch it initially and fails (maybe because `mvn` protocol is not known at that place).

laeubi:
@apeteri I have simplified the product a bit according to my test only a very minimal configuration is required to reproduce the issue

laeubi:
@apeteri the automatic rebase failed here can you please rebase manually?

apeteri:
The ECA check is complaining about @laeubi's e-mail address, not sure why 😕. My agreement looks to be still in effect.

laeubi:
Yeah don't mind that, I just want to see if with the rebase the build works fine :-)

laeubi:
Thanks for the test-case :-)

PR #301, opened 2021-09-22
Fix #261 BaselineValidator.validateAndReplace() fails with UnsupportedOperationException under Maven 3.8.2

### LINKED ISSUE #261: BaselineValidator.validateAndReplace() fails with UnsupportedOperationException under Maven 3.8.2
When using Maven 3.8.2 Tycho's `P2MetadataMojo` or `P2MetadataDefaultMojo` in version 2.4.0 (it is very likely the same for Tycho 2.5.0-SNAPSHOT because the class did not change) `BaselineValidator.validateAndReplace()` fails by throwing an `UnsupportedOperationException` when it attempts to remove an element from the list returned by `project.getAttachedArtifacts()` because the corresponding artifact is not present in the baseline.

As of Maven 3.8.2 `MavenProject.getAttachedArtifacts()` returns an unmodifiable List.
This was changed as part of the fix of [MNG-5868](https://issues.apache.org/jira/browse/MNG-5868) which was merged into the Maven master https://github.com/apache/maven/pull/358 and back-ported into the 3.8. stream as commit 66a43813770cd7dee9eb1ed0372e3f784ba0c541 (did not found the PR).

Unfortunately I did not found an alternative to remove attached artifacts from a MavenProjects.

laeubi:

### DIFF
--- a/tycho-p2/tycho-p2-plugin/src/main/java/org/eclipse/tycho/plugins/p2/BaselineValidator.java
+++ b/tycho-p2/tycho-p2-plugin/src/main/java/org/eclipse/tycho/plugins/p2/BaselineValidator.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012 Sonatype Inc. and others.
+ * Copyright (c) 2012, 2021 Sonatype Inc. and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -9,6 +9,7 @@
  *
  * Contributors:
  *    Sonatype Inc. - initial API and implementation
+ *    Christoph Läubrich - BaselineValidator.validateAndReplace() fails with UnsupportedOperationException under Maven 3.8.2
  *******************************************************************************/
 package org.eclipse.tycho.plugins.p2;
 
@@ -20,6 +21,7 @@ import static org.eclipse.tycho.plugins.p2.BaselineReplace.none;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -27,6 +29,7 @@ import java.util.ListIterator;
 import java.util.Map;
 import java.util.Map.Entry;
 
+import org.apache.commons.lang3.reflect.MethodUtils;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.plugin.MojoExecutionException;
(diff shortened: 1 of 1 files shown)


laeubi:
FYI @HannesWell

laeubi:
FYI @sbouchet

tycho-p2/tycho-p2-plugin/src/main/java/org/eclipse/tycho/plugins/p2/BaselineValidator.java
```diff
@@ -145,8 +148,21 @@ public String getDetailedMessage() {
...
-                                    iterator.remove();
+                                Artifact next = iterator.next();
+                                if (classifier.equals(next.getClassifier())) {
+                                    try {
+                                        iterator.remove();
+                                    } catch (UnsupportedOperationException e) {
+                                        ArrayList<Artifact> list = new ArrayList<>(attachedArtifacts);
+                                        list.remove(next);
+                                        try {
+                                            MethodUtils.invokeMethod(project, true, "setAttachedArtifacts", list);
```
HannesWell, line 159: Doesn't this cause an illegal reflective access error when the build is performed on a Java-17 VM?

laeubi: Then this workaround won't work of course, but tycho does not work in all combinations of build vm and maven version anyways.

HannesWell: But I think the combination of latest Java, latest Maven and latest Tycho is not some special niche configuration but one that should work. Especially because Users are always encouraged to use the latest versions, which I think is good. But then Tycho should support it.

laeubi: One could still use latest java to compile, but feel free to suggest a better alternative here :-)

HannesWell: Unfortunately I don't have a better solution immediately at hand.
I know this is not very helpful to find a solution for the moment, but I think it is a point to mention and to consider for the future.

laeubi: This mojo was designed in 2012 so many assumptions of it might not hold true for modern java, at least this allows to use it for some more time until we find a better solution e.g. convince maven to make the method public, provide an alternative approach to remove artifacts or even overhaul the BaseLineVlidator to use a different approach. Even though most eclipse-project use this, I have never seen a use of it before in any other tycho project so maybe this is also a bit eclipse-specific and the default ci-servers use java 11 to build afaik so there is some time to investigate for better solutions until then.

laeubi:
@mickaelistria @akurtakov any comments?

laeubi:
Cherry picked to 2.5.x as well

PR #306, opened 2021-09-24
More description in N&N auto of Bundle-License

jonahgraham:
BTW Thanks for adding a flag to control this - it means I can upgrade to Tycho 2.4 without having to bump every version number of every bundle in CDT. I hope this expanded N&N entry helps other resolve the issue more quickly.


laeubi:
Thanks for this. If you like please try out the [upcoming 2.5.x ](https://github.com/eclipse/tycho/discussions/295)release and report any issues.

Issue #303, opened 2021-09-22
M2E-core build fails with Uses-constraint-violations


HannesWell:
The Maven build of m2e-core fails when the latest Tycho 2.5.0-SNAPSHOT is used because the resolution of the Target-Platform fails because of `org.apache.felix.resolver.reason.ReasonException: Uses constraint violations`.

A draft PR in m2e outlines this issue: https://github.com/eclipse-m2e/m2e-core/pull/334:

```
[INFO] Fetching org.w3c.dom.smil_1.0.1.v200903091627.jar from https://download.eclipse.org/eclipse/updates/4.21/R-4.21-202109060500/plugins/ (21.24kB)
[ERROR] Internal error: java.lang.RuntimeException: org.osgi.framework.BundleException: Bundle org.eclipse.m2e.core.ui cannot be resolved:org.eclipse.m2e.core.ui [115]
[ERROR]   Unresolved requirement: Require-Bundle: org.eclipse.m2e.model.edit; bundle-version="[1.16.0,2.0.0)"; visibility:="reexport"
[ERROR]     -> Bundle-SymbolicName: org.eclipse.m2e.model.edit; bundle-version="1.18.2.qualifier"; singleton:="true"
[ERROR]        org.eclipse.m2e.model.edit [116]
[ERROR]          Unresolved requirement: Require-Bundle: org.eclipse.wst.xml.core; visibility:="reexport"
[ERROR]            -> Bundle-SymbolicName: org.eclipse.wst.xml.core; bundle-version="1.2.300.v202102110535"; singleton:="true"
[ERROR]               org.eclipse.wst.xml.core [105]
[ERROR]                 Unresolved requirement: Require-Bundle: org.eclipse.jem.util; bundle-version="[2.1.200,3.0.0)"; resolution:="optional"
[ERROR]                 Unresolved requirement: Require-Bundle: org.eclipse.wst.validation; bundle-version="[1.2.700,1.3.0)"; resolution:="optional"
[ERROR]                 Unresolved requirement: Require-Bundle: org.eclipse.wst.common.emfworkbench.integration; bundle-version="[1.2.100,1.3.0)"; resolution:="optional"
[ERROR]                 Bundle was not resolved because of a uses constraint violation.
[...]

PR #307, opened 2021-09-24
#303 M2E-core build fails with Uses-constraint-violations

laeubi:
Let resolver ignore 'uses' constraints to speedup resolving and prevent
failures for complex setups

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
@@ -42,6 +42,7 @@ import org.eclipse.osgi.container.ModuleContainerAdaptor;
 import org.eclipse.osgi.container.ModuleDatabase;
 import org.eclipse.osgi.container.ModuleRevision;
 import org.eclipse.osgi.container.ModuleRevisionBuilder;
+import org.eclipse.osgi.container.ModuleRevisionBuilder.GenericInfo;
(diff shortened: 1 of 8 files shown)


laeubi:
@mickaelistria @akurtakov @HannesWell @bjhargrave I have now implemented the "remove uses constraint" approach and was now able to derive a test case for this, to make the test fail one can run the following commandline:

```
tycho-its/projects/resolver.usesConstraintViolations$
     mvn clean install -Dtycho.equinox.resolver.batch.size=1 -Dtycho.equinox.resolver.uses=true
```

this forces the use of `uses` constraints and forces a batch-size of `1` what makes the resolver fail. I'll leave it to your decisions if we should merge this for `2.5.x` or if this is to risky but like to propose integrate this for `2.6.x` as with my testing speeds up resolving the classpath considerably but still prevents the `Bundle was not resolved because of a uses constraint violation` problem reliable.

mickaelistria:
IMO, we should try to merge it for 2.5, even if that implies we delay 2.5 by about a week.
It's very good that you managed to write a test.
p2 doesn't follow "uses" directive properly anyway, so when we reach the point of EquinoxResolver, it's not really something we should start taking too seriously. Maybe in the future, someone will report an issue with build-time dependencies missing a bundle which is supposed to comes from a "uses" directive, but we'll deal with this story separately, from p2 to EquinoxResolver.

laeubi:
Ok if no test failures occurs I'll merge this for 2.5 and master so we can respin release tests ASAP...

PR #333, opened 2021-10-12
Junit vintage 5.7

mickaelistria:


RELEASE_NOTES.md
```diff
@@ -12,6 +14,12 @@ Default for this mojo has been flipped from true to false as pack200 artifacts a
 <includePacked>true</includePacked>
 ```
 
+### Improved plain JUnit 5 suppoort
+
+Plain JUnit 5 tests now works fine without extra dependency on JUnit 4.
+
+Test bundles that mix JUnit 5 (with Jupiter engine) and JUnit 4 (with JUnit Vintage engine) will keep working; however, the support for newer JUnit versions is likely to not be provided. So it's recommended that users who mix JUnit versions either migrate everything to JUnit 5, or separate their test suites into 2 distinct bundles (that will use the different legacy JUnit 4 and newer Jupiter engines separately).
```
akurtakov, line 21: Such a requirement is too big for any migration. E.g. I think of platform test bundles they are still a mixture of JUnit 3 and 4 and when we manage to introduce 5 there is no chance that someone migrates all tests in given bundle to JUnit 5 . Creating more bundles and etc. is too much releng work to be seriously considered.

PR #384, opened 2021-11-18
Correctly use "rerunFailingTestsCount" property

Bananeweizen:
As somebody pointed out on the mailing list, the currently used
TestRequest constructor ignores the "rerunFailingTestsCount" property.
Another overload needs to be used instead.


mickaelistria:
Thanks!

PR #388, opened 2021-11-21
Describe different ways of profiling Tycho

Bananeweizen:


CONTRIBUTING.md
```diff
@@ -132,7 +132,23 @@ Note: Tycho always allows references to locally built artifacts, even if they ar
...
+
+You can [add a timestamp to each log line](https://blogs.itemis.com/en/in-a-nutshell-adding-timestamps-to-maven-log-output) produced by Maven. This is the most easy to apply, but you have to do the calculation of the runtime of different goals yourself.
+
+#### Add Maven profiler
+
+Download the [Maven profiler extension](https://github.com/jcgay/maven-profiler) and add it to your aggregator project. It will produce an HTML report for each goal.
+
+#### Yourkit YouMonitor
+
+[Yourkit YouMonitor](https://www.yourkit.com/youmonitor/) (not to be confused with Yourkit Profiler) can be used to measure the build time steps. It reports the timing for Maven mojos, Ant goals etc. You need to register it as a Java agent for your build. It allows easy comparison of multiple builds, therefore it's really nice for trying different optimizations and configurations. Be aware the free license is only available for local builds, not for CI servers.
```
laeubi, line 151: Can you add an example how one can run a tycho build locally with the free license?

CONTRIBUTING.md
```diff
@@ -132,7 +132,23 @@ Note: Tycho always allows references to locally built artifacts, even if they ar
...
+To understand where the build spends most of its time, you can try the following approaches:
+
+#### Add timestamps to Maven logging
+
+You can [add a timestamp to each log line](https://blogs.itemis.com/en/in-a-nutshell-adding-timestamps-to-maven-log-output) produced by Maven. This is the most easy to apply, but you have to do the calculation of the runtime of different goals yourself.
+
+#### Add Maven profiler
+
+Download the [Maven profiler extension](https://github.com/jcgay/maven-profiler) and add it to your aggregator project. It will produce an HTML report for each goal.
+
```
laeubi, line 148: Maybe we can describe here the "use extension way" from https://github.com/jcgay/maven-profiler#maven--33x ? This seems most suitable for tycho-builds.

laeubi (commented):
Thanks for the PR, can you adjust two minor things to make it easier for newcomers to use the suggested profiling options?

CONTRIBUTING.md
laeubi: I think we should recommend the extension way instead of download/manual install the plugin into maven.

Bananeweizen: I've incorporated all suggestions and described especially the YouMonitor usage as good as I remember. :) I used it last time maybe 2 years ago.

laeubi (approved):

PR #394, opened 2021-11-25
Fixes #342: Run surefire tests outside of lock

### LINKED ISSUE #342: Parallelization performance severely degraded since Tycho 1.7.0
We have been using Java 11 / Tycho 1.6.0 in our project since the release of Java 11. 

The build is quite lengthy and, without any special configuration, takes about 3 hours. Therefore we set the `-T 1C` command line option to build several projects in parallel. This logs warnings about unsupported parallelization but has always been working fine. Building in parallel drops the build time to 2 hours.

Anticipating the upcoming Java 17 support I tried upgrading the build system to Tycho 2.4.0 to test for any major roadblocks.

With Tycho 2.4.0 the unparallelized build takes the same 3 hours as with Tycho 1.6.0, but setting the parallelization option actually increases the build time to 3 hours and 47 minutes!

I tested with older versions of Tycho and it seems that the severe slowdown happens since version 1.7.0. I did not expect that - minor changes in the build time are fine, but I expected the build time to stay at least in the area of 2 hours.

||with `-T 1C`|without `-T 1C`|
|---|---|---|
|**Tycho 1.6.0**|2 hours|3 hours|
|**Tycho 1.7.0+**|~4 hours|3 hours|

Going from our current, parallelized, 2 hour build, that is an increase of almost 100 % - but even increasing the build time to 3 hours by building unparallelized is hardly an option.

Is there a problem with running the build parallelized since Tycho 1.7.0? Do I have to configure something differently?

Our maven call looks like this:
`mvn install -Dmaven.repo.local=<repository location> -P <profile name> -T 1C -fae`

The profile just determines which tests are executed and should not behave differently whether the `-T 1C` parameter is present or not.

k-spa:
Running tests while the lock is present leads to the tests being executed in series which severely degrades performance when the build is running with parallel maven projects.

### DIFF
--- a/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java
+++ b/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java
@@ -673,8 +673,8 @@ public abstract class AbstractTestMojo extends AbstractMojo {
             return;
         }
         if (shouldRun()) {
+            EquinoxInstallation equinoxTestRuntime;
             synchronized (LOCK) {
-                EquinoxInstallation equinoxTestRuntime;
                 if ("p2Installed".equals(testRuntime)) {
                     equinoxTestRuntime = createProvisionedInstallation();
                 } else if ("default".equals(testRuntime)) {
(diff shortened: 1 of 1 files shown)


laeubi (approved):

laeubi:
Thanks for the change. Just a note: You have created your PR from your forks master branch, it is often more desirable to create a branch for the feature and just name it e.g. issue_nnn_description that way you can for example always sync with the master repository if required.

PR #395, opened 2021-11-25
Investigate build time speedup

laeubi:
We might can get faster results by enabling parallel integration test builds.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -42,7 +42,7 @@ jobs:
         cp toolchains-gh.xml ~/.m2/toolchains.xml
         mvn -U -V -e clean install --file pom.xml
     - name: Run Integration Tests
-      run: mvn -U -V -e -Pits -T 1C -Dmaven.test.failure.ignore=true clean install --file tycho-its/pom.xml
+      run: mvn -U -V -e -Pits -Ditest-forkcount=1C -Dmaven.test.failure.ignore=true clean install --file tycho-its/pom.xml
     - name: Upload Test Results
       uses: actions/upload-artifact@v2
       with:
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -42,7 +42,7 @@ jobs:
         cp toolchains-gh.xml ~/.m2/toolchains.xml
         mvn -U -V -e clean install --file pom.xml
     - name: Run Integration Tests
-      run: mvn -U -V -e -Pits -Dmaven.test.failure.ignore=true clean install --file tycho-its/pom.xml
+      run: mvn -U -V -e -Pits -T 1C -Dmaven.test.failure.ignore=true clean install --file tycho-its/pom.xml
     - name: Upload Test Results
       uses: actions/upload-artifact@v2
       with:


akurtakov:
@laeubi  Is this the speedup you expected?

laeubi:
Not really, it seems either the parallel execution does not help much or Github is not giving more resources for our executors... as more projects moving to GH now it seems good to check if (and how) projects could acquire additional resources on GH infra.

laeubi:
I testes locally, normal build `Total time:  29:12 min`, with `-T 1C` I don't see direct speed-up most probably because all itests are in one module and maybe only separate modules are executed in parallel and it seems we have to enable that in code:
https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html

laeubi:
Using forkCount=1C on my local machine cut down the time by 1/3: `Total time:  10:34 min` but there are some failures so it seems not all test are yet ready for parallel execution. Anyways I created a PR to add a property to control the forkCount:
- https://github.com/eclipse/tycho/pull/412

Bananeweizen:
I've been playing with parallel testing in our company Tycho tests a lot in the past. I found it generally easier to enable parallelism in JUnit 5 than in Surefire (but admittedly, we also have UI test modules were Surefire parallelism would not even work)
Pros:
* parallelization also works when running in Eclipse
* parallelization can be disabled for single test classes/methods (and globally enabled by JUnit properties)
Cons:
* Need to migrate tests to JUnit 5 Jupiter, not using the Vintage engine.

If I get it right, the majority of tests is still JUnit 4, right? Are there any other obstacles prohibiting the migration of Tycho tests to Jupiter? If not, I might be able to help in converting them.

laeubi:
@Bananeweizen I'm not aware of any special thing that requires Juni4, actually the integration tests are rather "boring" and using the Maven Verifier under the hood. And as we are getting more and more tests (what is good) the verification times increase more and more (what is bad).

Then there are some unit-test in the different module, also here a parallel execution could help speedup things, so if you like to help here just go ahead :+1: 

Currently a typical full run (build+unit-tests+integration tests) takes ~ 40 minutes and thats too long for a good flow where smaller increments and fast feedback are describable.

laeubi:
I found an interesting extension here, not sure if everyone is aware of it already but still like to share it here:

http://takari.io/book/30-team-maven.html#takari-smart-builder

AhmetBilgin:
> I found an interesting extension here, not sure if everyone is aware of it already but still like to share it here:
> 
> http://takari.io/book/30-team-maven.html#takari-smart-builder

@laeubi I asked about this extension some time ago in the old tycho-user list: [see here](https://www.eclipse.org/lists/tycho-user/msg07405.html)
The conclusion was that this extension would have to be somehow integrated into tycho as I understood, but this exceeds my tycho knowledge.
I ended up doing full clean builds every time, as we encountered problems with the incremental build of maven and tycho (see the tycho-user post for details).

If someone can integrate this into tycho, it's probably you :)

Best regards,
Ahmet

laeubi:
@AhmetBilgin thanks for the reference, just to make sure we are talking about the same thing, the link point to the `takari-smart-builder` (that should support an improved parallel build and independent from tycho) while the news-post seems to reference `takari-incremental-java-compiler` or do I get something wrong?

AhmetBilgin:
@laeubi you're absolutely right. I missed that these are two different extensions.

akurtakov:
Should we drop this one? Or there is some plan?

laeubi:
@Bananeweizen has offered to tweak the test and we probably could enable parallel compile.

laeubi:
But I think we can close this PR

PR #402, opened 2021-11-29
Document parallel testing changes (fixes #342)

### LINKED ISSUE #342: Parallelization performance severely degraded since Tycho 1.7.0

Bananeweizen:


mickaelistria:
+1 for updating the doc.
However, on the long run, you may want to add some option to the TestMojo to prevent it from running parallel executions.

laeubi:
Isn't there already a flag (useUIHarness) for UI tests that might be used here?

mickaelistria:
UI tests usually can execute in parallel. It's more some specific tests or frameworks that require to have the window active. Bit tests can usually have other "locks", eg a DB, that can apply.
So useUIHarness doesn't fit as it's not fine-grain enough. I imagine maven-surefire-plugin may have some capabilities about it that Tycho could consume.

Bananeweizen:
Using that flag seems reasonable on first glance. However, you can also run SWTBot tests just to have thread-safe access to controls, without ever using the "click" features etc. Such tests would be fine. Also, I'm not a linux expert, but I would expect to be able to use different virtual screens there where parallel execution would be okay. That's why I would go for something more explicit (which would also be more easy to find in the documentation then).

Bananeweizen:
BTW: Is there some description available for how you committers generally work? Since I had no negative feedback about this PR and since it passed tests, I would now merge it. Or are self-merges of own PRs not okay?

laeubi:
I would then rather not add any options here but let either

- disable parallel execution for such test
- enhance the framework / or use "safe" variants (especially for SWT bot you can construct a bot that only works on a specific window regardless of "active" window)
- use some other kind of synchronization (e.g. lockfiles or similar)

laeubi:
> BTW: Is there some description available for how you committers generally work? Since I had no negative feedback about this PR and since it passed tests, I would now merge it. Or are self-merges of own PRs not okay?

If you are certain that your change is small enough without getting any feedback from others you can merge it, otherwise you can always request a review on it.

PR #411, opened 2021-12-05
Fix #331 Support pom dependecies in maven targets

### LINKED ISSUE #331: Support pom dependecies in maven targets
Starting with the next release m2e will support pom dependencies also see https://github.com/eclipse-m2e/m2e-core/issues/242 we should add support for this as well.

laeubi:
This add support for resolving pom-dependencies in maven targets.


mickaelistria:
Thanks! Please add a note about it to RELEASE_NOTES document.

PR #413, opened 2021-12-05
Avoid Maven info about missing resource encoding

Bananeweizen:
Without that setting the build has 3 occurrences of the following info.

[INFO] The encoding used to copy filtered properties files have not been
set. This means that the same encoding will be used to copy filtered
properties files as when copying other filtered resources. This might
not be what you want! Run your build with --debug to see which files
might be affected. Read more at
https://maven.apache.org/plugins/maven-resources-plugin/examples/filtering-properties-files.html


mickaelistria:
Thanks!

PR #415, opened 2021-12-05
upgrade Tycho Surefire JUnit Jupiter 5.8 to 5.8.2

Bananeweizen:
Use Jupiter 5.8.2 instead of 5.8.1 at Tycho Surefire runtime, when
selecting the junit58 provider.


laeubi:
@akurtakov is the license check covering this already or do we need to check this manually?

akurtakov:
Yes, it's covered.

mickaelistria:
> is the license check covering this already or do we need to check this manually?

I don't think the license check does cover artifacts other than plain "dependencies", here they're in maven-dependency-plugin configuration so they won't be considered.

akurtakov:
When I said covered I was speaking for junit 5.8.2 and friends being already covered from IP side of things. I haven't looked at the mechanics here.

PR #417, opened 2021-12-06
Fix #401 - Support nested targets

### LINKED ISSUE #401: Support nested targets
PDE supports [nested targets](https://bugs.eclipse.org/bugs/show_bug.cgi?id=577184) now, tycho should support this as well.

laeubi:
This adds support for nested targets.

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/target/TargetDefinitionResolver.java
+++ b/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/target/TargetDefinitionResolver.java
@@ -157,7 +157,7 @@ public final class TargetDefinitionResolver {
                 try {
                     fileLocation = new File(resolvePath).getCanonicalFile();
                 } catch (IOException e) {
-                    throw new ResolverException("I/O Error while resolve path " + resolvePath, e);
+                    throw new ResolverException("I/O Error while resolving path " + resolvePath, e);
                 }
                 if (fileLocation.exists()) {
                     FileTargetDefinitionContent fileRepositoryRolver = fileRepositories.computeIfAbsent(
@@ -177,7 +177,7 @@ public final class TargetDefinitionResolver {
                     }
                 } else {
                     logger.warn("Target location path '" + fileLocation.getAbsolutePath()
-                            + "' does not exist, target resoloution might be incomplete.");
+                            + "' does not exist, target resolution might be incomplete.");
                 }
             } else if (locationDefinition instanceof MavenGAVLocation) {
                 MavenGAVLocation location = (MavenGAVLocation) locationDefinition;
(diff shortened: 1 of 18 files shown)


laeubi:
FYI @vogella

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/target/TargetDefinitionResolver.java
```diff
@@ -148,10 +151,18 @@ TargetDefinitionContent resolveContentWithExceptions(TargetDefinition definition
...
                 PathLocation pathLocation = (PathLocation) locationDefinition;
-                File path = resolvePath(pathLocation.getPath(), definition);
-                if (path.exists()) {
+
+                String resolvePath = resolvePath(pathLocation.getPath(), definition);
+                File fileLocation;
+                try {
+                    fileLocation = new File(resolvePath).getCanonicalFile();
+                } catch (IOException e) {
+                    throw new ResolverException("I/O Error while resolve path " + resolvePath, e);
```
Bananeweizen, line 160:
```suggestion
                    throw new ResolverException("I/O Error while resolving path " + resolvePath, e);
```

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/target/TargetDefinitionResolver.java
```diff
@@ -165,7 +176,7 @@ TargetDefinitionContent resolveContentWithExceptions(TargetDefinition definition
                                 new LoggingProgressMonitor(logger)));
                     }
                 } else {
-                    logger.warn("Target location path '" + path.getAbsolutePath()
+                    logger.warn("Target location path '" + fileLocation.getAbsolutePath()
                             + "' does not exist, target resoloution might be incomplete.");
```
Bananeweizen, line 180:
```suggestion
                            + "' does not exist, target resolution might be incomplete.");
```

tycho-its/projects/target.references/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+	<version>0.0.1-SNAPSHOT</version>
+	<packaging>pom</packaging>
+
+	<modules>
+		<module>test.bundle</module>
+		<module>target.refs</module>
+	</modules>
+
+	<properties>
+		<tycho-version>2.6.0-SNAPSHOT</tycho-version>
```
Bananeweizen, line 24: I still haven't understood why this is defined again and again in many ITs (and with older versions than 2.6 in some cases). Shouldn't all tests point to the current development version for testing (by inheriting from the root parent of tycho)? What am I missing there?

Bananeweizen (commented):
I don't really understand most of the changes (due to my lacking knowledge, not because of the changes). Generally it looks okay to me.

tycho-its/projects/target.references/pom.xml
laeubi: The` tycho-version` is overridden in the actual i-test run with the current build version so in that value is then irrelevant in this case as it is never used.
But if you are open the pom.xml in eclipse m2e of course complains that `tycho-version` is undefined and all kind of error markers are shown and code-completion does not work. Thus it is defined in (some but not all) of the itest mostly with the version that was recent at time of writing the test, it can always be adjusted to the latest version of course.

Just for convenience, it allows to right-click and Choose `Run As > Maven` install to test without a full itest run or specifying the version on the commandline.

laeubi:
> I don't really understand most of the changes (due to my lacking knowledge, not because of the changes). Generally it looks okay to me.

Thanks for your suggestions, I had to refactor out an xml-lib tycho previously used to pare target files to move the parsing code to  common location so don't mind it looks a bit crowded.

Issue #410, opened 2021-12-05
Lots of deprecation errors in tycho build 'Please fix your build!'


laeubi:
I'm using maven 3.6.3 and get a lot of

```
[ERROR] 

Maven dependencies of Maven Plugins should be in provided scope.
Please make sure that all your dependencies declared in POM whose group ID is
org.apache.maven have set '<scope>provided</scope>' as well.
In the future this error will break the build.

The following dependencies are in wrong scope:
 * org.apache.maven:maven-plugin-api:jar:3.8.4:compile
 * org.apache.maven:maven-model:jar:3.8.4:compile
 * org.apache.maven:maven-artifact:jar:3.8.4:compile
 * org.apache.maven:maven-core:jar:3.8.4:compile
 * org.apache.maven:maven-settings:jar:3.8.4:compile
 * org.apache.maven:maven-settings-builder:jar:3.8.4:compile
 * org.apache.maven:maven-builder-support:jar:3.8.4:compile
 * org.apache.maven:maven-repository-metadata:jar:3.8.4:compile
 * org.apache.maven:maven-model-builder:jar:3.8.4:compile
 * org.apache.maven:maven-resolver-provider:jar:3.8.4:compile
 * org.apache.maven:maven-compat:jar:3.8.4:compile

Please fix your build!
```

laeubi:
@akurtakov is this intentional or do we have a problem in tycho?

akurtakov:
Seems that this is coming from https://issues.apache.org/jira/projects/MPLUGIN/issues/MPLUGIN-370?filter=allissues and maven-plugin-tools was updated with https://github.com/eclipse/tycho/commit/a4b5add36e88a74bc0bfd30e60bff05a37b7d885 . So this is really new error/warning coming from Maven developers. Seems like these dependencies will have to be with scope=provided

Bananeweizen:
The problem is, at least for me locally just adding those scopes leads to compile errors for running "mvn clean install", e.g. in tycho-surefire-plugin. I have compared the effective POM and see no other difference than the one line adding the scope. I'm rather clueless why it doesn't compile.

PR #441, opened 2021-12-16
Fix warning from maven-plugin-plugin

akurtakov:
It comes as:
'''
[ERROR]
Maven dependencies of Maven Plugins should be in provided scope.
Please make sure that all your dependencies declared in POM whose group
ID is
org.apache.maven have set '<scope>provided</scope>' as well.
In the future this error will break the build.
'''

so just reading the log makes it look like an error which it isn't
(yet!).

There are couple more staying due to maven-surefire-plugin and
maven-gpg-plugin having this bad practice and thus transitively showing
in Tycho build too.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,6 +99,7 @@
 			<dependency>
 				<groupId>org.apache.maven</groupId>
 				<artifactId>maven-plugin-api</artifactId>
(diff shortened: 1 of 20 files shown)


laeubi:
See also 
- #410

PR #450, opened 2021-12-20
Fix typos

Bananeweizen:
Some of these are in public class/field names, but those have only been
added in this SNAPSHOT version, so it is safe to rename them.


laeubi:
> Some of these are in public class/field names

As far as I know there is no restriction regarding "API compatibility" in Tycho so no need to worry here at all :+1:

PR #461, opened 2021-12-31
Fix #460 Delay classpath resolution to the compile phase

### LINKED ISSUE #460: Delay classpath resolution to the compile phase
Currently classpath resolution is performed as part of the project setup in the `TychoResolver.resolveProject(MavenSession, MavenProject, List<ReactorProject>)` this has the disadvantage that

- it is probably done unnecessarily (e.g. if I call `mvn validate` and probably `mvn clean`)
- we need to handle parallel builds while otherwise maven would do the job for us
- it is not possible to consider sources for the class-path that are part of the regular maven build (might be related to this discussion https://github.com/eclipse/tycho/discussions/434)
- it is not possible to consider non-tycho projects that provide additional artifacts (eg. from `felix-bundle-plugin`)

looking at the code of `OsgiBundleProject.resolveClassPath(MavenSession, MavenProject)` it seems not strictly necessary to run this code at this early phase...

The result of this operation is set into the following keys:

- `CTX_ECLIPSE_PLUGIN_CLASSPATH`
- `CTX_ECLIPSE_PLUGIN_STRICT_BOOTCLASSPATH_ACCESSRULES`
- `CTX_ECLIPSE_PLUGIN_BOOTCLASSPATH_EXTRA_ACCESSRULES`

interestingly these are only used inside OsgiBundleProject and/or Tycho compiler mojo but not accessed at the early stage, so it seems valid to init them when first accessed.

laeubi:

### DIFF
--- a/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
+++ b/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
@@ -69,7 +69,6 @@ import org.eclipse.tycho.classpath.ClasspathEntry.AccessRule;
 import org.eclipse.tycho.classpath.JavaCompilerConfiguration;
 import org.eclipse.tycho.classpath.SourcepathEntry;
 import org.eclipse.tycho.core.BundleProject;
-import org.eclipse.tycho.core.TychoConstants;
 import org.eclipse.tycho.core.TychoProject;
 import org.eclipse.tycho.core.dotClasspath.JREClasspathEntry;
 import org.eclipse.tycho.core.dotClasspath.M2ClasspathVariable;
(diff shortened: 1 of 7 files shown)


laeubi:
Example output that shows that classpath resolution is delayed until compile phase:

```
[INFO] ---------------------------[ eclipse-plugin ]---------------------------
[INFO] 
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ tycho.bundle ---
[INFO] Deleting /home/christoph/git/laeubi/tycho/tycho-its/projects/mixed.reactor/tycho.bundle/target
[INFO] 
[INFO] --- tycho-packaging-plugin:2.6.0-SNAPSHOT:build-qualifier (default-build-qualifier) @ tycho.bundle ---
[INFO] The project's OSGi version is 1.0.0.202112310830
[INFO] 
[INFO] --- tycho-packaging-plugin:2.6.0-SNAPSHOT:validate-id (default-validate-id) @ tycho.bundle ---
[INFO] 
[INFO] --- tycho-packaging-plugin:2.6.0-SNAPSHOT:validate-version (default-validate-version) @ tycho.bundle ---
[INFO] 
[INFO] --- target-platform-configuration:2.6.0-SNAPSHOT:target-platform (default-target-platform) @ tycho.bundle ---
[INFO] 
[INFO] --- maven-resources-plugin:2.4.3:resources (default-resources) @ tycho.bundle ---
[INFO] skip non existing resourceDirectory /home/christoph/git/laeubi/tycho/tycho-its/projects/mixed.reactor/tycho.bundle/src/main/resources
[INFO] 
[INFO] --- tycho-compiler-plugin:2.6.0-SNAPSHOT:compile (default-compile) @ tycho.bundle ---
[INFO] Resolving class path of tycho.bundle...
```

laeubi:
I have added a `validate-classpath` mojo just in case one want to still perform the class-path-computation in the `verify` phase.
This change also proofs that the additional resolving via the Equinox Resolver is only necessary for classpath computation and not for project order.

PR #465, opened 2022-01-01
add validate-classpath to the initialize phase of default life-cycle

laeubi:
FYI @cdietrich this should make the test/xtext pass without additional configuration.

### DIFF
--- a/tycho-maven-plugin/src/main/resources/META-INF/plexus/components.xml
+++ b/tycho-maven-plugin/src/main/resources/META-INF/plexus/components.xml
@@ -18,7 +18,8 @@
                 org.eclipse.tycho:tycho-packaging-plugin:${project.version}:validate-version
               </validate>
               <initialize>
-                org.eclipse.tycho:target-platform-configuration:${project.version}:target-platform
+                org.eclipse.tycho:target-platform-configuration:${project.version}:target-platform,
+                org.eclipse.tycho:tycho-compiler-plugin:${project.version}:validate-classpath
               </initialize>
               <process-resources>
                 org.apache.maven.plugins:maven-resources-plugin:${resources-plugin.version}:resources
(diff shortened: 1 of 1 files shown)


cdietrich:
will rebase

Bananeweizen:
@laeubi Could this lead to performance decrease due to resolving the classpath multiple times or do you think that doesn't matter?

laeubi:
> @laeubi Could this lead to performance decrease due to resolving the classpath multiple times or do you think that doesn't matter?

The classpath is always only resolved once for a given project.

sratz:
> > @laeubi Could this lead to performance decrease due to resolving the classpath multiple times or do you think that doesn't matter?
> 
> The classpath is always only resolved once for a given project.

Actually, it looks like this still costs us about 800ms per maven project in our project containing 700 artifacts.

Is there a way to disable this?

laeubi:
> Is there a way to disable this?

You can simply not compile your code ... in all other cases you require to resolve the classpath of course.

laeubi:
No this is not a regression, why should it *not* compute the classpath? This does not make any sense. The time was simply before taken on another place (in the project setup phase)..

sratz:
Yes, exactly. I thought something was done twice, but it's not. So no issue with this pull request.

But something else has changed in 2.7.0 that slows down our build, that has do do with classpath computation / compilation.

laeubi:
> But something else has changed in 2.7.0 that slows down our build, that has do do with classpath computation / compilation.

Actually it should be faster if you enable parallel execution (e.g. `-T1C`)

sratz:
> Actually it should be faster if you enable parallel execution (e.g. `-T1C`)

Things are already parallelized heavily due to heavy load on Jenkins.

If possible it would be nice to get the root cause for this slowdown and not just throw more computing power at it.

laeubi:
> If possible it would be nice to get the root cause for this slowdown and not just throw more computing power at it.

Without any measures for "slowdown" its hard to guess. In any way if there is no "free processing power" parallelizing won't help, so you either have processing power laying around and can improve this. For example with -T1C I could speed up my serial builds (compared to 2.6.0) by 50%-70% ...

mickaelistria:
@sratz Can you please open a separate bug for the apparent perf regression iwth more details? Discussing it on a close and not necessarily related issue isn't best, it seems worth a dedicated ticket.

sratz:
> @sratz Can you please open a separate bug for the apparent perf regression iwth more details? Discussing it on a close and not necessarily related issue isn't best, it seems worth a dedicated ticket.

Yes, I have opened #719.

PR #468, opened 2022-01-02
Allow to resolve with missing requirements and record those

laeubi:

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/util/resolution/ResolutionData.java
+++ b/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/util/resolution/ResolutionData.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012 SAP AG and others.
+ * Copyright (c) 2012, 2022 SAP AG and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -9,6 +9,7 @@
  *
  * Contributors:
  *    SAP AG - initial API and implementation
+ *    Christoph Läubrich - #462 - Delay Pom considered items to the final Target Platform calculation
  *******************************************************************************/
 package org.eclipse.tycho.p2.util.resolution;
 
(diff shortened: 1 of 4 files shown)


mickaelistria:
Please try to enrich commit message for such important change. This seems to have the capability to alter behavior quite noticeably and the commit message doesn't explicit the value provided by this change.

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/util/resolution/ResolutionData.java
```diff
@@ -30,4 +31,14 @@
...
+
+    /**
+     * @return <code>true</code> if the resolve operation should fail if there are missing
+     *         requirements
+     */
+    boolean failOnMissingRequirements();
+
+    void addMissingRequirement(IRequirement requirement);
+
+    Collection<IRequirement> getMissingRequirements();
```
mickaelistria, line 43: This is not appropriate as part of the ResolutionData interface IMO. ResolutionData is to configure the resolution and shouldn't incorporate information about resolution results as the ResolutionData can be reused across multiple builds. Please rework it.

mickaelistria:
You introduce some code here without any concrete consumer (I don't see anywhere in the code where `setFailOnMissing(true)` in invoked so this change doesn't bring value at the moment. It would be nice to just not merge such thing greedily without a more complete story of how this is supposed to be used and how it's supposed to serve end-users (eg with an integration test highlighting the value).

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/util/resolution/ResolutionData.java
laeubi: > ResolutionData can be reused across multiple builds

There are only 3 locations where it is used, in all those locations the ResolutionData is disposed after the method call.

laeubi:
> Please try to enrich commit message for such important change. This seems to have the capability to alter behavior quite noticeably and the commit message doesn't explicit the value provided by this change.

This does not alter any behavior without explicitly enabling it, so everything just works as before.

This is a prequisite for https://github.com/eclipse/tycho/pull/470 and I just don't wanted to complicate this story more as it already is, thus I extracted some of the changes from that.

mickaelistria:
> thus I extracted some of the changes from that.

My concern is more about merging such changes early without guarantee the whoke story works. IMO such change should be merged only when the "functional" patch that does create value is approved, otherwise the net impact is just that technical debt grows without increasing value.

laeubi:
> > thus I extracted some of the changes from that.
> 
> My concern is more about merging such changes early without guarantee the whoke story works. 

I have a working example for this and tested it with some larger build-setups, anyways we already have some demands in the past for a similar feature so I'm very certain we can use it in one way or the other.

> IMO such change should be merged only when the "functional" patch that does create value is approved, otherwise the net impact is just that technical debt grows without increasing value.

we can easily revert this change if necessary and as it does not require existing code to change its impact is very low.

PR #469, opened 2022-01-02
Add a debug(msg, exception) method to the maven logger

laeubi:

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.core.shared/src/main/java/org/eclipse/tycho/core/shared/MultiLineLogger.java
+++ b/tycho-bundles/org.eclipse.tycho.core.shared/src/main/java/org/eclipse/tycho/core/shared/MultiLineLogger.java
@@ -89,6 +89,19 @@ public class MultiLineLogger implements MavenLogger {
     }
 
     @Override
+    public void debug(String message, Throwable cause) {
+        String[] messageLines = message.split("\n");
+        for (int ix = 0; ix < messageLines.length - 1; ix++) {
+            String messageLine = messageLines[ix];
+            delegate.debug(messageLine);
+        }
+        if (messageLines.length > 0) {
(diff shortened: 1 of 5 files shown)


mickaelistria:
Can you please update some existing piece of code to use this new method so it does not seem unused?

laeubi:
This is used here now: https://github.com/eclipse/tycho/pull/468/files#diff-4df1b0d100c2af5db6fcd56209b07e2ce96ebbf71b1b3d9ff125cc9029f2dcaaR107

mickaelistria:
Sure but in general, make sure now methods are submitted with a change that leverage them in the existing code, or incorporate it in PRs which leverage them if current code cannot be updated to use them.

laeubi:
In this case there was a bit chicken-egg problem, I could have submitted the other PR with an incomplete message and change it afterwards again but this does not seems sufficient given that its quite obvious what is missing here.

tycho-bundles/org.eclipse.tycho.core.shared/src/main/java/org/eclipse/tycho/core/shared/MultiLineLogger.java
```diff
@@ -88,6 +88,19 @@ public void debug(String message) {
         }
     }
 
+    @Override
+    public void debug(String message, Throwable cause) {
+        String[] messageLines = message.split("\n");
```
Bananeweizen, line 93: what about `"\r?\n"` to catch both linux and windows?

laeubi: I just copied that from the existing implementations. If we want to change her we should change this for the other methods as well.
Beside that, the purpose of multi-line logger is a bit strange to me especially if it comes to parallel builds it could soon become quite confusing.

PR #472, opened 2022-01-02
Fix #471 Consider JAVA_HOME in tycho-document-bundle-plugin

### LINKED ISSUE #471: tycho-document-bundle-plugin tests depend on PATH
After lots of trial and error I've just found one reason for failing integration tests. The document-bundle-plugin tests run javadoc by just invoking the javadoc executable without using any path. Therefore whatever is first on the local PATH environment variable will be called. Since there are subtle differences between Java 8 (what I had on the path) and Java 11 Javadoc, the verification groovy script fails with some files not found, some files not matching content and finally even NoClassDefFoundError (for the Doclet).

What is a good way to avoid that? I can imagine these solutions:
* Somehow get the toolchain definition from maven (how?) and use that path to create a FQN
* verify the generated files to contain "Generated by javadoc (11" (which they have as HTML comment)

Any better ideas?

Bananeweizen:
Use also JAVA_HOME to find a suitable javadoc executable. That logic is
similar to what plexus JavacCompiler does.

Tested locally on Windows. Not sure how to have an actual integration test for this change.

### DIFF
--- a/tycho-extras/tycho-document-bundle-plugin/src/main/java/org/eclipse/tycho/extras/docbundle/JavadocRunner.java
+++ b/tycho-extras/tycho-document-bundle-plugin/src/main/java/org/eclipse/tycho/extras/docbundle/JavadocRunner.java
@@ -33,294 +33,312 @@ import org.eclipse.tycho.core.osgitools.BundleReader;
 import org.eclipse.tycho.core.osgitools.OsgiManifest;
 
 public class JavadocRunner {
-    private File output;
+	private File output;
 
-    private ToolchainManager toolchainManager;
+	private ToolchainManager toolchainManager;
 
-    private MavenSession session;
+	private MavenSession session;
 
-    private Set<File> sourceFolders = Collections.<File> emptySet();
+	private Set<File> sourceFolders = Collections.<File>emptySet();
 
-    private Set<File> manifestFiles = Collections.<File> emptySet();
+	private Set<File> manifestFiles = Collections.<File>emptySet();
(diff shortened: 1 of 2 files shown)


tycho-extras/tycho-document-bundle-plugin/src/main/java/org/eclipse/tycho/extras/docbundle/JavadocRunner.java
```diff
@@ -33,294 +33,312 @@
...
+			javadocFromJavaHome += ".exe";
+		}
+
+		log.debug("Testing javadoc from java.home = " + javadocFromJavaHome);
+
+		if (new File(javadocFromJavaHome).canExecute()) {
+			return javadocFromJavaHome;
+		}
+
+		// use JAVA_HOME
```
Bananeweizen, line 303: lines 303 to 317 are the actual change. Not sure what lead to showing everything as diff.

laeubi: never mind, I think this is due to changed formatting settings.

laeubi (approved):
Looks good to me and as it does not break any existing code and is more close to `maven-javadoc-plugin` I think it is fine.

As this slightly can change behavior we probably want to add a hint to the release notes, but I'm not sure how widely used this mojo actually is.

mickaelistria:
Thanks!

PR #473, opened 2022-01-03
Use Maven coordinates are used when possible in dependencies

mickaelistria:
`mvn dependency:list`

### DIFF
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/target/TargetPlatformLocationsTest.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/target/TargetPlatformLocationsTest.java
@@ -42,6 +42,16 @@ public class TargetPlatformLocationsTest extends AbstractTychoIntegrationTest {
 	}
 
 	@Test
+	public void testMavenArtifactHaveMavenDepsCoordinates() throws Exception {
+		Verifier verifier = getVerifier("target.maven", false, true);
+		verifier.executeGoal("dependency:list");
+		verifier.verifyErrorFreeLog();
+		verifier.verifyTextInLog("commons-lang:commons-lang:jar:2.4:system"); // this is a weak assert, should
+																				// be
+		// improved
+	}
+
+	@Test
 	public void testMavenLocationMulti() throws Exception {
 		Verifier verifier = getVerifier("target.mavenMulti", false, true);
 		verifier.executeGoal("verify");


tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java
```diff
@@ -127,7 +133,24 @@ private boolean requiresExternalDependencies() {
     }
 
     private Dependency createSystemScopeDependency(ArtifactKey artifactKey, File location) {
-        return createSystemScopeDependency(artifactKey, P2_GROUPID_PREFIX + artifactKey.getType(), location);
+        if (session != null && location.getAbsolutePath().startsWith(session.getLocalRepository().getBasedir())) {
+            String suffix = location.getAbsolutePath().substring(session.getLocalRepository().getBasedir().length());
```
laeubi, line 137: I'm not sure if this is really a good idea, under a given path there could be stored multiple artifacts of different type, so should we not at least the the Dependency type also? What about the classifiers?

tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java
```diff
@@ -127,7 +133,24 @@ private boolean requiresExternalDependencies() {
...
+            List<String> segments = new ArrayList<>(Arrays.asList(suffix.split(File.separator)));
+            segments.removeIf(String::isEmpty);
```
laeubi, line 139: Won't something like 'Path' be better suited to examine segment instead of string splitting? or even file.getParent().getName() ...

tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java
laeubi: And if we are only for the main artifact, maybe it would be good to check for the .pom file related to the main artifact is present? One could even think about reading the Model (see org.eclipse.tycho.osgi.configuration.MavenDependenciesResolverConfigurer.loadModel(File) for example) and read the information from their?

mickaelistria: I agree it's not best, but it's better and IMO good enough for the moment and the concrete requirements that have led to this development (eg https://github.com/eclipse/dash-licenses/issues/125#issuecomment-996562928 )

> And if we are only for the main artifact, maybe it would be good to check for the .pom file related to the main artifact is present? One could even think about reading the Model (see org.eclipse.tycho.osgi.configuration.MavenDependenciesResolverConfigurer.loadModel(File) for example) and read the information from their?

What issue does it fix or what benefit does it bring?

tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java
mickaelistria: Yes, I'll try to rework this.

tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java
laeubi: Just consider the following case:

1. You have Bundle A deployed at maven together with its corresponding source bundle AS.
2. Bundle B Require-Bundle A+AS (just as an example not that I want to recommend that)

Now you will get a duplicated dependency to A instead of two dependencies to A+AS


Another case:

1. Tycho caches several files in the local maven repro (in `.cache`) 
2. So any file that is located there will be assumed to be a maven artifact
3. example: `.m2/repository/.cache/tycho/org.eclipse.pde.ds1_2.lib-1.0.100.v20180228-1600.jar/annotations.jar` will be considered as G=.cache A=tycho, V=org.eclipse.pde.ds1_2.lib-1.0.100.v20180228-1600.jar

that's why I thing at least some basic sanity checking is required here.

mickaelistria: > Now you will get a duplicated dependency to A instead of two dependencies to A+AS

Same GAV but different system path. However, I agree and will try to include support for classifier

> example: .m2/repository/.cache/tycho/org.eclipse.pde.ds1_2.lib-1.0.100.v20180228-1600.jar/annotations.jar will be considered as G=.cache A=tycho, V=org.eclipse.pde.ds1_2.lib-1.0.100.v20180228-1600.jar

That's a good example, I didn't realize that. Thanks. I'll try to improve that.

laeubi: > Same GAV but different system path. However, I agree and will try to include support for classifier

I don't know if this is actually valid in maven. GAV is always considered to be unique and required as far as I can see.

mickaelistria: They're also supposed to be existing and to map to repo artifacts; and yet with Tycho the GAV meant nothing to Maven. So it's not really worse ;)

PR #483, opened 2022-01-03
Fix #482 - add a system option to enter debug mode for itests

### LINKED ISSUE #482: Provide a commandline option to attach a debugger for the integration tests
Currently its a bit strange as one needs to modify the code, recompile tycho and then reverse this, we should use a system property instead for this.

laeubi:


mickaelistria:
how is the `tycho.mvnDebug` property passed from CLI to the tests?

laeubi:
Example call:

```
mvn clean install 
  -Pits
  -Dtest=<...>
  -Dtycho.mvnDebug
```

mickaelistria:
So integration tests do inherit parent Maven process parameters as system properties?

laeubi:
> So integration tests do inherit parent Maven process parameters as system properties?

At least for our integration test build.

mickaelistria:
OK, I wasn't aware of it, that's a good thing it's working that way, and a good addition you're making. Please merge!

Issue #443, opened 2021-12-16
Use regular Maven coordinates -when possible- for dependencies


mickaelistria:
Currently, running `mvn dependency:list` for a Tycho module lists all dependencies under the form `p2-installable-unit:org.domain.pluginId:version` (eg `p2-installable-unit:org.mockito.mockito-core:4.1.0`). The scope is usually `system` and the absolute path to file is set.
In case where the dependency is known to have Maven coordinates, it would be better to reuse the coordinates of this dependency to showcase it's the "official" artifact from Maven repo. So in case of Mockito, it would become `org.mockito:mockito-core:4.1.0`. p2 metadata are expected to contain the necessary info to map a p2 artifact/unit back to the Maven coordinates they're generated from, those data should then be used in `P2ResolverImpl` when calling `result.addArtifact` to not hardcode usage of `ArtifactType.TYPE_INSTALLABLE_UNIT` and let Maven coordinates be used in place.
I expect that as long as the scope remains system and path is hardcoded, this shouldn't be too disruptive.

akurtakov:
It's closely related to https://github.com/eclipse/dash-licenses/issues/125 and this might fix the dash-licenses bug even.

mickaelistria:
One hotspot would be `MavenDependencyInject#injectMavenDependencies()`

laeubi:
Just one thing that came into my mind, given we have a repository of type p2 specified, it should be possible to even resolve `p2-installable-unit:<unit>:<version>` in a regular maven build using a core extension in `.mvn` folder. That would then download the unit from the repository and set the appropriate path before the actual build start.

mickaelistria:
Some incremental work can be done to ease implementation of this feature:
* Remove `P2ResolutionResults.Entry` and use `ArtifactDescriptor` instead in `P2ResolutionResult`
* Add a `getMavenCoordinates()` method, returning a `GAV` to `ArtifactDescriptor`
* Make `P2DependencyResolver` set the Maven coordinates for given `ArtifactDescriptor` according to p2 metadata
* Make `MavenDependencyInject#injectMavenDependencies()` consume the GAV when available to populate dependencies instead of using the ArtifactKey (which is p2 more than Maven).

Issue #433, opened 2021-12-14
Run plugins before Maven-Tycho build


LuluDavid:
Hi, I am creating an issue because I am working currently with maven-tycho-plugin for a bunch of OSGI bundles, and I have a small problem to include preliminary instructions in my maven build before tycho's dependency resolution.  
The problem is : 
- Tycho (or potentially plexus) ties its resolution to the validate phase (as described in [this file](https://github.com/eclipse/tycho/blob/master/tycho-maven-plugin/src/main/resources/META-INF/plexus/components.xml), which is the first executed phase
- I am packaging my bundles as eclipse-plugin (so the resolution is done for all of them)
- All further plugins executions tied to the validate phase are executed afterwards (in accordance to Maven's lifecycle description)  

So in my case, as I want to download a jar that is referenced locally in one of my bundle's classpath before Tycho's dependency resolution, it does not work for the first build (but works for the second one as the jar is available then).  

I found [this issue](https://tycho-user.eclipse.narkive.com/9eUeEcjk/running-maven-plugin-before-tycho-packaging-plugin-0-25-0-build-qualifier-default-build-qualifier) that references a similar problem, but the concerned pom is packaged as pom and I tried overriding tycho-maven-plugin's executions to tie all validate-related ids (org.eclipse.tycho:tycho-packaging-plugin:${project.version}:build-qualifier,org.eclipse.tycho:tycho-packaging-plugin:${project.version}:validate-id, org.eclipse.tycho:tycho-packaging-plugin:${project.version}:validate-version) to the initialize phase instead : 
```xml
<pluginManagement>
      ...
      <plugin>
	      <groupId>org.eclipse.tycho</groupId>
	      <artifactId>tycho-maven-plugin</artifactId>
	      <version>${tycho.version}</version>
	      <extensions>true</extensions>
	      <executions>
		      <execution>
			      <id>org.eclipse.tycho:tycho-packaging-plugin:${project.version}:build-qualifier</id>
			      <phase>initialize</phase>
		      </execution>
		      <execution>
			      <id>org.eclipse.tycho:tycho-packaging-plugin:${project.version}:validate-id</id>
			      <phase>initialize</phase>
		      </execution>
		      <execution>
			      <id>org.eclipse.tycho:tycho-packaging-plugin:${project.version}:validate-version</id>
			      <phase>initialize</phase>
		      </execution>
	      </executions>
      </plugin>
</pluginManagement>
```
Without further success.  

[...]

PR #485, opened 2022-01-03
#443 - publish maven GAVC info into the P2 meta-data

laeubi:
This publishes the info of the original artifact into the P2 metdata.

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/impl/publisher/MavenPropertiesAdvice.java
+++ b/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/impl/publisher/MavenPropertiesAdvice.java
@@ -41,6 +41,14 @@ public class MavenPropertiesAdvice implements IPropertyAdvice {
         }
     }
 
+    public MavenPropertiesAdvice(String groupId, String artifactId, String version, String classifier,
+            String extension) {
+        this(groupId, artifactId, version, classifier);
+        if (extension != null && !extension.isEmpty()) {
+            properties.put(RepositoryLayoutHelper.PROP_EXTENSION, extension);
+        }
+    }
+
     @Override
     public Map<String, String> getArtifactProperties(IInstallableUnit iu, IArtifactDescriptor descriptor) {
         // workaround Bug 539672
--- a/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/resolver/MavenTargetDefinitionContent.java
+++ b/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/resolver/MavenTargetDefinitionContent.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020, 2021 Christoph Läubrich and others.
+ * Copyright (c) 2020, 2022 Christoph Läubrich and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -57,6 +57,7 @@ import org.eclipse.tycho.core.shared.DependencyResolutionException;
 import org.eclipse.tycho.core.shared.MavenDependenciesResolver;
 import org.eclipse.tycho.core.shared.MavenLogger;
 import org.eclipse.tycho.core.shared.MavenModelFacade;
+import org.eclipse.tycho.p2.impl.publisher.MavenPropertiesAdvice;
 import org.eclipse.tycho.p2.metadata.IArtifactFacade;
 import org.eclipse.tycho.p2.repository.GAV;
 import org.eclipse.tycho.p2.target.TargetDefinitionContent;
@@ -218,11 +219,14 @@ public class MavenTargetDefinitionContent implements TargetDefinitionContent {
                                     logger.debug("The following manifest was generated for this artifact:\r\n"
                                             + wrappedArtifact.getGeneratedManifest());
                                 }
-                                unit = publish(BundlesAction.createBundleDescription(tempFile), tempFile);
+                                unit = publish(BundlesAction.createBundleDescription(tempFile), tempFile,
+                                        /*
+                                         * do not propagate maven artifacts info for wrapped bundles
+                                         */ null);
                                 symbolicName = wrappedArtifact.getWrappedBsn();
                                 bundleVersion = wrappedArtifact.getWrappedVersion();
                             } else {
-                                unit = publish(bundleDescription, bundleLocation);
+                                unit = publish(bundleDescription, bundleLocation, mavenArtifact);
                             }
                             bundles.add(unit);
                             if (logger.isDebugEnabled()) {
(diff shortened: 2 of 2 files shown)


mickaelistria:
I don't think this is enough to cover #443, the properties would be set, but the MavenDependencyInjector wouldn't consume them, would it?

laeubi:
> but the MavenDependencyInjector wouldn't consume them

If it currently does not do this it should do that. At least that are seems to be the way how it is published inside the repository. I don't have a test-case at hand, but at least it should then show up in a published P2-Update-site...

laeubi:
I have put a breakpoint at `org.eclipse.tycho.core.maven.MavenDependencyInjector.addDependency(ArtifactDescriptor, String)` and can see that the IU has the neccesary properties at this point, lets see if we can made it available there....

mickaelistria:
See #473 for a test case of what's expected here.

laeubi:
@mickaelistria do you think we can merge this as a first valuable step? I could take a look at MavenDependencyInjector afterwards.

mickaelistria:
I don't think it's worth merging it if there is no consumer for it.

laeubi:
e.g the` remap-artifacts-to-m2-repo` mojo already consumes this information.

mickaelistria:
> e.g the remap-artifacts-to-m2-repo mojo already consumes this information.

It consumes it from artifact repository, and I believe the tweak you're adding here only affects the metadata repository. Artifacts are already holding information about Maven coordinates, I'm not sure (really not sure, not against) that adding Maven coordinates on installable units/metadata is a good idea as 1 unit can lead to different artifacts.

laeubi:
Sorry for confusion, you are right. Anyways there is one IU per bundle artifact and tycho already stores this information there, just copied from an arbitrary site generated by tycho:

```
 <unit id='org.eclipse.equinox.device' version='1.1.100.v20210212-1143' singleton='false' generation='2'>
      <update id='org.eclipse.equinox.device' range='[0.0.0,1.1.100.v20210212-1143)' severity='0'/>
      <properties size='8'>
        <property name='df_LT.bundleVendor' value='Eclipse.org - Equinox'/>
        <property name='df_LT.bundleName' value='Device Access Service'/>
        <property name='org.eclipse.equinox.p2.name' value='%bundleName'/>
        <property name='org.eclipse.equinox.p2.provider' value='%bundleVendor'/>
        <property name='org.eclipse.equinox.p2.bundle.localization' value='plugin'/>
        <property name='maven-groupId' value='org.eclipse.equinox'/>
        <property name='maven-artifactId' value='org.eclipse.equinox.device'/>
        <property name='maven-version' value='1.1.100-SNAPSHOT'/>
      </properties>
      <provides size='6'>
        <provided namespace='org.eclipse.equinox.p2.iu' name='org.eclipse.equinox.device' version='1.1.100.v20210212-1143'/>
        <provided namespace='osgi.bundle' name='org.eclipse.equinox.device' version='1.1.100.v20210212-1143'/>
        <provided namespace='java.package' name='org.eclipse.equinox.device' version='0.0.0'/>
        <provided namespace='osgi.identity' name='org.eclipse.equinox.device' version='1.1.100.v20210212-1143'>
          <properties size='1'>
            <property name='type' value='osgi.bundle'/>
          </properties>
        </provided>
        <provided namespace='org.eclipse.equinox.p2.eclipse.type' name='bundle' version='1.0.0'/>
        <provided namespace='org.eclipse.equinox.p2.localization' name='df_LT' version='1.0.0'/>
      </provides>
      <requires size='6'>
        <required namespace='java.package' name='org.osgi.framework' range='1.2.0'/>
        <required namespace='java.package' name='org.osgi.service.device' range='[1.1.0,1.2.0)'/>
        <required namespace='java.package' name='org.osgi.service.log' range='1.2.0'/>
        <required namespace='java.package' name='org.osgi.util.tracker' range='0.0.0'/>
        <required namespace='java.package' name='org.eclipse.osgi.util' range='0.0.0'/>
        <requiredProperties namespace='osgi.ee' match='(&amp;(osgi.ee=JavaSE)(version=1.8))'>
          <description>
            org.eclipse.equinox.device
          </description>
        </requiredProperties>
      </requires>
      <artifacts size='1'>
        <artifact classifier='osgi.bundle' id='org.eclipse.equinox.device' version='1.1.100.v20210212-1143'/>
      </artifacts>
      <touchpoint id='org.eclipse.equinox.p2.osgi' version='1.0.0'/>
      <touchpointData size='1'>
        <instructions size='1'>
          <instruction key='manifest'>
            Bundle-SymbolicName: org.eclipse.equinox.device&#xA;Bundle-Version: 1.1.100.v20210212-1143
          </instruction>
        </instructions>
      </touchpointData>
    </unit>
```

or take a look at `P2GeneratorImpl.getPublisherAdvice(IArtifactFacade, PublisherOptions)` so this aproach is quite "standard" tycho and probably this information is later on copied to the artifacts.xml as well....

laeubi:
@mickaelistria can we merge this given that obviously it is already the case in tycho for other parts to publish that information to the IU?

mickaelistria:
we have no test nor use-case for it so far. So I'd rather adding code and behavior to maintain without value to balance it.

laeubi:
org.eclipse.tycho.p2.publisher.P2GeneratorImplTest explicitly test this behavior for the generator so there is/was a use that was strong enough to put this into a central part of Tycho...

laeubi:
Also see org.eclipse.tycho.repository.testutil.ArtifactPropertiesMatchers.containsGAV(String, String, String, String)

mickaelistria:
If the test is there and is already green, what does this change bring compared to current state?
If it allows to remove some other pieces of code, or performs better, solves some bug, or whatever substantial benefit, good; but here I just see code addition and am still struggling to find out what's the gain.

laeubi:
> good; but here I just see code addition and am still struggling to find out what's the gain.

It simply aligns with how Tycho behave filling a gap currently not discovered by any test. I'm just a bit curious your change about PGP was also added without any additional test or "gain" in tycho, so how is this different?
So following your argumentation it should haven't been added at all (it even introduce new public configuration properties without any documentation)...

mickaelistria:
PGP support is a gain for Tycho because there is/was a concrete need from consumers to allow adding PGP signatures to artifact metadata. Although there is no test, there was a clear request and use-case to cover, and this was adding a clearly identified necessary feature for consumers.
Here I do not see what problem it does resolve, or what actual requirement does it fulfill, it seems more like the reason to do it is "because we can", but if it is so, it's not sufficient to balance the maintenance effort.
Please convince me this change is valuable and generate value for a real use-case and I'll happily merge it. What is the "gap"? What problem does it cause? What can't we do without this change that we'll now be able to do with it?...

mickaelistria:
Can you please just merge this with #487 so they can be reviewed together? I'm loosing track and having both patches in the same PR would allow to highlight the value and facilitate review.

laeubi:
> What is the "gap"?

I don't know why it was added at the first place, but if other tycho components do add this information when they work with maven items  its seem valid to do it here also. And it adds value to the story for #433 so we better can discover the origin of an artifact.

> What problem does it cause?

Artifacts for maven target locations might be seen in the P2 metadata as if the where actually locally build units and not derived from an maven artifact (e.g. like we already do with pom=consider artifacts where this meta-data is already generated the same way as here)

> What can't we do without this change that we'll now be able to do with it?...

I have already shown in #487 how it could be used to reliable recover maven GAV without fiddling around with the storage location of an artifact (e.g. if we consume one from an update-site that does *not* store the information on the *right* path so we can assume it is/was a maven artifact).

> Can you please just merge this with #487 so they can be reviewed together? 

But as this is not limited to Maven Target Location artifacts but *any* IU (e.g. even ones already published in P2 repositories) I think these are separate changes.

PR #501, opened 2022-01-04
ResolutionArguments#hashcode is not stable

### LINKED ISSUE #496: ResolutionArguments#hashcode is not stable
In parallel builds ResolutionArguments#hasCode returns different results over time this leads to the problem that the target is computed multiple times concurrently see:

- #493 
- #494

laeubi:
This fixes #496 and adds some toString/Output to ease debugging such problems.


mickaelistria:
No objection here. I think we can suppose that executionEnvironments are unique just by id actually, since they're often stored in a map.

laeubi:
Cherry-Pciked to 2.6.x branch as well.

PR #504, opened 2022-01-04
Fix #502 - TargetDefinitionUtil / UpdateTargetMojo should not be allowed to modify the internal state of the target.

### LINKED ISSUE #502: TargetDefinitionUtil / UpdateTargetMojo should not be allowed to modify the internal state of the target
TargetDefinitionUtil / UpdateTargetMojo both currently access (and modify!) the internal state of a target file object to update a file.
This should be avoided under all circumstance as it complicates the code and is error prone.

laeubi:


laeubi:
@mickaelistria do you like to review this? All tests are green so I would merge this if there are no concerns.

mickaelistria:
I don't think it's risky and a very quick look at the general approach brings no concern to me. So if all tests pass, feel free to merge.

PR #506, opened 2022-01-04
Issue 484

mickaelistria:


laeubi:
@mickaelistria actually I really like to get rid of the BlackboardRepository and there is a TODO in the code about it, do have found out why the Director needs this? Is this some legacy code? Or should we add some more documentation to the BlackboardRepository stuff do document whats going on?

laeubi:
Okay maybe I have misunderstood it does the "get rid of" todo means we should get rid of blackboard key(=not use it at all), or was it meant to say its implementation was/is incomplete? I really curious about that as I have stumbled too often about this TODO....

tycho-bundles/org.eclipse.tycho.p2.resolver.impl.test/src/test/java/org/eclipse/tycho/p2/target/TestResolverFactory.java
```diff
@@ -47,6 +59,39 @@ public TestResolverFactory(MavenLogger logger) {
...
         targetDefinitionResolverService.setMavenContext(mavenContext);
+        targetDefinitionResolverService.setMavenDependenciesResolver(new MavenDependenciesResolver() {
+
+            @Override
+            public Collection<?> resolve(String groupId, String artifactId, String version, String packaging,
+                    String classifier, String dependencyScope, int depth,
+                    Collection<MavenArtifactRepositoryReference> additionalRepositories, Object session)
+                    throws DependencyResolutionException {
+                // This is supposed to mimic Maven repo returning an artifact
+                String relativePath = groupId.replace('.', File.separatorChar) + File.separator + artifactId
```
laeubi, line 70: Can we use GAV(....) here as it already has a method to construct a maven repo path?

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/target/TargetPlatformFactoryImpl.java
```diff
@@ -186,18 +182,26 @@ public P2TargetPlatform createTargetPlatform(TargetPlatformConfigurationStub tpC
...
                 filter, //
                 localMetadataRepository, //
-                createExternalArtifactProvider(completeRepositories, targetFileContent, pomDependencyArtifactRepo,
-                        tpConfiguration.getIncludePackedArtifacts()), //
+                externalArtifactFileProvider, //
                 localArtifactRepository, //
                 includeLocalMavenRepo, //
                 logger);
 
+        RepositoryBlackboardKey blackboardKey = RepositoryBlackboardKey
```
laeubi, line 199: Can we add a little description what this blackboardkey is for and what it does (beside that some test fails if it is not there...). Is it some legacy code we might remove in the future?

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/target/TargetPlatformFactoryImpl.java
```diff
@@ -307,28 +311,60 @@ private IMetadataRepository loadMetadataRepository(MavenRepositoryLocation locat
         }
     }
 
+    private static final class SortedRepositories {
```
laeubi, line 314: can you add a little comment by what mean it is sorted? I assume its local first, then remote or something?

laeubi (commented):
Looks good and if it resolves the issue of mirroring local items that's a great improvement.

mickaelistria:
> actually I really like to get rid of the BlackboardRepository and there is a TODO in the code about it

Me too, one of my favorite things.

> do have found out why the Director needs this? Is this some legacy code? Or should we add some more documentation to the BlackboardRepository stuff do document whats going on?

The blackboard repository allows to keep the project TargetPlatform artifact repositories in memory and maps it with a URL of form `file:/resolution-context-artifacts@%2Fpath%2Fto%2Fproject`, that the director mojo can then use to retrieve the dependencies. This is partly documented on `RepositoryBlackboardKey#forResolutionContextArtifacts`.

By "get rid of the todo" I mean fix the implementation as suggested (ie resolution context now contains the whole Target Platform artifacts and not only the pom-dependencies.

laeubi:
But why can't we pass the memory reference some other way? Is this a restriction of the org.eclipse.equinox.internal.p2.director.app.DirectorApplication.initializeRepositories() being private? Interestingly metadataManager is already protected so one should be able to set this directly...

Anyways we should merge this and take a look at the blackboard repository later, can you create an issue that describes why we currently need this (e.g. because of DirectorApplication restriction or something else...)

mickaelistria:
> But why can't we pass the memory reference some other way?

DirectorApplication wants a URL, so this BlackboardRepositoryKey builds an URL which resolves to an ArtifactRepositoryBlackboard containing the target platform.

tycho-bundles/org.eclipse.tycho.p2.resolver.impl.test/src/test/java/org/eclipse/tycho/p2/target/TestResolverFactory.java
mickaelistria: The only GAV accessible here doesn't have method to build a path as far as I could see. Or did I miss something>

PR #513, opened 2022-01-06
Added integration test to check maven caching of artifacts from target platform

jstroleny:
It adds integration test to verify that maven/directory artifacts from target platform are not cached/copied automatically into m2 repo (on location `.m2/repository/p2/osgi/bundle`), but that they are referenced instead, see https://github.com/eclipse/tycho/issues/124 and https://github.com/eclipse/tycho/issues/124#issuecomment-1006455052 for more details.

The test is checking that maven caching is not applied by running two consecutive maven builds, where target artifact is modified between these two builds which has effect to cause compilation error. If there would be maven caching applied for artifact, the second build will pass.

This also fixes related `testDirectoryLocation` test which had missing `@Test` annotation and thus was not executed.


laeubi:
Thanks for your contribution @jstroleny can you make sure that you signed the ECA with the email address used in your commit?

mickaelistria:
Thank you!

PR #523, opened 2022-01-09
Fix #486 - Remove LocalDependencyResolver

### LINKED ISSUE #486: Remove org.eclipse.tycho.core.osgitools.targetplatform.LocalDependencyResolver
currently there is a "p2" resolver and a "local" resolver.

The local resolver uses deprecated code, and is only used for `-Dtycho.targetPlatform` described as:

> tycho.targetPlatform is deprecated and WILL be removed in the next Tycho version.

laeubi:


tycho-core/src/test/java/org/eclipse/tycho/core/testcomponents/StubP2DependencyResolver.java
```diff
@@ -1,21 +0,0 @@
...
- *******************************************************************************/
-package org.eclipse.tycho.core.testcomponents;
-
-import org.codehaus.plexus.component.annotations.Component;
-import org.eclipse.tycho.core.DependencyResolver;
-import org.eclipse.tycho.core.osgitools.targetplatform.LocalDependencyResolver;
-
-// TODO romove me as part of TYCHO-527
-@Component(role = DependencyResolver.class, hint = "p2", instantiationStrategy = "per-lookup")
-public class StubP2DependencyResolver extends LocalDependencyResolver {
```
laeubi, line 19: @mickaelistria it seems the tests use this StubP2 resolver that actually was a LocalDependencyResolver, do you know why this was choosen instead of using the "real" p2 resolver?

mickaelistria: No idea sorry. If we can get rid of stubs without pitfalls, I'm all for it.

laeubi: I think the problem is that P2 requires tycho core and thus tycho core cant depend on P2 for test-purpose. That's a bit undesirable as we actually test something other than the P2 resolver here.

I'll check how much we can stub this or if I can just move tests around....

laeubi:
This is nastier than I thought I'll close this and try to apply changes in smaller bunches to see what actually is breaking and if we can get rid of some quirks iteratively...

Issue #519, opened 2022-01-07
Provide better feedback to the user about the cause of a failed resoloution


laeubi:
Issue https://github.com/eclipse/tycho/issues/517 has outlined a problem that we can't expect Tycho users to be OSGi-Experts (or P2 error message experts).

Currently we output only the plain problem description P2 give, while this is suitable for in-deepth analysis it is often not so obvious  what is the _solution_ for this.

P2 resolver returns so called `Explanation` objects that can give a more detailed understanding whats actually is missing.
We should try to enhance tycho to inspect those and give better user advice.

mickaelistria:
>  we can't expect Tycho users to be OSGi-Experts (or P2 error message experts).

At some point, Tycho users are plugin developers and will need to understand p2 and OSGi ;) Instead of tweaking Tycho to override/improve those messages and taking more responsibility, we should consider either improving them in p2 directly, or building in Eclipse help a page explaining how to process those p2 messages and let Tycho show the link when an error happens.

laeubi:
> At some point, Tycho users are plugin developers and will need to understand p2 and OSGi

Sure, but if we can instead of print just a big list of dependency chain and one needs to inspect the error message something like:

```
[ERROR] Project requires Java-SE17 but your current target is Java-SE11
[ERROR] Bundle XYZ requires package a.b.c but it is not found in the target nor the reactor
<maybe even print a list of available packages on debug level>
>> print the usual list here <<
```

would certainly give better tooling experience (a bit more similar to what PDE offers you in the UI) and make people consider tycho even more useful.

> Eclipse help a page

The error message already mentions https://wiki.eclipse.org/Tycho/Dependency_Resolution_Troubleshooting but that seem not to help people much to find the cause fast enough.

mickaelistria:
To be honest, I'm not convinced that people will be more comfortable if we change error messages. I just think it's natural laziness of developers to not try to process error messages of different formats than the ones they're used to. So it's a vain fight IMO, and just redirecting to this page is enough.
People who use Tycho need to get familiar with p2 anyway, IMO it's fine to have them struggling when reading their 1st error message, they learn from it and are more efficient later, and not only with Tycho but with everything p2-based in general.

laeubi:
> To be honest, I'm not convinced that people will be more comfortable if we change error messages.

I just would add additional messages. P2 messages are often too verbose, whats good for detailed analysis, and too technical to really allow developers to get a fast feedback whats wrong.

[...]

PR #528, opened 2022-01-10
#519 Provide better feedback to the user about the cause of a failed resolution

laeubi:
This is just a showcase for the EE case, example output is:

```
[INFO] Resolving dependencies of MavenProject: tycho-its-project.compiler.jdt.bree.java8:c.j.b.j.bundle:1.0.0 @ tycho-its/projects/compiler.missing.bree/c.j.b.j.bundle/pom.xml
[ERROR] c.j.b.j.bundle 1.0.0 requires Execution Environment that matches (&(osgi.ee=JavaSE)(version=17)) but the current resolution context uses [a.jre.javase 11.0.0]
[INFO] {osgi.os=linux, osgi.ws=gtk, org.eclipse.update.install.features=true, osgi.arch=x86_64}
[ERROR]   Software being installed: c.j.b.j.bundle 1.0.0
[ERROR]   Missing requirement: c.j.b.j.bundle 1.0.0 requires 'osgi.ee; (&(osgi.ee=JavaSE)(version=17))' but it could not be found
[ERROR] 
[ERROR] See https://wiki.eclipse.org/Tycho/Dependency_Resolution_Troubleshooting for help.
[ERROR] Cannot resolve dependencies of MavenProject: tycho-its-project.compiler.jdt.bree.java8:c.j.b.j.bundle:1.0.0 @ projects/compiler.missing.bree/c.j.b.j.bundle/pom.xml: See log for details -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MavenExecutionException
```

As opposed to the plain P2 message: 
> requires 'osgi.ee; (&(osgi.ee=JavaSE)(version=17))' but it could not be found

Tycho now can more specifically tell the user what actually is in use:
> requires Execution Environment that matches (&(osgi.ee=JavaSE)(version=17)) but the current resolution context uses [a.jre.javase 11.0.0]

I also think that enhancing such a message in P2 itself would be hard to archive as it not only requires to examine the missing requirement but also what EE where provided and/or are available.


mickaelistria:
What happens if one set executionEnvironment=none to enforce usage of JustJ, and tries to build a product that requires Java 17 while only providing JustJ 16?

laeubi:
> What happens if one set executionEnvironment=none to enforce usage of JustJ, and tries to build a product that requires Java 17 while only providing JustJ 16?

I have no idea, is there an example project that demonstrate this use-case?

mickaelistria:
>  I have no idea, is there an example project that demonstrate this use-case?

There is a justj test case in tycho-its. It can maybe be tweaked it to make the issue happen.

laeubi:
Maybe just focus on the obvious cases first?
I can think about several ways to further improve this, some include enhancing P2 but I do not want to put too much effort into this if it is considered not useful and thus just give some examples. If we have agreed on the general aspects we could go into more details I think.

mickaelistria:
> Maybe just focus on the obvious cases first?

I just want to ensure those real use-cases (eg EPP uses it) would not lead to a worse result (eg a NPE instead of a usable error message).

laeubi:
> > Maybe just focus on the obvious cases first?
> 
> I just want to ensure those real use-cases (eg EPP uses it) would not lead to a worse result (eg a NPE instead of a usable error message).

The code does not harm there (resolution simply succeeds) but we get strange error messages in this case from the mojos, I have added a test case here and will try to improve that case.

mickaelistria:
No objection, thank you for testing with JustJ. Feel free to merge.

PR #530, opened 2022-01-11
Fixes #476 - drop "fixSWT" workaround

### LINKED ISSUE #476: drop "fixSWT"
We should simply drop this as the depending bug has be resolved for more than 5 years:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=381167

laeubi:


laeubi:
@akosyakov there is one test P2ResolverTest.testBundleUsesSWT that seems to changes in behavior without the SWT fix, is this to be expected? Should we remove the test?

akurtakov:
I have pushed an extra change via #531 . I don't think I can push to this one, can I?

laeubi:
I have allowed "maintainer access" to that branch so you should be able to commit directly to it:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork

PR #532, opened 2022-01-11
Apply changes auto done by pde to target files.

akurtakov:


akurtakov:
@laeubi as pde changes these I thought it's good idea to push the changes but tests are failing. What should we do here?

laeubi:
I would not change existing integration tests as these also test legacy behavior for backward compatibilty to old format.
Also if you update make sure to use the latest m2e-snapshot as some never features are not present in the current release.

akurtakov:
> I would not change existing integration tests as these also test legacy behavior for backward compatibilty to old format. Also if you update make sure to use the latest m2e-snapshot as some never features are not present in the current release.

Problem with keeping them is that if one just opens the target platform page these files are changed and one has to manually revert changes in them.

laeubi:
I know, but it should not be necessary to open them at all I think. Don't know if we can mark them to ignore changes with some git property?

laeubi:
BTW: the changes in `tycho-bundles-target.target` should be fine and the only one to work with on a regular basis.

tycho-its/projects/target.maven.autofeature/test.target
```diff
@@ -17,10 +17,6 @@
 			</dependencies>
 		</location>
 		<location includeSource="true" missingManifest="generate" type="Maven">
-
```
laeubi, line 20: This change seems to indicate that not the latest m2e-snapshot is used here and what makes the test fail

tycho-bundles/org.eclipse.tycho.p2.resolver.impl.test/resources/targetresolver/mavenDep.target
```diff
@@ -2,7 +2,7 @@
 <?pde version="3.8"?>
 <target name="mavenDep.targe">
 	<locations>
-		<location includeDependencyDepth="none" includeDependencyScope="compile" includeSource="false" missingManifest="ignore" type="Maven">
+		<location includeDependencyScope="compile" missingManifest="ignore" type="Maven">
```
laeubi, line 5: This removes the latest changes from m2e so maybe another reason why the test fails.

laeubi (changes requested):
I would propose that we only change the tycho-bundles.target file and leave the other ones alone.
These other targets are essentially test-resources and should never be touched/modified.

akurtakov:
OK, I'm giving up on this one as I clearly haven't followed deep enough and don't have latest snapshots of everything.

PR #533, opened 2022-01-11
Throw exception instead of warning for tycho.targetPlatform

laeubi:
This throws an error instead of print a warning and shows that some tests in Tycho still depend on the deprecated property.

I tried to move the stuff to test see https://github.com/laeubi/tycho/tree/486_move_to_test and poked around a bit with the test but can't get it to work. If you have an idea how to fix certain tests that would be appreciated.

Just keep in mind that `MavenDependencyCollector` and `EclipseInstallationLayout` also are eligible for removal as they are only required by the Local resolver, so any tests that fail for that classes could safely be ignored...

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/resolver/DefaultDependencyResolverFactory.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/resolver/DefaultDependencyResolverFactory.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2011 Sonatype Inc. and others.
+ * Copyright (c) 2008, 2022 Sonatype Inc. and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -7,11 +7,11 @@
  *
  * Contributors:
  *    Sonatype Inc. - initial API and implementation
+ *    Christoph Läubrich - #486 - Remove org.eclipse.tycho.core.osgitools.targetplatform.LocalDependencyResolver  
  *******************************************************************************/
 package org.eclipse.tycho.core.resolver;
 
-import java.io.File;
-import java.io.IOException;
+import java.util.Objects;
 import java.util.Properties;
 
 import org.apache.maven.project.MavenProject;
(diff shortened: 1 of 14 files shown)


laeubi:
The integration test TYCHO418pomDependencyConsider also uses the old aproach.

laeubi:
I have now:

1. converted the warning into an error so that any public calls get an error
2. deprecated dependent code and method
3. adjusted the code so that existing test can still use the local resolver, we can fix them step by step and it is up to us to get this done, afterwards it should be safe to remove it

If the build succeeds I plan to merge this as a first increment on the way to get rid of the local resolver. 

First step would be to fix all tests outside of tycho-core, it might be suitable to replace it with a `file://` repository reference. If that's done, the code should be able to merge everything into the test scope and we can further investigate how to get rid of the StubResolver...

mickaelistria:
Can you please update PR and commits (maybe by squashing them) to describe in which case is the warning replaced by the exception? Seems like it's related to some property; which one? I'm missing context here.

laeubi:
> Can you please update PR and commits (maybe by squashing them) to describe in which case is the warning replaced by the exception? Seems like it's related to some property; which one? I'm missing context here.

You can just take a look at the very first commit https://github.com/eclipse/tycho/pull/533/commits/ec19ed7eccc470dc48ba635d84dadaf558546e1f as this is the essential user-visible change. All others are just adjustments to Tycho test code as it currently still relies on this feature.

tycho-core/src/main/java/org/eclipse/tycho/core/resolver/DefaultDependencyResolverFactory.java
```diff
@@ -43,42 +42,25 @@ public DependencyResolver lookupDependencyResolver(MavenProject project) {
...
-        if (property != null) {
-            logger.warn("-Dtycho.targetPlatform is deprecated and WILL be removed in the next Tycho version.");
-
-            File location = new File(property);
-            if (!location.exists() || !location.isDirectory()) {
-                throw new RuntimeException("Invalid target platform location: " + property);
-            }
-
+        if (properties.getProperty("tycho.targetPlatform") != null) {
+            throw new RuntimeException("-Dtycho.targetPlatform was deprecated and IS REMOVED in this tycho version.");
```
laeubi, line 46: @mickaelistria this is the actual interesting, user-visible change.

laeubi:
All test are passing now so this is ready to merge,

mickaelistria:
> You can just ...

With the amount of code/bugs/reviews I'm reading daily, those "you can just" are much more expensive than what you seem to estimate. More explicit message on commits and PRs really make things easier for reviews.

laeubi:
> > You can just ...
> 
> With the amount of code/bugs/reviews I'm reading daily, those "you can just" are much more expensive than what you seem to estimate. More explicit message on commits and PRs really make things easier for reviews.

I have already merged the commits and have added a comment for the key part of this change.

mickaelistria:
Thanks

PR #555, opened 2022-01-22
Provide tycho specific org.apache.maven.graph.GraphBuilder for project dependency resolution 

laeubi:
This is a WIP to evaluate the use of the maven `GraphBuilder` to compute tycho project dependencies.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -493,6 +493,7 @@
 
 		<module>tycho-gpg-plugin</module>
 		<module>tycho-extras</module>
+		<module>tycho-build</module>
 	</modules>
 
 	<profiles>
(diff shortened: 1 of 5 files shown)


laeubi:
@sratz I just added a small component to get familiar with the `GraphBuilder`. I compiled it and run a regular maven build, but the component is never called (only TychoMavenLifecycleParticipant is called).

Tycho is marked as an extension in the `pom.xml`:
```
<plugin>
	<artifactId>tycho-maven-plugin</artifactId>
	<extensions>true</extensions>
</plugin>
```

I assume this feature requires tycho to be used as an [core-extension](https://maven.apache.org/docs/3.3.1/release-notes.html#Core_Extensions), or did I miss something?

laeubi:
BTW I was able to use this as a core-extension already, just wanted to make sure that a core-extension is the only option here.

laeubi:
I'll close the PR for now as much more things need to be resolved before this could effectively be reviewed/discussed.

PR #559, opened 2022-01-23
Issue 557 do not use LegacySupport

laeubi:


laeubi:
@mickaelistria any concerns? the interesting change is in https://github.com/eclipse/tycho/pull/559/files#diff-967225f1c2e7946d98e4d10a403f37d605099cdeb2139af5ca6c7850ff5ab077R38 where we could get rid of the LegacySupport the rest is more or less some refactoring and adjustments so the code is on the right places.

mickaelistria:
No concern, the change look beneficial and code cleaner with it and all tests are fine, so feel free to merge it if you're confident enough it's safe to do so.

Issue #556, opened 2022-01-22
Make tycho aviable as a core-extension


laeubi:
Currently one has to enable Tycho in the following way:

```
<plugin>
  <groupId>org.eclipse.tycho</groupId>
  <artifactId>tycho-maven-plugin</artifactId>
  <version>${tycho.version}</version>
  <extensions>true</extensions>
</plugin>
```

additionally one can also enable tycho-pomless in `.mvn/extensions.xml`

```
<extension>
  <groupId>org.eclipse.tycho.extras</groupId>
  <artifactId>tycho-pomless</artifactId>
  <version>2.7.0-SNAPSHOT</version>
</extension>
```

With the new feature of mixed reactor builds, and in the light of #551 and #48 it seems more profitable to enable tycho as a core-extension with the following properties:

- it is enough to add in  `.mvn/extensions.xml` to use tycho packaging types and resoloution
```
<extension>
  <groupId>org.eclipse.tycho</groupId>
  <artifactId>tycho-build</artifactId>
  <version>2.7.0-SNAPSHOT</version>
</extension>
```
- tycho-pomless is implicitly included (it is already possible to mix pomless+pom defined projects).

There is currently just one restriction the core-extensions do not allow to provide the version as a system-property, we might want to enhance maven in a way that one can use it that way.

This will allow tycho to participate even more close with maven as before and ee might even make this the only option for 3.0.0 to use tycho as a core extension. Individual mojos for the build itself can still be configured through the pom.xml

akurtakov:
My issue with extension usage is that I never found how to use SNAPSHOT if not previously available in local m2 repo. For that reason in Eclipse Platform we build using 2.7.0-SNAPSHOT but tycho-pomless is set to last release (2.6.0) as there is no way to add plugin repository for extension.

mickaelistria:
Wouldn't it be possible and simpler to have pomless generating pom the other way round, ie adding the tycho extensions to generated pom files instead?

laeubi:
> My issue with extension usage is that I never found how to use SNAPSHOT if not previously available in local m2 repo. For that reason in Eclipse Platform we build using 2.7.0-SNAPSHOT but tycho-pomless is set to last release (2.6.0) as there is no way to add plugin repository for extension.

For testing tycho-snapshots you need to specify this in `settings.xml` this is described here: https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots

[...]

Issue #148, opened 2021-06-09
move tycho-pomless from 'extras' to 'core' and drop 'tycho-pomgenerator:generate-poms' mojo


laeubi:
tycho-pomless has become a fully fledged solution to build pde projects and thus I think its time to remove it's 'extra' nature.

This won't harm much, as the only thing required when upgrading from a previous version is to adjust the version **and** group id once. 
On the other side it would be good to remove the `tycho-pomgenerator:generate-poms` mojo as it is not actively developed/maintained see the [recent news-group discussion](https://www.eclipse.org/lists/tycho-user/msg09099.html). Instead we should suggest to use a pomless-build with `-Dpolyglot.dump.pom=pom.xml` if one needs to generate `pom.xml` files. That way we only need to maintain tycho-pomless what gives much more benefits on its own beside generating pom.xml files.

laeubi:
It might be better to integrate pomless in a generic build extension as described in 
- #556

akurtakov:
With tycho-build extension added in 2.7  and tycho-pomgenerator:generate-poms removed I consider this one done.

PR #561, opened 2022-01-24
Add generic tycho-build core extension

laeubi:
- Issue #556 add a tycho-core-extension including pomless
- Issue #148 add a hint to the user to migrate to the generic extension
- Update integration tests to use the new tycho-core-extension

This does not change anything deep inside but prepares the move of tycho-pomless into tycho-core and will allow us to add maven-core extensions (of whatever kind) in a generic way.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,33 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## 2.7.0 (under development)
 
+### Tycho-Pomless will become a tycho-core extension
+
+Tycho pomless has started as a small experiment in tycho-extras. Over time it has grown to a fully-fledged solution to build pde-based artifacts with less effort and nearly zero additional configuration.
+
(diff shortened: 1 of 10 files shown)


laeubi:
@akurtakov strange enough the pomless itest now fail but I can execute them locally with an local snapshot, do you now if I need any special setup for the build to make it use the local build version? Or do we use an older/newer maven version on the jenkins?

laeubi:
The eror on the buildserver is:

> Plugin org.eclipse.tycho:tycho-build:2.7.0-SNAPSHOT or one of its dependencies could not be resolved: Could not find artifact org.eclipse.tycho:tycho-build:jar:2.7.0-SNAPSHOT

laeubi:
Okay, it was because the tycho-extras its are executed "to early" I moved them down and now everything works. If there are no concern I'd like to merge this.

mickaelistria:
No concern, feel free to merge at your own convenience. Please also add a note about it to RELEASE_NOTES.

laeubi:
> Please also add a note about it to RELEASE_NOTES.

done, I'll wait for the build and then merge this then.

PR #569, opened 2022-01-26
Generate additional Maven medata for OSGi bundles.

pzi-dsa:
In a [discussion](https://github.com/eclipse/tycho/discussions/435) last month, I suggested enhancing the metadata generation for Eclipse bundles, to better facilitate the Maven integration. 
More specifically, when deploying/fetching OSGi bundles from/to a Maven repository, the artifact dependencies should be preserved as much as possible.

For this, I have created a new Mojo in the _tycho-pomgenerator-plugin_. When enabled, the _generate-dependency-poms_ goal, creates pom.xml in the _target/generated-pom/_ directory- of the current reactor project, which is then included in the _META-INF/maven/..._ folder of the packaged jar.

This file contains all (reactor) dependencies of the artifact, which are required during runtime. For illustration, I have updated my 
[example project](https://github.com/pzi-dsa/tycho.example) to use this new functionality. E.g. a `mvn clean verify` now generates the following pom.xml for the _example.feature_ project:

```
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <groupId>example</groupId>
  <artifactId>example.feature</artifactId>
  <version>1.0.0-SNAPSHOT</version>
  <packaging>eclipse-feature</packaging>
  <dependencies>
    <dependency>
      <groupId>example</groupId>
      <artifactId>example.plugin2</artifactId>
      <version>1.0.0-SNAPSHOT</version>
      <type>eclipse-plugin</type>
    </dependency>
  </dependencies>
</project>
```

When deploying this artifact to a repository, the feature can then be added to the target platform (or as a Maven dependency in a pom) and will now make all of its dependent artifact available to the application as well, whereas previously, one would be required to add each individually.
In the example, _example.feature_ has a direct dependency on _example.plugin1_ and an indirect dependency on _example.plugin2_. Both of which are added to the classpath.

image

**Open Topics**

- It is currently not possible to consider dependencies that are not part of the current reactor build (e.g. artifacts from the target platform), because they do not contain a group and artifact id. Even if those dependencies are contributed using the m2e integration, this information is lost when converting them to OSGi bundles.
- I'm not satisfied with the name _generate-dependency-poms_ as it sounds too similar to the already existing _generate-poms_ goal. But I have yet to come up with a better one.


laeubi:
@pzi-dsa great that you started contributing stuff, please sign the ECA so we are able to accept your contribution.

Some notes on this topic: 

1. tycho-pomgenerator-plugin is really not used (or should not be used!) anymore, but its fine to place the code there we can later move it
2. There are recently some changes that should recover the maven infos from maven target locations so you might want to to retry that as well.

In general I think this should be a default feature and this data is always generated. I think there is even an option in Tycho to switch this on/off see : https://www.eclipse.org/tycho/sitedocs/tycho-packaging-plugin/package-plugin-mojo.html#archive

so maybe we can simply hook into this?

laeubi:
By the way it would be cool to include your example project as in [integration test in tycho](https://github.com/eclipse/tycho/wiki#providing-an-integration-test), that often is much more obvious on how a new feature behaves compared to the old state.

laeubi:
@pzi-dsa I think this code is better placed inside the `org.eclipse.tycho.packaging.PackagePluginMojo` and should respect the `archive.isAddMavenDescriptor()` setting instead of be an separate mojo.

tycho-pomgenerator-plugin/src/main/java/org/eclipse/tycho/pomgenerator/AbstractGeneratePomsMojo.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * Location of directory with template pom.xml file. pom.xml templates will be looked at this
+     * directory first, default templates will be used if template directory and the template itself
+     * does not exist.
+     * 
+     * See src/main/resources/templates for the list of supported template files.
+     */
+    @Parameter(property = "templatesDir", defaultValue = "${basedir}/pom-templates")
+    private File templatesDir;
```
laeubi, line 31: If I understand the code in org.apache.maven.archiver.MavenArchiver.createArchive(MavenSession, MavenProject, MavenArchiveConfiguration) corectly it is supposed that the pom itself is included.

What do you think instead of reading a template, simply reading the pom.xml itself and extending those?

pzi-dsa:
> I think this code is better placed inside the org.eclipse.tycho.packaging.PackagePluginMojo and should respect the archive.isAddMavenDescriptor() setting instead of be an separate mojo.

@laeubi  I was just about to suggest something similar. I'll have to do a little bit of refactoring, given that I reuse some functions of the pomgenerator-plugin, but I don't see any major issues.
Though I would prefer keeping a clear separation between the pom generation and the packaging and therefore create a separate Mojo. Perhaps something along the lines of GenerateMetadataMojo, which is executes directly before the packaging?
Especially since the metadata is generated for both plugins and features.

> There are recently some changes that should recover the maven infos from maven target locations so you might want to to retry that as well.

Great news! From what I could tell, Tycho substitutes the group id with "p2.eclipse-plugin". Is this a constant? In which case I could simply filter out all dependencies with this id.

> By the way it would be cool to include your example project as in integration test in tycho, that often is much more obvious on how a new feature behaves compared to the old state.

Sure, I can set up the integration test tomorrow.

laeubi:
> I'll have to do a little bit of refactoring, given that I reuse some functions of the pomgenerator-plugin, but I don't see any major issues.

Just copy relevant codes then, we want to get rid of pomgenerator-plugin soon, so we don't want to create new dependencies.



> Though I would prefer keeping a clear separation between the pom generation and the packaging and therefore create a separate Mojo.

It should just respect the above setting I think then.

> Tycho substitutes the group id with "p2.eclipse-plugin".

Actually you can filter anything starting with "p2." see org.eclipse.tycho.core.maven.MavenDependencyInjector.P2_GROUPID_PREFIX

mickaelistria:
>  Tycho substitutes the group id with "p2.eclipse-plugin".

That's not tree any more in Tycho 2.7.0. Tycho will use the groupId that can be found in p2 metadata, usually matching the groupId of a potential upstream artifact on Maven Central.

tycho-pomgenerator-plugin/src/main/java/org/eclipse/tycho/pomgenerator/AbstractGeneratePomsMojo.java
pzi-dsa: So in case of a pomless build, this would simply point to .polyglot.META-INF, right?
I decided against it, because it contains the <parent>...</parent> entry, pointing to the connector pom, which I don't want to keep in the metadata.
Otherwise the existing pom is almost identical to the one I create. And since it is already there, reusing it sounds like a good idea.

laeubi: If you want to support pomless the correct way to obtain the pom-**model** is to use something like:

```
List<ModelProcessor> lookupList = container.lookupList(ModelProcessor.class);
            for (ModelProcessor processor : lookupList) {
                File pom = processor.locatePom(projectBaseFolder);
                if (pom != null && pom.exists()) {
                    Map<String, Object> options = new HashMap<>(1);
                    options.put(ModelProcessor.SOURCE, new FileModelSource(pom));
                    Model project model = parentPom.getReader().read(pom, options);
                }
            }
```

of course you then could strip unwanted content from the model. To write it out you can use `ModelWriter`

`writer = container.lookup(ModelWriter.class, "xml");`

laeubi: Beside that it might be better to use `org.apache.maven.project.MavenProject.getModel()` as a basis e.g. write it out, then read it again with `ModelReader`, modify it and write the result again...

laeubi:
> > Tycho substitutes the group id with "p2.eclipse-plugin".
> 
> That's not tree any more in Tycho 2.7.0. Tycho will use the groupId that can be found in p2 metadata, usually matching the groupId of a potential upstream artifact on Maven Central.

That's what I wanted to mention above. Because of this, I think it might be enough to write the model out again (filtering some stuff that starts with p2....) instead of generating a new model....

But even the p2. info could be useful! So if maven did not complain, why not keep even that information?

mickaelistria:
> So if maven did not complain, why not keep even that information?

They allow to integrate much better with downstream tools that rely on dependency:list (eg license tools or all other forms of dependency analysis) as it allows to show the mapping between p2 and Maven artifacts. It allows to better integrate with the whole Maven ecosystem relying on GAVs, and this is more valuable than the opportunities brought by p2 coordinates.

laeubi:
I meant to keep al p2 infos that arenot already mapped to maven coordinates in contrast to filter everything out that start with p2.

mickaelistria:
I'm not sure I understand the question/concern, but in Tycho 2.7, the p2.* groupId is still used for deps that don't have Maven GAV info, nothing is filtered out as far as I know.

laeubi:
@pzi-dsa mentioned here to filter our p2. prefixed items from the generated pom.xml info!

laeubi:
@pzi-dsa What do you think about the following:

1. we keep your mojo but move it to the tycho-packaging plugin
2. It is *not* enabled by default
3. in the tycho package plugin we write out the current pom model (I'll try to come up with an example/draft) if your new mojo is not enabled

That way we offer users two choices:

1. Use the default if they like it
2. Use a template if they don't like it and could configure the additional mojo

This seems the most flexible approach to me, what do you think?

tycho-pomgenerator-plugin/src/main/java/org/eclipse/tycho/pomgenerator/GenerateDependencyPomsMojo.java
```diff
@@ -0,0 +1,201 @@
...
+        writePom(outDir.toFile(), model);
+        // Update the path of the pom.xml in the project
+        project.setPomFile(pom.toFile());
+    }
+
+    /**
+     * Calculates the direct dependencies of the current reactor project. I.e. the minimal amount of
+     * dependencies required to include all (transitive) dependencies.
+     *
+     * #TODO project.getArtifacts() should be used instead. However, it returns an empty set, even
```
laeubi, line 134: As dependency:list is working correctly as @mickaelistria has mentioned, you might want to take a look how it is done there?

tycho-pomgenerator-plugin/src/main/java/org/eclipse/tycho/pomgenerator/AbstractGeneratePomsMojo.java
laeubi: Its even easier: The project.getFile() seems to always point to the "real" pom file. In case of pomless to the `.polyglot.META-INF` file so it seems one could simply use this file (if present) as a base for the generated pom.

tycho-pomgenerator-plugin/src/main/java/org/eclipse/tycho/pomgenerator/GenerateDependencyPomsMojo.java
laeubi: by the way its

```
List<Dependency> list = Objects.requireNonNullElse(mavenProject.getDependencies(),
						Collections.emptyList());
```

what you are most probably want to use instead of `project.getArtifacts()` that is only filled if you resolve the project through maven I guess.

laeubi:
@pzi-dsa I have now prepared the [package-mojo parts](https://github.com/eclipse/tycho/pull/573) I think we should base your contribution on top of that so people can generate a custom pom if the feel the default one does not suffice.

pzi-dsa:
Good morning @laeubi 
> I have now prepared the package-mojo parts

So by default, the dependencies are now always added to the already existing pom.xml (assuming that the addMavenDescriptor flag is set)?

> That way we offer users two choices:
> 
> 1. Use the default if they like it
> 2. Use a template if they don't like it and could configure the additional mojo
> 
> This seems the most flexible approach to me, what do you think?

I agree, that sounds like the best solution.
In which case I would rewrite my Mojo to delegate the task of generating a custom pom to a template language like FreeMarker, where the template path is provided by the user.

laeubi:
> So by default, the dependencies are now always added to the already existing pom.xml (assuming that the addMavenDescriptor flag is set)?

correct. If your mojo is in place we might want to check for an existing file to skipp the defaults... or should we simply extend the template in favour of the original pom?

So one could generate a templatePom with whatever technique, then we read it and add the dependencies?

> In which case I would rewrite my Mojo to delegate the task of generating a custom pom to a template language like FreeMarker, where the template path is provided by the user.

I'm wondering if this is necessary and one could not simply use maven to replace variables in the template file?

laeubi:
@pzi-dsa the other change is merged, you might rebase your change and take a look how dependencies are found.

pzi-dsa:
I've moved the Mojo to the to the tycho-package-plugin and changed its name to GenerateMetadataMojo.

In case a handwritten pom exists in a Maven project, the dependencies are updated in the same fashion as in the PackagePluginMojo and PackageFeatureMojo.
Otherwise a custom pom is generated using FreeMarker and a user-specific template.

Remaining tasks are to disable the default behaviour, in case my Mojo is active and to create a proper integration tests.

laeubi:
@pzi-dsa can try out the current [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) build if that works for you with the default pom?

laeubi:
Thanks for the feedback, I wonder how maven do this in general?
e.g. I think we can replace the parent reference (is it useful for anything?) and we probably should propagate correct scoped dependencies (e.g. either compile or test).

laeubi:
I have merged a fix now that at least the reactor scoped dependencies are there with correct scopes. For others there is the problem that if we do not make them system scoped, then maven tries to download them ( see #585 ) and even if we won't mind that extra download now, the maven-info in P2 are not always reliable.

So my plan is to ignore the P2-Maven info if it is not specially flagged (e.g. because they originate already from the maven repo of the build) as a first step. We can then add a configuration option what "repository sources" (e.g. defaulting to `central`) are acceptable to be replaced with pure maven dependencies. On the package mojo we can even further narrow this down and even allow "fake" items to be replaced (maybe filtering by group-id.

laeubi:
I have now created another patch that will make everything from maven (e.g. pom-dependency or pom-location item) a regular maven dependency:

 ```
   <dependency>
      <groupId>org.osgi</groupId>
      <artifactId>org.osgi.service.jdbc</artifactId>
      <version>1.0.0</version>
      <scope>compile</scope>
    </dependency>
```

Everything from p2 with unclear state is inserted as an provided optional dependency:
```
    <dependency>
      <artifactId>com.ibm.icu</artifactId>
      <version>67.1.0.v20200706-1749</version>
      <type>eclipse-plugin</type>
      <scope>provided</scope>
      <optional>true</optional>
    </dependency>
```

I think that's the best we can do right now. If my suggested change at maven (https://github.com/apache/maven/pull/667) will be accepted we could offer more (e.g. add a maven-plugin that resolves P2 dependencies in plain maven build ...).

**Comment by pzi-dsa** [comment]
> Thanks for the feedback, I wonder how maven do this in general?

After looking at other organizations, at least Apache and Google deploy their parent poms, as well. It's not a major problem, as it is resolved by simply deploying the aggregator poms as well. I'd keep the parent dependency, as it might contain useful information for e.g. license tools.

> I think that's the best we can do right now

I've just tested the latest build and so far it's working perfectly. I'll have a closer look at it, but unless there's an issue showing up, it would make this pull request obsolete.

One oddity I noticed, though: When deploying the artifact with the maven-deploy-plugin, all of the dependencies are lost, even though they exist in the pom of the bundled jar as well as the pom that is written to the target folder.

I assume that the deployment is done using the pom specified via MavenProject#getFile(). Since the tycho-packaging-plugin works on a clone of the project, the updated dependencies are never propagated to successive plugins and are therefore omitted.

**Comment by pzi-dsa** [comment]
i.e. I would replace the following lines in the AbstractTychoPackagingMojo
```
MavenProject mavenProject = project.clone(); // don't alter the original project!
mavenProject.setFile(pomFile);
```

with 

```
mavenProject.setPomFile(pomFile);
```

**Comment by laeubi** [comment]
Replacing the pom location sounds waiting for issues to happen, as the pom will never be perfect at the first place. Maybe if we only do pomless builds it might be possible.

I think we have to better integrate with `mvn deploy` ... maybe we would need a special tycho-mojo bind to the deploy phase as well.

If you can came up with a small example + run instructions that would be helpful.

**Comment by laeubi** [comment]
The https://blog.frankel.ch/maven-flatten-plugin/ uses a flag for replacing the file or keeping it seperate, so that might be an option as well.

**Comment by pzi-dsa** [comment]
> If you can came up with a small example + run instructions that would be helpful.

I have updated by [example project](https://github.com/pzi-dsa/tycho.example) to use the latest nightly build. The issue can be reproduced by executing `mvn clean install`.

For example, the pom file in both the target folder and the jar of plugin2 contains the expected entry:
```
  <dependencies>
    <dependency>
      <groupId>example</groupId>
      <artifactId>example.plugin1</artifactId>
      <version>1.0.0-SNAPSHOT</version>
      <type>eclipse-plugin</type>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
  </dependencies>
```

But the pom in the local m2 repository does not. So the problem already occurs in the `install` phase, not the `deploy` phase.
I also had a quick look at the maven-flatten-plugin, and from what I can tell, they also use mavenProject.setPomFile(...) to redirect the path to the modified pom.

I agree that the safest solution is to add a mojo to the `install` phase, such that the correct pom is copied to the (local) repository.

pzi-dsa:
I've tested your modification and can confirm that it is working.
@laeubi Thank you a lot for all your help :)

PR #573, opened 2022-01-27
Fix #572 - Insert dynamic dependencies into the jar included pom

### LINKED ISSUE #572: Insert dynamic dependencies into the jar included pom
Currently the pom is included but it is missing the automatically generated dependencies.

laeubi:
This generates a pom file based on the original and add all dynamically added dependencies from Tycho into the model.

This should be a perquisite of 

- https://github.com/eclipse/tycho/pull/569

where the resulting model could be specified with a template in case someone don't want the default or has special needs.


laeubi:
@mickaelistria any concerns? If not I'd like to merge this soon.

mickaelistria:
No objection to merge as soon as tests are OK (on Eclipse CI)

PR #577, opened 2022-01-28
Support Maven option --also-make-dependents, and --resume-from for project dependency resolution

laeubi:
This is a basic prototype for https://github.com/eclipse/tycho/issues/48 and https://github.com/eclipse/tycho/issues/551 and also a showcase that Tycho could use P2 without the need of running it inside an OSGi-Framework (FYI @merks).

With this change (and enabling the [tycho-build extension](https://github.com/eclipse/tycho/blob/master/RELEASE_NOTES.md#tycho-pomless-will-become-a-tycho-core-extension)) the test cases from https://github.com/eclipse/tycho/pull/564 succeed.

As there where some discussion why we don't _"just fix this after such a long time"_, I'd like to give some insights about the efforts put into this prototype up to now (and its not complete but should work for common cases):

Beside the discussions and analysis in https://github.com/eclipse/tycho/issues/48 with the useful hint/example code in https://github.com/eclipse/tycho/issues/551 (thanks @sratz) and the provided test-cases in https://github.com/eclipse/tycho/pull/564 (thanks @joeshannon) the following changes and refactoring where a perquisite:

- https://github.com/eclipse/tycho/commit/e2cecb599d30ae8f110da92e92909ff50d0d79aa (271 changed LOCs)
- https://github.com/eclipse/tycho/commit/53ac9ec27529ec5e6f1353b3c622407cf38200b0 (177 changed LOCs)
- https://github.com/eclipse/tycho/commit/f22018598a4c6b2986d120b03a0102abd8bdf692 (308 changed LOCs)

To make this sustainable in the future the following changes in external projects where triggered and PRs provided:

- https://github.com/apache/maven/pull/665
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=578380
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=578394
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=578365

**Altogether I approximately spend 20 hrs of development time with coding, writing issues, reviewing, analysis and test.**

And as mentioned this is a prototype and it is not complete, there is more work to do to make this complete and more useful (e.g. at the moment project dependencies are resolved twice once for graph building and once in the regular build).

Lust but not least I think we can integrate this at the current state in Tycho if there are no concern ( @mickaelistria @akurtakov ).

If this is crucial to someones business/workflow and likes to push the development forward in that area a [sponsoring](https://github.com/sponsors/laeubi?frequency=one-time) would allow me to assign more time-slots on these particular issue. Of course one can always contact me directly or contribute upon this initial prototype.

### DIFF
--- a/tycho-build/pom.xml
+++ b/tycho-build/pom.xml
@@ -31,6 +31,84 @@
 			<version>${project.version}</version>
 		</dependency>
 		<dependency>
+			<groupId>org.eclipse.tycho</groupId>
+			<artifactId>org.eclipse.tycho.core.shared</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+		<dependency>
+			<groupId>org.eclipse.tycho</groupId>
+			<artifactId>org.eclipse.tycho.p2.maven.repository</artifactId>
+			<version>2.7.0-SNAPSHOT</version>
+		</dependency>
+		<!--
+		<dependency>
+			<groupId>org.eclipse.tycho</groupId>
+			<artifactId>tycho-core</artifactId>
+			<version>${project.version}</version>
+		</dependency>
+-->
+		<dependency>
+			<groupId>org.eclipse.platform</groupId>
+			<artifactId>org.eclipse.equinox.p2.core</artifactId>
+			<version>2.6.100</version>
+		</dependency>
+
+		<dependency>
+			<groupId>org.eclipse.platform</groupId>
+			<artifactId>org.eclipse.osgi</artifactId>
+			<version>3.17.100</version>
+		</dependency>
+
+		<dependency>
+			<groupId>org.eclipse.platform</groupId>
+			<artifactId>org.eclipse.equinox.p2.publisher</artifactId>
+			<version>1.6.200</version>
+		</dependency>
+
+		<dependency>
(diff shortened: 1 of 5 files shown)


mickaelistria:
That's a great step forward. I would personally not see any issue with merging such iteration as long as it doesn't cause any regression. As the tycho-build extension is new, it's probably very safe to start using it here.
So the thing that does fix #564 is addition of the GraphBuilder, right? Is the "p2 running outside of OSGi framework" a mandatory requirement to get the GraphBuilder? How strong is the coupling between those 2 major changes? As you can guess, if we can have them separated, it would be preferred for better tracking (and future debugging).

laeubi:
> Is the "p2 running outside of OSGi framework" a mandatory requirement to get the GraphBuilder?

I'm just calling code of P2 directly from maven as opposed to Tycho firing up an OSGi framework. This would not work so well here as we are at a very early phase, and even classloaders seem separated compared to later maven stages. So it is not really something one can "split off".

laeubi:
> As the tycho-build extension is new, it's probably very safe to start using it here.

Yes actually if not enabled code will never be called from there... and if enabled the tycho-code should never get aware of the extension code.

PR #582, opened 2022-01-28
Fix #581  Use the correct scope when injecting dependencies into the maven model

### LINKED ISSUE #581: Use the correct scope when injecting dependencies into the maven model
As reported in

- https://github.com/eclipse/tycho/pull/569

we currently set a system/dependent scope that could confuse maven tools. We should set an appropriate scope type depending on the compiles/test flags.

laeubi:
Try to use the actual dependency scope.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java
@@ -8,7 +8,8 @@
  * Contributors:
  *    Sonatype Inc. - initial API and implementation
  *    SAP AG - inject nested class path elements into maven model (TYCHO-483)
- *    Christoph Läubrich - Issue #443 - Use regular Maven coordinates -when possible- for dependencies 
+ *    Christoph Läubrich    - Issue #443 - Use regular Maven coordinates -when possible- for dependencies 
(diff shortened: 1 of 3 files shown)


tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java
```diff
@@ -151,28 +159,39 @@ private Dependency createSystemScopeDependency(ArtifactDescriptor descriptor, St
...
+            if (Artifact.SCOPE_SYSTEM.equals(dependency.getScope())) {
+                dependency.setSystemPath(location.getAbsolutePath());
+            }
         } else {
             ArtifactKey artifactKey = descriptor.getKey();
             dependency.setGroupId(groupId);
             dependency.setArtifactId(artifactKey.getId());
             dependency.setVersion(artifactKey.getVersion());
             dependency.setScope(Artifact.SCOPE_SYSTEM);
+            dependency.setType(artifactKey.getType());
```
laeubi, line 181: Here we propagate the actual artifact-type instead of the default (jar) this seems more semantic to me and we should just register a ArtifactHandler for the Tycho types (maybe in a later change, depending on how it is required for system-scoped items).

laeubi:
This is ready to be merged, even though it does only fix the scope for reactor dependencies this is better thank nothing, target provided items are still with scope=system as we first need to have support from #585 and #588 to make this more smooth...

PR #584, opened 2022-01-29
Fix #583 - TychoMavenLifecycleParticipant must clone the maven session before spinning new threads

### LINKED ISSUE #583: TychoMavenLifecycleParticipant must clone the maven session before sinning new threads
TychoMavenLifecycleParticipant currently spinning new threads and passing the MavenSession to them.

As MavenSession has only a single field for getCurrentProject() and the tycho realm helper modify that value it seems not safe to do so. We should simply clone the session to get a private copy for the threads.

laeubi:


mickaelistria:
It would be nice to add more explanation in commit messages about why is such change necessary. Eg the reason given on the issue would be good to add on commit message.

PR #586, opened 2022-01-29
Provide a Tycho Workspace Reader

laeubi:


akurtakov:
What is the status of this one? It helps finishing PRs so we get less waiting on builds.

laeubi:
This needs release of maven 3.8.5 to be finished.

laeubi:
@akurtakov maybe we can exclude PRs with a certain label (or simply ones marked as WIP) from the eclipse infra ci build?

laeubi:
@akurtakov I have enabled to skip draft PRs from the build for now, I think this is suitable as we have GH action as a first check.

akurtakov:
What is the status here? The dash licensetool patch is merged and the longer PRs stay in the queue the less likely it is to be finished and at the end become just noise in the system.

laeubi:
This waits for https://github.com/apache/maven/pull/695 and maven 3.8.6 release.

laeubi:
The maven release containing the fix is currently planned "something around 2022-05-08"

laeubi:
Maven 3.8.6 is out!

laeubi:
@akurtakov any idea? When running in "full" builds one get:

> Error:  org.eclipse.tycho.plugins.p2.extras.MirrorMojoTest.testMirrorWithPlatformFilter  Time elapsed: 0.195 s  <<< ERROR!
> java.lang.SecurityException: class "org.apache.commons.io.output.NullOutputStream"'s signer information does not match signer information of other classes in the same package
> 	at org.eclipse.tycho.plugins.p2.extras.MirrorMojoTest.setUp(MirrorMojoTest.java:50)

but running the `tycho-extras/tycho-p2-extras-plugin` alone succeeds...

akurtakov:
My first idea would be that commons.io classes are loaded from 2 different jars - one signed the other not or signed with different certificates.

laeubi:
That's really strange, I replaced commons i/o here with plain java, lets see if that helps...

laeubi:
Strange, now that I once run the extras build alone the full build suceeds...

akurtakov:
Something I learned from years on Fedora Java stack - using libraries is great as it gives you years of tested code and also this has to be well planned and aligned in the *whole stack* as otherwise you wish to have written that single function yourself.

Issue #106, opened 2021-04-29
ArtifactType should be an enum


laeubi:
ArtifactType is currently a collection of string types, we should try to convert it into an enum

PR #589, opened 2022-01-30
Issue #581 - use a repository id to identify 'real' maven dependencies

laeubi:

### DIFF
--- a/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/AbstractTychoPackagingMojo.java
+++ b/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/AbstractTychoPackagingMojo.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 
+import org.apache.maven.artifact.Artifact;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.model.Model;
@@ -172,6 +173,10 @@ public abstract class AbstractTychoPackagingMojo extends AbstractMojo {
 		for (Dependency dep : list) {
 			Dependency copy = dep.clone();
 			copy.setSystemPath(null);
+			if (dep.getGroupId().startsWith("p2.eclipse.") && Artifact.SCOPE_SYSTEM.equals(dep.getScope())) {
+				copy.setOptional(true);
+				copy.setScope(Artifact.SCOPE_PROVIDED);
+			}
 			dependencies.add(copy);
 		}
 		modelWriter.write(pomFile, null, projectModel);
--- a/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/MavenDependencyDescriptor.java
+++ b/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/MavenDependencyDescriptor.java
@@ -42,4 +42,10 @@ public interface MavenDependencyDescriptor {
      * @return the version of the dependency
      */
     String getVersion();
+
+    /**
+     * 
+     * @return the id of the repository this artifact is located in
+     */
+    String getRepository();
 }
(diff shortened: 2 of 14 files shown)


tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/AbstractTychoPackagingMojo.java
```diff
@@ -172,6 +173,10 @@ protected MavenProject updatePom(String finalName) throws IOException {
 		for (Dependency dep : list) {
 			Dependency copy = dep.clone();
 			copy.setSystemPath(null);
+			if (dep.getGroupId().startsWith("p2.eclipse.") && Artifact.SCOPE_SYSTEM.equals(dep.getScope())) {
+				copy.setOptional(true);
+				copy.setScope(Artifact.SCOPE_PROVIDED);
+			}
```
mickaelistria, line 179: Isn't it/shouldn't it be part of a separate commit?
Couldn't it be then that instead of a repository id, we could then assume that everything that's not "provided" comes either from reactor or repository?

mickaelistria:
As mentioned on another bug, I'm not sure we want to make the repository-id leak everywhere as it's not reliable (it's mutable, moveable, duplicat-able...) content. However, I think the scope is something we can rely on more safely.

tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/AbstractTychoPackagingMojo.java
laeubi: > Isn't it/shouldn't it be part of a separate commit?
Maybe yes, I just want to make the big picture work here :-)

> Couldn't it be then that instead of a repository id, we could then assume that everything that's not "provided" comes either from reactor or repository?

I hope we can over the time make more and more items with the real scope, and maybe provided would become a valid scope then also even though we currently only use compile+test+system, so my goal is to be as specific as possible here.

laeubi:
> As mentioned on another bug, I'm not sure we want to make the repository-id leak everywhere as it's not reliable (it's mutable, moveable, duplicat-able...) content. However, I think the scope is something we can rely on more safely.

The scope is nothing we can carry over in the p2-data as an artififact might be referenced in different projects with different scopes.

I agree that the repository-id is somewhat weak, but it at least indicates that the artifact was consumed one from a repository with that id (most often from `central` what is a default repository in maven and thus well known).

As mentioned in the comments, the longer goal would be to make this more configurable, so for example one might only want things converted that are from `central` or `nexus` or ... and as these ids match to the repository configuration of maven build, we can even query that repository for the artifact and skip if its not there.

mickaelistria:
> The scope is nothing we can carry over in the p2-data as an artififact might be referenced in different projects with different scopes.

That's right, I got confused.

> I agree that the repository-id is somewhat weak, but it at least indicates that the artifact was consumed one from a repository with that id (most often from central what is a default repository in maven and thus well known).

OK

> As mentioned in the comments, the longer goal would be to make this more configurable, so for example one might only want things converted that are from central or nexus or ... and as these ids match to the repository configuration of maven build, we can even query that repository for the artifact and skip if its not there.

I really hope that we can keep Tycho simple enough so that people wouldn't need such fine grain configuration!

Anyway, I'm fine with this patch if you want to merge it, and I now acknowledge that storing repository-id is not perfect but not too bad either.

laeubi:
> I really hope that we can keep Tycho simple enough so that people wouldn't need such fine grain configuration!

I use tycho with nearly zero configuration and actually that's the goal, its just that there are some cases where people need to comply to some "rules" so I'm always try to guess what *might* be necessary here in the future if some one demands it.

For me the important think is at the moment that we can express "this artifact was consumed at once from any remote repository and if you would add this your build will probably succeed in resolve it from there" versus "these artifact was probably never deployed at any maven repository at all" (beside the local one...).

laeubi:
Great I hope the last integreation test will now also pass, that reminds me of 
- #106 

we still have some issues about confusing packageing types with artifact types...

PR #591, opened 2022-01-30
Java workaround

MarcMil:
Prevents
An unexpected error occurred while launching the test runtime (process returned error code 13). Command-line used to launch the sub-process was null -Dosgi.noShutdown=false

Similar to https://github.com/talios/clojure-maven-plugin/blob/clojure-maven-plugin-1.7.1/src/main/java/com/theoryinpractise/clojure/AbstractClojureCompilerMojo.java#L247


laeubi:
@MarcMil would you mind the ECA so we can accept your code contribution?

MarcMil:
I've rewritten the history so that the commit happened with the correct E-Mail address, which already signed the ECA

laeubi (approved):
I think this is not that bad what do you think @mickaelistria

mickaelistria (changes requested):
Thanks. We'll merge it when CI has successfully ran all tests.

mickaelistria (approved):

laeubi:
Funny enough I run into exactly this issue right now while debugging an integration test :+1:

laeubi:
I can now see that it does not fail with 'null' but with 'java' even though I can run 'java' on my terminal without an issue. Is this supposed to work this way?

MarcMil:
@laeubi Interesting. Is System.getenv("PATH") containing the path to the JVM's bin folder as well? Maybe the PATH variable is set differently for the maven process for whatever reason. Maybe you need some more magic to find and use the absolute path to the JVM's bin folder.

PR #595, opened 2022-01-31
Switch IT tests to run against 2021-12 for latest platform.

akurtakov:


mickaelistria:
tycho-its/projects/target.restriction.filter/pom.xml contains the following block
```xml
<filter>
	<type> eclipse-plugin </type>
	<id> org.eclipse.osgi </id>  <!-- extra whitespace is not needed; it is just here to make the test harder -->
	<restrictTo>
		<versionRange>
			[3.12,3.13)
		</versionRange>
	</restrictTo>
</filter>				
```
which, with a newer .target, will exclude org.eclipse.osgi bundle from resolution; and org.eclipse.osgi bundle is the bundle that provides the EE capability. You may need to update this version range to latest version of org.eclipse.org (or maybe another one as this test seems about enforcing an older version when there are multiple ones available).

akurtakov:
Thanks Mickael! So I'm switching the tests to compare 2021-06 and 2021-12 in order to have minor version diff for that bundle.

PR #599, opened 2022-01-31
If no toolchain is found, try JAVA_HOME as a first guess

laeubi:
This is a follow-up of https://github.com/eclipse/tycho/pull/591


mickaelistria:
I earlier merged a "bad" version of this patch, which doesn't seem like it could cause a regression. So I'm going for the same criterion here: if you think it's safe, feel free to merge.

PR #604, opened 2022-01-31
Fix #602 - Align pom-generation with flatten plugin

### LINKED ISSUE #602: Align pom-generation with flatten plugin
Currently the pom is generates as part of the package plugin. We better should do this in a different phase, so it is possible to modify the pom if desired.

We could get some ideas from the [flatten-maven-plugin](https://www.mojohaus.org/flatten-maven-plugin/flatten-mojo.html), it should even be possible to use this together with tycho, this would give much more flexibility.

laeubi:
This adds a new mojo to the packe plugin that behave similar to the maven [flatten:flatten](https://www.mojohaus.org/flatten-maven-plugin/flatten-mojo.html) goal. it binds to the `prepare-package` phase so it is possible to further adjust/modify the pom (e.g. with the maven-flatten-plugin) before it gets packed into the jar or deployed.

### DIFF
--- a/tycho-maven-plugin/src/main/resources/META-INF/plexus/components.xml
+++ b/tycho-maven-plugin/src/main/resources/META-INF/plexus/components.xml
@@ -36,6 +36,9 @@
               <test>
                 org.apache.maven.plugins:maven-surefire-plugin:${surefire-plugin.version}:test
               </test>
+              <prepare-package>
+               org.eclipse.tycho:tycho-packaging-plugin:${project.version}:update-consumer-pom
+              </prepare-package>
               <package>
                 org.eclipse.tycho:tycho-packaging-plugin:${project.version}:package-plugin,
                 org.eclipse.tycho:tycho-p2-plugin:${project.version}:p2-metadata-default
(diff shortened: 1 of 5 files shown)


laeubi:
@mickaelistria any concerns? the feature can be disabled by setting `<skipPomGeneration>true</skipPomGeneration>` but is on by default. Users of jgit-timestamp provider may have to add `.tycho-pom.xml` to their ignore list.

mickaelistria:
I think it's good if we can have a dedicated mojo for that. Just a suggestion about wording: we're talking about the "consumer" pom right? The one that's generated and would be installed/deployed? If so, I suggest we explicit this term in to update-consumer-pom and UpdateConsumerPomMojo to clarify things here and ensure we cut some possible path of misinterpretation.

laeubi:
> The one that's generated and would be installed/deployed?

I don't know whats the right 'naming' but it takes the "project" pom and that is used later to pack inside the jar and/or deploy as .pom file at maven central.

mickaelistria:
OK, so I think it's what Maven calls "consumer pom", from https://foojay.io/today/flatten-maven-plugin/ "the consumer POM, that is published to Maven Central", so let's adopt this naming to embrace how Maven calls things.

laeubi:
> so let's adopt this naming to embrace how Maven calls things.

Done!

mickaelistria (approved):
Thanks for the update. Feel free to merge as soon as tests are OK with it.

laeubi:
Sad enough the CI server seems down :-(

https://ci.eclipse.org/tycho/

> Application is not available
> 
> The application is currently not serving requests at this endpoint. It may not have been started or is still starting.

PR #606, opened 2022-02-01
Add missin gheader, speed up the test

laeubi:
FYI @joeshannon let me know if the header is correct and if it improves the buildtime for you.


laeubi:
Total time decreased from `44m 52s` to `37m 18s` now.

PR #608, opened 2022-02-01
TychoGraphBuilder: -amd should pick up dependencies of dependants

sratz:
`-amd` in general is of little practical use in Tycho build environments that typically contain downstream features + updatesites that depend on practically everything.

To make it somewhat useful (i.e. get something buildable in the first place out of it), `-amd` should pick up dependencies of dependents.

Also add a few more integration tests.

Change-Id: I0538fb5bd285c8d2d4132e95c4d44d09c5b2026c

### DIFF
--- a/tycho-build/src/main/java/org/eclipse/tycho/build/TychoGraphBuilder.java
+++ b/tycho-build/src/main/java/org/eclipse/tycho/build/TychoGraphBuilder.java
@@ -220,9 +220,9 @@ public class TychoGraphBuilder extends DefaultGraphBuilder {
 								.distinct()//
 								.peek(project -> loggerAdapter.debug(" + add downstream project '" + project.getName()
 										+ "' of project '" + projectRequest.mavenProject.getName() + "'..."))//
-								// make behaviors are both false here as projectDependenciesMap includes transitive already
+								// request dependencies of dependants, otherwise, -amd would not be able to produce a satisfiable build graph
 								.forEach(
-										project -> queue.add(new ProjectRequest(project, false, false, projectRequest)));
+										project -> queue.add(new ProjectRequest(project, false, true, projectRequest)));
 					}
 				}
 			}
--- a/tycho-its/projects/reactor.makeBehaviour/bundle1a/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/reactor.makeBehaviour/bundle1a/META-INF/MANIFEST.MF
@@ -1,0 +1,7 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Bundle 1a
+Bundle-SymbolicName: bundle1a;singleton:=true
+Bundle-Version: 1.0.0.qualifier
+Bundle-RequiredExecutionEnvironment: J2SE-1.5
+Require-Bundle: bundle1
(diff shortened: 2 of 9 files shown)


tycho-build/src/main/java/org/eclipse/tycho/build/TychoGraphBuilder.java
```diff
@@ -222,7 +222,7 @@
 										+ "' of project '" + projectRequest.mavenProject.getName() + "'..."))//
 								// make behaviors are both false here as projectDependenciesMap includes transitive already
```
laeubi, line 223: The comment here seems outdated

laeubi (changes requested):

tycho-build/src/main/java/org/eclipse/tycho/build/TychoGraphBuilder.java
sratz: Done

laeubi (approved):

joeshannon:
I'm not sure about this as I think it will imply a behaviour difference between vanilla maven vs tycho which may be unexpected by users.

In general for non-linear dependency relations, both -am and -amd would typically have to be used as far as I understand it.

I will just upload a demo vanilla maven project to demonstrate.

laeubi:
@joeshannon I don't know as its a bit unclear how it is supposed to work.
From a user perspective I would find it rather strange if I have to add all the transitive projects manually, but maybe that how it is supposed to work?

joeshannon:
Yes I have to admit I am also not hugely experienced with using these options in practice outside of simple builds so I am happy to concede to other views as to how this should behave. I was approaching this by attempting to match the behaviour of a pure maven demo. [Here is that demo](https://github.com/joeshannon/simple-maven-reactor). I'll look at adding the additional modules as presented in this PR in a bit.

Whether we want to modify this behaviour in Tycho to enhance common use cases is another question I guess.

laeubi:
If we want to say that[ this document is normative](https://blog.sonatype.com/2009/10/maven-tips-and-tricks-advanced-reactor-options/) they state:

> If you wanted to run a portion of the larger build, you would use the -pl or --projects option with the -am or --also-make option. When you specify a project with the -am option, Maven will build all of the projects that the specified project depends upon (either directly or indirectly). Maven will examine the list of projects and walk down the dependency tree, finding all of the projects that it needs to build.

So if `-pl A` and specify `-am` would require that i build A and everything A depends on transitively this should work without adding anything beside the computed dependencies, as Tycho already computes a set of all transitive requirements. As mentioned above, for this we should even be able to skip computation of *all* projects dependencies but shortcircuit to only compute them for `A` (even we need to examine all others for their metadata).

> the -amd or --also-make-dependents option configures Maven to build a project and any project that depends on that project. When using --also-make-dependents, Maven will examine all of the projects in our reactor to find projects that depend on a particular project. It will automatically build those projects and nothing else.

Given that `-pl A` and specify `-amd` should only add _direct_ dependencies but not transitive ones! Even though I have no idea for what this is useful as it only will work for 'flat' structures, this will require to compute *all* project dependencies to be computed to find out who needs `A` but don't add anything else.

The interesting part would be (and that's what is not mentioned here) if we specify *both* options, should we only add dependencies of specified  projects and what depends on them but not the dependent dependencies ...?

laeubi:
@sratz it seems one test is still failing do you plan to take a look at this?

laeubi (approved):

laeubi:
Thanks for the testcase + fix

PR #609, opened 2022-02-01
Fix #588 resolve the extension from the artifact type

### LINKED ISSUE #588: Tycho should provide ArtifactHandlers for its types
Currently they are mapped by the default artifact handler, but we better should implement custom types to have more control how maven handle these.

I have created https://issues.apache.org/jira/browse/MNG-7403 for this.

laeubi:
Currently Tycho uses a confusing mixture of "artifact-types" and "extensions". With this change, the extension for a given artifact type is looked up in the maven artifacthandler registry instead.

The whole code is rather deeply baked into tycho so we need some backward-compat handling to make it work, but we will be more flexible and possible remove some of them in the future.

This will especially become handy if we want better expression of tycho dependencies in the maven model.


laeubi:
@mickaelistria all test pass so I'D like to merge this any concerns?

PR #618, opened 2022-02-02
Improve multiple exports project to expose bug with JDK 17

parttimenerd:


laeubi:
@parttimenerd can you explain a bit more whats the background of this change is?

parttimenerd:
Essentially this triggers a bug in the tycho compiler with multiple `--add-exports` options on JDK 17.
The original test project did not use the exported package thereby hiding real access problems and not testing whether the exporting really worked.

akurtakov:
Before looking at your patch would you please sign ECA?

parttimenerd:
I'm trying too.

tycho-its/projects/compiler.extraExports.java17/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,4 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-SymbolicName: compilerExtraExports
+Bundle-Version: 1.0.0.qualifier
```
laeubi, line 4: It should have a Bundle-EE of Jse17?

parttimenerd: I'm confused

parttimenerd: What do you mean?

laeubi: Lets see first how the build behaves.

tycho-its/projects/compiler.extraExports.java17/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+				<version>${tycho-version}</version>
+				<extensions>true</extensions>
+			</plugin>
+			<plugin>
+			  	<groupId>org.eclipse.tycho</groupId>
+				<artifactId>tycho-compiler-plugin</artifactId>
+				<version>${tycho-version}</version>
+				<configuration>
+					<configuration>
+						<useJDK>JavaSE-17</useJDK>
```
laeubi, line 32:
```suggestion
						 <useJDK>BREE</useJDK>
```

tycho-its/projects/compiler.extraExports.java17/META-INF/MANIFEST.MF
laeubi, line 4:
```suggestion
Bundle-Version: 1.0.0.qualifier
Bundle-RequiredExecutionEnvironment: JavaSE-17
```

laeubi (commented):
See review comments

laeubi:
@mickaelistria @akurtakov any idea why this succeeds even if no real jdk17 jvm is there?

akurtakov:
What's with the failing ECA validation here?

parttimenerd:
That is probably the reason why it failed.

tycho-its/src/test/java/org/eclipse/tycho/test/compiler/CompilerExtraExportsJava17.java
```diff
@@ -0,0 +1,28 @@
...
+import org.apache.maven.it.Verifier;
+import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
+import org.junit.Test;
+
+public class CompilerExtraExportsJava17 extends AbstractTychoIntegrationTest {
+
+    @Test
+    public void testExtraExports() throws Exception {
+        Verifier verifier = getVerifier("compiler.extraExports.java17", false);
+        verifier.executeGoal("compile");
```
laeubi, line 25: You probably want to verifier.verifyErrorFreeLog() assertion here...

tycho-its/src/test/java/org/eclipse/tycho/test/compiler/CompilerExtraExportsJava17.java
```diff
@@ -0,0 +1,28 @@
...
+ *     Red Hat Inc. - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.tycho.test.compiler;
+
+import org.apache.maven.it.Verifier;
+import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
+import org.junit.Test;
+
+public class CompilerExtraExportsJava17 extends AbstractTychoIntegrationTest {
```
laeubi, line 20:
```suggestion
public class CompilerExtraExportsJava17Test extends AbstractTychoIntegrationTest {
```

According to surefire matching rules I think all test classes must end with 'Test'

parttimenerd: I just copied the `CompilerExtraExports` test class which also does end with `Test`. So maybe I should change the other names too?

laeubi: If you go here:

https://github.com/eclipse/tycho/pull/618/checks?check_run_id=5056657496

expand  the second "raw output" and search for the test/testname it is not found so I think both tests are simply never executed by the CI ... I think because of the name not ending in `Test`

parttimenerd: Good, then I'm going to make these renames in this PR today.

tycho-its/src/test/java/org/eclipse/tycho/test/compiler/CompilerExtraExportsJava17.java
parttimenerd: Bit I want this in the other test case too?

laeubi: I always add this as it does not harm to assert there are no (other) errors beside a build-failure.

parttimenerd: I take this as a yes for the other test case (the non Java 17 one).

laeubi:
@parttimenerd could you take a look at the compile errors?

https://github.com/eclipse/tycho/runs/5069764604?check_suite_focus=true

parttimenerd:
The errors should be fixed

laeubi:
> The errors should be fixed

Just one small issue it seems [CompilerExcludeTest](https://github.com/eclipse/tycho/pull/618/files#annotation_2701731228) uses `tycho.version` instead of `tycho-version`

parttimenerd:
Oh, I forgot it. I'm working on it now.

parttimenerd:
> Beside that I see org.eclipse.tycho.test.compiler.CompilerExtraExportsJava17Test ‑ testExtraExports executed but it seems to succeed here.

That confuses me too.

parttimenerd:
With JDK17 `mvn -Dtycho-version=2.6.0 test` fails for the extraExports project, but not for the new project.

laeubi:
Lets see if they pass the build, if yes I'd like to merge this as we clearly have an error here that tests are not executed. I'll then investigate why it makes a difference on ci or on commandline call (e.g. why it does not errors).
Can you try to run the test with J11 and see if it complains due to missing Java17 at all?

parttimenerd:
First: the difference it that it assumes, without the useJDK, that the OSGi profile is JavaSE-11, which seems to trip it up (`-target 11 -source 11` gets passed to JDT).

> Can you try to run the test with J11 and see if it complains due to missing Java17 at all?
Both projects work on Java 11

laeubi:
> Both projects work on Java 11

Yeah but are there any warnings? e.g. with `<useJDK>BREE</useJDK>` and `Bundle-RequiredExecutionEnvironment: JavaSE-17` I would expect it to fail when compiling it on a jdk-11 without further configuration.

parttimenerd:
```
[WARNING] No explicit target runtime environment configuration. Build is platform dependent.
[INFO] Computing target platform for MavenProject: tycho.its:compilerExtraExportsJava17:1.0.0-SNAPSHOT @ /Users/I560383/code/tests/tycho/tycho-its/projects/compiler.extraExports.java17/pom.xml
[WARNING] No system packages found in profile nor toolchain for JavaSE-17, using current JRE system packages.
```

laeubi:
@parttimenerd can you suqash all the commits and rebase them on master?

laeubi:
@parttimenerd could you please rebase this PR as we have released an now tycho 3.0 is the current version? Hopefulyl this will fix the last remaining failed test.

laeubi:
Great lets see if it finally succeeds :-)

PR #620, opened 2022-02-02
Let licensecheck fail if not every dep is approved

akurtakov:


akurtakov:
@laeubi you were right that it should fail by default so everyone checks it! 
https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/1929 and https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/1928 look like mishandled metadata so it should not be failing soon.
Eitherway I'm pushing this one so we don't get more non-approved libs used.

laeubi:
:+1:

laeubi:
Just wondering isn'T sat4j used by P2 already? Or how it got ino tycho?

akurtakov:
p2 should update to sat4j 2.3.6 too but I would rather wait for next stream so we use the maven artifact directly.

laeubi:
Thanks for the explanation should we do anything to mitigate this e.g. exclude/pin versions in pom.xml?

mickaelistria:
> should we do anything to mitigate this e.g. exclude/pin versions in pom.xml?

It's actually not an issue to use a newer version, on the contrary. The issue is more that the approved version is an older one, and the proper remediation is to get a later version approved as things are already happening.

akurtakov:
New versions of sat4j jars approved. Licensecheck is back to succeeding.

laeubi:
Thanks for taking care!

PR #621, opened 2022-02-02
Only generate the consumer-pom if it is explicitly requested or required

laeubi:


laeubi:
@akurtakov it seems the check now works :-)

akurtakov:
> @akurtakov it seems the check now works :-)

At least that :)

PR #641, opened 2022-02-10
Updated the description of the new goal tycho-surefire:integration-test

LorenzoBettini:
I found the description of the new goal tycho-surefire:integration-test in the RELEASE notes a bit confusing and with several errors.

First of all, the RELEASE notes used to mention tycho-failsafe, which does not exist at all. They were also mentioning MOJOs, but maybe it's better to use the term `goals`?

I also tried to summarize the differences (see the "Summarizing..."), which were not clear to me. That's what I understand from taking a look at the code, but I might be wrong.

Is there an example project showing this new goal, by the way?

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -164,12 +164,12 @@ It is now possible to resolve the JDT 'JUnit Classpath Container', for this do t
 For an example take a look at the [integration tests](https://github.com/eclipse/tycho/tree/master/tycho-its/projects/compiler.junitcontainer/junit4-in-bundle)
 
 ### [Execute unit-tests with eclipse-plugin packaging](https://bugs.eclipse.org/bugs/show_bug.cgi?id=572420)
-Before unit-tests are only executed for eclipse-test-plugin packaging types. Beside that it was only possible to execute them as part of **tycho-surefire** (what executes them inside an OSGi runtime) in the integration-test phase (making them actually some kind of integration test).
+Previously, unit-tests were only executed for `eclipse-test-plugin` packaging types. Besides that, it was only possible to execute them as part of the **tycho-surefire:test** goal (which executes them inside an OSGi runtime) in the `integration-test` phase (making them actually some kind of integration tests).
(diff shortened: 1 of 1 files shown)


laeubi (commented):
You can find an example here:

https://github.com/eclipse/tycho/tree/master/tycho-its/projects/surefire.combinedtests/bundle.test

The term mojo references a goal see: https://maven.apache.org/guides/introduction/introduction-to-plugins.html#what-is-a-mojo-and-why-the-h-is-it-named-mojo

If you scroll donw here https://maven.apache.org/surefire/maven-failsafe-plugin/ you will find some more information about how fail safe integration tests are supposed to work in maven.

LorenzoBettini:
> You can find an example here:
> 
> https://github.com/eclipse/tycho/tree/master/tycho-its/projects/surefire.combinedtests/bundle.test
> 

OK, thanks!

> The term mojo references a goal see: https://maven.apache.org/guides/introduction/introduction-to-plugins.html#what-is-a-mojo-and-why-the-h-is-it-named-mojo
> 

Yes, I know, but usually when you speak about something bound to a Maven phase you usually use the term "goal" instead of MOJO, which is an implementation detail. At least, that's what I usually read...

> If you scroll donw here https://maven.apache.org/surefire/maven-failsafe-plugin/ you will find some more information about how fail safe integration tests are supposed to work in maven.

I know that one as well, indeed I simply mentioned that in the PR.

In any case, do I need to modify somehow the PR? Note that a few terms were wrong anyway ;)

laeubi:
> Yes, I know, but usually when you speak about something bound to a Maven phase you usually use the term "goal" instead of MOJO, which is an implementation detail. At least, that's what I usually read...

Maybe yes, at least a goal could be bound to different phases and are named Mojo.Name in the code, in the end I personally don't care much about how it is named :-)

> In any case, do I need to modify somehow the PR? Note that a few terms were wrong anyway ;)

If you are fine with it we can merge this right now.

PR #654, opened 2022-02-15
NPE in MirrorMojo when using target platform as source

mbooth101:
Source is allowed to be unset when sourcing bundles from the target
platform (i.e. when targetPlatformAsSource is set) so just add a
null check and continue.


laeubi:
Cherry picked to 2.7.0 as well

PR #668, opened 2022-02-21
[Tycho-Build] remove duplicated dependency

HannesWell:
When building Tycho I get a warning about a duplicated dependency `org.eclipse.tycho:org.eclipse.tycho.p2.maven.repository` one with specified version and one with a variable `${project.version}`. I think the variable version should be used instead.

I actually discovered this on the 2.7 branch but I expect the warning to persist on the master as well.


akurtakov:
Thanks!

PR #673, opened 2022-02-21
Fix #670 - Significant target-resolution runtime regression

### LINKED ISSUE #670: Significant target-resolution runtime regression
As reported in https://github.com/eclipse/tycho/discussions/651#discussioncomment-2218409 the initial resolution of the target platform is magnitudes slower with the current Tycho `2.7.0-SNAPSHOT` compared to `2.6.0`, in my case ~30min compared to ~30sec before.

I have the impression that each dependency of each plug-in being built is fully fetched from a mirror again, even if it was already fetched for another plug-in which has the same dependency in the same Maven build. Therefore the runtime becomes linear to the number of references to dependencies and not only linear to the cardinality of the set of dependencies (which would be relatable).
Furthermore the dependencies should not be re-downloaded on each build which happens often (or maybe always, I cannot say that clearly from the print-out).

Please find attached a relativly small reproducer that contains three identical plug-ins that have the same dependencies:
[tycho-2-7.zip](https://github.com/eclipse/tycho/files/8108483/tycho-2-7.zip)

laeubi:

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.p2.maven.repository/src/main/java/org/eclipse/tycho/repository/local/LocalArtifactRepository.java
+++ b/tycho-bundles/org.eclipse.tycho.p2.maven.repository/src/main/java/org/eclipse/tycho/repository/local/LocalArtifactRepository.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2021 Sonatype Inc. and others.
+ * Copyright (c) 2008, 2022 Sonatype Inc. and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -9,10 +9,9 @@
  *
  * Contributors:
  *    Sonatype Inc. - initial API and implementation
+ *    Christoph Läubrich - #670 - Issue  Significant target-resolution runtime regression 
  *******************************************************************************/
 package org.eclipse.tycho.repository.local;
-
-import static java.util.stream.Collectors.toSet;
 
 import java.io.BufferedOutputStream;
 import java.io.File;
@@ -27,6 +26,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.Collectors;
 
 import org.eclipse.equinox.p2.core.IProvisioningAgent;
 import org.eclipse.equinox.p2.metadata.IArtifactKey;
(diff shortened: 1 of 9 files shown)


laeubi:
With locally applying https://github.com/eclipse/tycho/pull/682 the integration test pass with this change so it seems an issue that the itest is not executed with the tycho version of the current build.

laeubi:
@HannesWell @akurtakov I hope the remaining test failure is now fixed, if all tests pass i'd like to merge this as a first improvement and look at the SWT issue reported in the issue.

HannesWell:
> @HannesWell @akurtakov I hope the remaining test failure is now fixed, if all tests pass i'd like to merge this as a first improvement and look at the SWT issue reported in the issue.

The state from yesterday afternoon worked perfectly fine for my case. From comparing the latest force-push with its predecessor I assume it was just a rebase, so I am fine with merging this. But I can test it this evening (merged or from this branch).

laeubi:
@akurtakov do you have any idea what might be wrong here? The problem sees that the "local-repo" could not be read but I don't get how it is created (or not created) in a faulty way...

laeubi:
I also added SWT bundle as per @akurtakov mentioned case, it seems the current patch also fixes this...

HannesWell:
I just tested the current state back-ported to the 2.7 branch and it worked perfectly fine for my case.

akurtakov:
> @akurtakov do you have any idea what might be wrong here? The problem sees that the "local-repo" could not be read but I don't get how it is created (or not created) in a faulty way...

I just managed to look at it locally. Reproduced locally by doing:
* `mvn clean install -DskipTests=true` in tycho top level
* cd tycho-extras/tycho-p2-extrast-plugin
* `mvn clean verify -Pits`

Fails with '''Caused by: org.eclipse.equinox.p2.core.ProvisionException: No repository found at file:/home/akurtakov/git/tycho/tycho-extras/tycho-p2-extras-plugin/target/local-repo/.''' from build.log file. I haven't investigated further what happens.

akurtakov:
@laeubi Look at invoker.properties file. It defines 11 different scenarios to be tested with details about build result to be expected, profiles to be enabled, project to use as test root and etc.

laeubi:
@akurtakov Thanks for taking a look, what I don't get is that `tycho-p2-extras-plugin/target/local-repo/` is the *maven* repository used to isolate the integration test... so what has tycho to do with that? But I'll try to investigate further in this area. Sadly it seems ther is no way to just list all commands the invokerplugin would issue as far as I can see... So one has to build them from the properties files what is kind of cumbersome I don't even see that there is a way to se *what* call fails...

laeubi:
What I have done now:

Remove all but one item from `tycho/tycho-extras/tycho-p2-extras-plugin/src/it/baseline/invoker.properties`

```
# Build baseline
invoker.goals.1 = verify -DforceContextQualifier=original
invoker.buildResult.1 = success
```

`mvn clean verify -Pits` fails ...

Now I perform the following steps:

1. Copy the folder `tycho/tycho-extras/tycho-p2-extras-plugin/src/it/baseline` to temp folder
2. `/tmp/baseline$ mvn -Dtycho-extras-version=3.0.0-SNAPSHOT -Dtycho-version=3.0.0-SNAPSHOT verify -DforceContextQualifier=original`

The build succeeds without an error ...

laeubi:
I compared this with the 2.7.x line and the only difference in the "local-repo" is that for 2.7.x there is an empty '.meta/p2-artifacts.properties' file while with my change there is no '.meta/p2-artifacts.properties' why this cause the error is a bit strange as P2 most probably does not know anything about this file...

laeubi:
I think the crucial part is here:

[`org.eclipse.tycho.repository.local.LocalArtifactRepositoryFactory.load(URI, int, IProgressMonitor)`](https://github.com/eclipse/tycho/blob/6689f646354ada88d84e46981da93155e04fcd3a/tycho-bundles/org.eclipse.tycho.p2.maven.repository/src/main/java/org/eclipse/tycho/repository/local/LocalArtifactRepositoryFactory.java#L40-L49)

where it explicitly checks for the presence of the file.

akurtakov:
I confirm this fixed the repeated downloads when building eclipse.platform.releng

laeubi:
I have cherry picked this to the 2.7.x branch, so we hopefully will get an official build "soon" (about 1 1/2 hr).

HannesWell:
> I have cherry picked this to the 2.7.x branch, so we hopefully will get an official build "soon" (about 1 1/2 hr).

Great! I will test it again as soon as it is available.

PR #674, opened 2022-02-21
[Oomph-Setup] add missing/remove obsolete projects and remove mavendev

HannesWell:
This adapts Tycho's Oomph-setup to recent project changes and adds the new projects (`tycho-build` and `tycho-gpg)` and removes `tycho-releng` and moves `tycho-its` to its place in the otherwise alpabetically sorted list of projects.

Furthermore the `com.ifedorenko.m2e.mavendev.feature` is removed from the P2-Director since it uses internal API of M2E that has been removed recently (https://github.com/eclipse-m2e/m2e-core/pull/548) and therefore makes it impossible to launch Maven builds from the IDE.

### DIFF
--- a/setup/Tycho.setup
+++ b/setup/Tycho.setup
@@ -90,8 +90,6 @@
     <requirement
         name="org.sonatype.m2e.plexus.annotations.feature.feature.group"/>
     <requirement
-        name="com.ifedorenko.m2e.mavendev.feature.feature.group"/>
-    <requirement
         name="com.ianbrandt.tools.m2e.mdp.feature.feature.group"/>
     <requirement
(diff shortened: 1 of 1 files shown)


setup/Tycho.setup
```diff
@@ -305,13 +317,6 @@
           rootFolder="${git.cloneTycho.location}/tycho-testing-harness"/>
       <description>Import Maven Project</description>
     </setupTask>
-    <setupTask
```
laeubi, line 308: Why not import integration tests?

HannesWell: The integration tests have just been moved upwards because they were the only import not in alphabetical order.

PR #680, opened 2022-02-22
Testcase for Issue #673

laeubi:
FYI @HannesWell I think I was able to extract a minimal test-case for the problem, lets see what the CI build think about it...

### DIFF
--- a/tycho-its/projects/p2Repository.duplicateDownload/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/p2Repository.duplicateDownload/META-INF/MANIFEST.MF
@@ -1,0 +1,8 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Plugin 1
+Bundle-SymbolicName: issue-670
+Bundle-Version: 1.0.0.qualifier
+Automatic-Module-Name: plugin.1
+Bundle-RequiredExecutionEnvironment: JavaSE-11
+Require-Bundle: com.google.guava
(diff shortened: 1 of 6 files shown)


HannesWell:
Do I read correctly from the test-case that you achieve to avoid re-downloads in subsequent builds?

laeubi:
> Do I read correctly from the test-case that you achieve to avoid re-downloads in subsequent builds?

Yes, it should download it once when upgrading from Tycho 2.6 > Tycho 2.7 as if you have an empty m2 repository and afterwards do not download it again. There are some nasty cases all over the code (pack200, source generation, local artifacts installed, feature-root files, ...) so I hope to have cached them all correctly now and the integration tests run fine...

HannesWell:
> > Do I read correctly from the test-case that you achieve to avoid re-downloads in subsequent builds?
> 
> Yes, it should download it once when upgrading from Tycho 2.6 > Tycho 2.7 as if you have an empty m2 repository and afterwards do not download it again. There are some nasty cases all over the code (pack200, source generation, local artifacts installed, feature-root files, ...) so I hope to have cached them all correctly now and the integration tests run fine...

That's great news! 🎉
One re-download is totally fine. I will just test the current state of your PR.

HannesWell:
Isn't the issue #670? #673 is the fix of the issue.

laeubi:
@akurtakov can you check if you can enhance the testcase to reproduce your issue?

akurtakov:
https://github.com/akurtakov/tycho/commit/d7ea5fd7b8b98c0ef9b650b3096d78c537ffeb57 shows the issue with downloading swt always. There are others but I'm not going to try covering them as they are probably caused by same root cause.

laeubi:
Yes one example is enough, I'll try to take a look at it.

PR #682, opened 2022-02-23
Fix #681 move integration tests to the itest module.

### LINKED ISSUE #681: Move tycho-source-feature-plugin its to tycho-its
tycho-source-feature-plugin uses groovy and the maven-invoker-plugin to execute integration test. This is a bit alien on how we do integration testing with tycho and should be migrated to better integrate with the current layout.

This also leads to the situation that currently those tests are not executed as part of the "integrationtest check" of GH actions.

laeubi:


laeubi:
@akurtakov WDYT?

laeubi:
e.g in this example: https://ci.eclipse.org/tycho/job/tycho-github/job/PR-673/12/ it says "No Test failures" but the build actually has failed due to test failures.... then I have to search the logs for the groovy output and try to guess whats going on (no idea if I can debug groovy tests without groovy install...). Beside that, I have no idea why the tests fail, (no build output as with our regular itest).

akurtakov:
I still think it's better to have IT tests per module. Aggregating tests in different module makes it harder to find out which tests are for given module. Yes, as you pointed out, I can run given test only but I have to know it's exact name and noone knows the names of the tests for each module. 
I fully get that there are tests which test multiple modules in the same time which do belong in tycho-its though.
With all that said - it's second best to have the tests in tycho-its than lose them when we delete tycho-source-feature-plugin.
GH action should be changed to run all ITs  as I would strongly oppose moving the ITs for other projects that are not up for removal.

laeubi:
Another problem seems that these tests are using the tycho snapshot repository, but should actually use the current build-output.

laeubi:
> I still think it's better to have IT tests per module.

If it is possible yes, but I don't see how this could work for tycho as we don'T have modules that can run standalone. AFAIK only the "tycho-extra" has some integration tests in the module and reference the tycho-snapshots repos so are always one build behind (or I just don't get how they work).

laeubi:
@akurtakov it seems the licence tool fails on the builserver can you take a look whats wrong there?
```
1) [Guice/ErrorInjectingMethod]: RuntimeException: ConnectException
  at ClearlyDefinedSupport.bootstrap(ClearlyDefinedSupport.java:179)
  at LicenseToolModule.configure(LicenseToolModule.java:43)
  at LicenseChecker.clearlyDefinedService(LicenseChecker.java:24)
      \_ for field clearlyDefinedService
  at LicenseToolModule.configure(LicenseToolModule.java:41)
```

PR #691, opened 2022-02-25
Update Jenkinsfile

### LINKED ISSUE #689: Builds are triggered twice if waiting for the nex build executor
Currently each change at master triggers a new build of all PR. If there are not enough executors, and one pushes another commit to the master, a new job is started, waiting for a new executor and so on, this can easily stack up.

Maybe we could simply disable concurrent builds like described here: https://stackoverflow.com/a/41516843/9503281

@mickaelistria and @akurtakov what do you think?

laeubi:
Fix #689

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,6 +2,7 @@ pipeline {
 	options {
 		timeout(time: 180, unit: 'MINUTES')
 		buildDiscarder(logRotator(numToKeepStr:'10'))
+		disableConcurrentBuilds(abortPrevious: true)
 	}
 	agent {
 		label "centos-8"


laeubi:
Lets see if that works :-)

laeubi:
@akurtakov I think we first should get this done to disable the concurrent builds... from that the build started successful, I assume that at least the syntax is supported so far :-)

akurtakov:
So does it mean it will not have concurrent builds of given PR or just that only one PR can be built at given time? If the later that would mean huge delays.

laeubi:
This should schedule only one build per branch.

Given that we only have two executors there could only be two concurrent PRs at the same time...

akurtakov:
That would be best solution right now.

laeubi:
Here is an example where a new build replaced an old running one:

https://ci.eclipse.org/tycho/job/tycho-github/view/change-requests/job/PR-662/7/

Issue #713, opened 2022-03-03
Wrong error message for failure creating config.ini


ajohnson1:
I think this would generate the message:

>Error creating .eclipseproduct file.

when the failure was creating config.ini

https://github.com/eclipse/tycho/blob/645e591f40ffb1b6e8492b9c1c6fac2907d186a9/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/ProductExportMojo.java#L453

PR #717, opened 2022-03-03
Fix error message in exception for config.ini

ajohnson1:
Fixes issue #713
Indicate the actual file with the problem


mickaelistria:
Thanks!

Issue #590, opened 2022-01-30
Make P2ArtifactRepositoryLayout actually lookup artifacts


laeubi:
currently the `P2ArtifactRepositoryLayout` only provides a disabled repository. We should implement it that way that it is possible to resolve `p2.` prefixed artifacts e.g in the form of:

```
<dependency>
  <groupId>p2.eclipse.plugin</groupId>
  <artifactId>org.eclipse.osgi</artifactId>
  <version>3.17.0.v20210823-1805</version>
  <type>eclipse-plugin</type>
</dependency>
```

if I add a repository of layout p2 it should be able to resolve this to the correct item.

mickaelistria:
If we keep type `eclipse-plugin` or other tycho-specific types, we can just ignore the groupId and filter on the type.

laeubi:
Yes that's the idea here. in the above example one would simply search for an IU with `id=org.eclipse.osgi` and `version= 3.17.0.v20210823-1805` of `namespace=osgi.bundle` (what we might can support as an alternative type name)

laeubi:
Sadly maven does not support version-range resolving through custom layouts, I've opened https://issues.apache.org/jira/browse/MNG-7421

laeubi:
We might be able to generate a https://maven.apache.org/ref/3.2.5/maven-repository-metadata/ for a P2 repository with tycho.

PR #718, opened 2022-03-03
 Make P2ArtifactRepositoryLayout actually lookup artifacts

laeubi:
Currently the P2ArtifactRepositoryLayout only provides a disabled repository. This implement it that way that it is possible to resolve p2. prefixed artifacts e.g in the form of:

<dependency>
  <artifactId>org.eclipse.osgi</artifactId>
  <version>3.17.0.v20210823-1805</version>
  <type>eclipse-plugin</type>
</dependency>

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,12 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## 4.0.0 (under development)
 
+#### P2 repository now actually lookup artifacts
+
+Previously the P2 repository was just a "marker" for Tycho to know what repositories are there and was actually disabled.
+From now on P2 repositories can be used to actually look up artifacts even if not a Tycho build is executed, this becomes useful if you want to consume a P2 item in a regular maven build.
+A full example can be found here: https://github.com/eclipse/tycho/tree/master/tycho-its/projects/p2mavenDependency/pom.xml
+
 ### Migration guide 3.x > 4.x
 
 #### Properties for tycho-surefire-plugin's 'useUIThread' and 'useUIHarness' parameters
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/LoggerProgressMonitor.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/LoggerProgressMonitor.java
@@ -1,0 +1,75 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Christoph Läubrich and others.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ * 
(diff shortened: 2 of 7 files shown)


mickaelistria:
Please allow me some more time to think about it. As it can be the trigger of more important change, I'd rather take some time to feel comfortable with it than rushing it in as modifying is often harder.

mickaelistria:
Can you please include an itest to showcase how this can be used?
Would one who use the tycho-maven-plugin extension also need to include this plugin as build extension or is the inclusion implicit?

laeubi:
> Can you please include an itest to showcase how this can be used? 

For example it could be used like this (**this does not currently works with this change**):
```
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">

	<groupId>org.eclipse.tycho-its</groupId>
	<artifactId>p2mavendependency</artifactId>
	<version>1.0.0</version>

	<packaging>jar</packaging>

	<properties>
		<tycho-version>3.0.0-SNAPSHOT</tycho-version>
		<p2-repo>https://download.eclipse.org/tools/orbit/downloads/2022-03/</p2-repo>
	</properties>

	<build>
		<plugins>
			<plugin>
				<artifactId>p2-maven-plugin</artifactId>
				<extensions>true</extensions>
			</plugin>
		</plugins>
	</build>

	<repositories>
		<repository>
			<id>p2-data-repo</id>
			<layout>p2</layout>
			<url>${p2-repo}</url>
		</repository>
	</repositories>

	<dependencies>
		<dependency>
			<groupId>p2.pluginx</groupId>
			<artifactId>org.slf4j.api</artifactId>
			<version>1.7.30</version>
			<type>eclipse-plugin</type>
		</dependency>
	</dependencies>

</project>
```

> Would one who use the tycho-maven-plugin extension also need to include this plugin as build extension or is the inclusion implicit?

no for tycho users nothing would change.

mickaelistria:
> this does not currently works with this change

What is missing for such thing to work? Can making it work be in the scope of this change?

> no for tycho users nothing would change.

Good.

mickaelistria:
My concerns is more that adding another plugin is there is currently no functional value isn't beneficial: more modules => more complexity. So I would rather wait for #590 to be working before we merge this refactoring. Those can be different commits on the same PRs.

laeubi:
Sure, this might just take some time until fixes from P2 buble up the chain to tycho but we are probably not in a rush here...

laeubi:
BTW this will probably allow Tycho to consume e.g. P2 from I-Builds rather than waiting for the release/deployment to maven central.

laeubi:
This works now locally on my machine but is currently blocked by [Bug 579072](https://bugs.eclipse.org/bugs/show_bug.cgi?id=579072).

laeubi:
@merks this shows usage of P2 repositories without OSGi running.

laeubi:
Closing this as the plan is to get rid of P2 as much as possible what will require to rework the concept of artifact downloads as well.

PR #698, opened 2022-02-28
Increase bootstramp version of Tycho to 2.7.0

vogella:
Without this setting the local Tycho build fails for me very early with
errors bundlelocation not found

PR #722, opened 2022-03-07
Fix #708 - Ignore local items for tycho builds by default

### LINKED ISSUE #708: Ignore local items for tycho builds by default
As described in #696 sometime releases and snapshots diverge. One problem might be locally build units.

We should change the build and deploy instructions to ignore locally build units, as a default this can the be override if needed (e.g. testing a custom P2 distribution).

laeubi:
Any concerns? This should make local build much more reproducible in case of someone also do some platform development.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,1 @@
+-Dtycho.localArtifacts=ignore
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -125,7 +125,7 @@ Tycho makes heavy use of p2 functionality. Therefore it may be useful to try out
 1. Get the p2 sources (see [p2 project information](http://projects.eclipse.org/projects/rt.equinox.p2/developer))
 2. Make changes in the p2 sources, **(!) don't forget to increase the version of that bundle otherwise your changes will be overwritten with the current release version (!)**
 3. Build the changed p2 bundles individually with <tt>mvn clean install -Pbuild-individual-bundles</tt> (see [Equinox/p2/Build](https://wiki.eclipse.org/Equinox/p2/Build) for more information)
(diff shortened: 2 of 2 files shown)


mickaelistria:
I think changing that would suddenly become inconsistent with how Maven behaves. Maven users expect than building A and then building B (that depends on A) will get B seeing the recent changes from A. I don't think it's a good thing to change it by default, unless Maven itself changes it.

laeubi:
this is only for the tycho build (actually tycho-bundles build) so users are not really affected.

mickaelistria:
Still, people who contribute to Tycho would expect to have it work normally and react to change in upstream deps.
If this saves time on CI, this flag would better be set on Jenkinsfile & various GitHub worklows.

mickaelistria:
Still, it's consistent with Maven which IMO matters more than some people being surprised that modifying their local repo can modify their builds.

laeubi:
> Still, it's consistent with Maven which IMO matters more than some people being surprised that modifying their local repo can modify their builds.

I don't agree here. Tycho always chooses the highest version, while maven only matches the *exact* version for local artifact.

mickaelistria:
> Tycho always chooses the highest version, 

Nope, Tycho picks the version that's suggested by p2, which is not necessarily the highest version and actually happens to be the version with the smallest dependency tree in practice.

> maven only matches the exact version for local artifact.

Nope. If your project has multiple transitive dependencies on an artifact, with those dependencies requiring different versions, than Maven will choose one version for it. What Maven will choose depends on what you previously installed in your local Maven repo.

laeubi:
I still think all of this is not relevant for tycho here and harms more than it helps. Anyways if there so so many concern I would simply drop this, but would the ask to work at #698 with some priority to get reliable releases.

akurtakov:
So should this one be closed?

Issue #719, opened 2022-03-04
Performance regression in 2.7.0 / 3.0.0 (maybe caused by #626?)


sratz:
I've noticed that in 2.7.0 the dependency and classpath resolution phases are slower than in 2.6.0.

For a non-parallelized build this adds about 7-10 minutes to our build time (727 artifacts):
2.6.0: ~ 25 minutes
2.7.0: ~ 35 minutes

I am currently trying to get some more concrete numbers with hot caches and populated ~/.m2/repository.

First bisecting show that this could be caused by #626.

sratz:
So I did some testing using 2.7.0 vs. 2.7.0 + reverted 774e135844b20b46ef71b6de4463126aa0f6da9b from #626:
Same machine, first a build using `mvn clean verify` to warm up, then a 2nd run using
`mvn clean verify -o` to force offline mode and work on caches only:

Here are the numbers from the 2nd runs:
```
                              2.7.0       2.7.1-SNAPSHOT (2.7.0 + reverted 774e135844b20b46ef71b6de4463126aa0f6da9b)
dependency resolution phase   00:08:17    00:03:33
build phase (clean verify)    00:19:25    00:18:58
--
total                         00:27:42    00:22:31
```

laeubi:
Can you set a breakpoints to see how many units are queried? I think we might use some caching because most of the units should be the same for all projects.

laeubi:
@sratz can you check if that reduces the execution times in your case?

https://github.com/eclipse/tycho/pull/721

Beside that, you should really try to enable parallel execution, as we have enabled parallel compile again in Tycho 2.7 this should reduce the build times significantly if building a lot of modules.

sratz:
Already better with #721, but it still bottlenecks in the loop over addMatchingFragments().

I've added another improvement on top of your change in #724.

With that, he performance is almost on par again with 2.6.0.

PR #724, opened 2022-03-07
Issue #719 Performance regression with fragment additons to classpath #2

sratz:
Based on #721 with additional optimization.

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/resolver/P2ResolverFactoryImpl.java
+++ b/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/resolver/P2ResolverFactoryImpl.java
@@ -19,8 +19,6 @@ import java.util.AbstractMap.SimpleEntry;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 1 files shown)


laeubi:
@sratz thanks for this improvement, I also noticed that there is an "resolved IU" that has attached fragments so we probably can use that in the future, but I think for now with these enhancements we should be fine as a first step :+1:

sratz:
Branch should be up-to-date.

I'm trying to make sense of the test failures and if/why they are caused by this change.

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/resolver/P2ResolverFactoryImpl.java
```diff
@@ -134,33 +132,34 @@ public P2ResolverImpl createResolver(MavenLogger logger) {
...
-            Set<IInstallableUnit> dependencyFragments, IInstallableUnit unitToMatch) {
-        Iterator<Entry<IInstallableUnit, IRequiredCapability>> iterator = fragmentsList.iterator();
-        while (iterator.hasNext()) {
-            Entry<IInstallableUnit, IRequiredCapability> fragment = iterator.next();
-            if (fragment.getValue().isMatch(unitToMatch)) {
-                dependencyFragments.add(fragment.getKey());
-                iterator.remove();
-            }
-        }
+        Map<String, List<IInstallableUnit>> availableIUsById = availableIUs.stream()//
```
laeubi, line 136:
```suggestion
        Map<String, List<IInstallableUnit>> availableIUsById = resolvedUnits.stream()//
```

sratz: Oops :)

PR #737, opened 2022-03-08
Remove deprecated packaging types

mickaelistria:


tycho-p2/tycho-p2-facade/src/main/java/org/eclipse/tycho/p2/resolver/P2DependencyResolver.java
```diff
@@ -449,7 +449,7 @@ private DependencyArtifacts doResolveDependencies(MavenSession session, MavenPro
     private boolean isAllowConflictingDependencies(MavenProject project, TargetPlatformConfiguration configuration) {
         String packaging = project.getPackaging();
 
-        if (PackagingType.TYPE_ECLIPSE_UPDATE_SITE.equals(packaging)
+        if (PackagingType.TYPE_ECLIPSE_REPOSITORY.equals(packaging)
```
laeubi, line 452: I wonder if this should not be removed here as the `TYPE_ECLIPSE_REPOSITORY` was previously not included here?

tycho-release/tycho-versions-plugin/src/main/java/org/eclipse/tycho/versions/manipulation/CategoryXmlManipulator.java
```diff
@@ -126,4 +125,11 @@ public void writeMetadata(ProjectMetadata project) throws IOException {
         // this manipulator does not add any restrictions on version changes allowed for eclipse-repository projects
         return null;
     }
+
+    private static String rewriteFeatureUrl(String url, VersionChange change) {
```
laeubi, line 129: Is this related?

mickaelistria: Not really, it could be extracted into a separate patch

akurtakov:
This PR tries to achieve too much at once thus I'll be trying to handle things separately so I can actually merge them rather then constantly add to the pile. E.g. #865

laeubi:
> This PR tries to achieve too much at once thus I'll be trying to handle things separately so I can actually merge them rather then constantly add to the pile. E.g. #865

Good idea, I have just boot up the build-node to speed up PR build queue, should we then close this one for now?

akurtakov:
Yeah, let's close it.

PR #738, opened 2022-03-08
Warning against switching between Tycho versions

hd42:
Include warning against switching between Tycho versions in release notes.


mickaelistria:
Thank you!

PR #739, opened 2022-03-08
Fix #736 TestPluginMojo should have an option for concurrency level

### LINKED ISSUE #736: TestPluginMojo should have an option for concurrency level
According to the discussion below it would be useful to have an option to control the overall (build) concurrency level of the execution.

### Discussed in https://github.com/eclipse/tycho/discussions/735

<div type='discussions-op-text'>

<sup>Originally posted by **sratz** March  8, 2022</sup>
I've experimented with parallel builds using mvn -T1C and Tycho 2.7.0.

While this does give a significant performance boost, it also introduces some instability:

We have a lot of unit tests that do UI testing, e.g. verifying that certain controls have focus / are disabled / etc.

These tests become flaky when multiple test processes are executed simultaneously:

- Different Eclipse instances popping up and stealing focus
- Heavy load extends net test runtime per artifact leading to different timings

Is there a way to have -T1C only affect the dependency resolution/compile/package phases, but keep the test/verifiy/integration-test phases -- or specifically invocations of tycho-surefire-plugin -- linearized?</div>

laeubi:


tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/ConcurrencyLock.java
```diff
@@ -0,0 +1,44 @@
...
+    public AutoCloseable aquire(int concurrencyLevel) throws InterruptedException {
+        if (concurrencyLevel < 1) {
+            return () -> {
+                //nothing to do
+            };
+        }
+        int aquire = 1;
+        synchronized (this) {
+            if (semaphore == null) {
+                semaphore = new Semaphore(concurrencyLevel);
```
sratz, line 31: the semaphore is initialized with the first value of concurrencyLevel.

This all only works correctly when all projects in the reactor share the same value for concurrencyLevel, i.e. is typically configured top-level.

Would it make sense to add a consistency check (remember this first value of concurrencyLevel, compare the current one with the initial one, bail out if mismatching)?

tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java
```diff
@@ -674,18 +683,26 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
-                            + "' is unkown. Allowed values: 'default', 'p2Installed'.");
+                            + "' is unknown. Allowed values: 'default', 'p2Installed'.");
                 }
             }
             if (equinoxTestRuntime != null) {
-                runTest(equinoxTestRuntime);
+                try (AutoCloseable runLock = CONCURRENCY_LOCK.aquire(reactorConcurrencyLevel)) {
+                    runTest(equinoxTestRuntime);
+                } catch (InterruptedException e) {
+                    return;
```
sratz, line 700: Thread.interrupt()?

tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/ConcurrencyLock.java
sratz: Ah, never mind. I didn't fully understand earlier that this is accounted for in the `else` part.

laeubi: As mentioned in the Release notes and javadoc, only the smallest number of all configured values in the reactor will be applied as it is otherwise impossible to ensure a consistent state. So one could see the configured value as an upper-bound but not of an guarantee. That's similar to requesting a parallel build on a single core processor where effectively only one CPU could be used even if a higher degree of concurrency is requested.

PR #745, opened 2022-03-09
Inject dependencies for all Tycho types

mickaelistria:
Getting a proper dependency:list can always be useful.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java
@@ -14,6 +14,7 @@
 package org.eclipse.tycho.core.maven;
 
 import java.io.File;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


jan-pesta:
Would be possible to get this fix into 2.7.5 + build. 

We have tool which is needed to have resolved dependencies for eclise-feature to be able to perform needed steps. 

This is caused by fix for Bug 567760.

I am unable to use 3.0.x yet

laeubi:
> Would be possible to get this fix into 2.7.5 + build.

Regarding back-ports please see https://github.com/eclipse-tycho/tycho/blob/master/CONTRIBUTING.md#backporting for details, but this is very unlikely to happen if not driven by a funding imitative as it requires a complete release for an outdated branch.

> I am unable to use 3.0.x yet

It might be easier to lift that, actually there is no reason for not using 3.x and if there are any problems better these get fixed there then relying on old 2.x line.

laeubi:
> I would like to use 3.0.x but it is forcing to use Java 17, which is not possible at my project now, we still need to stick to 11 for some time.

Tycho needs to run the build with Java 17, your code can still use Java 11 (or even Java 1.8).

PR #752, opened 2022-03-11
Example of a custom P2 transport using hc4 with peristent and 404 aware cache.

laeubi:
@mickaelistria @akurtakov as discussed I extracted my custom transport into a tycho use case example. To enable it, one needs to add  `-Dtycho.p2.transport=tycho` to the commandline. This improves the performance also by caching 404 and break out early, not authentication is currently supported and of course there might be still some edge cases and general improvement so this is just a POC, for example there is no cache eviction.

What also would improve the situation if P2 could gather more context from the Transport, e.g. currently if one downloads from a different mirror then the cache is not hit, obviously it would be great if the transport would know "this is a mirror request of URL x" so it could simply store the file under the original URL and if there is already something stored don't need to bother the mirror server at all.


laeubi:
Probably we could even thing to use standard Java 11 client: https://openjdk.java.net/groups/net/httpclient/intro.html

mickaelistria:
> Probably we could even thing to use standard Java 11 client:

That would be ideal.

laeubi:
> > Probably we could even thing to use standard Java 11 client:
> 
> That would be ideal.

only problem is that is does not support caching out-of-the-box as far as I understand...

laeubi:
In favor of #798

PR #753, opened 2022-03-11
Add feature with souce bundles to mixed.reactor IT test

akurtakov:


laeubi:
@mickaelistria @akurtakov I wonder how we should handle this, I see two options:

1. We could automatically handle this (probably taking `sourceMode` of the target into account)
2. We could require to add/configure the "source-bundle" mojo to explicitly enable this.

akurtakov:
Automatic sounds better to me.

laeubi:
By the way it seems that including a pom-deried bundle does not work at all for a feature even removing the source reference fails with:  org.eclipse.tycho.artifacts.DependencyResolutionException: eclipse-plugin artifact with ID "felix.bundle" and version matching "0.0.0" was not found in the target platform.

laeubi:
Al right it works, but the name of the bundle was wrong, I improved the error message in https://github.com/eclipse/tycho/pull/756 to say now:
xecution default-package-feature of goal org.eclipse.tycho:tycho-packaging-plugin:3.0.0-SNAPSHOT:package-feature failed: eclipse-plugin artifact with ID "felix.bundle" and version matching "0.0.0" was not found in the target platform, **did you probably mean: org.eclipse.tycho.it.felix.bundle:1.0.0.SNAPSHOT** -> [Help 1]

laeubi:
It seems maven is not generating sources by default so we probably should require it is enabled like this[ as described here](https://maven.apache.org/plugin-developers/cookbook/attach-source-javadoc-artifacts.html):
```

    <plugin>
      <groupId>org.apache.maven.plugins</groupId>
      <artifactId>maven-source-plugin</artifactId>
      <executions>
        <execution>
          <id>attach-sources</id>
          <goals>
            <goal>jar</goal>
          </goals>
        </execution>
      </executions>
    </plugin>
```

laeubi:
I have added support for this and instructions on how to use it here: https://github.com/eclipse/tycho/pull/758

laeubi:
@akurtakov do you like to try out if you can adjust your PR according to the new documentation here: https://github.com/eclipse/tycho/blob/master/RELEASE_NOTES.md#automatic-generation-of-pde-source-bundles-for-pom-first-bundles

laeubi:
@akurtakov I think it would be good to enhance this to a working example so we could merge it.

PR #758, opened 2022-03-11
Fix #757 - add a mojo to support generation of PDE headers

### LINKED ISSUE #757: Provide a mojo to generate pde headers for a plain source jar
Now we have mixed reactor builds there is a demand to use pom-first bundles inside PDE. As PDE requires some special headers we need a way to generate them in a standard maven build.

laeubi:
This is the very first working solution. We might want to improve later on the following things:

1. Check if we can hook into the `maven-source-plugin` to not need to explicitly configure `generate-pde-source-header` to run after the `maven-source-plugin`. e.g. we can call this automatic if the `plugin-source` is enabled
2. Check if we can call the necessary generation of P2 metadata automatically.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,66 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## 3.0.0 (under development)
 
+### Automatic generation of PDE source bundles for pom-first bundles
+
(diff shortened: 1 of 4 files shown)


laeubi:
@HannesWell maybe also interesting for you ...

HannesWell:
> @HannesWell maybe also interesting for you ...

It is indeed. Thanks!

PR #782, opened 2022-03-16
Fix IT tests to run with Java 17

akurtakov:
Java 17 no longer supports 1.6 bytecode target so have to update to
newer maven-compiler-plugin and change source/target values where set.
XTend 2.25 also has issues with Java 17 which are fixed by moving to
2.26.
Custom profile IT test had to provide newer JavaSE osgi.ee capability as
the old 3.4.2 eclipse is no longer starting with Java 17.


laeubi:
> Java 17 no longer supports 1.6 bytecode target so have to update to
> newer maven-compiler-plugin and change source/target values where set.

@akurtakov I have added a ticket [here](https://github.com/eclipse/tycho/issues/522) that we probably should drop support for any java < 9 for tycho 3.x what do you think?

Issue #726, opened 2022-03-07
Failing tests on maven 3.8.5


laeubi:
The following integration tests fail on maven 3.8.5
```
[ERROR] Errors: 
[ERROR]   MultiplatformProductTest.exportProduct:27 » Verification Error in execution: [...
[ERROR]   PomDependencySystemScopedTest.testSystemScopedDependenciesIgnored:32 » Verification
[ERROR]   TestNGBundleTest.test:32 » Verification Error in execution: [ERROR] 'dependenc...
[ERROR]   Tycho154BundleJarTest.test:26 » Verification Error in execution: [ERROR] 'depe...

```

laeubi:
The full error is for example:

Error in execution: [ERROR] 'dependencies.dependency.systemPath' for org.eclipse.tycho.tycho-its.tycho-154:bundle:eclipse-plugin:simple.jar is missing. @
	at org.eclipse.tycho.test.tycho154.Tycho154BundleJarTest.test(Tycho154BundleJarTest.java:26)

so probably this is a tycho issue.

mickaelistria:
If this is not a Tycho issue, then is it a Maven issue?

laeubi:
Haven investigated yet any further, we might can have a gh action for maven 3.8.5 as described on the malinglist by Mikael Barbero:

> curl -sSJOLf https://dist.apache.org/repos/dist/dev/maven/maven-3/3.8.5/binaries/apache-maven-3.8.5-bin.tar.gz
> $ tar zxf apache-maven-3.8.5-bin.tar.gz 
> $ export PATH="$(pwd)/apache-maven-3.8.5/bin:${PATH}"
> $ mvn -v 

https://docs.github.com/en/actions/using-github-hosted-runners/customizing-github-hosted-runners

but the eror message suggest that a systempath is not set, so maybe maven has ignored this before and now print it as an error ...

mickaelistria:
Could those failures be related to this regression in Maven: https://issues.apache.org/jira/browse/MNG-7432 ?

laeubi:
Maybe yes, maybe no :-)
I'm currently a bit short in time, maybe I can try that out next week, at laest there is a Quickfix so if you check that out and compile a local maven you could try if that makes a difference.

mickaelistria:
This seems fixed thanks to your extra commit for #789

PR #783, opened 2022-03-16
Run IT tests with same Maven version as dependencies.

akurtakov:


akurtakov:
@laeubi Any idea what these failures are about? Loooks like Tycho is facing issues with Maven 3.8.5.

laeubi:
> @laeubi Any idea what these failures are about? Loooks like Tycho is facing issues with Maven 3.8.5.

I have not yet investigated these failures as I haven't found how to enable maven 3.8.5 for the itest, but now they seem to be executed I could probably take a look whats wrong with them. See also #726 I think there is additional error output or maven is more strict about system scoped dependencies now (and that's probably an error of tycho).

mickaelistria:
This patch is more than correct and this Maven 3.8.5 story shows that this patch is actually a must-have for proper testing of Maven (that wasn't done properly earlier). So let's merge, even if it shows some issues with Maven 3.8.5.
I think we'll have to revert the move to Maven 3.8.5 anyway.

Issue #711, opened 2022-03-03
Create a p2-mavenplugin


laeubi:
Currently we use P2 embedded inside OSGi, but the tycho-build extension already have made some efforts to use P2 standalone in a maven context.

We should further follow this path to make P2 more accessible to tycho, probably even maven some of the mojos that do not really interact with tycho so the plugin could be used independent from tycho, e.g. for the [dash-licenses](https://github.com/eclipse/dash-licenses) (@waynebeaton ) tool.

Because of that, I'd like to propose the name "p2-maven-plugin" as we already have "tycho-maven-plugin" the scope of this should be:

1. resolve solutions for a given set of IUs
2. read/write/update repositories
3. provide tasks similar to the "ant-based-tasks" (e.g. mirror application and alike)

In the end this will save us a lot of headaches and probably we could finally even drop the embedded OSGi runtime. A side effect (but not a goal) might be that we can drop the ant-tasks completely from platform decreasing the maintenance effort there so p2 can focus on its core functionality.

mickaelistria:
Do you have a concrete use story in mind that could benefit from this?
I think the main point is to allow p2 to work without OSGi, so it's mostly development in p2, isn't it?

laeubi:
e.g. https://github.com/eclipse/tycho/issues/590 would benefit from it as well. the p2 repository layout is not really "tycho" but would make is possible to use p2-update sites in a regular maven build as well. So that is my first "milestone" here.
Even more I'd like to have target platform resolution happen in the maven reactor instead of the OSGi runtime so we do not pas back-and-forth so much stuff and it could be performed as part of the tycho-build extension.

mickaelistria:
> would make is possible to use p2-update sites in a regular maven build as well. 

OK. that's an interesting story, however I don't think a dedicated mojo is necessary right now to achieve that. But it's all up to you ;)

laeubi:
We have the p2-plugin and use t already on some places so I think this could be closed, further work can use more dedicated tickets.

PR #787, opened 2022-03-17
Issue #711 create a p2-maven plugin

laeubi:

### DIFF
--- a/p2-maven-plugin/pom.xml
+++ b/p2-maven-plugin/pom.xml
@@ -1,78 +1,57 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ - Copyright (c) 2022 Christoph Läubrich and others.
+ - All rights reserved. This program and the accompanying materials
+ - are made available under the terms of the Eclipse Public License v1.0
+ - which accompanies this distribution, and is available at
+ - http://www.eclipse.org/legal/epl-v10.html
+ -
+ - Contributors:
+ -    Christoph Läubrich - initial API and implementation
+ -->
+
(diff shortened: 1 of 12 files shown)


laeubi:
@mickaelistria this already improves two things:

1. p2 (maven) dependencies are now managed in the `p2-maven-plugin` instead of the `tycho-build`
2. The P2 repository layout is defined in `p2-maven-plugin` instead of `tycho-core`

Further moves are all blocked by current p2 work (e.g. https://github.com/eclipse-equinox/p2/issues/20 ) and it becomes hard to keep this up-to date for a long time, so I'd like to merge this in the current state.

Probably it would be possible to move all the transport and custom tycho p2 stuff here as well, but this most probably would require more bundles using package imports, I'll try to investigate and open appropriate issues later on.

mickaelistria:
That looks interesting. However, we cannot use the `org.eclipse.p2` namespace for packages. Tycho packages have to start with `org.eclipse.tycho` as per Eclipse Development Process specification.

laeubi:
I could move it to whatever (just have to make sure it does not clash with any existing package) but we also have `org.eclipse.sisu` in tycho repo so probably there are a few exceptions from that rule?

mickaelistria:
There are exceptions, or issues, some because of history (code moved between projects and there is an agreement to keep package name). But for new code, it's always best to have `org.eclipse.<project>` as package prefix.

laeubi:
Maybe we can make an exception here if we kindly ask the p2 team to use the prefix `org.eclipse.p2.maven` (as this is not a bundle it should not be an issue anyways), because the idea is to have this only use p2-stuff separated from Tycho it seems odd to name it with something 'tycho' in it then... Probably we can even migrate the plugin to p2 but that seems to add a lot of inflexibility.... probably sometimes later when there is not much changes anymore?

laeubi:
> p2 will never host Maven plugins

Why not? I think there is no technical reason as many projects offer a maven/grade/... plugin.

> f you want to start a new project for this code, feel free

Not sure if that's a bit oversized ... anyways I opened https://github.com/eclipse-equinox/p2/issues/23 as this would be a perquisite for any project depending on p2 not be delayed to much in case of changes.

mickaelistria (approved):

PR #794, opened 2022-03-20
Fix #792 - Support exclusion of certain dependencies from pom dependency consideration

### LINKED ISSUE #792: Support exclusion of certain dependencies from pom dependency consideration
We currently have the ability to add additional bundles for tests, but we do not have a way to *exclude* something, e.g. if something is pulled in by P2 but causes problems.

One example is https://mvnrepository.com/artifact/com.google.code.findbugs/jsr305/3.0.2 that if added as a maven dependency with `pom=consider` is chosen as a provider for the package `javax.annotation` and the causes issue because it does not contains the `@PostConstruct` for example.

laeubi:


target-platform-configuration/src/main/java/org/eclipse/tycho/target/TargetPlatformConfigurationMojo.java
```diff
@@ -1,7 +1,21 @@
+/*******************************************************************************
```
laeubi, line 1: @mickaelistria I guessed the copyright header from the git-history, could you confirm that this is correct?

mickaelistria: Thanks, it's correct.
Shame on me for missing it earlier!

Issue #797, opened 2022-03-23
Implement a caching P2 transport


laeubi:
Currently certain access to files (mostly p2.index) are not cached making tycho quite slow. Another issue are non-existing index files if p2 tries to query to find out where the meta-data resides.

laeubi:
It seems this has increased the buildtime of the unit test a lot, I'll take a look if this is just because of now we have cold-caches or if there are other issues, probably we need to disable the transport by default (currently enabled) until it is clear what causing this, lets wait for the next build ...

laeubi:
Build times are back to normal

laeubi:
@jcompagner do you like to try it out? https://github.com/eclipse/tycho/blob/master/RELEASE_NOTES.md#improved-p2-transport-for-more-efficiently-http-cache-handling-and-improved-offline-mode

If you set `-Dtycho.p2.transport.min-cache-minutes=40320` Tycho should only contact update-site servers once a month, so if your targets are really static this should improve build times and save you from server outages.

jcompagner:
i will have a look (but busy in the comming 2 weeks with our own release, then i am back on master where this kind of stuff is in)
i guess i need then the nighly build right (to get 3.0)
what if i don't set that -Dtycho.p2.transport.min-cache-minutes=40320 ? what is the default value?

laeubi:
default is one hour.

PR #798, opened 2022-03-23
Issue #797 - add a custom transport based on URL#openStream() with caching of 404 responses

laeubi:
This is a draft on how Tycho could use a custom transport. This also is a showcase on how an alternate P2 transport might work, to investigate what is required and give a bigger picture for possible P2 enhancements.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,31 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## 3.0.0 (under development)
 
+### Improved P2 transport for more efficiently http-cache handling and improved offline mode
+
+P2 default transport is more designed as a weak cache that assumes the user is always online.
+While for an IDE that might be sufficient as target resolution is only performed once in a while and updates are triggered by explicit user request, for tycho this does not work well:
+
+- Builds are often trigger on each code change, requiring repeated target resolution
+- Builds might be asked to run in an offline mode
+- If there is a temporary server outrage one might to fallback to the previous state for this build instead of fail completely
+- Build times are often a rare resource one don't want to waste waiting for servers, bandwidth might even be limited or you have to pay for it
+
+Because of this, Tycho now includes a brand new caching P2 transport that allows advanced aching, offline handling and fallback to cache in case of server failures. The transport is enabled by default so nothing has to be done, just in case you want the old behavior you can set `-D=tycho.p2.transport=ecf` beside that the following properties might be interesting:
+
+#### Force cache-revalidation
+If you run maven with the `-U` switch Tycho revalidates the cache, this is useful if you have changed an updatesite in an unusual way (e.g. adding new index files) as tycho now also caches not found items to speed-up certain scenarios where many non existing files are queried.
+
+#### Configure minimum caching age
+
+Some servers don't provide you with sufficient caching information, for this purpose, tychy by default assumes a minimum caching age of one hour. You can switch this off, or configure a longer delay by using `-Dtycho.p2.transport.min-cache-minutes=<desired minimum in minutes>`.
+Choosing a sensible value could greatly improve your build times and lower banYdwith usage. If your build contains a mixture of released and 'snapshot' sites you have the following options:
+
+1. Consider adding a mirror to your settings.xml for the snapshot page that point to a file-local copy (e.g. output of another build)
+2. Configure the webserver of your snapshot site with the `Cache-Control: must-revalidate` header in which case tycho ignores any minimum age
+3. Use `-Dtycho.p2.transport.min-cache-minutes=0` this will still improve the time to resolve the target
+
+
 ### Automatic generation of PDE source bundles for pom-first bundles
 
 PDE requires some special headers to detect a bundle as a "Source Bundle", there is now a new mojo `tycho-source-plugin:generate-pde-source-header` that support this, it requires the following configuration:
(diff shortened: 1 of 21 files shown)


laeubi:
For measurement purpose i use the following simple project and add some statistics:
```
<?xml version="1.0" encoding="utf-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
    xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">

    <groupId>com.test</groupId>
    <artifactId>p2cache</artifactId>
    <version>1.0.0</version>
    <packaging>eclipse-plugin</packaging>

    <properties>
        <tycho.version>3.0.0-SNAPSHOT</tycho.version>
    </properties>

    <repositories>
        <repository>
            <id>eclipse</id>
            <layout>p2</layout>
            <url>https://download.eclipse.org/releases/2022-03/</url>
        </repository>
        <repository>
            <id>eclipse-link</id>
            <layout>p2</layout>
            <url>https://download.eclipse.org/rt/eclipselink/updates/</url>
        </repository>
    </repositories>
    
    <build>
        <plugins>
            <plugin>
                <artifactId>tycho-maven-plugin</artifactId>
                <extensions>true</extensions>
            </plugin>
        </plugins>
    </build>
</project>
```

This results in: total 35 seconds for 104 requests where 56 not found response.

laeubi:
The same with a simple etag based cache requires on a hot-cache only 7 seconds, this is mostly because poor caching performance see https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1039

mickaelistria:
Overall, I don't think we should add into Tycho a transport that's not maintained by p2. I see only pitfalls in maintaining multiple strategies for the same thing in different projects.
Would it be possible that you implement such a transport directly in p2 now?

laeubi:
If we have a working example in tycho and there is conses in P2 team I probably could think about migrating the code to P2 but most likeley this will be a longer process as there is some strong demand to not change much in P2 at the moment...

mickaelistria:
There is a consensus in the p2 team already that a solution without ECF is desired. There may be some demand to not change too much in p2 at once, but Tycho shouldn't be treated as p2's sandbox and p2 has to evolve. This transport can be an extra transport in p2, not opt-in by default (until we remove ECF) and Tycho could then be configured to pick one transport over the other.

laeubi:
> There is a consensus in the p2 team already that a solution without ECF is desired

At least in the tickets I opened it was rather controversial to even change small things in that area...

> There may be some demand to not change too much in p2 at once

I think it won't work well without changing "too much" and I'm not sure if I can afford all the discussions and required changes (we would need to adjust in other areas also for sure as ECF is somehow referenced directly internally) without any immediate benefit. Especially we would need to add some "hooks" to support tycho what most likely will soon enter the discussion loop why P2 needs to change and if not tycho could simply implement it ...

Even if it could be solved in the 2022-06 release this is three month away and I'd like to benefit from it as soon as possible as 

1. it will greatly reduce build times
2. Will support true offline mode making current workarounds in tycho obsolete
3. Will allow defeat bad caching behaviors by a user configurable forced caching delay ( @jcompagner  has already asked for it long time ago)
4. I'm a bit tired of all these "please don't break user X that might relies on Y but we never know and are all concerned it ends in a big desaster" :-\

laeubi:
This is now feature complete.

Final measurements for my example on second run to ensure caches are filled:

1. Standard P2 transport: 48 seconds
2. Tycho transport with disabled min-cache: 47 seconds
3. Tycho transport with 1hr min cache: 5 seconds

laeubi:
@mickaelistria @akurtakov all checks are green, I would like to merge this after squashing/rebase so we can get feedback as soon as possible?

mickaelistria:
Can you please create a new bundle from it with the necessary bits only, in order to increase change that this code one day gets moved to p2? In the current form, it seems like we add maintenance (and value) in Tycho without any positive effect on p2; while conceptually p2 is supposed to be the right place to fix things. So let's ensure we don't start building too much of this logic in a way that cannot then be backported to p2.

laeubi:
> Can you please create a new bundle from it with the necessary bits only

This is not really suitable to be placed in a new bundle as it needs to interact with maven/tycho here.

> it seems like we add maintenance (and value) in Tycho without any positive effect on p2

As we do on many other places as well. The problem is that tycho requirements are often not p2 requirements...

> too much of this logic in a way that cannot then be backported to p2

It would be a lot of work to really integrate this in P2 as it is not just prting a few classes but would require to adjust P2 at just a few more places and as long as P2 wan't sto maintan strickt backward compatibility this is even more challenging without any gain (from tycho point of view...).

mickaelistria:
OK, feel free to merge.

mickaelistria:
Actually, before we merge it, can't we try removin https://github.com/eclipse/tycho/blob/b8183fd42099c701ce6fe0fe36e77e524fa3e899/tycho-bundles/tycho-bundles-external/tycho-bundles-external.product#L43..L49 at the same time?

laeubi:
I'll try that, but PC directly references ecf on some places ...

laeubi:
Hm... this won't work as we then need to remove ecf completely and there is no way for the user to choose ecf as an alternative way... So I think we would need to keep this for a while.

mickaelistria:
OK.

PR #813, opened 2022-03-29
Update maven.yml

laeubi:
Cache by base ref

### DIFF
--- a/.github/workflows/licensecheck.yml
+++ b/.github/workflows/licensecheck.yml
@@ -29,9 +29,9 @@ jobs:
       uses: actions/cache@v3
       with:
         path: ~/.m2/repository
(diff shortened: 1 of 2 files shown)


akurtakov:
As my GH actions-fu is not that good - what is gained that way? Don't hesitate to put the reply in the git commit too for further reference. Note that GH is still new for eclipse community so every bit of knowledge matters.

laeubi:
This is that each branch get its own cache, nothing one needs to care about most of the time, but as the m2 cache of Tycho 2.6.x is incompatible with Tycho 2.7.x+ (what is used internally when building 2.7.x) I try to separate these two caches.

laeubi:
by the way this should actually not be necessary so this is just an experiment...

laeubi:
Obsolete

PR #816, opened 2022-03-30
Update extensions.xml

laeubi:
Maven 3.8.5 finally supports a placeholder here.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,7 @@ pipeline {
 		label "centos-8"
 	}
 	tools {
-		maven 'apache-maven-latest'
+		maven 'apache-maven-3.8.5'
 		jdk 'openjdk-jdk11-latest'
 	}
 	stages {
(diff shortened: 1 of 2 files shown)


laeubi:
On the jenkis `apache-maven-latest` is still 3.8.4 so we have to wait with this a bit more... but the GH action already shows that this will work then.

laeubi:
> On the jenkis `apache-maven-latest` is still 3.8.4 so we have to wait with this a bit more...

It obviously would be possible to just set `apache-maven-3.8.5` as a requirements here...

akurtakov:
Could https://ci.eclipse.org/tycho/job/tycho-github/view/change-requests/job/PR-866/12/console be related ?

mickaelistria:
> Could https://ci.eclipse.org/tycho/job/tycho-github/view/change-requests/job/PR-866/12/console be related ?

Nope. I enabled option to cleanup workspace before checkout (from Jenkins job config), and it seems to work better.

laeubi:
Sadly sometimes the downloads get wrecked because of network problems. The builds produce a huge amount of network traffic, for the buildnode I have setup I see a total of 28 GB already transferred the last days...

PR #817, opened 2022-03-30
Update Jenkinsfile

laeubi:
Exclude MD files and github workflow changes from triggering a build


akurtakov:
Is this one still planned/needed?

laeubi:
Actually it would be useful but it seems I have to find out the correct syntax for this. The idea is that we do *not* trigger a CI build if the github actions changes or if only documentation is changed as this is wasted time then...

akurtakov:
This one conflicts already. Should we keep it?

laeubi:
I'll close this for now, seems there is not much interest...

PR #824, opened 2022-04-01
Update maven.yml

laeubi:
add -T1C for integration test


laeubi:
One test seems have issues with that (other is a timeout), but speed improvement is quite low.

Issue #697, opened 2022-02-28
Failed to resolve dependencies with Tycho 2.7.0 for custom repositories


vogella:
With Tycho 2.7.0 we see errors for resolving artifacts from internal Nexus servers.

[ERROR] Failed to execute goal on project a: Could not resolve dependencies for project a: Could not find artifact b:4.20.1 in central (https://repo.maven.apache.org/maven2) -> [Help 1]

It should go to the configured URL.
```
<repositories>
			<repository>
				<id>nexus</id>
				<url>https://nexus.client.de/repository/secure-public</url>
			</repository>
		</repositories>
```
@HannesWell IIRC you implemented this. Does it still work for you with Tycho 2.7.0?

laeubi:
I think you should provide a bit more context here:

- Is this a snippet from a target, a pom or a settings.xml?
- are there any additional errors/warning in the log
- ...

vogella:
> Is this a snippet from a target, a pom or a settings.xml?

Target. 
> are there any additional errors/warning in the log

No.

We reverted this build to Tycho 2.6.0 and had to delete the content of .mvn/repository after going back to 2.6.0. 
Unfortunately I'm not aware of other open Nexus server which I could use to build a standalone example which does not require internal network access. Is there any?

vogella:
~/git/tycho/tycho-its/projects/target.mavenMulti seems to contains a IT test but does currently not work as it pointsto               <tycho-version>2.5.0-SNAPSHOT</tycho-version>

But if I change it to 2.7.0 it still works, looks to me that is downloads part from Maven central and parts from the other repo. 

````
[WARNING] No explicit target runtime environment configuration. Build is platform dependent.
[DEBUG] [main] Computing preliminary target platform for MavenProject: org.eclipse.tycho.itests:test.bundle:0.0.1-SNAPSHOT @ C:\Users\vogella\git\tycho\tycho-its\projects\target.mavenRepos\test.bundle\pom.xml
[DEBUG] Using execution environment 'JavaSE-17' configured in Currently running profile, newer than configured profile (JavaSE-11) from [Bundle-RequiredExecutionEnvironment (unique entry)]
[DEBUG] Currently running JRE matches source level for JavaSE-17; current JRE system packages are used.
[DEBUG] Adding target definition file "file:/C:/Users/vogella/git/tycho/tycho-its/projects/target.mavenRepos/test.bundle/../test.target"
[...]

PR #827, opened 2022-04-01
issue #697 reproducer

pzi-dsa:
The test case attempts to build two plugins with dependencies from an external (Nexus) repository.

The first plugin depends on a Maven artifact which is rebundled and builds successfully.
The second plugin depends on an OSGi bundle which can't be resolved during the build.

The issue has been introduced with Tycho 2.7.0.

### DIFF
--- a/tycho-its/projects/issue697/bundle1/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/issue697/bundle1/META-INF/MANIFEST.MF
@@ -1,0 +1,8 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: bundle
+Bundle-SymbolicName: bundle
+Bundle-Version: 1.0.0.qualifier
+Bundle-RequiredExecutionEnvironment: JavaSE-11
+Automatic-Module-Name: bundle
+Require-Bundle: wrapped.de.monticore.monticore-runtime;bundle-version="7.2.0"
--- a/tycho-its/projects/issue697/bundle1/build.properties
+++ b/tycho-its/projects/issue697/bundle1/build.properties
@@ -1,0 +1,3 @@
+source.. = src/
+bin.includes = META-INF/,\
+               .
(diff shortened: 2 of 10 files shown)


pzi-dsa:
> Then when the target retries it might has already failed the attempt and did not try again with the different repositories (probably a bug in maven).

I agree, it seems like the problem only appears, whenever dependencies are resolved via Maven. I assume the problem only appeared with 2.7.0 because of the addition of Maven metadata to be bundled artifacts. Since then Maven is used directly, in favor of Tycho.

> [DEBUG] Resolving artifact de.monticore:dsltool-api:jar:0.0.1 from [central (https://repo.maven.apache.org/maven2, default, releases)]

I noticed that the nexus repository is not included in the brackets. Instead it looks like Maven only uses central, default and releases to find the artifact. Is Maven aware of the repositories specified in the target platform definition?

During separate tests, I replaced the URL in either the *central*, *default* or *releases* repository with the one from the nexus repository. In all cases, Tycho/Maven was then able to find all required dependencies. My guess is that either the resolver instance used by Maven is created before Tycho adds the repositories from the target platform to it and is then not properly updated, or the repositories aren't added to begin with.

laeubi:
> I noticed that the nexus repository is not included in the brackets. Instead it looks like Maven only uses central, default and releases to find the artifact. Is Maven aware of the repositories specified in the target platform definition?

No because target platform resolution has nothing to do with maven resolution, even though both use the same resolver!
I have to check if thats a problem how tycho calls the resolver or if that is a bug in maven.

laeubi:
Okay what happens here is the following:

The target is resolved and there everything works fine, but maven records the repository that was used to resolve the artifact:
```
dsltool-api-0.0.1.jar>rwth=
dsltool-api-0.0.1.pom>rwth=
```
Then we add this to the maven model as a regular dependency and maven tries resolve this as well. Because the repository is not configured, maven think something has changed and do an update with maven central, where it is not found.

So what would work is to configure all repositories as well for maven as for the target, but that's kind of undesirable...

There are some ways to handle this, but I'm not sure each has some drawbacks:

1. we could mark the dependency as optional, while this works its a bit a hack because it will simply ignore the failure
2. we could add all repositories from the target to the maven model, but for sure this might also be undesirable because now we enhance the maven resolution with things from the target and possibly mix different target location items 
3. We could assemble a local artifact repository and only add resolved items from the target

laeubi:
@pzi-dsa thanks for the reproducer, I have merged this and fixed it with https://github.com/eclipse/tycho/pull/855

PR #831, opened 2022-04-02
Test custom jenkins build node

laeubi:
This is just for testing the custom build node


laeubi:
JDk8 configuration is missing, beside that with ~1hr run time the performance is quite okay, I think it would be useful to probably exclude the node from building the master branches and focus on PR building...

laeubi:
Custom Build-Node can now fully build tycho with the restriction that it is unable to deploy snapshots, I have opened tickets at EF to improve the labels for the executors so we can better control this in the future.

Issue #845, opened 2022-04-05
Feature restrictions are not taken into account when using emptyVersion


laeubi:
Given a feature like this:

```
<?xml version="1.0" encoding="UTF-8"?>
<feature
      id="com.test.sample.feature"
      version="1.0.0">
   <includes id="com.test.base.feature" version="0.0.0"/>
   <requires>
      <import feature="com.test.base.feature" match="compatible" version="1.0.0"/>
   </requires>
</feature>
```

[according to the docs](http://help.eclipse.org/stable/topic/org.eclipse.platform.doc.isv/reference/misc/feature_manifest.html) a match rule of `match="compatible"` means:

> dependent plug-in version must be at least at the version specified, or at a higher service level or minor level (major version level must equal the specified version).

So here one would expect that only a version < 2 is considered a valid match, but Tycho still tries to resolve e.g. a version 2.0 and fails in this case if not all requirements for 2.0 are fulfilled, there is an example project here: https://github.com/and-k/tycho-issue-new

As a workaround one could specify `<includes id="com.test.base.feature" version="1.0.0"/>` but this duplicates the information from the restriction therefore it seems to be related to the use of the `emptyVersion` here.

FYI @and-k

laeubi:
The error one sees here is:

> [ERROR] Cannot resolve project dependencies:
> [ERROR]   Software being installed: com.test.sample.product 1.0.0.qualifier
> [ERROR]   Missing requirement: com.test.base.feature.feature.group 2.0.0 requires 'org.eclipse.equinox.p2.iu; org.apache.activemq.activemq-core [5.4.0,5.4.0]' but it could not be found
> [ERROR]   Cannot satisfy dependency: com.test.sample.feature.feature.group 1.0.0 depends on: org.eclipse.equinox.p2.iu; com.test.base.feature.feature.group [2.0.0,2.0.0]
> [ERROR]   Cannot satisfy dependency: com.test.sample.product 1.0.0.qualifier depends on: org.eclipse.equinox.p2.iu; com.test.sample.feature.feature.group [1.0.0,1.0.1)

The important part is `com.test.base.feature.feature.group 2.0.0 requires 'org.eclipse.equinox.p2.iu; org.apache.activemq.activemq-core [5.4.0,5.4.0]' but it could not be found`where one can see that the **version 2** is chosen for `com.test.base.feature.feature` (what has an unsolvable requirement here on intention).

[...]

PR #846, opened 2022-04-05
Consider restrictions when resolving the build version of depending features/plugins #845

laeubi:


laeubi:
FYI @and-k

PR #859, opened 2022-04-07
Speed up P2InstalledRuntimeTest.

akurtakov:
It's the slowest test in our builds as of today. These changes combined
cut 20+ sec on my machine (from almost 80 sec to a bit under 60).


akurtakov:
Using less of Oxygen repo and newer maven plugins reduces network traffic as more things are already downloaded by previous modules/tests is what I noticed. Oxygen usage seems to add a lot - probably because it's a composite repo containing R, SR1 and SR2 repos thus delaying resolving.

laeubi:
And most noticeable it uses `archive.eclipse` ... good catch, reminds me of https://github.com/eclipse/tycho/issues/594 we should simply drop all usages of such old updatesite for tycho 3.x I think whenever possible.

PR #871, opened 2022-04-09
Ensure latest P2 repo for tests is set the same

akurtakov:
And in only one place. Currently we had a mix of 2020-03 and 2022-03.


laeubi:
Maybe we can also define a default property in the test (like `tycho-version`) named `eclipse-latest` and then try to use this consitently all over the place? currently there is a mixture of different "namings"

PR #882, opened 2022-04-12
Support Eclipse features via m2 repositories

pzi-dsa:
Currently all Maven artifacts are treated as OSGi bundles.

If one wants to assemble a product or build features including other features, those have to either be part of the same reactor build or be contributed via a p2 repository.

This commit checks whether an artifact contains a 'feature.xml'. If so, it is treated as a feature, otherwise as a bundle.


mickaelistria:
That's a good patch! Would you be able to also add a test case that covers it, to prevent from further regressions?

pzi-dsa:
I've added a test case which tries to build a feature depending on JGit. In case of a regression, I expect an error of the form:
`Missing requirement: feature.feature.group 0.0.1.qualifier requires 'org.eclipse.equinox.p2.iu; org.eclipse.jgit.feature.group 0.0.0' but it could not be found`

> Do you plan to add support for this in m2e as well?

I'm not overly familiar with the m2e code, but sure, I'll try to create a pull request there, as well.

mickaelistria:
Thank you!

vogella:
For the future: Can you add a little bit more descriptive commit messages? "[Detect Eclipse features.]" is not really sufficient to see what the purpose of the commit was.

PR #886, opened 2022-04-13
Fix #885 - 'includePackedArtifacts' must be automatically disabled when running with an incompatible vm

### LINKED ISSUE #885: [2.7.1] 'includePackedArtifacts' must be automatically disabled when running with an incompatible vm
p2artifacts.xml looks like this:

```
<?xml version='1.0' encoding='UTF-8'?>
<?artifactRepository version='1.1.0'?>
<artifacts size='2'>
  <artifact classifier='osgi.bundle' id='com.google.guava' version='21.0.0.v20170206-1425'>
    <processing size='1'>
      <step id='org.eclipse.equinox.p2.processing.Pack200Unpacker' required='true'/>
    </processing>
    <properties size='11'>
      <property name='artifact.size' value='692436'/>
      <property name='download.size' value='692436'/>
      <property name='format' value='packed'/>
      <property name='maven-version' value='21.0.0-SNAPSHOT'/>
      <property name='maven-artifactId' value='com.google.guava'/>
      <property name='maven-extension' value='jar.pack.gz'/>
      <property name='maven-groupId' value='org.eclipse.orbit.bundles'/>
      <property name='maven-classifier' value='pack200'/>
      <property name='download.md5' value='7724a7147b6f138046a78076ed2c7810'/>
      <property name='download.checksum.md5' value='7724a7147b6f138046a78076ed2c7810'/>
      <property name='download.checksum.sha-256' value='04105b986e1eb37211f0e42698f1648a2e0cf4ba00282c4932c441386006bb6a'/>
    </properties>
  </artifact>
  <artifact classifier='osgi.bundle' id='com.google.guava' version='21.0.0.v20170206-1425'>
    <properties size='10'>
      <property name='artifact.size' value='2675014'/>
      <property name='download.size' value='2675014'/>
      <property name='download.md5' value='d45d6cd464786a3dfbaf11a4bb718fb1'/>
      <property name='download.checksum.md5' value='d45d6cd464786a3dfbaf11a4bb718fb1'/>
      <property name='download.checksum.sha-256' value='910e145dee0bc3cfcbcc1fa2b80e259a608ec042f9027877271e0a1c6ad7002b'/>
      <property name='maven-version' value='21.0.0-SNAPSHOT'/>
      <property name='maven-artifactId' value='com.google.guava'/>
      <property name='maven-extension' value='jar'/>
      <property name='maven-groupId' value='org.eclipse.orbit.bundles'/>
      <property name='maven-type' value='eclipse-plugin'/>
    </properties>
  </artifact>
</artifacts>
```
and a file `com.google.guava-21.0.0-SNAPSHOT-pack200.jar.pack.gz` is placed in the cache location. under `org/eclipse/orbit/bundles/com.google.guava/21.0.0-SNAPSHOT`

Two things are to note here:

1. as the file seem to originate from a p2-repository it should be placed under `osgi/bundle/com.google.guava` (this is also where the jar is found)
2. the pack200 should not be used at all for java > 14

### Discussed in https://github.com/eclipse/tycho/discussions/877

<div type='discussions-op-text'>

<sup>Originally posted by **sewe** April 11, 2022</sup>
Hi,

recently (after a bigger Tycho upgrade), I get the following warning at the beginning of my build during target platform resolution:

```
[WARNING] eclipse-plugin:com.google.guava:21.0.0.v20170206-1425: /home/sewe/.m2/repository/org/eclipse/orbit/bundles/com.google.guava/21.0.0-SNAPSHOT/com.google.guava-21.0.0-SNAPSHOT.jar can not be represented in Maven model and will not be visible to non-OSGi aware Maven plugins.
```

And indeed, a latter builds step fails because of this:

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:2.7.1:validate-classpath (default-validate-classpath) on project com.teamscale.ide: Execution default-validate-classpath of goal org.eclipse.tycho:tycho-compiler-plugin:2.7.1:validate-classpath failed: bundleLocation not found: /home/sewe/.m2/repository/org/eclipse/orbit/bundles/com.google.guava/21.0.0-SNAPSHOT/com.google.guava-21.0.0-SNAPSHOT.jar -> [Help 1]

org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:2.7.1:validate-classpath (default-validate-classpath) on project com.teamscale.ide: Execution default-validate-classpath of goal org.eclipse.tycho:tycho-compiler-plugin:2.7.1:validate-classpath failed: bundleLocation not found: /home/sewe/.m2/repository/org/eclipse/orbit/bundles/com.google.guava/21.0.0-SNAPSHOT/com.google.guava-21.0.0-SNAPSHOT.jar

Caused by: org.apache.maven.plugin.PluginExecutionException: Execution default-validate-classpath of goal org.eclipse.tycho:tycho-compiler-plugin:2.7.1:validate-classpath failed: bundleLocation not found: /home/sewe/.m2/repository/org/eclipse/orbit/bundles/com.google.guava/21.0.0-SNAPSHOT/com.google.guava-21.0.0-SNAPSHOT.jar

Caused by: java.lang.IllegalArgumentException: bundleLocation not found: /home/sewe/.m2/repository/org/eclipse/orbit/bundles/com.google.guava/21.0.0-SNAPSHOT/com.google.guava-21.0.0-SNAPSHOT.jar
```

What is really strange, however, and might hint at a bug in Tycho is that the very first build after a `rm -r ~/.m2/repository` succeeds, but executing the same `mvn` command again fails with the above warning and error. (Yes, not even the warning is shown during the initial run against the empty repository.)

AFAICT, this happens with Tycho 2.5.0, 2.6.0, 2.7.0, and 2.7.1.

Is there any advice on how to fix this or at least diagose this? I am not sure where the 21.0.0-SNAPSHOT version even comes from.

Best wishes,

Andreas Sewe</div>

laeubi:
This ignores the setting and emits a warning when running on JDK >= 14.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,7 @@ This page describes the noteworthy improvements provided by each release of Ecli
 ## 2.7.2
 Fixes:
 - [2.7.1][regression] Neither raw version nor format was specified #876 
+- [2.7.1] 'includePackedArtifacts' must be automatically disabled when running with an incompatible vm #885 
 
 ## 2.7.1
 
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/resolver/DefaultTargetPlatformConfigurationReader.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/resolver/DefaultTargetPlatformConfigurationReader.java
@@ -9,6 +9,7 @@
  *    Sonatype Inc. - initial API and implementation
  *    Christoph Läubrich -  [Bug 461284] - Improve discovery and attach of .target files in eclipse-target-definition
  *                          [Bug 567098] - pomDependencies=consider should wrap non-osgi jars
+ *                          [Issue #885] - 'includePackedArtifacts' must be automatically disabled when running with an incompatible vm
  *******************************************************************************/
 package org.eclipse.tycho.core.resolver;
 
@@ -167,7 +168,12 @@ public class DefaultTargetPlatformConfigurationReader {
         if (value == null) {
             return;
         }
-        result.setIncludePackedArtifacts(Boolean.parseBoolean(value));
+        boolean include = Boolean.parseBoolean(value);
+        if (include && Runtime.version().feature() >= 14) {
+            logger.warn("Pack200 is not supported when running on Java 14 and higher");
+            return;
+        }
+        result.setIncludePackedArtifacts(include);
     }
 
     private void setTargetDefinitionIncludeSources(TargetPlatformConfiguration result, Xpp3Dom configuration)


mickaelistria:
@laeubi I see more and more effort toward 2.7.x branch, and I believe it's adding some risk on actual capability to deliver a 3.0.0 soon-ish. Is it necessary to fix this kind of issues on the branch? Can't we instead drive users to use 3.0.0-SNAPSHOT, or to prepare them by just telling them to remove such deprecated flag in their configurations?

laeubi:
Actually finding out what was the problem here is more changeling as the fix itself so I think its better to fix than to analyze and instruct people here. Beside that I think to work more uninterrupted on 3.0.0 we need a stable 2.7x release.

mickaelistria:
> Beside that I think to work more uninterrupted on 3.0.0 we need a stable 2.7x release.

I disagree with that. There will always be something to fix or improve in older releases, at some point, stopping this work is the only way to move forward and focus on future releases. People who have real need for fixes in 2.7 can contribute them. Initially, we expected to release 3.0 in end of spring, with all the 2.7 work, this has become unrealistic and 3.0 is now more likely to happen in the end of summer.

mickaelistria:
I would say that as committers who target a 3.0, priority should be given to reviewing PRs, not to create new ones against older branches. But I'm not anyone boss, it's all up to you. It's just a reminder that every work a committer agrees to do against 2.7 does delay 3.0. Particularly for issues that are already fixed in 3.0...

sewe:
FWIW, works for my reproducer. :+1: 

Also, thank you very much @laeubi for contributing this to the 2.7.x line. While I understand the desire to move towards 3.0.0 ASAP, not everyone is able to always use the latest and greatest; unfortunately, sometimes regressions get in the way. In our case, we had to jump "over" multiple minor versions of Tycho because of this. So patch releases like 2.7.1 or 2.7.2 are much appreciated.

laeubi:
This should appear soon in the official [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) when the build is done:
https://ci.eclipse.org/tycho/job/tycho-github/job/tycho-2.7.x/30/

So if you like to try that out and verify it is fixed that would be welcome, if you likes the back-ports of critical fixes consider say [thank you](https://github.com/sponsors/laeubi?frequency=one-time) or even [sponsor further development ](https://github.com/sponsors/laeubi?frequency=recurring) of Tycho.

PR #904, opened 2022-04-21
Stop dealing with Java 16

akurtakov:
It's not an LTS version and thus already dead.

### DIFF
--- a/tycho-core/src/test/java/org/eclipse/tycho/core/ee/StandardExecutionEnvironmentTest.java
+++ b/tycho-core/src/test/java/org/eclipse/tycho/core/ee/StandardExecutionEnvironmentTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2020 SAP AG and others.
+ * Copyright (c) 2010, 2022 SAP AG and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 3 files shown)


laeubi:
Is this really about Java 16? Or Java 6?

laeubi:
Okay I was confused because of the other change then :-)

BTW I always wonder why we remove support for java versions, it seems it does not harm to just keep them there as it is just some constants/switch cases...

tycho-core/src/test/java/org/eclipse/tycho/core/ee/StandardExecutionEnvironmentTest.java
```diff
@@ -61,8 +60,6 @@ public void setUp() throws Exception {
                 new SilentLog());
         javaSE17Environment = ExecutionEnvironmentUtils.getExecutionEnvironment("JavaSE-17", null, null,
                 new SilentLog());
-        javaSE16Environment = ExecutionEnvironmentUtils.getExecutionEnvironment("JavaSE-16", null, null,
```
laeubi, line 64: Maybe instead of removing old version we can have 
`ExecutionEnvironmentUtils.getUnsupportedExecutionEnvironment("JavaSE-16");`
method that prints out an error message "<profilename> is no longer supported" or something?

akurtakov: Let's not complicate things. Dead is dead.

PR #911, opened 2022-04-23
Use specific P2 types

HannesWell:
While going through the code I noticed that there are a few Interfaces that use the generic-wildcard (?) or `Object` instead of the specific P2 types.
Since we are currently preparing a new Major version I want to proposed to use the specific types to get the usual benefits of static types. Or are there reasons to not depend on p2 in the affected modules that speak against it?
Affected are:
- org.eclipse.tycho.core.shared
- org.eclipse.tycho.embedder.shared
- org.eclipse.tycho.p2.resolver.shared

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.p2.resolver.impl.test/src/test/java/org/eclipse/tycho/p2/impl/test/ArtifactMock.java
+++ b/tycho-bundles/org.eclipse.tycho.p2.resolver.impl.test/src/test/java/org/eclipse/tycho/p2/impl/test/ArtifactMock.java
@@ -16,6 +16,7 @@ import java.io.File;
 import java.util.LinkedHashSet;
 import java.util.Set;
 
+import org.eclipse.equinox.p2.metadata.IInstallableUnit;
 import org.eclipse.tycho.IDependencyMetadata;
 import org.eclipse.tycho.IDependencyMetadata.DependencyMetadataType;
 import org.eclipse.tycho.p2.metadata.IArtifactFacade;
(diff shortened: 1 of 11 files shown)


laeubi:
This is because these P2 types only live in the (embedded) OSGi container but cannot be used in the "maven world" even though I'm working on making that possible it is nothing currently possible.

HannesWell:
Ah crap. I should have asked before.
Do you think you will be successful in the foreseeable future so it is worth to keep this around or should I abandon it?

laeubi:
It is more a marathon than a sprint ...
I'm currently try two approaches:

1. do more of P2 stuff inside tycho "maven" and outside OSGi, thats why at least now the P2 classes are available in the maven mojos (but incompatible classloaders) and we can use (string, Int, ...) constants already, also the tycho-build extension uses the Slicer already outside of OSGi
2. Try to convert more P2 bundles to use import package instead of require bundle so we can export the "maven" class realm to the OSGi as system packages see https://github.com/eclipse-equinox/p2/issues/22

akurtakov:
I'll convert this one to draft so it doesn't get build constantly.

laeubi:
@HannesWell as this will likeley not happen in the near future I'll close the PR, we can reopen it once we made more progress in that area!

HannesWell:
> @HannesWell as this will likeley not happen in the near future I'll close the PR, we can reopen it once we made more progress in that area!

As https://github.com/eclipse-tycho/tycho/pull/1178 has showed it seem to be possible now. I noticed there are some changes from this PR left, so lets see what works now.

HannesWell:
Any idea why no tests have been found by the GH build? The Jenkins build looks fine. All tests seem to be executed.

HannesWell:
> Don't know, but the Test-Plugin was updated in the meantime maybe this influenced this?

Just triggered new builds, so lets see what happens. I also deleted the two comments of the test results. Maybe that was also an issue.

laeubi:
Actually when the Jenkins build success we should be fine here ...

HannesWell:
> Actually when the Jenkins build success we should be fine here ...

It succeeded: https://ci.eclipse.org/tycho/job/tycho-github/job/PR-911/1/
So should we just merge it now or await the completion since I started it now?

laeubi:
What ever you prefer ;-)

HannesWell:
> What ever you prefer ;-)

Then lets merge it. There are enough other things to build. 😄

Issue #611, opened 2022-02-02
Support setting CI-Friendly-Versions in tycho-build-extension


laeubi:
As discussed in https://github.com/eclipse/tycho/issues/603 CI-Friendly-Versions might be of high value for ease building consistent versions.

We should add support for automatically setting the following items:

- `{revision}`  initialized with what currently is the `DefaultBuildTimestampProvider`
- `{sha1}` initialized with what currently is the `JGitBuildTimestampProvider`

this could be done in the following way:

1. We add a new LifeCycleParticipant
2. We read from a `tycho.config` file two properties: `cifriendly.revision` and `cifriendly.sha1` if they are `true` we set the corresponding value in the maven session properties in `afterSessionStarts`
3. Everything else stays as is, with the exception that we skip generation of new qualifiers in `BuildQualifierMojo` if there is already a value
4. For Tycho 3.x we can try to overhaul the concept of `BuildTimestampProvider` so it could be used more flexible, e.g. one might want to have `cifriendly.revision` not true/false but having a hint for a specific provider to lookup in the PlexusContainer (would require changing the `BuildTimestampProvider` interface) passing config values and so on

laeubi:
I have investigated here and thing with current maven this is to cumbersome.
I therefore have opened https://github.com/apache/maven/pull/674 and will postpone this for 3.0 for now.

PR #914, opened 2022-04-24
Issue  #611 Support setting CI-Friendly-Versions in tycho-build-extension

laeubi:
Add a first implementation and integration test.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,62 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## 3.0.0 (under development)
 
+### Enhanced Support for Maven CI Friendly Versions
+
+Starting with Maven 3.8.5 Tycho now supports an enhanced form of the [Maven CI Friendly Versions](https://maven.apache.org/maven-ci-friendly.html) beside the standard properties names one could also use:
+
+- releaseVersion
+- major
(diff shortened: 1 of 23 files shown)


laeubi:
FYI @mickaelistria @akurtakov this will make it possible to have easier deployments of stuff with a simple `mvn deploy -Dtycho.buildqualifier.format=yyyyMMddHHmm` where the -SNAPSHOT is then replaced with `.<buildqualifier>` in the maven build (and hopefully then also in the P2 metadata with valid references), I have only tested this with the very basic integrationtest example but like to merge this as a starting point so we can start to play around with it, I'll add some documentation and see if I can enable git-timestamps easily and then it should be fine as a first iteration I think if there are no concerns.

**One important note:** this strictly requires Maven 3.8.5+ and even if we seem to enforce this in the pom I don't see that maven complains when running with an older version...

laeubi:
Documentation and Jgit support added!

laeubi:
This also includes some cleanup and enhancements to JGIt + Build timestamp, I might try to eplit these from the current change...

laeubi:
Trying to split things actually do not make this more readable as some changes are only inspired by the new functionality, if the build succeeds I'd plan to merge this.

mickaelistria:
That's a major progress. Thanks for that!

PR #920, opened 2022-04-27
Drop tests that need Java 8 explicitly to succeed

### LINKED ISSUE #522: Drop official support for non modular vms
We currently have one test that explicitly requires a Java-8 VM in toolchains.

As Java 8 is EOL for a long time and we are about to switch to Java 17 maybe at end of year I think we should drop official support for non modular VMs.

That would mean we

1. remove the test that currently requires Java 8
2. do no longer support bugs/regressions that target Java 8 unless a patch is provided that do not require us to maintain legacy code

akurtakov:
Requiring Java 8 is no longer sane requirement now that there are 2 more
Java LTS versions.
Fixes #522 .


akurtakov:
Unrelated test failure.

laeubi:
@akurtakov when you next time see this we might better want to use https://keyserver.ubuntu.com/ instead of the https://pgp.mit.edu as the later is incredible slow.

Maybe we even can one time use an eclipse one instead: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/797

akurtakov:
Thanks for the hint! I hope to not forget it :)

laeubi:
If you like and are on a flow you can do that right now :-)

PR #933, opened 2022-05-04
p2-maven-site includes bundles in the repository

laeubi:


laeubi:
FYI @reckart the problem could be reproduced with the tycho test- suite as well.

PR #939, opened 2022-05-05
[eclipse/tycho#937] reproducer for 'does not have an expanded version issue

cdietrich:
[eclipse/tycho#937] reproducer for 'does not have an expanded version issue

### DIFF
--- a/tycho-its/projects/tycho937/pom.xml
+++ b/tycho-its/projects/tycho937/pom.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/maven-v4_0_0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<groupId>tycho937</groupId>
+	<artifactId>tycho937-parent</artifactId>
+	<packaging>pom</packaging>
+	<version>2.0.0-SNAPSHOT</version>
+
+	<properties>
+		<tycho-qualifier>yyyyMMdd-HHmm</tycho-qualifier>
+	</properties>
+
+	<modules>
+		<module>bundle</module>
+	</modules>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>tycho-maven-plugin</artifactId>
+				<version>${tycho-version}</version>
+				<extensions>true</extensions>
+			</plugin>
+			<plugin>
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>tycho-packaging-plugin</artifactId>
+				<version>${tycho-version}</version>
+				<configuration>
+					<format>${tycho-qualifier}</format>
+					<strictVersions>true</strictVersions>
+				</configuration>
+			</plugin>
+		</plugins>
+	</build>
+</project>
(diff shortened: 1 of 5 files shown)


cdietrich:
@laeubi if i run this locally i get
[ERROR] The build could not read 1 project -> [Help 1]
org.apache.maven.project.ProjectBuildingException: Some problems were encountered while processing the POMs:
[ERROR] Unknown packaging: eclipse-plugin @ line 7, column 13
how can i avoid it
and if i run the complete build it does not fail.
so i have no clue what causes the problem :(

cdietrich:
is it possible that it is the use of multiple different versions?

laeubi:
If you run it from commandline make sure to pass `-Dtycho-version=....`

laeubi:
> and if i run the complete build it does not fail.

Have you passed the `-Pits` to maven? But be aware this is running very long. So you can ask surefirey to only execute your test with `-Dtest=dull.class.name`

cdietrich:
`-Dtest=dull.class.name -Pits` wont work, cause not all modules have this test

cdietrich:
with mvn clean install -Pits -Dtest=Tycho937Test -DfailIfNoTests=false
i am getting the same problem about unkown packagig.
gonna configure tycho plugin too

laeubi:
> `-Dtest=dull.class.name -Pits` wont work, cause not all modules have this test

You should execute this in the `tycho-its` folder.

tycho-its/projects/tycho937/pom.xml
```diff
@@ -0,0 +1,27 @@
...
+		<tycho-qualifier>yyyyMMdd-HHmm</tycho-qualifier>
+	</properties>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>tycho-packaging-plugin</artifactId>
+				<version>${tycho-version}</version>
+				<configuration>
```
laeubi, line 20: I think you need to add the Tycho plugin here not only the packing plugin!

cdietrich:
yes now test is green so it does not reproduce the problem :(

tycho-its/projects/tycho937/pom.xml
cdietrich: yes now test is green, so it does not show the problem

laeubi:
> yes now test is green so it does not reproduce the problem :(

The testfailure is there the build succeeds see https://github.com/eclipse/tycho/runs/6302785138

tycho-its/projects/tycho937/pom.xml
cdietrich: is it possible jenkins and local(mac) behave differently?

cdietrich:
looks like a jenkins problem
https://ci.eclipse.org/mwe/job/mwe2/job/cd_tycho30stripDown/
is also red but green on my local machine

laeubi:
As mentioned in your example you don't enable tycho and thus the packaging type is unknown.

cdietrich:
@laeubi i can finally reproduce.
updated the pr
javadoc is also involved

laeubi:
Thanks I'll try to take a look at this!

laeubi:
Interestingly this is cause by the **order** of goals:

- `mvn clean javadoc:aggregate-jar verify` -> fails
- `mvn clean verify javadoc:aggregate-jar` -> works

cdietrich:
hmmm since when are goals order dependent

laeubi:
> hmmm since when are goals order dependent

just try `package clean` (will clean your build output after build finished!)

laeubi:
Merged as part of https://github.com/eclipse/tycho/pull/941 thanks for the reproducer

PR #944, opened 2022-05-06
Improve handling / logging of baseline replacements

laeubi:
- use more dynamic default value
- remove the lock
- if baselines checks are enforced, error if no repository is defined
- give better log message in case of missing repositories

Fix https://github.com/eclipse/tycho/issues/943

### DIFF
--- a/tycho-p2/tycho-p2-plugin/src/main/java/org/eclipse/tycho/plugins/p2/BaselineValidator.java
+++ b/tycho-p2/tycho-p2-plugin/src/main/java/org/eclipse/tycho/plugins/p2/BaselineValidator.java
@@ -74,131 +74,147 @@ public class BaselineValidator {
     public Map<String, IP2Artifact> validateAndReplace(MavenProject project, MojoExecution execution,
             Map<String, IP2Artifact> reactorMetadata, List<Repository> baselineRepositories, BaselineMode baselineMode,
             BaselineReplace baselineReplace) throws IOException, MojoExecutionException {
+        if (baselineMode == disable) {
+            return reactorMetadata;
+        }
+        if (baselineRepositories == null || baselineRepositories.isEmpty()) {
+            if (baselineMode == fail || baselineMode == failCommon) {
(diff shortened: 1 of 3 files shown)


mickaelistria:
The commit doesn't fix the issue described in commit message, please rephrase commit message.

laeubi:
> The commit doesn't fix the issue described in commit message, please rephrase commit message.

Why not? It fixes for me (hopefully) the issue I have described that there are strange errors and I have no clue whats wrong beside a not very informative message.

mickaelistria:
See comment on bug message, you're looking at the wrong component and what you suggest here wouldn't fix anything. Please retitle to highlight it's some code or functional improvement, or abandon.

laeubi:
I changed the title if you think its good for anything, my point is that I think I had not been guessing around if the messages are more clear ...

mickaelistria:
>  if you think its good for anything

Yes, good commit messages matter to me (and many others) in several workflows and are a vital for project health when trying to lookup complex and old issues. A bad history makes it harder to resolve those.

> my point is that I think I had not been guessing around if the messages are more clear ...

Sure, and the new commit message is better at telling that; while the previous one was telling something totally different from what it's actually doing. So it's good now.

laeubi:
@mickaelistria would you like to review the code? do you think this is still useful?

mickaelistria:
>  do you think this is still useful?

The code or my review? ;)
I had a very quick look and I'm fine with the overall changes, so if you think it will help in the future, please merge.

laeubi:
At least it causes no regression in the test suite :-D

laeubi:
I'll postpone this as it seem that the actual issue is #947 and if no one else have demands here seem to only partially improve here.
e.g. BND has a https://github.com/bndtools/bnd/blob/master/maven/bnd-baseline-maven-plugin/README.md#full-reporting option to output additional details about the 'why' that would be useful here but as pointed out not (yes) possible due to the current structure (we might need to save some more details from previous stages in the project key-value-store for this).

PR #967, opened 2022-05-19
Ignore .gitignore and .gitatributes file types from packaging

akurtakov:
Caused by https://github.com/codehaus-plexus/plexus-utils/pull/174 .
Fixes https://github.com/eclipse/tycho/issues/966 .


akurtakov:
Speedy merge so we can get a fixed snapshot for platform M3 release.

PR #970, opened 2022-05-22
[Setup] Simplify Maven project import

HannesWell:
As a side effect this now also imports newly added projects.

@merks do you think it would useful to add a `maxDepth` parameter (or similar) to the `(Maven)ProjectImportTask` to limit the search depth if `locateNestedProjects` is true?
This would avoid the workaround using child predicates to prevent deeper nested test-projects to be imported.

### DIFF
--- a/setup/Tycho.setup
+++ b/setup/Tycho.setup
@@ -143,172 +143,22 @@
       xsi:type="setup:CompoundTask"
       name="Maven Project Import">
     <setupTask
-        xsi:type="maven:MavenImportTask">
-      <sourceLocator
-          rootFolder="${git.cloneTycho.location}/tycho-bundles/tycho-bundles-target"
-          locateNestedProjects="true"/>
-      <description>Import Target Platform Project</description>
-    </setupTask>
-    <setupTask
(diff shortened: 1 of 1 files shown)


merks:
I'm not sure how generally useful that would be, but it looks easy to implement, in SourceLocator I assume.

HannesWell:
> I'm not sure how generally useful that would be, but it looks easy to implement, in SourceLocator I assume.

At least Tycho and m2e could use it. But I think it would be useful in general for projects that contain nested test projects that are not intended to be imported into the workspace.
For most Eclipse projects the relevant projects have nesting depth 1 or 2 and test-projects that are nested deeper.
So using a SourceLocator with `locateNestedProjects` and `maxDepth=2` would help there.

I had a quick look into `ProjectImportTaskImpl` and `MavenProjectImportTaskImpl` and I suspect it cannot be implemented generically for both in `SourceLocator` and has to be done for each ProjectImportTask.
But since it is not urgent for me at the moment I will come back later to that, when there is a higher need, unless someone else wants to do it before. :)

PR #976, opened 2022-05-26
Try to remove the boot delegation from the profiles

laeubi:


akurtakov:
Please put more details in the commit message like why it's done? what are the implications, benefits, etc. ? I really hope that at some point Tycho will be able to stop shipping these profiles much like Equinox no longer needs them but never paid more attention to try understand where and how they are used.

laeubi:
I just wanted to see if this breaks something obvious here... seams at least the itest are all running fine. I checked a bit about getting rid of the profiles and noticed that P2 at least has stopped adding this for modular vms so we probably should also do the same here as a first step.

PR #982, opened 2022-06-01
Remove ancient TODOs and simplify javadoc in DirectorMojo

### LINKED ISSUE #949: Rename org.eclipse.tycho.plugins.p2.director.DirectorMojo
It has a TODO 

https://github.com/eclipse/tycho/blob/8dda882a3d77ecba8601f9bba5d9a2ecbc4f6695/tycho-p2/tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/p2/director/DirectorMojo.java#L39-L41

Should we rename it with Tycho 3.x?

akurtakov:
Fixes #949 .


akurtakov:
Unrelated test failure.

PR #984, opened 2022-06-01
Example of converting ReactorRepositoryManagerImpl into a DS component

laeubi:
FYI @akurtakov 

Let me know if there is anything unclear but actually its quite simple :-)


akurtakov:
Thanks, I'll look into it.

akurtakov:
I would like to merge that one, is it ok?

akurtakov:
I just now noticed that the project fails to compile in the IDE. Does it work for you just fine?

laeubi:
What kind of compilation failure do you see? This currently requires m2e + pde to work (unless we add plain m2e support for it).

akurtakov:
"Import org.osgi.service.component.annotations.Activate cannot be resolved" and all other annotations from that package fail similarly. I don't see any dependency in the pom nor anything ds specific in .project/.classpath providing these either so how does m2e/pde figure it out?

laeubi:
Can you do `Project > Settings > DS` and check that "Add DS to the classpath" is checked there?

akurtakov:
image
Looks like it.

laeubi:
Do you see the annotations.jar in the classpath? And is the PDE nature enabled for that project?
grafik

akurtakov:
ds builder is missing. I'll push a patch adding it.

laeubi:
> Actually, .project/.classpath files are ignored and I don't see how to fix it without them.

Yes, the are ignored I think m2e must be "fixed" here to detect the presence of the preference file and then add the appropriate builder. But @mickaelistria  is more familiar with the "import magic" so I can't tell if it is hard to archive, for now you could add it manually, or we can force add the necessary files.

akurtakov:
I think I found the solution. Please look at the patch and confirm that the new xml file should be ignored.

PR #985, opened 2022-06-01
Fix failing CiFriendlyVersionsTest

akurtakov:
It expects the date of last commit to be of given format but the exact
date is not constant as this is the git repo.
Change the test to verify that the qualifier is of 'yyyyMM' format
instead of checking exact value.


akurtakov:
@laeubi What do you think about this change? It's not so exact check but it would at least verify the correct qualifier format still.

PR #990, opened 2022-06-02
Fix having ds annotations on path.

akurtakov:
Ignore generated file. Continuation for  #984

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/OSGI-INF/.gitignore
+++ b/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/OSGI-INF/.gitignore
@@ -1,0 +1,1 @@
+/org.eclipse.tycho.p2.manager.ReactorRepositoryManagerImpl.xml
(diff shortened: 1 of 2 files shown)


tycho-bundles/org.eclipse.tycho.p2.resolver.impl/build.properties
```diff
@@ -15,3 +15,4 @@ bin.includes = META-INF/,\
                about.html,\
                plugin.properties
 source.. = src/main/java/
+additional.bundles = org.eclipse.pde.ds.lib
```
laeubi, line 18: this should really not be nessesary

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/OSGI-INF/.gitignore
```diff
@@ -0,0 +1 @@
+/org.eclipse.tycho.p2.manager.ReactorRepositoryManagerImpl.xml
```
laeubi, line 1: yes we can ignore this file as it is automatically generated

tycho-bundles/org.eclipse.tycho.p2.resolver.impl/build.properties
laeubi: If the DS builder is enabled, it should add this automatically here and no warning should be shown.

akurtakov: OK. I'll still push this one so we unbreak things for now. Once m2e/pde/whatever gets fixed let's remove it.

laeubi: You can better then add an import package (with optional) instead.

akurtakov: Let's see what happens.

Issue #676, opened 2022-02-22
Remove tycho.mode=maven


laeubi:
We should no longer support `tycho.mode=maven` it was often used to package an update-site in a regular maven build or to build "mixed reactors", for both we already have better solutions now.

mickaelistria:
Isn't m2e still using it?

laeubi:
m2e build is currently using this but the goal is to get rid of it with tycho 2.7

HannesWell:
> m2e build is currently using this but the goal is to get rid of it with tycho 2.7

That's right. But I think Tycho should still support cases like the current first build step of m2e. These are edge-cases no question, but it can be handy to build a project with a partly different configuration to for example generate some resource in advance.

I suspect the problem to be that the module containing the target-platform definition does not participate in the build, but I have not verified that.
Given that is the case one could either try to add the TP module to the build or Tycho would not fail in case the TP-module is not found and would just print a warning instead. Alternatively maybe it could work to disable the target-platform-configuration, but I don't think this is possible yet (because that goal is actually never executed), isn't it? But in the end this just re-introduces `tycho.mode=maven` in another way.

HannesWell:
It indeed worked to add the tp-module in the `tycho.mode=maven` build. To not interfere with the regular build I added it in the profile that is activated only when `tycho.mode=maven` was used, please see the second commit in https://github.com/eclipse-m2e/m2e-core/pull/589/
This has the positive effect that only the profile has to be activated and the additional `tycho.mode=maven` property can be skipped, which makes the build command simpler. I think this is an 'workaround' for the better, since I expect a profile to be used in cases similar to the one of M2E anyway.

The only problem left is, that it does not work when the `MANIFEST.MF` was not yet generated and fails with the exception below.

Do you think the project setup could fall back to the pom.xml (which I think has to be present in such cases)?

```
[INFO] Scanning for projects...
[ERROR] Internal error: org.eclipse.tycho.core.osgitools.OsgiManifestParserException: Exception parsing OSGi MANIFEST C:\dev\git\eclipse.m2e-core\m2e-maven-runtime\org.eclipse.m2e.archetype.common\META-INF\MANIFEST.MF: Manifest file not found -> [Help 1]
[...]

PR #993, opened 2022-06-03
Revert "Remove tycho.mode=maven"

akurtakov:
Reverts eclipse/tycho#981

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -59,9 +59,6 @@ Please note that we use another new feature from Maven 3.8.5 here, where you can
 
 You can now control your Tycho version for `.mvn/extensions.xml` and your `pom.xml` in one place and still override it on the commandline with `-Dtycho-version=...`
 
-### Support for -Dtycho.mode=maven removed
-
-The property has never been properly documented and excluding tycho resolution is risky if one is not 100% sure what the effects are (e.g  using it with maven-dependency-plugin will result will many dependency not being listed at all). Clean phase is manually excluded by Tycho as we clearly know that it doesn't need resolution, if there is the need for more such exclusions they should be done similarly.
 
 ### Support for non-modular JVMs no longer top tier
 
--- a/tycho-bundles/tycho-bundles-external/pom.xml
+++ b/tycho-bundles/tycho-bundles-external/pom.xml
@@ -31,7 +31,7 @@
 		1. Run "mvn clean install" in tycho-bundles folder. This is necessary to populate local repo with information
 		   about org.eclipse.tycho.noopsecurity bundle
 		2. If necessary, change p2 repository url in tycho-bundles/pom.xml
-		3. In this folder, run mvn -Dtycho-version=0.11.0-SNAPSHOT org.sonatype.tycho.extras:tycho-version-bump-plugin:0.11.0-SNAPSHOT:update-product
+		3. In this folder, run mvn -Dtycho.mode=maven -Dtycho-version=0.11.0-SNAPSHOT org.sonatype.tycho.extras:tycho-version-bump-plugin:0.11.0-SNAPSHOT:update-product
 		4. Edit tycho-p2-runtime.product and remove version lock on org.eclipse.tycho.noopsecurity bundle
 
 		Luckily, we don't need to include new p2 version too often.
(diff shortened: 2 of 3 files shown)


thannerme:
Hi @akurtakov 

I switched from tycho 2.2.0 to 2.7.4. but somehow the tycho.mode=maven flag does not take effect any more. As far as I see in this PR, it should however again for 2.7.4?

akurtakov:
It is not removed so should be working. Can you share reproducer for what is not working for you?

thannerme:
a bit hard to reproduce in few lines. I execute following command, which now (using new tycho version now, command stayed the same) does the dependency analysis.. but I will check on my side also again then.

mvn -e -f pom.xml -B --no-transfer-progress -s /home/jenkins/workspace/xxxxx -Dstyle.color=always org.eclipse.tycho:tycho-versions-plugin:set-version -Pall -Dtycho.mode=maven -DnewVersion=xx -Dartifacts=root

thannerme:
or is there another possibility to achieve skipping the dependency resolution? (cause you thought about removing the flag)

thannerme:
@laeubi   Do you maybe have an idea or workaround? (I read https://github.com/eclipse-tycho/tycho/issues/676) 

in short: 
- i switched from tycho 2.2.0 to 2.7.4 
- use org.eclipse.tycho:tycho-versions-plugin:set-version  as only goal
- with Dtycho.mode=maven set

but with 2.7.4 now... dependency analysis is done nonetheless..

laeubi:
I'm not aware of any workaround, `tycho.mode=maven` only skips the inital project resolving but don't take place at other places (e.g we are moving some work from that inital phase to the mojo phases) so it seems that `org.eclipse.tycho:tycho-versions-plugin:set-version` runs in a later phase that includes the resolving of project dependencies.

thannerme:
ok. thx! but then this should not relate to migration from 2.2.0 to 2.7.4 I guess..

laeubi:
`tycho.mode=maven` is nowhere documented, nor covered by any tests, so actually *anything* can happen between releases. If you think something is wrong the best way is provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue and then we can decide if/what can be done here.

thannerme:
ok, thanks again. I will  do some more tests/ wait for tycho 3.0.0 and maybe come back to you then

thannerme:
it seems to be related to the new available artifactID "tycho-build" in the extension.xml. 
I changed it when I saw the warning/info "org.eclipse.tycho.extras:tycho-pomless build extension will be replaced in a future version of Tycho by the new org.eclipse.tycho:tycho-build extension" in the logs.

I just changed it back.. and guess will stay with it as long as we use version < 3.x

laeubi:
> it seems to be related to the new available artifactID "tycho-build" in the extension.xml.

Yes it is disabled there by intention.

laeubi:
> I just changed it back.. and guess will stay with it as long as we use version < 3.x

Could you try out the current [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) build and use the option  `-Dtycho.resolver.classic=false`?

thannerme:
and using the artifactID tycho-build? this still does the dependency analysis

laeubi:
> using the artifactID tycho-build and tycho-version set to 2.7.5-SNAPSHOT?

You need Tycho 3.0.0-SNAPSHOT for you tycho-maven-plugin, the tycho-build could be 2.7.5 or 3.0.0-SNAPSHOT (just take care about special instructions about using SNAPSHOTS in maven extensions).

Beside that, I think it would be good to provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue or a small reproducer project.

laeubi:
@thannerme I have now [created a fix](https://github.com/eclipse-tycho/tycho/pull/1323) that detects if a dependency resolution is required or not for the requested Mojos of a project, so once the PR is merged, it should work for the `org.eclipse.tycho:tycho-versions-plugin:set-version` to use:

`mvn org.eclipse.tycho:tycho-versions-plugin:set-version -Dtycho.resolver.classic=false ... other arguments ....`

this will then still resolve direct dependencies between projects (but that should be really fast) but skip the expensive target platform resolution.

thannerme:
Hi @laeubi 

thanks a lot! This seems to work  (I tested with 2.7.5 and artifactID tycho-build)

laeubi:
Not that I'm aware of ... thats just standard maven behavior.

thannerme:
I tried it by leaving it out and setting the plugin in the project's plugin configuration as described here (I thought that should also work)
https://ci.eclipse.org/tycho/job/tycho-sitedocs/lastSuccessfulBuild/artifact/target/staging/tycho-release/tycho-versions-plugin/plugin-info.html

akurtakov:
Can further discussions/questions/etc be done separately to not hijack this PR history. https://github.com/eclipse-tycho/tycho/discussions is good place for it.

Issue #198, opened 2021-07-21
Support JUnit @DisplayName


DariaMarkaryan:
Hi. I'm using JUnit for run tests and it turns out that the @DisplayName does not work with tycho-surefire. I don't really care if I can use this particular annotation, it is important for me to change the names of the suites and test cases. Is there any way to do this?

mickaelistria:
I think it's a bug, @DisplayName is supposed to work.
Please provide a minimal example licensed under EPLv2, which we could try to include into Tycho test suite.

FedorovaDarya:
I'm using tycho-surefire 2.3.0 and JUnit 5.7.2
Test class just for an example:
```
package <...>;

import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.DisplayName;

@DisplayName("Custom name")
public class MyTests {

    @Test
    @DisplayName("My test name")
    public void uglyTestName(){
        System.out.println("Do something");
    }
}

```
It generates report like that:
```
<?xml version="1.0" encoding="UTF-8"?>

-<testsuite failures="0" skipped="0" errors="0" tests="1" time="5.18" name="org.tests.MyTests" xsi:noNamespaceSchemaLocation="https://maven.apache.org/surefire/maven-surefire-plugin/xsd/surefire-test-report.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
<properties/>
<testcase time="5.153" name="uglyTestName" classname="org.test.tests.MyTests"/>
</testsuite>
```

FedorovaDarya:
And one additional question: is this possible to include logs into passed test cases?

ernstblechaJKU:
The behaviour of the maven-surefire-plugin changed some time ago; as far as i remeber there is a difference between JUnit4 and JUnit5 when it comes to handling display-name;

The new version of the surefire-plugin requiers additional configuration (see https://maven.apache.org/surefire/maven-surefire-plugin/examples/junit-platform.html#Surefire_Extensions_and_Reports_Configuration_for_.40DisplayName ). So far i wasn't able to figure out how to set these options when using tycho-surefire?

joeshannon:
I wonder if this would be possible using the providerProperties on the Tycho mojo?

However I've just seen that support for this was added to maven-surefire in version 3.0.0-M4 but Tycho currently uses 2.22.2 so that would probably have to be upgraded first I think.

joeshannon:
[...]

PR #1008, opened 2022-06-09
Support JUnit5 @DisplayName

joeshannon:
Issue #198

### DIFF
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/surefire/JUnit5Test.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/surefire/JUnit5Test.java
@@ -30,13 +30,12 @@ public class JUnit5Test extends AbstractTychoIntegrationTest {
 		verifier.verifyErrorFreeLog();
 		String projectBasedir = verifier.getBasedir();
 		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit4Test", "testWithJUnit4");
-		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit5Test",
-				"myFirstJUnit5Test(TestInfo)");
-		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit5Test",
-				"parameterizedJUnit5Test(String)[1]");
-		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit5Test",
-				"parameterizedJUnit5Test(String)[2]");
-		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit5Test", "repeatedJUnit5Test", 3);
+		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit5Test", "My 1st JUnit 5 test!");
+		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit5Test", "[1] one");
+		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit5Test", "[2] two");
+		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit5Test", "repetition 1 of 3");
+		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit5Test", "repetition 2 of 3");
+		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit5Test", "repetition 3 of 3");
 		// make sure test tagged as 'slow' was skipped
 		assertNumberOfSuccessfulTests(projectBasedir, "bundle.test.JUnit5Test", 6);
 	}
@@ -49,8 +48,7 @@ public class JUnit5Test extends AbstractTychoIntegrationTest {
 		verifier.verifyErrorFreeLog();
 		String projectBasedir = verifier.getBasedir();
 		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit4Test", "testWithJUnit4");
-		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit54Test",
-				"myFirstJUnit54Test(TestInfo)");
+		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit54Test", "My 1st JUnit 5.4 test!");
 		// make sure test tagged as 'slow' was skipped
 		assertNumberOfSuccessfulTests(projectBasedir, "bundle.test.JUnit54Test", 1);
 	}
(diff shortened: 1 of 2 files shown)


tycho-its/src/test/java/org/eclipse/tycho/test/surefire/JUnit5Test.java
```diff
@@ -75,10 +79,22 @@ public void testJUnit56Runner() throws Exception {
...
-				"myFirstJUnit56Test(TestInfo)");
+		assertTestMethodWasSuccessfullyExecuted(projectBasedir, "bundle.test.JUnit56Test", "My 1st JUnit 5.6 test!");
 		// make sure test tagged as 'slow' was skipped
-		assertNumberOfSuccessfulTests(projectBasedir, "bundle.test.JUnit56Test", 1);
+		assertNumberOfSuccessfulTests(projectBasedir, "bundle.test.JUnit56Test", 4);
+	}
+
+	@Test
+	public void testJUnit5Reporting() throws Exception {
+		var verifier = getVerifier("/surefire.junit56/bundle.test", false);
```
joeshannon, line 89: For now I've modified the existing `/surefire.junit56/bundle.test` but might be cleaner with its own project to avoid changing the existing test?

tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java
```diff
@@ -432,6 +432,9 @@ public abstract class AbstractTestMojo extends AbstractMojo {
     @Parameter(property = "surefire.rerunFailingTestsCount", defaultValue = "0")
     private Integer rerunFailingTestsCount;
 
+    @Parameter(defaultValue = "default")
+    private String tychoReportingMode;
```
joeshannon, line 436: Maybe should be called `surefireReportingMode` but as it's Tycho specific I thought best to prefix with Tycho as I wasn't sure if it was also passed into Surefire too but looking at `OsgiSurefireBooter` now, it probably isn't.

mickaelistria: Is this necessary to have it an option here? Can't we just pick the best one according to current context?
More options means more user support and more maintenance care. So if there is no obvious reason to have one, it's better to avoid them.

joeshannon: Yes it would be simpler.

 It's just in case we needed the flexibility of using e.g. the default reporting with JUnit5 (as is the current behaviour with tycho-surefire and maven-surefire). I'm not sure if there really is a requirement for that though, and the parameter in this form will not support the full flexibility of the maven-surefire plugin anyway.

tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/java/org/eclipse/tycho/surefire/osgibooter/OsgiSurefireBooter.java
```diff
@@ -217,6 +220,46 @@ private static Map<String, String> extractProviderProperties(Properties surefire
         return providerProps;
     }
 
+    private static SurefireConsoleOutputReporter getSurefireConsoleOutputReporter(String tychoReportingMode) {
+        if (tychoReportingMode.equals("JUnit5")) {
```
Bananeweizen, line 224: I'm not sure if there is any convention for Maven POM properties, but I find it rather confusing to have a property value that is not lowercase or camelCase. This will become a source of error for users, therefore I would just write it as "junit5" in the documentation instead.
Independent of the actual predefined value, wouldn't it make sense to compare case insensitive here?

tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/java/org/eclipse/tycho/surefire/osgibooter/OsgiSurefireBooter.java
```diff
@@ -217,6 +220,46 @@ private static Map<String, String> extractProviderProperties(Properties surefire
...
+            return jUnit5ConsoleOutputReporter;
+        }
+        SurefireConsoleOutputReporter consoleOutputReporter = new SurefireConsoleOutputReporter();
+        //consoleOutputReporter.setDisable(true); // storing console output causes OOM, see https://github.com/eclipse/tycho/issues/879 & https://issues.apache.org/jira/browse/SUREFIRE-1845
+        return consoleOutputReporter;
+    }
+
+    private static SurefireStatelessReporter getSurefireStatelessReporter(String tychoReportingMode,
+            boolean disableXmlReport, String version) {
+        if (tychoReportingMode.equals("JUnit5")) {
```
Bananeweizen, line 238: same here

tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/java/org/eclipse/tycho/surefire/osgibooter/OsgiSurefireBooter.java
```diff
@@ -217,6 +220,46 @@ private static Map<String, String> extractProviderProperties(Properties surefire
...
+            jUnit5Xml30StatelessReporter.setUsePhrasedTestCaseMethodName(true);
+            jUnit5Xml30StatelessReporter.setUsePhrasedTestSuiteClassName(true);
+            return jUnit5Xml30StatelessReporter;
+        }
+        return new SurefireStatelessReporter(disableXmlReport, version);
+    }
+
+    private static SurefireStatelessTestsetInfoReporter getSurefireStatelessTestsetInfoReporter(
+            String tychoReportingMode) {
+        if (tychoReportingMode.equals("JUnit5")) {
```
Bananeweizen, line 252: and here

tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/java/org/eclipse/tycho/surefire/osgibooter/OsgiSurefireBooter.java
joeshannon: Thanks for the comment. Yes, if we actually do need to add a new parameter, lowercase would probably be better.

joeshannon:
The new parameter is now gone and instead the reporting configured based on the provider class.

The integration tests already included uses of `ParameterizedTest`, `DisplayName` and `RepeatedTest` so I haven't added any new tests, just modified the existing ones.

laeubi:
@Bananeweizen @mickaelistria can you please finish review this? anything to do here or can we merge this?

laeubi:
Anything left to do here? Can we merge this?

mickaelistria (approved):
I didn't review thoroughly and don't really plan to do it shortly, but a quick glance is enough to make me support a merge as soon as all tests are successful

laeubi:
@joeshannon do you want to rebase and quash your commits?

joeshannon:
Thanks for looking. Yes I will do now.

laeubi:
@joeshannon thanks for improving the Junit5 support!

PR #1012, opened 2022-06-10
Don't fail if trying to set same custom EE more than once.

akurtakov:
If building with useJDK=BREE and there is no jdk for given version
properly set compile fails with "Cannot set full specification for a
custom profile more than once" as there could be more than one plugin
setting it e.g. tycho-compiler-plugin and
tycho-p2-extras-plugin:compare-version-with-baselines .
This change creates temp EE and compare its values to the existing one
and if they match continue happily.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/ee/ExecutionEnvironmentConfigurationImpl.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/ee/ExecutionEnvironmentConfigurationImpl.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2014 SAP SE and others.
+ * Copyright (c) 2012, 2022 SAP SE and others.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
@@ -107,8 +107,7 @@ public class ExecutionEnvironmentConfigurationImpl implements ExecutionEnvironme
         }
         String profileName = getProfileName();
         Collection<String> profileNames = ExecutionEnvironmentUtils.getProfileNames(toolchainManager, session, logger);
-        boolean profileExists = profileNames
-                .contains(profileName);
+        boolean profileExists = profileNames.contains(profileName);
         if (!profileExists && ignoredByResolver) {
             throw new BuildFailureException(
                     "When using a custom execution environment profile, resolveWithExecutionEnvironmentConstraints must not be set to false");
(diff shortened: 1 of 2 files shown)


akurtakov:
@laeubi I found it only cause SWT GH actions/verification builds ( e.g. https://github.com/eclipse-platform/eclipse.platform.swt/runs/6836461583?check_suite_focus=true ) are not properly setting toolchain.xml which led to this failure. At the very minimum the error is cryptic and tycho should fail better, although I believe it's fine to just continue as this patch does.

akurtakov:
Patch is not ready as order in systempackages/properties values is not guaranteed thus comparison claims these are different . But the idea where I aim at should be clear. Please advise whether there is better way to do it.

laeubi:
> I found it only cause SWT GH actions/verification builds ( e.g. https://github.com/eclipse-platform/eclipse.platform.swt/runs/6836461583?check_suite_focus=true ) are not properly setting toolchain.xml which led to this failure

Can you explain what is/was setup wrong? Can we reproduce this with a itest?

> At the very minimum the error is cryptic and tycho should fail better, 

Yeah I stumbled from time to time over this and have already adjusted this a bit, but never understand what this "custom profile could not be set twice" should really mean :-\

> although I believe it's fine to just continue as this patch does.

Good idea, maybe @mickaelistria can tell a bit about the background of this.

> as order in systempackages/properties values is not guaranteed thus comparison claims these are different

Maybe just put them in a Map/Set and then compare these?

akurtakov:
Issue happens if toolchains.xml is there but doesn't contain Java-11 definition (as SWT requires) and -Dcompare-version-with-baselines.skip=false is passed . Note that if you don't enable baseline comparison maven succeeds as this plugin tried to set the EE for second time.

laeubi:
> @laeubi why do you mark it as ready? Have you fixed the order of packages in properties in some other commit?

I just wanted to see if CI build succeeds :-)

So the build passes and it seems we have no validation for this :-\

mickaelistria:
I'm fresh on this issue so I may be missing a few things, here are some general questions

> If building with useJDK=BREE and there is no jdk for given version properly set compile fails with "Cannot set full specification for a custom profile more than once"

So the error message is incorrect in 1st place? 
Or is the root issue that "no matching profile" is treated as "custom profile"?

laeubi:
@akurtakov sadly my tycho workspace is messed up due to a bug in the reference search and I can'T found where the toolchain is searched/custom profile created but `ExecutionEnvironmentUtils#getExecutionEnvironment` or `ExecutionEnvironmentUtils#getToolchainFor` might be useful here.

akurtakov:
Would you please handle that?

mickaelistria:
I'm on it.

mickaelistria:
I think #1019 should be a more correct fix.

PR #1019, opened 2022-06-13
Fix custom profile detection

mickaelistria:
When we ask for a profile that is not part of toolchain, we can still
get another compatible environment, which is not custom. The check
verifies where we get an environment without caring if it's a perfect
match.


tycho-core/src/main/java/org/eclipse/tycho/core/ee/ExecutionEnvironmentConfigurationImpl.java
```diff
@@ -106,14 +106,20 @@ public boolean isCustomProfile() {
             return false;
         }
         String profileName = getProfileName();
-        Collection<String> profileNames = ExecutionEnvironmentUtils.getProfileNames(toolchainManager, session, logger);
-        boolean profileExists = profileNames
-                .contains(profileName);
-        if (!profileExists && ignoredByResolver) {
+        try {
+            boolean profileExists = ExecutionEnvironmentUtils.getExecutionEnvironment(profileName, toolchainManager,
```
laeubi, line 110: This method should never return null

laeubi (approved):
LGTM

akurtakov:
@mickaelistria Should I be troubled by the fact that I see:
```
[INFO] --- tycho-compiler-plugin:3.0.0-SNAPSHOT:compile (default-compile) @ org.eclipse.swt.fragments.localbuild ---
[WARNING] Using JavaSE-17 to fulfill requested profile of JavaSE-11 this might lead to faulty dependency resolution, consider define a suitable JDK in the toolchains.xml
[WARNING] Using JavaSE-17 to fulfill requested profile of JavaSE-11 this might lead to faulty dependency resolution, consider define a suitable JDK in the toolchains.xml
[WARNING] Using JavaSE-17 to fulfill requested profile of JavaSE-11 this might lead to faulty dependency resolution, consider define a suitable JDK in the toolchains.xml
[WARNING] Using JavaSE-17 to fulfill requested profile of JavaSE-11 this might lead to faulty dependency resolution, consider define a suitable JDK in the toolchains.xml
[WARNING] Using JavaSE-17 to fulfill requested profile of JavaSE-11 this might lead to faulty dependency resolution, consider define a suitable JDK in the toolchains.xml
[WARNING] Using JavaSE-17 to fulfill requested profile of JavaSE-11 this might lead to faulty dependency resolution, consider define a suitable JDK in the toolchains.xml
```

I don't think same warning should be shown so many times.

mickaelistria:
> Should I be troubled by the fact that I see:

Yes, the goal of the warning is to trouble people so they consider improving things on their end.

> I don't think same warning should be shown so many times.

Right, I think it's another issue though, so I suggest that if current patch works for you and test pass, we merge and look at the multiple warning after.

akurtakov:
> > I don't think same warning should be shown so many times.
> 
> Currently it is printed on each lookup, but if you think that is not suitable you an do print it only on the first lookup.

Please do so. Esp with warnings detected and displayed on jenkins/gh/etc. that would make numbers go crazy.

akurtakov:
Merging this one as it's an improvement. @mickaelistria I would appreciate if you improve the warnings.

kjsmita6:
Using Tycho 3.0.1 I am getting those warnings a lot, about 3 times per project, and with a reactor with 70 projects... Is there an open issue to improve this? Note I'm not on the latest snapshot so if it was fixed there already then my apologies.

`[WARNING] Using JavaSE-19 to fulfill requested profile of JavaSE-17 this might lead to faulty dependency resolution, consider define a suitable JDK in the toolchains.xml`
I also don't even have Java19 on my system so not sure where it's getting this. The java on my path is 17 so I would think it should be able to find it.

laeubi:
> I also don't even have Java19 on my system so not sure where it's getting this. 

Please check your toolchains

> . The java on my path is 17 so I would think it should be able to find it.

Tycho do not considder the path.

PR #1037, opened 2022-06-16
Fix compilation of demos.

akurtakov:
* removed the one relying on deprecated eclipse-application
* updated the rest to use newer deps, targets, plugins, bytecode and
etc.

Fixes https://github.com/eclipse/tycho/issues/1036


akurtakov:
Not waiting for the verification build as demoes are not compiled by it.

PR #1039, opened 2022-06-16
Support bnd processing in pomless builds

laeubi:
Fixes https://github.com/eclipse/tycho/issues/1038


laeubi:
Added an integration test showing BND generates JPMS infos for a pomless bundle.

PR #1042, opened 2022-06-18
Run tests against 2022-06

akurtakov:


akurtakov:
@laeubi Do I understand correctly that your latest changes (already in master) are supposed to fix the failing resolution tests?

laeubi:
Actually I though so... but strangely the (automatic) rebase has not triggered a new build ...

akurtakov:
@laeubi Any idea what else is still needed here?

laeubi:
Seems we have some strange EE requirements in the target/tests here but I have no immediate idea how to solve this:

> [DEBUG] Class DummyTest matches the current filter.
> [DEBUG] No solution found because the problem is unsatisfiable.: [Unable to satisfy dependency from javax.activation 1.2.2.v20201119-1642 to osgi.ee; (&(osgi.ee=JavaSE)(version=9.0)).; Unable to satisfy dependency from org.osgi.service.prefs 1.1.2.202109301733 to osgi.ee; (&(osgi.ee=JavaSE/compact1)(version=1.8)).; Unable to satisfy dependency from org.osgi.util.function 1.2.0.202109301733 to osgi.ee; (&(osgi.ee=JavaSE/compact1)(version=1.8)).; Unable to satisfy dependency from org.osgi.util.promise 1.2.0.202109301733 to osgi.ee; (&(osgi.ee=JavaSE/compact1)(version=1.8)).; No solution found because the problem is unsatisfiable.]
> [ERROR] org.osgi.service.prefs 1.1.2.202109301733 requires Execution Environment that matches (&(osgi.ee=JavaSE/compact1)(version=1.8)) but the current resolution context uses []
> [INFO] {osgi.os=linux, osgi.ws=gtk, org.eclipse.update.install.features=true, osgi.arch=x86_64}
> [ERROR] Cannot resolve project dependencies:
> [ERROR]   Software being installed: org.eclipse.core.runtime 3.25.0.v20220506-1157
> [ERROR]   Missing requirement: org.osgi.service.prefs 1.1.2.202109301733 requires 'osgi.ee; (&(osgi.ee=JavaSE/compact1)(version=1.8))' but it could not be found
> [ERROR]   Cannot satisfy dependency: org.eclipse.core.runtime 3.25.0.v20220506-1157 depends on: osgi.bundle; org.eclipse.equinox.preferences [3.10.0,4.0.0)
> [ERROR]   Cannot satisfy dependency: org.eclipse.equinox.preferences 3.10.0.v20220503-1634 depends on: osgi.bundle; org.osgi.service.prefs [1.1.0,1.2.0)

Maybe the `osgi.ee; (&(osgi.ee=JavaSE/compact1)(version=1.8))` is not recognized anymore now?

laeubi:
Is this test special? `JavaSE/compact1` should be[ part of equinox](https://github.com/eclipse-equinox/equinox/blob/master/bundles/org.eclipse.osgi/JavaSE_compact1-1.8.profile) would it be expected that Tycho finds it?

PR #1048, opened 2022-06-19
[Maven-Target] Consider extra-repositories when fetching source-jars

HannesWell:
Cherry-pick of https://github.com/eclipse/tycho/pull/1020 to the `tycho-2.7.x` maintenance branch.

@laeubi is this correct?

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/resolver/MavenTargetDefinitionContent.java
+++ b/tycho-bundles/org.eclipse.tycho.p2.resolver.impl/src/main/java/org/eclipse/tycho/p2/resolver/MavenTargetDefinitionContent.java
@@ -18,7 +18,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Iterator;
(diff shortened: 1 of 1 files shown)


laeubi:
yep, I somehow broken the GH action cache, no idea if one can clear it manually so lets wait for the jenkins ci.

HannesWell:
> yep, I somehow broken the GH action cache, no idea if one can clear it manually so lets wait for the jenkins ci.

Not yet but it looks like this is planned:
- https://github.com/actions/cache/issues/2
- https://github.com/github/roadmap/issues/502

This will likely be part of the Cache-API:
https://docs.github.com/en/rest/actions/cache
https://github.com/actions/cache/issues/632

So that we can create a manually trigger-able workflow to clear the cache.

laeubi:
If you like you can add an extra char to the cache key, but I think we should not invest too much time for the GH action and a maintenance release... this is mostly for PGP, getting BND in and some trivial fixes like yours.

HannesWell:
> If you like you can add an extra char to the cache key, but I think we should not invest too much time for the GH action and a maintenance release... this is mostly for PGP, getting BND in and some trivial fixes like yours.

That would work as well.

IIRC a TP is also used for Tycho's build, isn't it? If so *.target files should be  considered for the cache-key as well.
I plan to add the enhanced license-check to Tycho soon (hopefully today) and can then do both in PR.
This should then automatically lead to a new cache-key.

laeubi:
> This should then automatically lead to a new cache-key.

Cache key is already per branch so no need to worry for master...

laeubi:
Thanks for backporting.

PR #1049, opened 2022-06-19
[Build] Use eclipse/dash-licenses' reusable licenses-check workflow

HannesWell:
The workflow checks if all licenses are vetted on each push to the master branch and on each new (push to a) PR.
Furthermore committers can request a review of unvetted licenses by adding the following comment to a PR (open or closed):
```
/request-license-review
```
If a committer requests a license-review the license vetting summary files created by the dash-tools plug-in are attached to the workflow. So if you are interested in the summary you can comment `/request-license-review` and download the summary, even if you know all licenses are vetted. The workflow handles this case gracefully.

The workflow works quite well for m2e, but for the near future I cannot foresee version updates so in order to test it more we need to use it in more projects.

@akurtakov, @mickaelistria if you agree I can open a corresponding EFN help-desk issue to ask for inclusion of the GITLAB_API token for tycho.

In the meantime this workflow can be used already. Only the license review requests will be rejected.
At the moment the workflow does not run on pushes to the specified branches, but I have already opened a PR to address this, which hopefully gets merged soon: https://github.com/eclipse/dash-licenses/pull/177

Additionally with this PR *.target files are considered when computing the dependency cache key.

PR #1061, opened 2022-06-21
Remove now obsolet license-tool-plugin calls in Jenkins and repos

HannesWell:
Because Tycho now uses the eclipse/dash-licenses' reusable licenses-check GH workflow (added via https://github.com/eclipse/tycho/pull/1049)
- calling the license-tool-plugin in the Jenkins build is not necessary any more (was actually already before since when a dedicated GH-workflow was used for that)
- the dash-licenses-snapshots repository can be removed (it is added by the workflow via a dedicated maven-settings file)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -570,16 +570,6 @@
 		</profile>
 	</profiles>
 
-	<pluginRepositories>
-		<pluginRepository>
-		  	<id>dash-licenses-snapshots</id>
-		  	<url>https://repo.eclipse.org/content/repositories/dash-licenses-snapshots/</url>
-		  	<snapshots>
-		  	  <enabled>true</enabled>
-		  	</snapshots>
-  		</pluginRepository>
-	</pluginRepositories>
-
 	<distributionManagement>
 		<site>
 			<id>tycho.site</id>
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,7 +24,7 @@ pipeline {
 	stages {
 		stage('Build') {
 			steps {
-				sh 'mvn --batch-mode -U -V -e clean install org.eclipse.dash:license-tool-plugin:license-check -Pits -Dmaven.repo.local=$WORKSPACE/.m2/repository'
+				sh 'mvn --batch-mode -U -V -e clean install -Pits -Dmaven.repo.local=$WORKSPACE/.m2/repository'
 			}
 			post {
 				always {


pom.xml
```diff
@@ -570,16 +570,6 @@
 		</profile>
 	</profiles>
 
-	<pluginRepositories>
```
akurtakov, line 573: Can we keep this one? It helps if one has to open cq manually and pruned his ~/.m2 repo. If/whed dash-license-tool is available in maven central that would no longer be needed.

HannesWell: Sure, if you still want to do that manually this can be kept.
But I want to remark, in case you just want to get the review summary you can just request a review (although all licenses are already vetted) and download the attached summary files.

HannesWell:
> Workflow run (with attached summary files): https://github.com/eclipse/tycho/actions/runs/2540738067

When you open the link you will find the `technology.tycho-license-vetting-summary` artifact attached to the workflow. This zip contains the summary and review-summary files. The latter is empty since now review was created/is pending and the former contains all artifacts found and their license status. 
Just an excerpt from this run:
```
maven/mavencentral/aopalliance/aopalliance/1.0, LicenseRef-Public-Domain, approved, CQ2918
maven/mavencentral/biz.aQute.bnd/biz.aQute.bnd.util/6.3.1, Apache-2.0 and EPL-2.0, approved, #2672
maven/mavencentral/biz.aQute.bnd/biz.aQute.bndlib/5.3.0, Apache-2.0 OR EPL-2.0, approved, clearlydefined
maven/mavencentral/biz.aQute.bnd/biz.aQute.bndlib/6.3.1, Apache-2.0, approved, #2674
maven/mavencentral/com.google.code.findbugs/jsr305/3.0.2, Apache-2.0, approved, #20
maven/mavencentral/com.google.errorprone/error_prone_annotations/2.1.3, Apache-2.0, approved, #805
...
```
Just in case that is why you would want to runt the license check locally. :)

pom.xml
akurtakov: I was thinking of it as some temp measure till things are proven but seeing your responsiveness we don't need this safeguard :)

HannesWell: Glad to hear to that. :)
Nevertheless in the worst case you can still revert this change and #1049.
But the more use cases we have the earlier bugs hopefully become visible.

PR #1064, opened 2022-06-22
Move Target Resoloution to the Maven build

laeubi:
Currently the target content is resolved as part of a
TychoMavenLifecycleParticipant before the regular maven build starts,
this has several drawbacks:

- this is "invisible" to common maven tools
- it requires massive amount of time even if nothing is executed later
on
- parallelize the build is limited by this preliminary step
- depending on data from previous build stages is hard to archive

Because of this, now this step is postponed to the initial validate
phase of the maven build of each project. Dependencies between projects
are computed with a much faster and lightweight approach instead.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java
@@ -19,9 +19,11 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -133,9 +135,29 @@ public final class MavenDependencyInjector {
                 ? collectProjectDependencies(artifact, scope) //
                 : collectExternalDependencies(artifact, scope);
         Model model = project.getModel();
+        Set<String> existing = model.getDependencies().stream().map(dep -> getKey(dep))
+                .collect(Collectors.toCollection(HashSet::new));
         for (Dependency dependency : dependencyList) {
-            model.addDependency(dependency);
+            if (existing.add(getKey(dependency))) {
+                model.addDependency(dependency);
+            }
         }
+        Map<String, MavenProject> projectReferences = project.getProjectReferences();
+        ReactorProject mavenProject = artifact.getMavenProject();
+        if (mavenProject != null && DefaultReactorProject.adapt(project) != mavenProject) {
+            String key = new StringBuilder().append(mavenProject.getGroupId()).append(':')
+                    .append(mavenProject.getArtifactId()).append(':').append(mavenProject.getVersion()).toString();
+            if (!projectReferences.containsKey(key)) {
+                logger.debug("Found a P2 dependency (" + artifact
+                        + ") that is not reflected in the maven model project references");
+            }
+        }
+    }
+
+    private static String getKey(Dependency dependency) {
+
+        return dependency.getGroupId() + ":" + dependency.getArtifactId() + ":" + dependency.getType() + ":"
+                + Objects.requireNonNullElse(dependency.getClassifier(), "");
     }
 
     private List<Dependency> collectExternalDependencies(ArtifactDescriptor artifact, String scope) {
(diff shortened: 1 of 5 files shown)


laeubi:
Lets see what all fails here now ... :-)

target-platform-configuration/src/main/java/org/eclipse/tycho/target/ValidateTargetPlatformMojo.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.Component;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.project.MavenProject;
+import org.eclipse.tycho.ReactorProject;
+import org.eclipse.tycho.core.osgitools.DefaultReactorProject;
+import org.eclipse.tycho.resolver.TychoResolver;
+
+@Mojo(name = "validate-target-platform", threadSafe = true)
```
mickaelistria, line 30: Please add javadoc so it appears in Maven plugin doc.

laeubi: Done, since this is an internal mojo it might be not that interesting anyways.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/MavenProjectDependencyProcessor.java
```diff
@@ -140,12 +157,74 @@ public Collection<IInstallableUnit> computeProjectDependencies(MavenProject proj
...
 		return resolved;
 	}
 
+	private static boolean isFragment(IInstallableUnit installableUnit) {
+		return getFragmentCapability(installableUnit).isPresent();
+	}
+
+	private static Optional<IProvidedCapability> getFragmentCapability(IInstallableUnit installableUnit) {
+		for (IProvidedCapability capability : installableUnit.getProvidedCapabilities()) {
+			String nameSpace = capability.getNamespace();
```
HannesWell, line 205: I wonder if it would in general lead to a performance improvement for P2 (and Tycho) if capabilities/requirements could be retrieved per namespace and would be stored internally per namespace, just like it is now done in Equinox for bundle requirements/capabilities. 
When switching to the Equinox-resolver Tycho first suffered from a massive increase of the initial target resolution just because of that: https://github.com/eclipse/tycho/issues/47

laeubi: The problem scope is much smaller here so it should not be an issue, nerveless a lookup by name could be useful I think, it is just a bit out of scope of this changes, as the first priority would be correctness and could later be optimized for speed :-)

HannesWell: > The problem scope is much smaller here so it should not be an issue, nerveless a lookup by name could be useful I think, it is just a bit out of scope of this changes, as the first priority would be correctness and could later be optimized for speed :-) SO if you like to propose such a change for P2 it would be welcome.

Yes this is definitely related to this change and just came into my mind when locking at the change.
I'm going to create an issue, however I did not yet check the P2 sources if this is in general useful.

laeubi:
One test currently fails due to 
- https://github.com/eclipse-equinox/p2/issues/99
- https://github.com/eclipse-equinox/p2/issues/100

laeubi:
I got quite far given that this quite drastically changes how Tycho works today, but think for making the final steps on this one would require an enhancement to maven so we do not need an extra mojo for this, thus I have opened https://issues.apache.org/jira/browse/MNG-7510

laeubi:
Still this could be a replacement for the `mode=maven`, but a bit twisted:

If one only executes direct maven mojos that do not require the project to be resolved (e.g. `clean`) this will be disabled by default. On the other hand, it might be required for some of those to explicitly enable the execution (e.g. for `dependency:list` to see Tycho p2 dependencies as well).

laeubi:
There is only one test that now shows a cyclic dependency, beside that everything works as intended 🥳

laeubi:
All issues are solved, tests are green.

My proposal would be the following:

1. If no more concerns, I'd like to merge this so we get this change "into the wild"
2. I'd like to have a "quite-period" of three days to see if there is a serve problem so we can easily revert if absolutely necessary (would this be ok for you @eclipse/technology-tycho-committers ?)
3. If there are minor issues or edge cases where this is causing issue, I have added a flag `tycho.resolver.classic` that could be set to enable the old resolver step while investigating / fixing the issue for things currently not covered.

akurtakov:
Sounds good to me. It's quite good time considering Platform M1 has just been released too.

PR #1081, opened 2022-06-27
Check that components declared in the manifest exits

laeubi:
Currently one can easily miss to either add a component or enable the
generation of the component. As this only found out at runtime or even
worse after a release, Tycho should check that all declared components
are present.

### DIFF
--- a/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackagePluginMojo.java
+++ b/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackagePluginMojo.java
@@ -48,7 +48,9 @@ import org.codehaus.plexus.component.annotations.Requirement;
 import org.eclipse.tycho.BuildProperties;
 import org.eclipse.tycho.ReactorProject;
 import org.eclipse.tycho.TychoConstants;
+import org.eclipse.tycho.core.TychoProject;
 import org.eclipse.tycho.core.osgitools.DefaultReactorProject;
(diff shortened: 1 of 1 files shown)


tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackagePluginMojo.java
```diff
@@ -61,266 +63,302 @@
...
-        }
-    }
+			return pluginFile;
+		} catch (IOException | ArchiverException | ManifestException | DependencyResolutionRequiredException e) {
+			throw new MojoExecutionException("Error assembling JAR", e);
+		}
+	}
 
-    private void updateManifest(File output) throws IOException, MojoExecutionException {
+	private void assertComponentExits(String component) throws MojoExecutionException {
```
Bananeweizen, line 282: typo exists

tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackagePluginMojo.java
```diff
@@ -61,266 +63,302 @@
...
+		File projectRelative = new File(project.getBasedir(), component);
+		if (projectRelative.isFile()) {
+			return;
+		}
+		File outputRelative = new File(project.getBuild().getOutputDirectory(), component);
+		if (outputRelative.isFile()) {
+			return;
+		}
+		String message = "Component definition" + component + " not found! Neither "
```
Bananeweizen, line 291: missing blank after definition

PR #1083, opened 2022-06-28
Improvements to tycho-its

akurtakov:
* Fix package names to be all lowercase
* Remove maven-shared-utils - it's being replaced in maven land too and
is used only in its. Removing it makes commons-io go to lower version as
it's transitive dependency and thus new methods used no longer exist.
Luckily, there is direct replacement in java.nio thus no need to pay
more attention to it.

### DIFF
--- a/tycho-its/pom.xml
+++ b/tycho-its/pom.xml
@@ -204,11 +204,6 @@
 			<version>2.2</version>
 		</dependency>
 		<dependency>
-			<groupId>org.apache.maven.shared</groupId>
-			<artifactId>maven-shared-utils</artifactId>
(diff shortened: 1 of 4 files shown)


laeubi:
> * Removing it makes commons-io go to lower version

Just in case, we simply can add the commons dependency directly.

akurtakov:
> > * Removing it makes commons-io go to lower version
> 
> Just in case, we simply can add the commons dependency directly.

I know but using standard Java APIs is better when we can IMHO thus better to not do extra work here.

laeubi:
We might need a JDT cleanup action for commons-io -> std java :-D

PR #1088, opened 2022-06-28
Skip DS check if wildcards are used

laeubi:
See https://github.com/eclipse/tycho/issues/1087


tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackagePluginMojo.java
```diff
@@ -252,8 +252,13 @@ private File createPluginJar() throws MojoExecutionException {
 				OsgiBundleProject bundleProject = (OsgiBundleProject) facet;
 				String components = bundleProject.getManifestValue("Service-Component", project);
 				if (components != null) {
-					for (String component : components.split(",\\s*")) {
-						assertComponentExists(component);
+					if (components.contains("*")) {
+						getLog().warn(
+								"Checking Service-Component header that contains wildcards is currently not supported!");
```
akurtakov, line 257: Do we have to warn? Tycho already throws so many warnings that even I no longer read them. IMHO having a //TODO here is better.

laeubi: Given that we hopefully could replace this soon I think it is okay.

akurtakov (approved):

PR #1091, opened 2022-06-29
Dependency updates and cleanups

HannesWell:
Clean-up/Unify some dependency versions and update some to hopefully hit some unvetted licenses and to be able to test the license-vetting workflow review requests.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,14 +74,10 @@
 		<min.jdk.version>11</min.jdk.version>
 
 		<!-- copy&paste from maven/pom.xml file. I wish import scope worked in 3.0 -->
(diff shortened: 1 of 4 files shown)


HannesWell:
To bad nothing unvetted found...
So this basically just unifies dependency versions and removes unused properties and updates to maven-verifier 1.8.0.
The latter is only used in tycho's testing plug-ins so this is not critical. If the build succeeds this should be fine.

HannesWell:
OK, looks like the update to `maven-verifier` 1.8.0 causes build-failures.
Any way I reverted that update, because this change was actually intended to find some unvetted licenses (which I didn't) to test the license-vetting workflow.

PR #1094, opened 2022-06-30
[Backport 2.7.x] Fix pom.model properties read for bundle-projects and add it-tests

HannesWell:
Backport of https://github.com/eclipse/tycho/pull/1090 to the 2.7.x branch.


laeubi:
@HannesWell if it is to cumbersome to include the itest I think it would be ok to just include the code changes as we have the test in the master already.

HannesWell:
> @HannesWell if it is to cumbersome to include the itest I think it would be ok to just include the code changes as we have the test in the master already.

Stupidly I mixed JUnit 4 and 5 assertions in that test case and JUnit-5 seems not to be available on this branch.
I changed it to use the JUnit-4 Assert. If that fixes the issue lets include the tests, if there are more complicated problems I agree with you and will only include the code changes.

laeubi:
@HannesWell because we use an older maven version here, you need to encode a fixed version for the tycho-build extension in the `extensions.xml`

PR #1101, opened 2022-07-01
Clean up and reduce file-system interactions of Tycho-Mappings

HannesWell:
This PR has the goal to clean-up/simplify Tycho's pomless Mappings a bit and to reduce the number of file-system interactions, for example by performing the check if a File exists after it has been checked if it even matches the name pattern. Interactions with the file system are usually very expensive an can require significant amount of time if done often.
That being said the changes applied are only from looking at the code without having done any measurements. So there might not be a noticeable improvement.

### DIFF
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
@@ -28,6 +28,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.function.Supplier;
 
 import org.apache.maven.model.InputLocation;
 import org.apache.maven.model.InputSource;
@@ -35,7 +36,6 @@ import org.apache.maven.model.Model;
 import org.apache.maven.model.Parent;
 import org.apache.maven.model.building.FileModelSource;
 import org.apache.maven.model.building.ModelProcessor;
-import org.apache.maven.model.io.ModelParseException;
 import org.apache.maven.model.io.ModelReader;
 import org.apache.maven.model.io.ModelWriter;
 import org.codehaus.plexus.PlexusContainer;
@@ -107,7 +107,7 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
     }
 
     @Override
-    public Model read(InputStream input, Map<String, ?> options) throws IOException, ModelParseException {
+    public Model read(InputStream input, Map<String, ?> options) throws IOException {
         String location = PolyglotModelUtil.getLocation(options);
         File file = new File(location);
         try (InputStreamReader stream = new InputStreamReader(input, getPrimaryArtifactCharset())) {
@@ -116,12 +116,11 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
     }
 
     @Override
-    public Model read(File input, Map<String, ?> options) throws IOException, ModelParseException {
+    public Model read(File input, Map<String, ?> options) throws IOException {
         File artifactFile = getRealArtifactFile(input);
-        if (artifactFile.exists()) {
-            if (artifactFile.isDirectory()) {
-                return read(new StringReader(""), input, options);
-            }
+        if (artifactFile.isDirectory()) {
+            return read(new StringReader(""), input, options);
+        } else if (artifactFile.isFile()) {
             try (InputStreamReader stream = new InputStreamReader(new FileInputStream(artifactFile),
                     getPrimaryArtifactCharset())) {
                 return read(stream, input, options);
@@ -133,13 +132,12 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
     }
 
     @Override
-    public Model read(Reader input, Map<String, ?> options) throws IOException, ModelParseException {
+    public Model read(Reader input, Map<String, ?> options) throws IOException {
         File file = new File(PolyglotModelUtil.getLocation(options));
         return read(input, file, options);
     }
 
-    private Model read(Reader artifactReader, File artifactFile, Map<String, ?> options)
-            throws ModelParseException, IOException {
+    private Model read(Reader artifactReader, File artifactFile, Map<String, ?> options) throws IOException {
         Model model = new Model();
         model.setModelVersion("4.0.0");
         model.setPackaging(getPackaging());
@@ -162,8 +160,7 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
         return polyglotArtifactFile;
     }
 
-    protected Parent findParent(File projectRoot, Map<String, ?> projectOptions)
-            throws ModelParseException, IOException {
+    protected Parent findParent(File projectRoot, Map<String, ?> projectOptions) throws IOException {
         Parent parent = (Parent) projectOptions.get(MODEL_PARENT);
         if (parent != null) {
             //if the parent is given by the options we don't need to search it!
@@ -224,10 +221,8 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
             List<ModelProcessor> lookupList = container.lookupList(ModelProcessor.class);
             for (ModelProcessor processor : lookupList) {
                 File pom = processor.locatePom(folder);
-                if (pom != null && pom.exists()) {
-                    if (reference == null || pom.getName().equals(nameHint)) {
-                        reference = new PomReference(pom, processor);
-                    }
+                if (pom != null && (reference == null || pom.getName().equals(nameHint)) && pom.exists()) {
+                    reference = new PomReference(pom, processor);
                 }
             }
         } catch (ComponentLookupException e) {
@@ -261,18 +256,47 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
         return StandardCharsets.UTF_8;
     }
 
-    protected abstract void initModel(Model model, Reader artifactReader, File artifactFile)
-            throws ModelParseException, IOException;
+    protected abstract void initModel(Model model, Reader artifactReader, File artifactFile) throws IOException;
 
     protected static Properties getBuildProperties(File dir) throws IOException {
+        return loadProperties(new File(dir, "build.properties"));
+    }
+
+    static Properties loadProperties(File propertiesPath) throws IOException {
         Properties properties = new Properties();
-        File buildProperties = new File(dir, "build.properties");
-        if (buildProperties.exists()) {
-            try (FileInputStream stream = new FileInputStream(buildProperties)) {
+        if (propertiesPath.isFile()) {
+            try (FileInputStream stream = new FileInputStream(propertiesPath)) {
                 properties.load(stream);
             }
         }
         return properties;
+    }
+
+    static Supplier<Properties> getPropertiesSupplier(File propertiesFile) {
+        return new Supplier<>() {
+            private Properties properties;
+
+            @Override
+            public Properties get() {
+                if (properties == null) {
+                    try {
+                        properties = loadProperties(propertiesFile);
+                    } catch (IOException e) { // ignore externalized data and try again the next time
+                        return new Properties();
+                    }
+                }
+                return properties;
+            }
+        };
+    }
+
+    static String localizedValue(String value, Supplier<Properties> properties) {
+        if (value != null && value.startsWith("%")) { //load translated value from properties
+            String key = value.substring(1);
+            String translation = properties.get().getProperty(key);
+            return translation != null && !translation.isEmpty() ? translation : key;
+        }
+        return value;
     }
 
     @Override
@@ -288,10 +312,8 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
     }
 
     protected Properties getEnhancementProperties(File file) throws IOException {
-        if (file.isDirectory()) {
-            return getBuildProperties(file);
-        }
-        return getBuildProperties(file.getParentFile());
+        File dir = file.isDirectory() ? file : file.getParentFile();
+        return getBuildProperties(dir);
     }
 
     private static void setLocation(Model model, File modelSource) {
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractXMLTychoMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractXMLTychoMapping.java
@@ -17,6 +17,10 @@ package org.eclipse.tycho.pomless;
 import java.io.File;
 import java.io.IOException;
 import java.io.Reader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
 
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
@@ -40,13 +44,11 @@ public abstract class AbstractXMLTychoMapping extends AbstractTychoMapping {
     private static final DocumentBuilderFactory FACTORY = DocumentBuilderFactory.newInstance();
 
     @Override
-    protected void initModel(Model model, Reader artifactReader, File artifactFile)
-            throws ModelParseException, IOException {
+    protected void initModel(Model model, Reader artifactReader, File artifactFile) throws IOException {
         initModelFromXML(model, parseXML(artifactReader, artifactFile.toURI().toASCIIString()), artifactFile);
     }
 
-    protected abstract void initModelFromXML(Model model, Element xml, File artifactFile)
-            throws ModelParseException, IOException;
+    protected abstract void initModelFromXML(Model model, Element xml, File artifactFile) throws IOException;
 
     protected static Element parseXML(Reader artifactReader, String documentURI) throws IOException {
         try {
@@ -95,5 +97,12 @@ public abstract class AbstractXMLTychoMapping extends AbstractTychoMapping {
     @Override
     public float getPriority() {
         return 1;
+    }
+
+    protected static Stream<File> filesWithExtension(Path directory, String extension) throws IOException {
+        Predicate<String> nameFilter = n -> !n.startsWith(".polyglot.") && n.endsWith(extension);
+        return Files.walk(directory, 1) //
+                .filter(p -> nameFilter.test(p.getFileName().toString())) // 
+                .filter(Files::isRegularFile).map(Path::toFile);
     }
 }
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
@@ -14,24 +14,24 @@
  *******************************************************************************/
 package org.eclipse.tycho.pomless;
 
+import static java.util.stream.Stream.concat;
+import static java.util.stream.Stream.of;
+
 import java.io.BufferedReader;
-import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileFilter;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
 import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.OutputStreamWriter;
 import java.io.Reader;
+import java.nio.file.Files;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.UUID;
+import java.util.stream.Stream;
 
 import org.apache.maven.model.Model;
-import org.apache.maven.model.io.ModelParseException;
 import org.codehaus.plexus.component.annotations.Component;
 import org.sonatype.maven.polyglot.mapping.Mapping;
 
@@ -75,16 +75,9 @@ public class TychoAggregatorMapping extends AbstractTychoMapping {
                 }
                 if (!modules.isEmpty()) {
                     file.deleteOnExit();
+                    Stream<CharSequence> lines = concat(of(TYCHO_AUTOMATIC_GENERATED_FILE_HEADER), modules.stream());
                     try {
-                        try (BufferedWriter writer = new BufferedWriter(
-                                new OutputStreamWriter(new FileOutputStream(file), getPrimaryArtifactCharset()))) {
-                            writer.write(TYCHO_AUTOMATIC_GENERATED_FILE_HEADER);
-                            writer.newLine();
-                            for (String module : modules) {
-                                writer.write(module);
-                                writer.newLine();
-                            }
-                        }
+                        Files.write(file.toPath(), lines::iterator, getPrimaryArtifactCharset());
                     } catch (IOException e) {
                         throw new RuntimeException("writing modules file failed", e);
                     }
@@ -103,15 +96,12 @@ public class TychoAggregatorMapping extends AbstractTychoMapping {
     }
 
     @Override
-    protected void initModel(Model model, Reader artifactReader, File artifactFile)
-            throws ModelParseException, IOException {
+    protected void initModel(Model model, Reader artifactReader, File artifactFile) throws IOException {
         logger.debug("Generate aggregator pom for " + artifactFile);
         try (BufferedReader reader = new BufferedReader(artifactReader)) {
-            String line;
-            while ((line = reader.readLine()) != null) {
-                if (line.startsWith("#") || line.trim().isEmpty()) {
-                    continue;
-                }
+            Stream<String> lines = reader.lines().filter(l -> !l.startsWith("#") && !l.isBlank()).map(String::strip);
+            for (Iterator<String> iterator = lines.iterator(); iterator.hasNext();) {
+                String line = iterator.next();
                 logger.debug("Adding module " + line);
                 model.getModules().add(line);
             }
@@ -126,14 +116,10 @@ public class TychoAggregatorMapping extends AbstractTychoMapping {
     }
 
     private boolean isCurrent(File file) {
-        try {
-            try (BufferedReader reader = new BufferedReader(
-                    new InputStreamReader(new FileInputStream(file), getPrimaryArtifactCharset()),
-                    TYCHO_AUTOMATIC_GENERATED_FILE_HEADER.length() * 2)) {
-                String readLine = reader.readLine();
-                return readLine == null || !readLine.startsWith(TYCHO_AUTOMATIC_GENERATED_FILE_HEADER_PREFIX)
-                        || readLine.equals(TYCHO_AUTOMATIC_GENERATED_FILE_HEADER);
-            }
+        try (var lines = Files.lines(file.toPath(), getPrimaryArtifactCharset())) {
+            String firstLine = lines.findFirst().orElse(null);
+            return firstLine == null || !firstLine.startsWith(TYCHO_AUTOMATIC_GENERATED_FILE_HEADER_PREFIX)
+                    || firstLine.equals(TYCHO_AUTOMATIC_GENERATED_FILE_HEADER);
         } catch (IOException e) {
             //can't be sure, assume it is not stale then...
         }
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
@@ -17,10 +17,10 @@ package org.eclipse.tycho.pomless;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.Reader;
 import java.util.Arrays;
 import java.util.Properties;
+import java.util.function.Supplier;
 import java.util.jar.Attributes;
 import java.util.jar.Manifest;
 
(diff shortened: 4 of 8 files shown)


tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
```diff
@@ -103,18 +95,13 @@ protected String getPackaging() {
...
-            throws ModelParseException, IOException {
+    protected void initModel(Model model, Reader artifactReader, File artifactFile) throws IOException {
         logger.debug("Generate aggregator pom for " + artifactFile);
         try (BufferedReader reader = new BufferedReader(artifactReader)) {
-            String line;
-            while ((line = reader.readLine()) != null) {
-                if (line.startsWith("#") || line.trim().isEmpty()) {
-                    continue;
-                }
+            reader.lines().filter(l -> !l.startsWith("#") && !l.isBlank()).map(String::strip).forEach(line -> {
```
Bananeweizen, line 101: You are aware of Oracles warning against side effects in forEach? A more clean functional style would be ...collect(Collectors.toList()) and then modules.addAll(collectedList).

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
```diff
@@ -261,18 +255,20 @@ protected Charset getPrimaryArtifactCharset() {
...
-        Properties properties = new Properties();
-        File buildProperties = new File(dir, "build.properties");
-        if (buildProperties.exists()) {
-            try (FileInputStream stream = new FileInputStream(buildProperties)) {
+        Properties buildProperties = new Properties();
+        loadProperties(new File(dir, "build.properties"), buildProperties);
+        return buildProperties;
+    }
+
+    static void loadProperties(File propertiesPath, Properties properties) throws IOException {
```
Bananeweizen, line 266: On quick glance it looks like you could always create the properties inside, since the input is always empty. I might be wrong, however.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
laeubi: @Bananeweizen maybe you can explain what specific problems you see here?

Bananeweizen: I don't see a specific problem in this piece of code, but it's easy to make it problematic. Just have someone extract reader.lines() to a method in the future, and then someone else add a ".parallel()" in that method. Boom! Or someone adds multi-threaded access to the collection that is modified in the forEach. Admitted, that's all far-fetched for this piece of code, but at least I have seen similar chains of refactorings leading to problems due to multiple people changing code independently too often already.
On the other hand, if you use the result.addAll(someStream.collect(toList()) version, it's fault proof for all those mentioned change scenarios and many more.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
HannesWell: In `TychoBundleMapping.initModel()` this is called multiple times with the same properties object in order to avoid multiple loads of the same file for each localized property.
However I reworked that part of the code and this method now creates and returns a Properties object on each invocation.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
laeubi: > reader.lines() to a method in the future, and then someone else add a ".parallel()" in that method. Boom!

What will boom? The only thing the javadoc mentions is that order is not guaranteed (something I won't expect anyways) but as module order is not relevant I thing it won't become a problem. That's really a bit hypothetical here.

HannesWell: Theoretical you are right, but I think in practice this is not likely a problem. If one chooses to use a parallel stream the entire stream-pipeline should be revised any way. But since that pipeline is not long I think that is something that will be noticed.
However I think it is also not harmful to change it in the suggested way, since creating and iterating over the List will likely not have any performance impact. So I changed it.

Bananeweizen (commented):
To me it looks good.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
```diff
@@ -261,20 +258,47 @@ protected Charset getPrimaryArtifactCharset() {
...
+    static Supplier<Properties> getPropertiesSupplier(File propertiesPath) {
+        FutureTask<Properties> t = new FutureTask<>(() -> loadProperties(propertiesPath));
+        return () -> {
+            try { // Load properties on first demand and cache result
+                t.run(); // No-Op if already running
+                return t.get();
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+            } catch (ExecutionException e1) { // ignore externalzied data
+            }
```
HannesWell, line 286: I just wondered if a `IllegalStateException` should be thrown here using the `ExecutionException`'s cause as its cause?
Before the `TychoBundleMapping` propagates exceptions while loading the localization while the `TychoFeatureMapping` ignored the externalized data when loading the properties failed with an exception.

I think it would be better to propagate the exception to not swallow the failure.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
```diff
@@ -261,20 +258,48 @@ protected Charset getPrimaryArtifactCharset() {
...
+        if (propertiesPath.isFile()) {
+            try (FileInputStream stream = new FileInputStream(propertiesPath)) {
                 properties.load(stream);
             }
         }
         return properties;
     }
 
+    static Supplier<Properties> getPropertiesSupplier(File propertiesFile) {
+        FutureTask<Properties> t = new FutureTask<>(() -> loadProperties(propertiesFile));
```
Bananeweizen, line 278: I haven't seen this pattern of using a FutureTask for "cached singletons" before (in fact, I've probably never used a FutureTask myself yet). Do you typically use that instead of double-checked locking, initialization-on-demand-holder, or just static maps for caches? I'm really curious to learn about that pattern, I definitely don't ask for changes.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
laeubi: Maybe then we better stick with a simple iteration instead of "stream" style at all...

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
laeubi: Localization is something one could assume as optional and thus I don't think a failure is appropriate.

HannesWell: I'm not sure if this is such a minor problem then.
For example many Eclipse Plugins use `%Bundle-Name` and `%Bundle-Vendor` in the Manifest headers for their name/vendor. Imagine the read fails on the release build (for whatever reason and for whatever reason it only fails for that file), then we publish a Plug-in that uses that variable name as its Bundle-Name/Vendor.
It probably won't break anything, but at least it is nasty. And as somebody who builds and publishes some bundles I would like to get an error.

Actually I would even go one step further and fail if the substitution for the variable is not present in order to detect errors where a variable is used but not specified.

laeubi: This is only used here to fill the maven info, the goal of the pomless build is not to validate things, this could be done in the actual build if required. Also internationalization could be part of a fragment as well.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
HannesWell: This is not a pattern I use regularly. I actually came up with this to implement a concurrent lazy computing cache.
It was something like:

```
  private static class LazyConcurrentCache<K, V> {
    private final Map<K, FutureTask<V>> map = new ConcurrentHashMap<>();

    private final Function<K, FutureTask<V>> mappingFunction;

    public LazyConcurrentCache(Function<K, V> function) {
      this.mappingFunction = k -> new FutureTask<>(() -> function.apply(k));
    }

    public V getValue(K key) throws InterruptedException {
      FutureTask<V> task = map.computeIfAbsent(key, mappingFunction);
      task.run();
    private static class LazyConcurrentCache<K, V> {
    private final Map<K, FutureTask<V>> map = new ConcurrentHashMap<>();

    private final Function<K, FutureTask<V>> mappingFunction;

    public LazyConcurrentCache(Function<K, V> function) {
      this.mappingFunction = k -> new FutureTask<>(() -> function.apply(k));
    }

    public V getValue(K key) throws InterruptedException {
      FutureTask<V> task = map.computeIfAbsent(key, mappingFunction);
      try {
        task.run();
        return task.get();
      } catch(InterruptedException e) {
        throw e;
      } catch(ExecutionException e) {
        Throwable cause = e.getCause();
        if(cause instanceof RuntimeException re) {
          throw re;
        }
        throw (Error) cause;
      }
    }
  }    try {
        return task.get();
      } catch(InterruptedException e) {
        throw e;
      } catch(ExecutionException e) {
        Throwable cause = e.getCause();
        if(cause instanceof RuntimeException re) {
          throw re;
        }
        throw (Error) cause;
      }
    }
  }
```
But in this case I mainly thought it would be a convenient replacement for a Mutable, which would avoid having a inital check if the value is present. But when thinking about it again, due to the extra exceptions using a `o.a.commons3.lang.mutable.Mutable` in the end it is simpler to understand and less fancy.

Your suggestion to use it as solution for the double-checked locking idiom seems interesting. Actually it should be a valid solution that does not suffer the problems of the `null-check, syncronized, null-check` approach. But if that is intended to be used more often I would maybe wrap that in a class that handles the different Exception types more gracefully. For example like in the following example.

The nested class can then be used like this
`new LazyValueHolder.LazyValueHolder<>(function).getValue()`
and only throws a `InterruptedException`. But even for that case default could be implemented (like return null, Optional.empty() or just propagate it) to make it more convenient.

```
    public static class LazyValueHolder<V, E extends Exception> {

        public static interface Callable2<V, E extends Exception> extends Callable<V> {
            V call() throws E;
        }

        private final FutureTask<V> task;

        public LazyValueHolder(Callable2<V, E> supplier) {
            this.task = new FutureTask<>(supplier);
        }

        @SuppressWarnings("unchecked")
        V getValue() throws E, InterruptedException {
            try {
                task.run();
                return task.get();
            } catch (InterruptedException e) {
                throw e;
            } catch (Exception e) {
                throw (E) e.getCause();
            }
        }

        public static class Runtime<V> extends LazyValueHolder<V, RuntimeException> {
            public Runtime(Supplier<V> supplier) {
                super(supplier::get);
            }
        }
    }
```

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
HannesWell: > This is only used here to fill the maven info, the goal of the pomless build is not to validate things, this could be done in the actual build if required. Also internationalization could be part of a fragment as well.

Indeed. I didn't consider that. Then we can treat this as optional.
But then localizedValue() should return null in case a translation is not found to not write `%whatever` into the pom?

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
HannesWell: > Maybe then we better stick with a simple iteration instead of "stream" style at all...

Probably, yes. What do you both think about the current approach?
This does not create a List and also ensures the iteration is done serially even if the stream-pipeline is parallelized.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
laeubi: The translation is applied as "best effort", in general the contract for pomless is that it tries to extract as much information as possible, if you follow the convention, if not one has to customize it (e.g. build.properties) or use `pom.xml`

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
laeubi: I just wanted to note that we here do not need high-performance to the edge parallel processing, I'm even not sure if project models are even read in parallel. So we should not over optimize here without actual proof that it is a performance bottleneck as it makes debugging such code much harder.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
laeubi: I think creating a list to add it to another list to account for "probably" future refactorings is actually wasted effort, a parallel processing of reading a file which applies trivial operations is even waste of time and most likely "slower", so if streams make the code
a) easier to understand
- or -
b) easier to debug

we should change it if it is just "because we can do it" we should not use it, especially if more complex exception handling is required, streams are an *option* not an requirement nor do they produce "better" or "faster" code.

HannesWell: Fully agree.
And I think the latest patch-set reflects that (unfortunately the discussion does not move with subsequent changes.
Just in case you missed that:
```
        try (BufferedReader reader = new BufferedReader(artifactReader)) {
            Stream<String> lines = reader.lines().filter(l -> !l.startsWith("#") && !l.isBlank()).map(String::strip);
            for (String line : (Iterable<String>) lines::iterator) {
                logger.debug("Adding module " + line);
                model.getModules().add(line);
```
Let me know if you think that's not fine.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
HannesWell: > The translation is applied as "best effort", in general the contract for pomless is that it tries to extract as much information as possible, if you follow the convention, if not one has to customize it (e.g. build.properties) or use `pom.xml`

Yes, agree as well. But do you think a value to translate (i.e. starting with a '%') should be kept if no translation is found?
In this case I think it would be better to assume no such value exists at all and stick with the default, since with "best effort" no actual value is found.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
HannesWell: > I just wanted to note that we here do not need high-performance to the edge parallel processing, I'm even not sure if project models are even read in parallel. So we should not over optimize here without actual proof that it is a performance bottleneck as it makes debugging such code much harder.

Yes sure. The displayed code, was just in response to @Bananeweizen's question and totally unrelated to this change. :)
I just used `FutureTask` for convenience, but changed it to use a `MutableObject` since it turned out to not be that convenient in the end. So this is plain serial programming. The only optimization is that the properties file is only load on the first translated value found.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
laeubi: This should already be the case that the value is retained but I'm not sure if maybe the code behaves like if emtpy value then no name is set at all...

HannesWell: At the moment `TychoBundleMapping` and `TychoFeatureMapping` simply uses the 'raw-value' of a property if it starts with `%` and there is no translation available.
It would IMHO be better to assume no value than, than using a value that actually has to be translated first.

laeubi: > At the moment `TychoBundleMapping` and `TychoFeatureMapping` simply uses the 'raw-value' of a property if it starts with `%` and there is no translation available.

That's how it is meant to be, the "key" is not a technical key but a "default value if no translation is available" see http://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#i3189742

HannesWell: OK. I changed it to return the 'key' (i.e. the raw-value without %-prefix) if no translation is found.
After checking again this was the behavior for bundles but not for features (which just used the raw-value as it was).

HannesWell:
@laeubi, @Bananeweizen are you now fine with this change? I think all remarks should be addressed.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
laeubi: Yes it might be inconsistent (aka not tested) here for features :-)

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
```diff
@@ -261,20 +258,45 @@ protected Charset getPrimaryArtifactCharset() {
...
+        if (propertiesPath.isFile()) {
+            try (FileInputStream stream = new FileInputStream(propertiesPath)) {
                 properties.load(stream);
             }
         }
         return properties;
     }
 
+    static Supplier<Properties> getPropertiesSupplier(File propertiesFile) {
+        Mutable<Properties> props = new MutableObject<>();
```
laeubi, line 278: Why do we need the Mutable stuff here? I think you can replace this with a simple inner class of a Supplier and use an instance field instead of wrapping the mutable in a lamda.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractXMLTychoMapping.java
```diff
@@ -96,4 +98,11 @@ protected static String getXMLAttributeValue(Element element, String attributeNa
     public float getPriority() {
         return 1;
     }
+
+    protected static Stream<File> filesWithExtension(File projectRoot, String extensions) throws IOException {
+        Predicate<String> nameFilter = n -> !n.startsWith(".polyglot.") && n.endsWith(extensions);
+        return Files.walk(projectRoot.toPath(), 1) //
```
laeubi, line 104: In the past I have seen problem with some obscure (windows) directories that could not be queried and listFiles() returning null, I wonder if `projectRoot.toPath()` is guaranteed to always return non `null`?

extensions --> extension ?
projectRoot --> directory?

Also one could think about passing a Path object instead of a File object here, and even though we need to have more file.toPath calls at the caller side, we might later on use more path API without converting back and forth.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
```diff
@@ -103,15 +95,11 @@ protected String getPackaging() {
...
+    protected void initModel(Model model, Reader artifactReader, File artifactFile) throws IOException {
         logger.debug("Generate aggregator pom for " + artifactFile);
         try (BufferedReader reader = new BufferedReader(artifactReader)) {
-            String line;
-            while ((line = reader.readLine()) != null) {
-                if (line.startsWith("#") || line.trim().isEmpty()) {
-                    continue;
-                }
+            Stream<String> lines = reader.lines().filter(l -> !l.startsWith("#") && !l.isBlank()).map(String::strip);
+            for (String line : (Iterable<String>) lines::iterator) {
```
laeubi, line 102: Why do we need a cast here? if its because of the lamda, why not calling lines.iterator()?

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
```diff
@@ -49,7 +49,7 @@ public class TychoBundleMapping extends AbstractTychoMapping {
     @Override
     protected boolean isValidLocation(String location) {
         File polyglotArtifactFile = new File(location);
-        if (polyglotArtifactFile.isDirectory() && polyglotArtifactFile.getName().equals(META_INF_DIRECTORY)) {
+        if (polyglotArtifactFile.getName().equals(META_INF_DIRECTORY) && polyglotArtifactFile.isDirectory()) {
             return new File(polyglotArtifactFile, MANIFEST_MF).exists();
```
laeubi, line 53: Maybe `.exists() -> isFile() `to be consitent to your other changes?

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
```diff
@@ -115,7 +111,7 @@ protected void initModel(Model model, Reader artifactReader, File artifactFile)
 
     @Override
     protected Properties getEnhancementProperties(File file) throws IOException {
-        //Look up build.properties in the project's root. The passed file points to the 'META-INF' folder.
+        //Look up build.properties in the project's root. The given file points to the 'META-INF' folder.
```
laeubi, line 114: maybe one should check here:

1. name of directory is META-INF
2. file is a directory

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoRepositoryMapping.java
```diff
@@ -125,22 +119,15 @@ protected boolean isValidLocation(String location) {
...
-            for (File file : productFiles) {
-                if (file.isFile()) {
-                    return file;
-                }
-            }
-        }
-        File category = new File(projectRoot, CATEGORY_XML);
-        if (category.exists()) {
-            return category;
+        Optional<File> repoFile = Optional.empty();
```
laeubi, line 122: why assign it here if we overwirte it in the next line? If it is for the exception case, better return the empty optional there. and move the return code into the try block

laeubi (changes requested):
I added some remarks now.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractXMLTychoMapping.java
HannesWell: > In the past I have seen problem with some obscure (windows) directories that could not be queried and listFiles() returning null, I wonder if `projectRoot.toPath()` is guaranteed to always return non `null`?

I assumed that. But now I checked the doc and nothing of `File.toPath()` and `FileSystem.getPath(String)` (which does the work), mentions any case where null is returned. Unfortunately my Eclipse does not show me any implementation of FileSystem so I cannot verify it at the moment.

> 
> extensions --> extension ? projectRoot --> directory?

Done, thanks.

> 
> Also one could think about passing a Path object instead of a File object here, and even though we need to have more file.toPath calls at the caller side, we might later on use more path API without converting back and forth.

Sure. Actually personally I prefer the Path API over the the File API, because many thinks are just more convenient and easier to make performant. However I didn't want to change everything just for my personal preference. And unfortunately the Maven API also uses File. But I can change the parameter. However changing the return value requires so many back-conversion at the callers that I think it is not profitable at the moment.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
HannesWell: Yes because of the lambda and you can only use an enhanced-for-loop with an Iterable, not with an iterator.
And without the cast, the 'assignment-target' of lines::iterator is not considered as an functional interface and therefore results in a compile error.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
HannesWell: Yes.
But actually this could be simplified to:
`return new File(location, META_INF_DIRECTORY + "/" + MANIFEST_MF).isFile();`
Which is also more performant because the File-System is only queried once.
To avoid the hard-coded line-delimiter, we can use this:
`return Files.isRegularFile(Path.of(location, META_INF_DIRECTORY, MANIFEST_MF));`

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
HannesWell: Done.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractXMLTychoMapping.java
laeubi: If its not in the API we don't need to handle it but [File.listFiles](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#listFiles()) says:

> Returns null if this abstract pathname does not denote a directory, or if an I/O error occurs.

I thin the "if an I/O error occurs" is covered here by throwing IOException already....

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoRepositoryMapping.java
HannesWell: The intention is to check the category.xml even if the product file search failed.
At least before the category.xml was checked if the file search returned an empty error or null (the File.listFiles() ) never throws.
I moved the assignment to the catch clause to make this more clear.
Is that ok?

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
laeubi: I would suggest simply use while with iterator then.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
laeubi: > just used `FutureTask` for convenience

Just a note, `CompleteableFuture` is more powerful than `FutureTask`

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractXMLTychoMapping.java
HannesWell: > I thin the "if an I/O error occurs" is covered here by throwing IOException already....

Yes and furthermore `.toPath()` should not perform actual file-system operations. IIRC it only throws if for example an invalid file-system is used or the path cannot be parsed. But then one usually cannot create a File object either.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
HannesWell: > Why do we need the Mutable stuff here? I think you can replace this with a simple inner class of a Supplier and use an instance field instead of wrapping the mutable in a lamda.

Good suggestion. A anonymous inner class makes this even simpler.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
HannesWell: A while loop would be the longest way to use that:
```
            Iterator<String> iterator = lines.iterator();
            while (iterator.hasNext()) {
                String string = iterator.next();
```
Using a for loop with an iterator at least only has two lines:
```
            for (Iterator<String> iterator = lines.iterator(); iterator.hasNext();) {
                String line = iterator.next();
```
But the current approach has only one. But if you prefer to stick to the classic way, I suggest to use the for-loop-with-iterator.

laeubi: > But the current approach has only one. But if you prefer to stick to the classic way, I suggest to use the for-loop-with-iterator.

Having less lines is at least not a goal here and if we can save creating a lambda + obscure cast we should do that.

laeubi:
Please let me know if you think I should review again, there is a "Re-request review" button next to reviewers for this :-)

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
HannesWell: > Having less lines is at least not a goal here and if we can save creating a lambda + obscure cast we should do that. You can even inline the lines variable to save a line and then we are equal in line count ;-)

That's right, but if you have less text to read (lines are IMHO not the ideal measure, since one can cheat and make it just a long line) one can also understand the code better. But I also understand that the cast+plus lambda is a unusual construct and can confuse people. It would be nice if the RHS of an enhanced for loop would behave like an assignment to a `Iterable<E>` variable. Then the case could have been omitted.

Anyway, I think the regular for loop with iterator is a good way to go. :)

HannesWell:
> Please let me know if you think I should review again, there is a "Re-request review" button next to reviewers for this :-)

Sure a review is always welcome. :) But I think I addressed all remarks from your previous review properly.

laeubi (approved):
Looks good, lets merge this when build succeeds.

HannesWell:
Thanks.

PR #1105, opened 2022-07-02
Embed m2e lifecycle-mapping-metadata in Tycho plug-ins

HannesWell:
The embedded metadata simply advice M2E to ignore executions of each Tycho Mojo.
In the Eclipse IDE Plug-in development support is PDE's duty and therefore Tycho Plug-ins have nothing to do.
Because the metadata embedded in a plugin are considered with lowest priority, a user can simply overrule this decision, if it is desired to execute a Tycho Plugin within the IDE.

Providing mappings for all Plug-ins is probably not necessary for most cases, but since theoretically one can explicitly bind a plug-in to any phase, it can also appear in a phase M2E is interested in (i.e. M2E is interested in all phases from validate until before test).

I used the simple app posted to https://github.com/eclipse-m2e/m2e-core/issues/830 to generate all mappings, so they should be complete (I verified that manually).

Fixes https://github.com/eclipse/tycho/issues/945

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,11 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## 3.0.0 (under development)
 
+### Eclipse M2E lifecycle-mapping metadata included
+
+All Tycho plugins are now shipped with embedded M2E lifecycle-mapping-metadata files.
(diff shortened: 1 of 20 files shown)


laeubi (approved):

HannesWell:
Thank you Christoph for your approval.

I just updated the xml files to use two-spaces instead of tabs for indentation, which is the default for the generated Xpp3-writers (I had changed that locally and forgot to revert that). For consistency I think we should stick with that default.
However for this white-space change it is not necessary to wait for a full build.
-> Merging

PR #1109, opened 2022-07-04
Update to Tycho release 2.7.4 and remove now obsolete M2E-connector

HannesWell:


HannesWell:
@laeubi any idea why the tests fail?

laeubi:
Seems like eclipse infra has (not yet) fetched some artifacts? GH action succeeds... if you pgrade the tycho version, can you also update the RELESE_NOTES by just copy the new entries 2.7.4 from the 2.7.x branch and add them to the master RELESE_NOTES.

HannesWell:
> Seems like eclipse infra has (not yet) fetched some artifacts? GH action succeeds... if you pgrade the tycho version, can you also update the RELESE_NOTES by just copy the new entries 2.7.4 from the 2.7.x branch and add them to the master RELESE_NOTES.

Just waiting a bit would be a simple fix. :)

Copied the release notes. Shouldn't I then remove the entry e.g. for `Eclipse M2E lifecycle-mapping metadata embedded` from the 3.0.0 section? Otherwise it would be duplicated. If yes are there more entries to remove?

laeubi:
I don't think they need to be removed.

PR #1117, opened 2022-07-08
Generify PomDependencyCollector

akurtakov:


laeubi:
@akurtakov sadly this will not (yet) work as the `IInstallableUnit` in the "Maven World" is not compatible with the one in "OSGi World", so don't be confused.

akurtakov:
Closing. It would be good if you can give some extra details what's going on here.

laeubi:
Tycho starts an OSGi Framework (that is the [tycho-bundle-product](https://github.com/eclipse/tycho/blob/master/tycho-bundles/tycho-bundles-external/tycho-bundles-external.product)) for the P2 parts and calls into this framework with this "shared" classes (what we export as system-packages), sadly we can currently not share the P2 classes itself, that's why "untyped" Collections are used here. It would be good if we can export the P2 API as well, but because P2 uses API+Impl in one bundle and heavily relies on require bundle this is not easy. Thus I already started to refactor P2 but that's quite hard as well (to not 'break' existing code).

Beside that, I have started to implement some of this functionality in "pure java" that could be used in the maven world as well, but currently these object can't 'cross the border'. This works quite well when it comes to the resolver itself, but the repository part its quite complex as it mangles a lot of concepts (extension points, OSGI service and Agent service).

So I hope that once there will be a point where we "meet in the middle" and can either join both worlds or even get rid of the embedded OSGi-Framework and then we can not only use generics here but throw away a lot of stuff that currently converts back-and-forth.

If you like to dig into this, you can try to export some packages of P2 into the OSGi world and look whats missing and how far we can get...

laeubi:
@akurtakov FYI related P2 issue: https://github.com/eclipse-equinox/p2/issues/22

PR #1120, opened 2022-07-10
Add includeAllSources property to include all available source bundles

laeubi:
This is a first draft to fix https://github.com/eclipse/tycho/issues/926 but currently misses an integration test to verify the functionality.

How it works:

1. Let the slicer do its job
2. Generate an IU that greedily fetches any source bundle using the usual naming scheme (@merks is there a more P2 way to find the source for a bundle than adding `.source` to the id?)
3. Perform another slicer operation and adds the result to the final result.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,11 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## 3.0.0 (under development)
 
+### Support for inclusion of all source bundles in an update-site
+
+The [tycho-p2-repository-plugin:2.7.0:assemble-repository](https://www.eclipse.org/tycho/sitedocs/tycho-p2/tycho-p2-repository-plugin/assemble-repository-mojo.html) now support a new property `includeAllSources` that,
+when enabled, includes any available source bundle in the resulting repository.
+
 ### Support for Eclipse-Products with mixed Features and Plugins
 
 Tycho now supports building _mixed_ Products. In mixed Products both the listed features and listed plug-ins are installed.
--- a/tycho-bundles/org.eclipse.tycho.p2.tools.impl/src/main/java/org/eclipse/tycho/p2/tools/mirroring/MirrorApplication.java
+++ b/tycho-bundles/org.eclipse.tycho.p2.tools.impl/src/main/java/org/eclipse/tycho/p2/tools/mirroring/MirrorApplication.java
@@ -13,11 +13,15 @@
 package org.eclipse.tycho.p2.tools.mirroring;
 
 import java.net.URI;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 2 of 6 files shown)


HannesWell:
Thank you @laeubi for working on this.
I can offer to work out an integration test, but I will be on vacation until Friday and likely won't have the time for that today.

Since I'm not familiar with the details I cannot assess them in detail but from what I understand it looks good

But I would like to clarify what sources are included exactly especially for different combinations of `includeAllSources` and `includeAllDependencies`.

With respect to `includeAllDependencies` the p2-repo created by Tycho has the following content:
- `includeAllDependencies=true`
-> All bundles/features listed in the category.xml are included plus all transitive requirements (bundles/features).
- `includeAllDependencies=false`
-> Only bundles/features listed in the category.xml are included and those bundles/features that are (transitively) included in the listed ones.

If one combines this with `includeAllSources` I would expect the following content:
- `includeAllDependencies=true` and `includeAllSources=true`
-> All transitive requirements (bundles/features) and the source of each feature/bundle in the repo is included.
- `includeAllDependencies=true` and `includeAllSources=false`
-> All transitive requirements (bundles/features) are included and sources are only included if they are _included_ (transitively) in any feature listed in the category.xml.
- `includeAllDependencies=false` and `includeAllSources=true`
-> Only bundles/features listed in the category.xml are included and those bundles/features that are (transitively) included in the listed ones and the source of each feature/bundle in the repo is included.

So simply `includeAllSources` just ensures that for each bundle/feature in the repo, the corresponding source artifact is added as well (if not yet included anyways), regardless of how that bundle/feature made it into the repo.
Is that the case?

At the moment we have `includeAllDependencies=true` and `includeAllSources=false` for the Eclipse-Platform repo and could then switch to `includeAllSources=true`.

laeubi:
> At the moment we have `includeAllDependencies=true` and `includeAllSources=false` for the Eclipse-Platform repo

I'm not sure if `includeAllDependencies=true` is used, but regardless of what is sued for every bundle contained in the final repository a source bundle will be included if found in the current target context.

akurtakov:
What is the status of this one ?

laeubi:
Code is there but @HannesWell wanted to provide a test-case for this so we can verify this works as intended.

HannesWell:
> Code is there but @HannesWell wanted to provide a test-case for this so we can verify this works as intended.

Sorry I forgot a bit about that. I try to provide one before my (other) vacation next week.

akurtakov:
@laeubi would you please rebase it on top of master and guide whether and how I can publish extra commit to this PR with an integration test.

laeubi:
> would you please rebase it on top of master

Will do this ...

> and guide whether and how I can publish extra commit to this PR with an integration test.

Its actually quite simple, branch from current master, add an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue and open a PR and we will see a failing test and thus proof a missing feature or regression. I can the cherry-pick that PR into my branch to proof my change fixes the test, so one could really work independently.

laeubi:
I plan to merge this once CI passes, we can enhance / add tests afterwards.

Issue #1124, opened 2022-07-11
Cycle now detected in SDK build


mickaelistria:
The eclipse.platform.releng.aggregator build is now failing with
```
[ERROR] [ERROR] The projects in the reactor contain a cyclic reference: Edge between 'Vertex{label='org.eclipse.e4:org.eclipse.e4.ui.workbench.renderers.swt.cocoa:0.12.600-SNAPSHOT'}' and 'Vertex{label='eclipse.platform.ui:org.eclipse.e4.ui.workbench.renderers.swt:0.15.500-SNAPSHOT'}' introduces to cycle in the graph eclipse.platform.ui:org.eclipse.e4.ui.workbench.renderers.swt:0.15.500-SNAPSHOT --> eclipse.platform.ui:org.eclipse.e4.ui.widgets:1.3.0-SNAPSHOT --> org.eclipse.e4:org.eclipse.e4.ui.workbench.renderers.swt.cocoa:0.12.600-SNAPSHOT --> eclipse.platform.ui:org.eclipse.e4.ui.workbench.renderers.swt:0.15.500-
```
I think it's a consequence of recent changes in the lifecycle.
The error seems actually correct. So we should first investigate whether this can be solved on Platform side.

mickaelistria:
See https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/PR-390/2/ for a full build showing the error.

laeubi:
> I think it's a consequence of recent changes in the lifecycle.

Yes, if you think this is wrong, it would be great to have a minimal example to investigate on this, while development I saw similar problems with the existing integration-tests when fragments involved, but I think the itest only cover one fragment and one host, so maybe we have here a case with multiple fragments but different hosts?

Can the aggregator job be run locally as well? It seem requiring a lot of submodules and setup of PGP stuff...

laeubi:
Just from the code, I think this might be because `org.eclipse.e4:org.eclipse.e4.ui.workbench.renderers.swt.cocoa` has

`Fragment-Host: org.eclipse.e4.ui.workbench.renderers.swt;bundle-version="[0.10.0,1.0.0)"`

and in the p2 inf additionally requires swt cocoa

```
requires.0.namespace = org.eclipse.equinox.p2.iu
requires.0.name = org.eclipse.swt.cocoa.macosx.x86_64
```

but that is just a guess here.

mickaelistria:
[...]

PR #1125, opened 2022-07-11
Make the classic resolver the default choice for now

laeubi:
This will for now make the classic resolver the default unless we find a good reproducer for the fragment problem.

@mickaelistria WDYT? This will allow the builds to succeed and we can enable the resolver to test individual builds, this might be less disturbing.

See  #1124


mickaelistria:
I don't think we need to make Tycho work immediately and can have Platform waiting a bit if there are good chances that we find a solution to this issue soon (eg tomorrow). If you are optimistic it will be fixed soon, I think we should keep the current issue and don't bother with such twaeks. If you think it requires more time, then yes, we'd better disable the change for the moment.

laeubi:
Yes, I think we first need to incooperate the test to cover the case of test-fragments and I need to dig a bit deeper how to solve this best (or: understand how the sat-solver resolves this case).

PR #1128, opened 2022-07-12
Reproducer for issue #1124

pzi-dsa:
Since we have the same issue in our project, here's a reproducer for the cycle.

I have 3 projects, plugin1, plugin2 and plugin2.tests with the following dependencies:

plugin1 -> plugin2
plugin2.test -> plugin1

Because plugin2.test is a fragment of plugin2, I think the dependencies of both plugins are merged, resulting in the cycle:
plugin1 -> plugin2
plugin2 -> plugin1

### DIFF
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/resolver/CycleTest.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/resolver/CycleTest.java
@@ -1,0 +1,30 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Patrick Ziegler and others.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Patrick Ziegler - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.tycho.test.resolver;
+
+import java.util.List;
+
+import org.apache.maven.it.Verifier;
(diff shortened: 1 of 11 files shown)


laeubi:
@pzi-dsa that's great (and simple) reproducer, lets see if the build can catch up with this, in the meanwhile can you check if we can reduce it even further, e.g. project files classpath and preferences are in most cases not needed for an integrationtest.

tycho-its/src/test/java/org/eclipse/tycho/test/resolver/CycleTest.java
```diff
@@ -0,0 +1,30 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Sonatype Inc. and others.
```
laeubi, line 2:
```suggestion
 * Copyright (c) 2022 Patrick Ziegler and others.
```

pzi-dsa:
> can you check if we can reduce it even further

Sure. That was just a quick & dirty reproducer for the issue. I'll wait for the build to complete to see whether the problem shows up, then I'll  clean up the test.

laeubi:
Maybe, but I really like to avoid binding it to the package type and it might be a problem with "classic" fragments as well, I think I'll only enable this for `Eclipse-ExtensibleAPI` bundles, and then one need to be careful unless I have implemented GrahBuilding to actually order the projects, maybe even maven can be enhanced with some kind of "ignore for dependency but take into account for ordering" flag :-)

Because the best would be if we omit any fragment from the build requirements and only attach them at resolve-time... I'll check if there is some way already for this.

laeubi (approved):
Looks good, as soon as the build only fails on this test we should merge it!

laeubi:
Thanks @pzi-dsa for this usefull test-case!

PR #1130, opened 2022-07-12
Use p2inf in DependencyComputerTest.testFragments

laeubi:
The DependencyComputerTest.testFragments simulates a build that includes
swt+jface but current SWT build uses a p2.inf to require additional
units. This currently makes this test behave a little bit different than
a actual swt build would behave. Because of this, this adds a p2.inf
file that is similar to what swt-build contains.


mickaelistria:
I don't think it's necessary as this test does actually test a particular subset of Tycho where p2 is not involved. For example, it does use `getSortedProjects()` with a variant that disable build plugins, so p2 is not even used here.

laeubi:
@mickaelistria do you think it would harm here? As it would better replicate what p2 does I'd like to include it if you don't mind but I think its fine to ommit it as well...

mickaelistria:
No objection, feel free to add it if you want; as long as the test is still capable of capturing regressions on the covered part, it's fine.

laeubi:
I'll close this for now and probably come back here later...

PR #1131, opened 2022-07-12
Do not depend on the topological order of projects in DependencyComputerTest

laeubi:
There are more similar places we should probably replace, this is just to show an example.


mickaelistria:
Yes, I think it's OK.

Issue #1127, opened 2022-07-12
Tycho seems no longer to be able to resolve dependencies from same reactor


cdietrich:
https://ci.eclipse.org/mwe/job/mwe2/job/cd_tycho30/155/console

started to fail today

https://ci.eclipse.org/mwe/job/mwe2/job/cd_tycho30/155/console
```
23:45:44  [ERROR] Failed to execute goal on project org.eclipse.emf.mwe2.language: Could not resolve dependencies for project org.eclipse.emf:org.eclipse.emf.mwe2.language:eclipse-plugin:2.14.0-SNAPSHOT: Could not find artifact org.eclipse.emf:org.eclipse.emf.mwe2.lib:jar:2.14.0-SNAPSHOT -> [Help 1]
```

but this is part of the reactor

side effect of https://github.com/eclipse/tycho/commit/c2757f6d35c4e5fb76d63c2ce71153702316af5c ?

can it also be this causes
https://github.com/itemis/xtext-reference-projects/runs/7295160649?check_suite_focus=true
?
(here is debug log: https://pipelines.actions.githubusercontent.com/serviceHosts/7f567f03-0c97-4d1b-94b0-cb817cff75a8/_apis/pipelines/1/runs/2447/signedlogcontent/32?urlExpires=2022-07-12T05%3A07%3A36.0896977Z&urlSigningMethod=HMACV1&urlSignature=4Ab1vQZQzKiTlv9%2BiCBPHFaqbq0EPAYVA%2Bnurgbd8dw%3D)

```
2022-07-12T05:00:43.6647688Z org.eclipse.tycho.core.osgitools.OsgiManifestParserException: Exception parsing OSGi MANIFEST /home/runner/work/xtext-reference-projects/xtext-reference-projects/domainmodel/2.28.0/org.eclipse.xtext.example.domainmodel.ide/target/com.googlecode.addjars.mojo.AddJarsMojo/META-INF/MANIFEST.MF: Manifest file not found
2022-07-12T05:00:43.6649343Z     at org.eclipse.tycho.core.osgitools.DefaultBundleReader.loadManifestFromDirectory (DefaultBundleReader.java:96)
2022-07-12T05:00:43.6650588Z     at org.eclipse.tycho.core.osgitools.DefaultBundleReader.doLoadManifest (DefaultBundleReader.java:63)
[...]

PR #1132, opened 2022-07-12
it for mwe part of #1127

cdietrich:
it for mwe part of #1127

@laeubi  unfortunately the same test that fails locally does not fail in test

### DIFF
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/compiler/MavenCompilerPluginTest.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/compiler/MavenCompilerPluginTest.java
@@ -18,6 +18,7 @@ import org.apache.maven.it.Verifier;
 import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
 import org.junit.Assert;
 import org.junit.Test;
+import java.util.List;
 
 public class MavenCompilerPluginTest extends AbstractTychoIntegrationTest {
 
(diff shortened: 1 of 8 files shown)


laeubi:
I set this ready for review so we execute the integration tests...

tycho-its/src/test/java/org/eclipse/tycho/test/compiler/MavenCompilerPluginTest.java
```diff
@@ -40,6 +40,15 @@ public void testAdditionalBundles() throws Exception {
         verifier.verifyErrorFreeLog();
     }
 
+
+    @Test
+    public void testAdditionalBundles2() throws Exception {
+        Verifier verifier = getVerifier("compiler.additional.bundles2", true);
```
laeubi, line 46:
```suggestion
        Verifier verifier = getVerifier("compiler.additional.bundles2", false, false);
```

Should be more suitable here, then you automatically ignore local artifacts.

cdietrich:
you can find the code i copy & pasted over here:
https://github.com/cdietrich/tycho1127_addjars_issue/tree/cd_mimicMweProblem
i basically just ported the groupid.
i dont know what is different in test

cdietrich:
ok, if i call install goal, it now fails

laeubi:
@cdietrich is the failure of testAdditionalBundles2 (org.eclipse.tycho.test.compiler.MavenCompilerPluginTest) with error:

https://github.com/eclipse/tycho/runs/7307088143

what you would expect here? If yes can you remove the 

`verifier.addCliOption("-Dtycho.resolver.classic=false -Dtycho.localArtifacts=ignore");`

so we can see that it succeeds on the current build? If yes we can merge this ...

cdietrich:
@laeubi yes, updated the pr. it now should become green again

laeubi:
Thanks for the additional test!

PR #1134, opened 2022-07-12
Improve the new resolver stage

laeubi:
Next try with enhanced test-suite ...

### DIFF
--- a/tycho-core/src/test/java/org/eclipse/tycho/core/test/TychoTest.java
+++ b/tycho-core/src/test/java/org/eclipse/tycho/core/test/TychoTest.java
@@ -12,10 +12,7 @@ package org.eclipse.tycho.core.test;
 
 import java.io.File;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.List;
-import java.util.Set;
-import java.util.stream.Collectors;
 
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Dependency;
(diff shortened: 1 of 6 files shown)


laeubi:
It seems all (existing) integration test are passing without the fragment attachment, so it was just the junit-tests that kidding me here :-(

tycho-core/src/test/java/org/eclipse/tycho/core/test/TychoTest.java
```diff
@@ -113,47 +110,48 @@ public void testProjectPriority() throws Exception {
...
-
-        // depends on host and because host has ExtensibleAPI also depends fragment and fragent2
-        MavenProject client = projects.stream().filter(p -> p.getArtifactId().equals("client")).findAny().get();
-        List<Dependency> clientDependencies = client.getModel().getDependencies();
-        assertEquals(4, clientDependencies.size());
-        assertEquals(Set.of("host", "fragment", "dep", "fragment2"),
-                clientDependencies.stream().map(Dependency::getArtifactId).collect(Collectors.toSet()));
-    }
+//    @Ignore
+//    public void testFragment() throws Exception {
```
laeubi, line 114: @mickaelistria I think this is a left-over from the old PDE based resolver and relies on the ordering of fragments by the ExtensibleAPI header, I think we can remove this:

1. We no longer use the old PDE used resolver but "upgraded" to new equinox one
2. the test uses the P2StubResolver what still uses the old code, so this does not really test what Tycho do in a build
3. ExtensibleAPI is nowhere else referenced, and even disabled in the p2StubResolver

laeubi:
FYI @cdietrich  your CycleTest also passes now. about the other failures in GH actin there seems a corrupted file I'll try to add a workaround here...

PR #1151, opened 2022-07-18
Change default and deprecate requireJREPackageImports, add release note

laeubi:
Fixes https://github.com/eclipse/tycho/issues/1136

### DIFF
--- a/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
+++ b/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
@@ -209,9 +209,24 @@ public abstract class AbstractOsgiCompilerMojo extends AbstractCompilerMojo
      * the design-time equivalent to the equinox runtime option
      * <a href="https://wiki.eclipse.org/Equinox_Boot_Delegation#The_solution"
      * >osgi.compatibility.bootdelegation</a>.
+     * 
+     * @deprecated OSGI requires all packages to be imported and support for
+     *             osgi.compatibility.bootdelegation will be removed in one of the next Tycho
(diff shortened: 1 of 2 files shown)


laeubi:
At least none of the integration test care here...

mickaelistria:
>   You requested to install 'osgi.bundle; org.bouncycastle.bcpg 0.0.0' but it could not be found

This should be fixed now.
If there is some risk of Tycho being broken, I'd rather wait after next I-Build is complete. We didn't have a good build of Platform for 4 days; having 1 proper one before changing Tycho would be useful.

tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
```diff
@@ -642,7 +657,14 @@ private void configureBootclasspathAccessRules(CompilerConfiguration compilerCon
             throws MojoExecutionException {
         List<AccessRule> accessRules = new ArrayList<>();
 
-        if (requireJREPackageImports) {
+        if (requireJREPackageImports != null) {
```
mickaelistria, line 660: boolean cannot be null, so warning would always show up.

mickaelistria (changes requested):

mickaelistria:
Actually, you don't need to merge this patch to test it on platform. Just submit a PR to the platform parent pom that set the existing flag to `true` and you'll see what happens.

laeubi:
> Actually, you don't need to merge this patch to test it on platform.

I'll run a aggregator local build, I assume everything is fixed now?

mickaelistria:
>  assume everything is fixed now?

Yes.

> I'll run a aggregator local build,

As you prefer, but sending a PR will trigger a full build without running tests on CI.

PR #1152, opened 2022-07-18
Implement the OSGi-Framework-Connect on top of Plexus

laeubi:
OSGi provides a new specification that allows powerful integration of
"outside" bundles into a framework.

Tycho can leverage this to have a seamless integration with the P2 code
that requires OSGi to run but still interact with the OSGi parts without
additional integration/conversion patterns.


mickaelistria:
At the moment, it's not clear how this change makes the integration simpler or more powerful. Can you please provide an example by the way?

laeubi:
Full example is here, where i extracted the generic parts from: https://github.com/eclipse/tycho/pull/718

The interesting part is here: https://github.com/eclipse/tycho/blob/f9dd2d1217d1294b43b9ec93cabd00ecb5224682/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryConnectorFactory.java#L71

```
IArtifactRepositoryManager repositoryManager = services.getProvisioningAgent()
						.getService(IArtifactRepositoryManager.class);
```

where from a maven mojo I can directly access services/object and there classes from within the connect-framework, while in current Tycho OSGi embedding they have distinct class-spaces and we have all these shared stuff that converts back and forth and also need to use anonymous collections.

laeubi:
e.g. https://github.com/eclipse/tycho/pull/1117 is a good example....

mickaelistria:
Does that imply that all occurrences of `Set<?/* IInstallableUnit */>` in Tycho code base will then be replaceable by `Set<IInstallableUnit>` and work everywhere?

laeubi:
> Does that imply that all occurrences of `Set<?/* IInstallableUnit */>` in Tycho code base will then be replaceable by `Set<IInstallableUnit>` and work everywhere?

As a first step, yes, furthermore we can actually remove many codes that abstract and convert back and fourth... but this will be some work and most likely will be done step by step.

laeubi:
This relates to https://github.com/eclipse/tycho/discussions/1156

laeubi:
And finally, if we can get rid of the other framework in Tycho, we can also get rid of the bundle-product and everything is defined in terms of maven dependencies, will make maintain them much easier.

mickaelistria:
Yes, the promise with this change is huge! It's really about half of the code in Tycho that might become useless, and a stronger typing everywhere that will prevent from a lot of bugs. This has been a dream for many years!

laeubi:
> This has been a dream for many years!

An finally dreams can become true thanks to the OSGi-R8-Framework-Connect :-)

I just wished I have discovered this a bit earlier before I have implemented it half way already :-(

PR #1168, opened 2022-07-19
Consider build.props 'pom.model.packaging' to determine test-plugins

HannesWell:
In order to specify if a Plug-in project, that does not follow the naming-convention, is a 'test'-project, one can specify in the build.properties file `tycho.pomless.testbundle=true|false`.

IMHO it would be more natural, if one could instead/also specify the packaging-type in the build.properties:
`pom.model.packaging = eclipse-test-plugin|eclipse-plugin`

What do the others think? This could be an alternative or even a replacement for `tycho.pomless.testbundle`. At the moment `pom.model.packaging` (defined by takari-polyglot) is not used by Tycho.

Additionally this moves the specification of all 'tycho.pomless-*' properties to AbstractTychoMapping, so that it is easier to get an overview.

### DIFF
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
@@ -52,8 +52,11 @@ import org.sonatype.maven.polyglot.mapping.Mapping;
  */
 public abstract class AbstractTychoMapping implements Mapping, ModelReader {
 
-    private static final String TYCHO_POMLESS_PARENT = "tycho.pomless.parent";
-    private static final String PARENT_POM_DEFAULT_VALUE = System.getProperty(TYCHO_POMLESS_PARENT, "..");
+    // All build.properties entries specifically considered by Tycho. Extends the list in Mapping interface
+    protected static final String TYCHO_POMLESS_PARENT_PROPERTY = "tycho.pomless.parent";
+    protected static final String TYCHO_POMLESS_AGGREGATOR_NAMES_PROPERTY = "tycho.pomless.aggregator.names";
+
+    private static final String PARENT_POM_DEFAULT_VALUE = System.getProperty(TYCHO_POMLESS_PARENT_PROPERTY, "..");
     private static final String QUALIFIER_SUFFIX = ".qualifier";
     private static final String MODEL_PARENT = "TychoMapping.model.parent";
 
@@ -168,7 +171,7 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
         }
         Properties buildProperties = getBuildProperties(projectRoot);
         // assumption parent pom must be physically located in parent directory if not given by build.properties
-        String parentRef = buildProperties.getProperty(TYCHO_POMLESS_PARENT, PARENT_POM_DEFAULT_VALUE);
+        String parentRef = buildProperties.getProperty(TYCHO_POMLESS_PARENT_PROPERTY, PARENT_POM_DEFAULT_VALUE);
         File fileOrFolder = new File(projectRoot, parentRef).getCanonicalFile();
         PomReference parentPom;
         if (fileOrFolder.isFile()) {
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
@@ -170,10 +170,12 @@ public class TychoBundleMapping extends AbstractTychoMapping {
 
     private boolean isTestBundle(String bundleSymbolicName, File bundleRoot) throws IOException {
         Properties buildProperties = getBuildProperties(bundleRoot);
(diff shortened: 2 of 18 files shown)


akurtakov:
I like the idea.
Bonus points for a test and some user documentation .

laeubi:
> is a 'test'-project, one can specify in the build.properties file `tycho.pomless.testbundle=true|false`.

This is from these days where there was no support for pomless (pom) properties.

> MHO it would be more natural, if one could instead/also specify the packaging-type in the build.properties:
> `pom.model.packaging = eclipst-test-plugin|eclipse-plugin`

why not

> At the moment `pom.model.packaging` (defined by takari-polyglot) is not used by Tycho.

Well this is a bit a chicken-egg-problem, because actually it *would* be used but *after* Tycho has computed the model and chosen if it is a test or not. So if we do not do any different thing beside the packaging type the resulting pom.model should get the packaging type adjusted automatically afterwards.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
```diff
@@ -52,8 +52,12 @@
  */
 public abstract class AbstractTychoMapping implements Mapping, ModelReader {
 
-    private static final String TYCHO_POMLESS_PARENT = "tycho.pomless.parent";
-    private static final String PARENT_POM_DEFAULT_VALUE = System.getProperty(TYCHO_POMLESS_PARENT, "..");
+    // All build.properties entries specifically considered by Tycho. Extends the list in Mapping interface
+    protected static final String TYCHO_POMLESS_PARENT_PROPERTY = "tycho.pomless.parent";
+    protected static final String TYCHO_POMLESS_AGGREGATOR_NAMES_PROPERTY = "tycho.pomless.aggregator.names";
+    protected static final String TYCHO_POMLESS_TESTBUNDLE_PROPERTY = "tycho.pomless.testbundle";
```
laeubi, line 58: Please remove the property and document it in the migration guide.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
```diff
@@ -170,11 +170,17 @@ private String getRequiredHeaderValue(String headerKey, Attributes headers, File
 
     private boolean isTestBundle(String bundleSymbolicName, File bundleRoot) throws IOException {
         Properties buildProperties = getBuildProperties(bundleRoot);
-        String property = buildProperties.getProperty("tycho.pomless.testbundle");
+        String property = buildProperties.getProperty(TYCHO_POMLESS_TESTBUNDLE_PROPERTY);
```
laeubi, line 173: As above, we should simply drop this code path and document it in the migration guide. It should be easy to perform a search+replace for anyone using it...

laeubi (changes requested):

HannesWell:
> > At the moment `pom.model.packaging` (defined by takari-polyglot) is not used by Tycho.
> 
> Well this is a bit a chicken-egg-problem, because actually it _would_ be used but _after_ Tycho has computed the model and chosen if it is a test or not. So if we do not do any different thing beside the packaging type the resulting pom.model should get the packaging type adjusted automatically afterwards.

Yes it is. I tested it and a `pom.model.packaging` entry overwrites anything else already. Therefore I have already changed the uses of `tycho.pomless.testbundle` in the Eclipse platform accordingly.


> I like the idea. Bonus points for a test and some user documentation .

Just added tests, but they maybe fail because I could not test them locally. My Tycho workspace is currently screwed since M2E does not import the `tycho-bundles` projects are Maven-projects and therefore sources are missing.

User doc was added here:
- https://github.com/eclipse/tycho/wiki/Tycho-Pomless#declare-test-plugins
- https://github.com/eclipse/tycho/wiki/Tycho-Pomless#specify-parent-path

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
HannesWell: Done.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
HannesWell: Removed it.
Documentation in the Release-notes? Or is there a dedicated guide? is still pending.
Have to mention that `pom.model.packaging = eclipse-test-plugin` can be used already.

laeubi:
Please add info the th Relase notes "Migration guide" section as well.

laeubi:
@HannesWell when you include the Migration notice I think this is fine to be merged.

HannesWell:
> @HannesWell when you include the Migration notice I think this is fine to be merged.

Thanks for the reminder. I added a notice and also (hopefully) fixed the tests.

laeubi (approved):
Lets merge once build completes!

HannesWell:
> Lets merge once build completes!

Build succeeded -> Merging.

Thank you for the reviews.

PR #1169, opened 2022-07-20
Rename sisu packages and API types

laeubi:


laeubi:
@akurtakov any hint for me? as soon as I rename the package the TargetPlatformMojo do not find the component anymore... and I don't have a clue why this should be related. I noticed that there are also Manifests generated but these seem fine... I even searched for `org.eclipse.sisu.equinox.embedder` in all test files and can't find a reference that explains this...

akurtakov:
Unfortunately I have never touched that code (it predates me) and have no clue. I'll try look into it later.

laeubi:
Thanks, I thing I'll then try to rename the stuff first and see if I find out why/what cause this... really strange, but this Plexus/Sisu Inject stuff is really bugging me....

laeubi:
I think the issue is a dependecy between sisu und tycho core... I opened https://github.com/eclipse/tycho/pull/1170 where i stepwhise renamed things...

laeubi:
@akurtakov I think i found the cause, it is because the maven-compat references sisu-inject, wich already contains a package `org.eclipse.sisu.osgi` and even reading https://eclipse.github.io/sisu.inject/ and looking at https://github.com/eclipse/sisu.inject/tree/master/org.eclipse.sisu.inject/src/org/eclipse/sisu/osgi I don't really get what this OSGi stuff is supposed to do (maybe if I run that inside an OSGi framework itself?) I wonder if it not would make sense to move the stuff from tycho to sisu (as both ar eclipse projects) but not sure about how active/focused/used sisu-inject is?? But for sure there was a reason to place this in Tycho in the past...

PR #1170, opened 2022-07-20
Rename the sisu-osgi interface to be more vendor-neutral

laeubi:
... but keep the package name for now as it seem to cuase issues ...


laeubi:
postponing for now...

PR #1171, opened 2022-07-20
Test+Fix https://github.com/eclipse/tycho/issues/1133

mickaelistria:


tycho-core/src/main/java/org/eclipse/tycho/osgi/configuration/MavenDependenciesResolverConfigurer.java
```diff
@@ -77,19 +76,18 @@ public Collection<?> resolve(String groupId, String artifactId, String version,
...
-                    logger.debug("[depth=" + trail.size() + ", scope matches =" + isValidScope(a, scopes) + "][" + a
-                            + "][" + trail.stream().collect(Collectors.joining(" >> ")) + "]");
-                }
-                if (trail.size() <= depth) {
-                    return isValidScope(a, scopes);
+                    logger.debug("[depth=" + trail.size() + ", scope=" + a.getScope() + "][" + a + "]["
+                            + trail.stream().collect(Collectors.joining(" >> ")) + "]");
                 }
-                return false;
+                return trail.size() <= depth;
```
laeubi, line 90: Why will it work to not filter for scopes anymore? That dosen't seem correct...

laeubi (changes requested):
I don't think that `CumulativeScopeArtifactFilter` is suitable here! From the javadoc:

>  Filter to only retain objects in the given scope or better. This implementation allows the accumulation of multiple scopes and their associated implied scopes

That's not what we want here, we want only explicitly given user scopes to be considered, not "better" or "implied" ones!

mickaelistria:
> don't think that CumulativeScopeArtifactFilter is suitable here! 

Yes it works as expected, unlike the current state.

> we want only explicitly given user scopes to be considered, not "better" or "implied" ones!

What we want is about the same result as `mvn dependency:tree`, eg if one use "test", in includes "compile". Maven dependency tree uses a similar filter.

mickaelistria:
As I'm trying to make the unit test more a unit one, and am facing difficulty towards that goal, it seems like the issue could be really specific to this commons-io bundle which has bom dependencies coming transitively and re-declared in other parts. So in the end, the symptom is that we don't receive the expected scope in the resolutionfilter for some artifacts. But I couldn't yet identify a "simple" scenario to reproduce that.
I think it might be a bug in Maven not setting the right scope here; but I'm not yet certain about that.
I'm going to remove the unit test, as at the moment there is not much that is a unit here.

laeubi:
> Yes it works as expected, unlike the current state.

No it only works for your particular case.

> What we want is about the same result as `mvn dependency:tree`, eg if one use "test", in includes "compile". Maven dependency tree uses a similar filter.

This is not how it is defined in m2e, this has nothing to do with dependency tree, the rule is that only that scopes are included that are explicitly given by the user.

mickaelistria:
> No it only works for your particular case.

It's not *my* case and it's a use-case that's going to affect all .target that want to include commons-io, so it's a tricly case, but still a legal and probably popular one. We cannot reject such use-cases just because you don't like them.

> This is not how it is defined in m2e, 

So I think in the current form, the current implementation, both in Tycho and m2e don't match basic user expectations and behave in a way that wouldn't be the expected one for regular users (who would typically expect things to behave like regular maven depedency resolution, as shown by dependency:tree). Usually, it doesn't cause any issue and works fine as it, but this could be improved and this use-case has just highlighted that.

laeubi:
> It's not _my_ case and it's a use-case that's going to affect all .target that want to include commons-io, so it's a tricly case, but still a legal and probably popular one. We cannot reject such use-cases just because you don't like them.

I din't say to reject the usecase, I just said that using the `CumulativeScopeArtifactFilter` only fixes it for *this* case, where there is no other dependency, so effectively you can fix this case by simply using `depth = 0` as well.

> So I think in the current form, the current implementation, both in Tycho and m2e 

I won't restart the discussion again, this was already decided and is simply because target + maven concerns are completely different so the user needs the choice. There is simply nothing to "improve" just because the implementation (either Tycho or maven) is flawed here, so as you always like to state, we can't break "the API" here.


> don't match basic user expectations and behave in a way that wouldn't be the expected one for regular users (who would typically expect things to behave like regular maven dependency resolution, as shown by dependency:tree). Usually, it doesn't cause any issue and works fine as it, 

Simply big No! `dependency:tree` is a completely different concern and has nothing to do with included target content and thus a user won't "expect" anything here. If you think you need a `dependency:tree` target location feel free to propose one, the maven target location do not work this way.

> but this could be improved and this use-case has just highlighted that.

This "use-case" shows nothing else but that there seems to be a bug for this particular case that must be fixed, there is no need to "reinvent the wheel" because of that.

mickaelistria:
I'm going to get rid of the ScopeFilter; but still I believe that the current behavior is not as good as it could be as users will actually expect dependency resolution here to honor scope like Maven does.

laeubi:
> but still I believe that the current behavior is not as good as it could be as users will actually expect dependency resolution here to honor scope like Maven does.

There is no "like Maven", because maven has different phases where different item are available see https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope

Just take **runtime** as a good example:

> This scope indicates that the dependency is not required for compilation, but is for execution. Maven includes a dependency with this scope in the runtime and test classpaths, but not the compile classpath.

So as targets do not have these different class scopes, this concept simply do not match and thus the simple rule is, that each artifact is included that has any scope that the user selected. That way you are able to include (and compile against) `runtime` dependencies, even though maven would **never include** them in the compile phase!

mickaelistria:
You seem to imply that I stated that only 1 scope should be allowed. That's not at all what I meant. What I state can be rephrased as "scope hierarchy and inheritance should be honored".

> So as targets do not have these different class scopes, this concept simply do not match 

I think what matches is that if I create a location of scope "test", I expect, just like everywhere in Maven that "compile" deps are included as well.

> Maven includes a dependency with this scope in the runtime and test classpaths, but not the compile classpath.

The CumulativeScopeFilter already implements that. The current change would still allow that.

> That way you are able to include (because you would need it to run the application) runtime dependencies in a target, even though maven would never include them in the compile phase and targets don't distinguish the case.

The CumulativeScopeFilter already implements that. The current change would still allow that.

laeubi:
> You seem to imply that I stated that only 1 scope should be allowed. That's not at all what I meant. What I state can be rephrased as "scope hierarchy and inheritance should be honored".

The simple thing is, that scopes are excluding each others, so one can't compute a "valid" definition that is consistent. 

Just take the `runtime` scope, what should happen if only that is selected? Per definition it excludes `compile` but includes `test` dependencies, but the test `scope` includes `compile`, so given there is only one state this means this can't be resolved without bring much confusion, while saying the scope you select is included (and nothing else) is much more easy and maps 1:1 what the user see in the resulting target content.

laeubi:
> I think what matches is that if I create a location of scope "test", I expect, just like everywhere in Maven that "compile" deps are included as well.

What if you want only include test-scoped dependencies but not their compile dependencies? `test` actually includes almost everything and there is no way in PDE to further limit that in a useful way.

Next assume I like only `compile`, but also want a `runtime` dependency (because I want to run my application of course), but as runtime implies `test` I again pull in everything....

Also please not that in maven there are `compile` / `test` / `runtime` **phases** and to that phases a given scope applies, PDE simply do not has such a concept.

mickaelistria:
> What if you want only include test-scoped dependencies but not their compile dependencies?

I don't think it's a realistic use case since Maven has not enabled it for many years.

> so given there is only one state this means this can't be resolved without bring much confusion, while saying the scope you select is included (and nothing else) is much more easy and maps 1:1 what the user see in the resulting target content.

Tycho is actually introducing new confusion by behaving differently as upstream Maven has been behaving for 15+ years.

> Just take the runtime scope, what should happen if only that is selected?

Maven uses AbstractScopeArtifactFilter to decide of what to do.

> Next assume I like only compile, but also want a runtime dependency (because I want to run my application of course), but as runtime implies test

According to AbstractScopeArtifactFilter, `runtime` does not imply `test`. It's the other way round.

> Also please not that in maven there are compile / test / runtime phases and to that phases a given scope applies, PDE simply do not has such a concept.

Scopes are filters, and how they inherit each other is defined by Maven. The current implementation of how they are used in .target does not honor Maven.

laeubi:
> I don't think it's a realistic use case since Maven has not enabled it for many years.

Again, this is not comparable to maven, and this usecase was brought by platform to m2e/tycho. And Maven **does** enable this please read the link: 

> Maven includes a dependency with this scope in the runtime and test classpaths, but not the compile classpath

So it is available for running your app, but not for compilation and thus it is (depending on the phase) included or not.

> Tycho is actually introducing new confusion by behaving differently as upstream Maven has been behaving for 15+ years.

Target definition is nothing "upstream maven", funny enough always when I state that Tycho behaves differently as "upstream Maven" it is claimed that "it has to work this way because Tycho is special", also Tycho is not in the lead here, this is a m2e feature that Tycho supports. The only _confusion_ is, that Platform-SDK does not include m2e and people starting to edit the file by hand and assuming things that where never promised, if I just use m2e and add your example dependency here, I can choose compile scope:
grafik

and I get exactly one dependency as a result, as expsected:
grafik

an there is no confusion at all. Simple as ABC, you can choose any combination you desire and you don't need to know what includes the other. If it do not work for Tycho this way it has to be fixed, as m2e is what set the standard here.

> Scopes are filters, and how they inherit each other is defined by Maven. The current implementation of how they are used in .target does not honor Maven.

If you want it "like Maven" add it as a dependency to the pom and use `pomDependecies=consider` and as explained, the "inheritance" is defined in the context of execution phases and targets do not have phases so this can never match.

mickaelistria:
I've updated the change so it doesn't use the ScopeFilter.

mickaelistria:
You're mixing things: phase do not have much relation to scopes. Phase are just an order of execution and group some mojos. The mojos may ask for dependencies of a given scope. A phase doesn't mandate a scope nor the other way round; it all depends on the mojos. I think you're confused because compile and test are also the name of some Maven phases, but it's something different than the compile and test scopes.
So most of your arguments involving phases are not really valid technically speaking.

To be honest I believe that no-one else than you would expect the Maven locations in .target to behave in a way that differs so much from the actual Maven dependency resolver and how scopes are implemented in Maven.
Maven resolver and maven-dependency-plugin which mostly is a facade for it are the reference. Diverging from a reference is adding confusion or complexity. If things are confusing unless one uses a particular tool, then they are still confusing by nature, the tool becomes lipstick on a pig. Usually, "you need a particular tool to make it work well and as expected" is a symptom of suboptimal choices and failure to grok actual users expectations.

laeubi:
> You're mixing things: phase do not have much relation to scopes. Phase are just an order of execution and group some mojos.

See https://maven.apache.org/developers/mojo-api-specification.html and search for `requiresDependencyResolution`, each mojo can declare the required scopes

> Flags this Mojo as requiring the dependencies in the specified class path to be resolved before it can execute. 

So there are different `classpathes` (namely `compile`, `runtime`, `compile+runtime` and `test`) and to confuse the usual developer these are named very similar to the names of the scopes, and map to a certain set of scopes included (see the mapping table there). So on each phase a Mojo is mapped to a certain set of scopes will be resolved and made available to the mojo. I first tried to use these, but as you can see some overlap, others are not select able (e.g. say you only want provided ones) and you cannot "exclude" thing, lets say you only want runtime.

This page summarize it in a good way already https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope 

> Dependency scope is used **to limit the transitivity** of a dependency and to determine when a **dependency is included in a classpath**.

Both concepts are hard to map, for **transitivity** we already have the depth attribute (and its was a bit hard to understand and cause confusion), for **classpath** we simply have no way to express it, as we do not wan't to make a classpath, we only want to include **artifacts** and the devloper can the choose if it should be used for compilation, only included in a feature/updatesite or used to run a test or product. *if* PDE/JDT would support more of the maven concept (currently its even not possible to mark a bundle as a test-requirement!) then we might have chosen otherwise, but as of today these maven-concept simply do not match to the target concept.

> To be honest I believe that no-one else than you would expect the Maven locations in .target to behave in a way that differs so much from the actual Maven dependency resolver and how scopes are implemented in Maven.

It uses the Maven resolver, the only difference is that we use a different filtering, that is both _flexible_ and easy to _explain_ but still _powerful_ (you can control even more than maven allows, e.g `import` scope) and is _not bound_ to maven definition (I assume most maven users do not know all these mapping tables)

> is a symptom of suboptimal choices and failure to grok actual users expectations.

This has been changed a while back (> 1 year) and until now no one has complained so I assume it is at least not that a bad choice and users are happy with that.

Just for reference I attaching a target used for testing the different combinations where you can see that selectively selecting combinations could make a difference if the user do not want to blow up its target content:
grafik

[maven_locations_test.target.zip](https://github.com/eclipse/tycho/files/9195342/maven_locations_test.target.zip)

laeubi (approved):
Looks good and having a way to test this in a unit test as well is great!

PR #1177, opened 2022-07-21
Issue 1159

laeubi:


laeubi:
@cdietrich thanks for the test-case!

PR #1178, opened 2022-07-21
Try using p2 types in both Maven and OSGi world

mickaelistria:


laeubi:
@mickaelistria this will take some time, I'm currently preparing the next changes for this ...

laeubi:
@mickaelistria @HannesWell @akurtakov such changes are now possible, just add the required maven-dependecy to the project to use e.g. `IInstallableUnit` in an interface.

HannesWell:
> @mickaelistria @HannesWell @akurtakov such changes are now possible, just add the required maven-dependecy to the project to use e.g. `IInstallableUnit` in an interface.

Sounds great. 👍🏽 
So https://github.com/eclipse/tycho/pull/911 should be feasible now? But this PR looks like doing the same like that. So I'm not sure If anything is left.

mickaelistria:
The current PR is a 1st step attempting to leverage class sharing, by focusing on having OSGi and Maven use the same IInstallableUnit, and removing some wildcard/Object in signatures. It's one step of many.
Further steps can include: removal of "wrapper" types, removal of some resolution methods, reusing some other p2 types (eg IMetadataRepository/IArtifactRepository...).

laeubi:
I think all of you already tried this in the past ;-)
As @mickaelistria  noted, there is a lot to do and I'm already struggle with #1215 because eclipse/equinox uses so many quirks its really a mess ...

mickaelistria:
> I think all of you already tried this in the past ;-)

Ooh yes, and the ones before us, and the ones before them, since Tycho was incepted ;) It's a huge achievement!

mickaelistria:
Holy sh**! This does actually work. It's (yet-another) Tycho dream becoming true! Thanks a lot @laeubi !

PR #1185, opened 2022-07-24
Execute tycho-ds-plugin for eclipse-test-plugins

HannesWell:
Projects with `Eclipse-test-plugin` packaging-type can have DS-components as well (e.g. for testing purposes).

@laeubi what do you think about enabling the `org.eclipse.tycho:tycho-ds-plugin` by default for them as well?
I have not tested this change yet, but I noticed it is missing in the build of https://github.com/eclipse-equinox/equinox/pull/68.

### DIFF
--- a/tycho-maven-plugin/src/main/resources/META-INF/plexus/components.xml
+++ b/tycho-maven-plugin/src/main/resources/META-INF/plexus/components.xml
@@ -106,6 +106,9 @@
               <compile>
                 org.eclipse.tycho:tycho-compiler-plugin:${project.version}:compile
               </compile>
+              <process-classes>
+              	org.eclipse.tycho:tycho-ds-plugin:${project.version}:declarative-services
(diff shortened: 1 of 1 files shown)


laeubi:
> @laeubi what do you think about enabling the `org.eclipse.tycho:tycho-ds-plugin` by default for them as well?

I actually don't care about the `test-plugin` packaging type, so if you think it is useful and all tests succeed why not ...

HannesWell:
> > @laeubi what do you think about enabling the `org.eclipse.tycho:tycho-ds-plugin` by default for them as well?
> 
> I actually don't care about the `test-plugin` packaging type, so if you think it is useful and all tests succeed why not ...

At least in the mentioned case I used it to build Attest case. The uses are probaly less often, but ist would simplify the registration of test components. And easily providing mocks for testing is as far as I know one purpose of DI/IOC.

PR #1196, opened 2022-07-26
Reproducer for issue 1093: misleading error message

howlger:
Plugin without Java code with a `build.properties` file containing:
```
bin.includes = META-INF/,\
               non-existing-file.txt
```
**Actual**: Build fails with the error message ...`/target/classes does not exist`

**Expected**: Build fails with an error message like ...`non-existing-file.txt does not exist`

### DIFF
--- a/tycho-its/projects/issue1093/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/issue1093/META-INF/MANIFEST.MF
@@ -1,0 +1,4 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-SymbolicName: issue1093
+Bundle-Version: 1.0.0.qualifier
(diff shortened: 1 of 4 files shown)


tycho-its/src/test/java/org/eclipse/tycho/test/issue1093/Issue1093.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.List;
+
+import static org.junit.Assert.fail;
+
+import org.apache.maven.it.VerificationException;
+import org.apache.maven.it.Verifier;
+import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
+import org.junit.Test;
+
+public class Issue1093 extends AbstractTychoIntegrationTest {
```
laeubi, line 22:
```suggestion
public class Issue1093Test extends AbstractTychoIntegrationTest {
```
The class should end with Test to be picked up by the surefire plugin.

laeubi:
I see the test now fails (as expsected) thanks @howlger I'll take a look whats wrong here now.

laeubi:
This is now included in https://github.com/eclipse/tycho/pull/1199

PR #1201, opened 2022-07-26
Fallback resolution to force keep uses

mickaelistria:
Uses is sometimes necessary for resolution to perform correctly. When
resolution ignoring uses fails, then fallback to full (longer)
resolution honoring uses which is more likely to succeed.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
@@ -26,7 +26,6 @@ import java.util.Set;
 import java.util.StringJoiner;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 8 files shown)


tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
```diff
@@ -81,31 +81,47 @@
...
-     */
-    private static final int THREAD_COUNT = Integer.getInteger("tycho.equinox.resolver.executor.threads", 1);
+        public ResolverConfiguration() {
+            // keep default values
+        }
+
+        public void forceKeepUses() {
+            if (!keepUses) {
+                keepUses = true;
+                batchSize = System.getProperty("tycho.equinox.resolver.batch.size", keepUses ? null : "1");
```
laeubi, line 96: I don't think we save anything when conditionally lookup the system property here, whats wrong with the static final ones?

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
```diff
@@ -194,15 +215,25 @@ public ModuleContainer newResolvedState(ReactorProject project, MavenSession mav
             return container;
         } finally {
             executorService.shutdownNow();
+            if (!config.keepUses) {
+                logger.info(FORCE_KEEP_USES);
+                config.forceKeepUses();
```
laeubi, line 220: If we really want to go with a config object, it should be final and unmodified and we create a new object with adjusted settings here...

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
```diff
@@ -81,31 +81,47 @@
...
-    private static final String BATCH_TIMEOUT = System.getProperty("tycho.equinox.resolver.batch.timeout",
-            String.valueOf(TimeUnit.SECONDS.toMillis(30)));
+    protected static final class ResolverConfiguration {
+        //The following properties are not supported for general usage and intended to experimental testing when developing Tycho, so use with care
 
-    /**
-     * Allow to adjust the default thread count used in resolver operations
-     */
-    private static final int THREAD_COUNT = Integer.getInteger("tycho.equinox.resolver.executor.threads", 1);
+        public ResolverConfiguration() {
```
laeubi, line 89: I think it would be better to keep the static defaults and pass them here at construction time into private final fields. Beside this the inner class should be extracted, we can the move all the static final fields there and have a default constructor wich uses these defaults then.

laeubi (changes requested):
Some minor adjustments would be good.

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
mickaelistria: you mean a record? Are we fine moving to newer Java allowing records? I don't remember if we agreed on something for Tycho?

laeubi: Just use a traditional final class with package protected final fields, getters or whats you preferred pattern, we can convert ti to records later on :-)

laeubi (approved):
Looks good

laeubi:
@mickaelistria I noticed that now the message is ALWAYS printed for platform projects, as we had no issues previously I suspect something is wrong here now can you check this?

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
```diff
@@ -194,15 +174,24 @@ public ModuleContainer newResolvedState(ReactorProject project, MavenSession mav
             return container;
         } finally {
             executorService.shutdownNow();
+            if (!config.keepUses) {
```
laeubi, line 177: I think this should not be in a finally but in a catch block?!

PR #1206, opened 2022-07-27
Exclude org.osgi.service.log / sanitize service return value

laeubi:
It seems bnd pulls in the problematic dependency that recently caused issues see:

- https://github.com/bndtools/bnd/issues/5335

because of that, we should exclude it in the pom


laeubi:
GH action works again, hurray!

PR #1220, opened 2022-07-30
More i installable unit

mickaelistria:


laeubi:
Should we merge this or is there more to do?

akurtakov:
If we see good Jenkins build let's do it.

PR #1228, opened 2022-08-01
Convert artifact handlers from xml to plexus-component

laeubi:
Currently we declare artifact handlers in the tycho-maven-plugin xml by
hand written code.

This has the drawback, that one always needs to include the
tycho-maven-plugin to use the artifact types, if one uses a tycho-mojo
standalone (e.g. eclipse-run-mojo) they are missing.

With declaring them in the code, next to the p2 plugin, they are
available for all mojos.


mickaelistria (approved):
Java is easier to maintain than XML and this reduces duplication of symbols.

PR #1229, opened 2022-08-01
Move the transport into the P2 plugin

laeubi:


laeubi:
Requires more smaller steps

PR #1231, opened 2022-08-01
Test the windows build

laeubi:


laeubi:
@pzi-dsa Any idea why this maven call on windows seem to fail on windows:

https://github.com/eclipse/tycho/runs/7625048441?check_suite_focus=true

It seems some who windows don't like the `-Dmaven.test.failure.ignore=true` but as we use this on other GH builds as well I'm a bit lost here...

pzi-dsa:
> It seems some who windows don't like the -Dmaven.test.failure.ignore=true but as we use this on other GH builds as well I'm a bit lost here...

I think this is a PowerShell issue rather than a Windows issue. At least that is my experience with Maven so far.
I'd suggest adding a space so that it's `-D maven.test.failure.ignore=true`

laeubi:
> I'd suggest adding a space so that it's `-D maven.test.failure.ignore=true`

Thanks I'll try that never know that it also accepts spaces here ...

cdietrich:
an alternative would be to quote the arg `'-Dmaven.test.failure.ignore=true' `

PR #1234, opened 2022-08-02
Create a preliminary apitools plugin

laeubi:
FYI @akurtakov @HannesWell 

I think this could be merged as a first step, I don't know much about what one would expects that such an `.api_description` file should contain, so if anyone later on likes to enhance this it would be good.


akurtakov:
Now that I thought a bit more about it it might be good to switch https://github.com/eclipse-pde/eclipse.pde/tree/master/apitools/org.eclipse.pde.api.tools.annotations/src/org/eclipse/pde/api/tools/annotations to switch to RetentionPolicy CLASS and skip the non-standart .api_description file. At least in theory that should allow us to make things work.

laeubi:
> to switch to RetentionPolicy CLASS and skip the non-standart .api_description file. At least in theory that should allow us to make things work.

Generating this file is not that much work so we still can keep this for now and we can later skip this if no longer required ..

laeubi:
At least this should replace the current "generate-api" part of the aggregator-pom.

HannesWell:
Look's good one the phone. I'm supprised this is so simple.

We could consider to skip the execution by default for eclipse-test-plugin type Projects?

HannesWell:
> > Look's good one the phone. I'm supprised this is so simple.
> 
> Because this is the simple part of the problem, but still will simplify setup

Yes definitivly.

If we use it in the Platform this makes 
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/419 more or less obsolet.

> 
> > We could consider to skip the execution by default for eclipse-test-plugin type Projects?
> 
> At least I do not plan any special "test plugin support" because there is no such thing neither in OSGi nor PDE .... The generator code even though checks if the api nature is enabled and if not simply do nothing...

Hmm yes, that's right.

As suggested in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/352#issuecomment-1159488042 respectivly the PR mentioned above. It should even be considered to generate the API description for Plugins without api-nature. But that is something to address in PDE.

mickaelistria:
Has anyone tried it on Platform?

laeubi:
> Has anyone tried it on Platform?

I'm not aware of that, so if you like to try it out ...

mickaelistria:
I created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/499 to see what happens then.

PR #1246, opened 2022-08-05
Fix tycho-dependency-tools-plugin is not listing all dependencies

laeubi:
The tycho-dependency-tools-plugin do not list items that are added by
the extra-classpath currently, this could lead to failing runs of the
API tools.


tycho-its/src/test/java/org/eclipse/tycho/test/compiler/CompilerClasspathEntryTest.java
```diff
@@ -16,12 +16,15 @@
...
+import java.util.List;
 
 import org.apache.maven.it.Verifier;
-import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
-import org.eclipse.tycho.test.util.EnvironmentUtil;
 import org.junit.Test;
 
+import src.main.java.org.eclipse.tycho.test.AbstractTychoIntegrationTest;
+import src.main.java.org.eclipse.tycho.test.util.EnvironmentUtil;
+
```
mickaelistria, line 27: Seems like adding more .classpath here and there would help ;)

PR #1264, opened 2022-08-11
Support arbitrary aggregator directories without extra setup

HannesWell:
Removes the need to specify a Set of allowed aggregator-directory names and simply scans all modules listed (transitively) in the root pom.xml for directly nested modules, if the directory is not any other kind of project (and has not pom.xml).

Update to the latest version of `io.takari.polyglot:polyglot-common` which reduces the calls to `Mapping.locatePom()` to the minimum. Polyglot-maven already only checks those directories listed as modules. But before it located a pom for all available mappings. Now it starts with the highest priority and tires those with lower once until a `Mapping.locatePom()` returns a result.
I proposed to change that: https://github.com/takari/polyglot-maven/pull/242
This PR requires that change (and therefore already has the not yet functional version update in the pom).

To minimize the number of file-system interactions the following distinct order of Tycho-Mappings is introduced:
- Bundle
- Feature
- pom.xml (provided by polyglot-maven)
- Repository (means Product or p2-repository)
- Target-Platform
- Aggregator

It is important that the Tycho-Feature-Mapping has a higher priority than the Pom-Mapping. Otherwise it can happen that the Pom-Reader is used to load the model from the `feature.xml` file and to populate the polyglot model, which would then be flawed. Until now we have been lucky that the `PolyglotModelManager` has sorted the default-pom mapping behind the feature mapping despite their same priority. Or is there another mechanism that reliable ensures that?

The remaining relations imposed by this order reflect the likelihood of a match (usually more occurring project types get a higher priority) and the cost of a match attempt (more expensive mappings get a lower priority), which results in the aggregator-mapping having the lowest priority.

With those changes AFAICT a directory should never be scanned unnecessarily. 

Additionally I update the description of the tycho-pomless pom a bit.

### DIFF
--- a/tycho-extras/tycho-pomless/pom.xml
+++ b/tycho-extras/tycho-pomless/pom.xml
@@ -23,8 +23,8 @@
   <artifactId>tycho-pomless</artifactId>
 
   <name>Tycho POM-less build extension</name>
-  <description>A build extension which allows to omit pom.xml files for bundles and features. 
-  Maven groupId is inherited from the mandatory parent pom in the parent directory, artifactId and version are derived from MANIFEST.MF and feature.xml.
+  <description>A build extension which allows to omit pom.xml files for all Eclipse PDE projects, Plug-ins(Bundles) and Features 
+  as well as Eclipse Products, Target-Platforms and p2-repository definitions (category.xml).
   </description>
 
   <build>
@@ -47,7 +47,7 @@
     <dependency>
       <groupId>io.takari.polyglot</groupId>
       <artifactId>polyglot-common</artifactId>
-      <version>0.4.8</version>
+      <version>0.4.9-SNAPSHOT</version>
     </dependency>
     <dependency>
       <groupId>org.eclipse.sisu</groupId>
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
@@ -23,8 +23,7 @@ import java.io.FileFilter;
 import java.io.IOException;
 import java.io.Reader;
 import java.nio.file.Files;
-import java.util.Arrays;
-import java.util.HashSet;
+import java.nio.file.Path;
 import java.util.Iterator;
 import java.util.Set;
 import java.util.TreeSet;
@@ -45,14 +44,22 @@ public class TychoAggregatorMapping extends AbstractTychoMapping {
 
     private static final String TYCHO_POM = "pom.tycho";
 
-    private static final Set<String> COMMON_NAMES = new HashSet<>(
-            Arrays.asList(System.getProperty(TYCHO_POMLESS_AGGREGATOR_NAMES_PROPERTY,
-                    "bundles,plugins,tests,features,sites,products,releng").split(",")));
+    @Override
+    protected String getPackaging() {
+        return "pom";
+    }
+
+    @Override
+    public float getPriority() {
+        return -10f;
+    }
 
     @Override
     protected boolean isValidLocation(String location) {
-        return location.endsWith(TYCHO_POM);
+        return Path.of(location).endsWith(TYCHO_POM);
     }
+
+    private static final ThreadLocal<Boolean> IS_FIRST = ThreadLocal.withInitial(() -> true);
 
     @Override
     protected File getPrimaryArtifact(File dir) {
@@ -60,7 +67,8 @@ public class TychoAggregatorMapping extends AbstractTychoMapping {
         if (file.exists() && isCurrent(file)) {
             return file;
         }
-        if (COMMON_NAMES.contains(dir.getName().toLowerCase())) {
+        if (IS_FIRST.get().booleanValue()) {
+            IS_FIRST.set(false);
             logger.debug("Scanning folder " + dir + " for modules");
             File[] subFolders = dir.listFiles((FileFilter) File::isDirectory);
             if (subFolders != null) {
@@ -73,6 +81,7 @@ public class TychoAggregatorMapping extends AbstractTychoMapping {
                         logger.debug("Found pom " + reference.getPomFile().getName() + " in subfolder " + name);
                     }
                 }
+                IS_FIRST.remove();
                 if (!modules.isEmpty()) {
                     file.deleteOnExit();
                     Stream<CharSequence> lines = concat(of(TYCHO_AUTOMATIC_GENERATED_FILE_HEADER), modules.stream());
@@ -85,14 +94,11 @@ public class TychoAggregatorMapping extends AbstractTychoMapping {
                 }
             }
         } else {
-            logger.debug("Skip folder " + dir + " because it does not match any common name " + COMMON_NAMES);
+            // For performance reasons only scan for directly nested (non-aggregator) modules.
+            // Otherwise folders without projects in an aggr-folder are fully traversed (e.g. a 'target'-folder within the aggregator).
+            logger.debug("Skip nested folder while scanning for modules " + dir);
         }
         return null;
-    }
-
-    @Override
-    protected String getPackaging() {
-        return "pom";
     }
 
     @Override
@@ -108,11 +114,6 @@ public class TychoAggregatorMapping extends AbstractTychoMapping {
             model.setArtifactId(artifactFile.getParentFile().getName());
             model.setName("[aggregator] " + model.getArtifactId());
         }
-    }
-
-    @Override
-    public float getPriority() {
-        return -10f;
     }
 
     private boolean isCurrent(File file) {
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoFeatureMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoFeatureMapping.java
@@ -17,6 +17,7 @@ package org.eclipse.tycho.pomless;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Path;
 import java.util.Properties;
 import java.util.function.Supplier;
 
@@ -32,6 +33,18 @@ public class TychoFeatureMapping extends AbstractXMLTychoMapping {
     private static final String NAME_PREFIX = "[feature] ";
     private static final String FEATURE_XML = "feature.xml";
     public static final String PACKAGING = "eclipse-feature";
+
+    @Override
+    protected String getPackaging() {
+        return PACKAGING;
+    }
+
+    @Override
+    public float getPriority() {
+        // It is important that this mapping has a higher priority than the 'XmlMapping' provided by polyglot. 
+        // Otherwise it can happen that the the pom.xml reader is used to read the feature.xml which will break the build!
+        return +1;
+    }
 
     @Override
     protected void initModelFromXML(Model model, Element xml, File artifactFile) throws IOException {
@@ -54,7 +67,7 @@ public class TychoFeatureMapping extends AbstractXMLTychoMapping {
 
     @Override
     protected boolean isValidLocation(String location) {
-        return location.endsWith(FEATURE_XML);
+        return Path.of(location).endsWith(FEATURE_XML);
     }
 
     @Override
@@ -64,11 +77,6 @@ public class TychoFeatureMapping extends AbstractXMLTychoMapping {
             return featureXml;
         }
         return null;
-    }
-
-    @Override
-    protected String getPackaging() {
-        return PACKAGING;
     }
 
 }
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/buildextension/PomlessTest.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/buildextension/PomlessTest.java
@@ -14,6 +14,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.apache.maven.it.Verifier;
 import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
@@ -36,9 +38,8 @@ public class PomlessTest extends AbstractTychoIntegrationTest {
 	public void testPomlessTestPluginDetection() throws Exception {
 
 		Verifier verifier = getVerifier("pomless-tests", false, true);
-		verifier.executeGoals(List.of("clean", "test"));
+		verifier.executeGoals(List.of("clean", "compile"));
 		verifier.verifyErrorFreeLog();
-		verifier.verifyTextInLog("");
 
 		Map<Path, ModelData> projectData = extractPomModelProperties(verifier);
 
@@ -76,6 +77,7 @@ public class PomlessTest extends AbstractTychoIntegrationTest {
 		// -> Plug-ins, Features, Products(Repos), Targets, Aggregators
 		// - tycho.pomless.parent is considered
 		// - explicit pom.xml is always preferred
+		// - arbitrary aggregator directories are supported without special setup
 
 		Verifier verifier = getVerifier("pomless-model", false, true);
 		verifier.executeGoals(List.of("clean", "package"));
@@ -130,6 +132,21 @@ public class PomlessTest extends AbstractTychoIntegrationTest {
 				"aggregator3-pomless");
 
 		assertThat(projectData, is(aMapWithSize(0))); // Ensure no more projects are found
+
+		// Ensure that only those aggregator-directories that are listed as modules in
+		// the parent (and don't have a pom) are scanned for nested modules.
+		List<String> logLines = verifier.loadFile(verifier.getBasedir(), verifier.getLogFileName(), false);
+		String logPrefix = "\\[\\w*] *";
+		Pattern scaned = Pattern.compile(logPrefix + "Scanning folder (.*) for modules");
+		Pattern skipped = Pattern.compile(logPrefix + "Skip nested folder while scanning for modules (.*)");
+		assertMentionedDirectories(List.of("bundles", "bundles-with-enhanced-parents"), scaned, logLines);
+		assertMentionedDirectories(List.of("target", "otherFolder", "target"), skipped, logLines);
+	}
+
+	private void assertMentionedDirectories(List<String> expected, Pattern filePattern, List<String> lines) {
+		List<String> scanned = lines.stream().map(filePattern::matcher).filter(Matcher::matches).map(m -> m.group(1))
+				.map(f -> new File(f).getName()).toList();
+		assertEquals(expected, scanned);
 	}
 
 	private static void assertProjectData(String path, Map<Path, ModelData> projectData, String expectedGAV,
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
@@ -54,7 +54,6 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
 
     // All build.properties entries specifically considered by Tycho. Extends the list in Mapping interface
     protected static final String TYCHO_POMLESS_PARENT_PROPERTY = "tycho.pomless.parent";
-    protected static final String TYCHO_POMLESS_AGGREGATOR_NAMES_PROPERTY = "tycho.pomless.aggregator.names";
 
     private static final String PARENT_POM_DEFAULT_VALUE = System.getProperty(TYCHO_POMLESS_PARENT_PROPERTY, "..");
     private static final String QUALIFIER_SUFFIX = ".qualifier";
@@ -231,11 +230,6 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
         } catch (ComponentLookupException e) {
         }
         return reference;
-    }
-
-    @Override
-    public float getPriority() {
-        return 0;
     }
 
     @Override
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
@@ -47,6 +47,16 @@ public class TychoBundleMapping extends AbstractTychoMapping {
     private static final String PACKAGING_TEST = "eclipse-test-plugin";
 
     @Override
+    protected String getPackaging() {
+        return PACKAGING;
+    }
+
+    @Override
+    public float getPriority() {
+        return +2;
+    }
+
+    @Override
     protected boolean isValidLocation(String location) {
         File polyglotFile = new File(location);
         return polyglotFile.getName().equals(META_INF_DIRECTORY) && new File(polyglotFile, MANIFEST_MF).isFile();
@@ -59,11 +69,6 @@ public class TychoBundleMapping extends AbstractTychoMapping {
             return metaInfDirectory;
         }
         return null;
-    }
-
-    @Override
-    protected String getPackaging() {
-        return PACKAGING;
     }
 
     @Override
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoRepositoryMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoRepositoryMapping.java
@@ -61,6 +61,11 @@ public class TychoRepositoryMapping extends AbstractXMLTychoMapping {
     }
 
     @Override
+    public float getPriority() {
+        return -1;
+    }
+
+    @Override
     protected void initModel(Model model, Reader artifactReader, File artifactFile) throws IOException {
         if (artifactFile.getName().endsWith(PRODUCT_EXTENSION)) {
             File projectRoot = artifactFile.getParentFile();
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoTargetMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoTargetMapping.java
@@ -38,6 +38,11 @@ public class TychoTargetMapping extends AbstractXMLTychoMapping {
     }
 
     @Override
+    public float getPriority() {
+        return -2;
+    }
+
+    @Override
     protected boolean isValidLocation(String location) {
         return location.endsWith(TARGET_EXTENSION);
     }
--- a/tycho-its/projects/pomless-model/.mvn/maven.config
+++ b/tycho-its/projects/pomless-model/.mvn/maven.config
@@ -1,2 +1,1 @@
 -Dtycho-version=3.0.0-SNAPSHOT
--Dtycho.pomless.aggregator.names=bundles,bundles-with-enhanced-parents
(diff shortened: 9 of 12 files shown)


tycho-extras/tycho-pomless/pom.xml
```diff
@@ -47,7 +47,7 @@
     <dependency>
       <groupId>io.takari.polyglot</groupId>
       <artifactId>polyglot-common</artifactId>
-      <version>0.4.8</version>
+      <version>0.4.9-SNAPSHOT</version>
```
laeubi, line 50: I don'T think we can use snapshot here, please ask for a release at polyglot

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoFeatureMapping.java
```diff
@@ -33,6 +34,18 @@ public class TychoFeatureMapping extends AbstractXMLTychoMapping {
...
+    @Override
+    protected String getPackaging() {
+        return PACKAGING;
+    }
+
+    @Override
+    public float getPriority() {
+        // It is important that this mapping has a higher priority than the 'XmlMapping' provided by polyglot. 
+        // Otherwise it can happen that the the pom.xml reader is used to read the feature.xml which will break the build!
+        return +1;
```
laeubi, line 46: Why should this ever happen?

tycho-its/src/test/java/org/eclipse/tycho/test/buildextension/PomlessTest.java
```diff
@@ -130,6 +132,21 @@ public void testPomlessModel() throws Exception {
 				"aggregator3-pomless");
 
 		assertThat(projectData, is(aMapWithSize(0))); // Ensure no more projects are found
+
+		// Ensure that only those aggregator-directories that are listed as modules in
+		// the parent (and don't have a pom) are scanned for nested modules.
+		List<String> logLines = verifier.loadFile(verifier.getBasedir(), verifier.getLogFileName(), false);
```
laeubi, line 138: I don't think this will work, it is not required that the module is defined in the parent, it could be any pom in any directory (just take https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/.github/aggregator.pom as an example)

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
```diff
@@ -45,22 +44,31 @@ public class TychoAggregatorMapping extends AbstractTychoMapping {
...
+    private static final ThreadLocal<Boolean> IS_FIRST = ThreadLocal.withInitial(() -> true);
+
     @Override
     protected File getPrimaryArtifact(File dir) {
         File file = new File(dir, TYCHO_POM);
         if (file.exists() && isCurrent(file)) {
             return file;
         }
-        if (COMMON_NAMES.contains(dir.getName().toLowerCase())) {
+        if (IS_FIRST.get().booleanValue()) {
```
laeubi, line 70: The mapping methods should **never** be statefull, we can't control, when and how often they are called!

laeubi (changes requested):
I think this has several problems and I don't think we should actually support this anyways.

1. `pom.xml` **must** have the highest priority, if a `pom.xml` is found Tycho should do nothing as it is otherwise not possible to override with a pom and the polyglot plugin heavily relies on the fact to in certain places.
2. Recursing into arbitrary (sub) directories is actually not desired (unless they contain a pom.xml or are explicitly mentioned) as it is common to move folders "out of the way" so they are not considered by Tycho pomless
3. last but not least the idea is to encourage people using a structured layout instead of throwing some random folders together

laeubi:
> The remaining relations imposed by this order reflect the likelihood of a match (usually more occurring project types get a higher priority) and the cost of a match attempt (more expensive mappings get a lower priority), which results in the aggregator-mapping having the lowest priority.

By the way I don't think this is a valid measure of ordering. The ordering should purely driven from the users POV and not any technical details especially when there are *multiple* matches, so lets say we have a bundle that contains a feature.xml, the order should reflect the precedence in this case we give 'bundle' a higher precedence than 'feature' and ignore the possible feature-nature of the project!

If one is concerned about the extra miliseconds one could simply use `pom.xml` files what is always faster than determine all that stuff dynamic on each build.

HannesWell:
> I think this has several problems and I don't think we should actually support this anyways.
> 
>     1. `pom.xml` **must** have the highest priority, if a `pom.xml` is found Tycho should do nothing as it is otherwise not possible to override with a pom and the polyglot plugin heavily relies on the fact to in certain places.

The new priorities just specify _reliably in which order the mappings are processed_. There is already an ordering but it is unspecified since all but the aggr-mapping have a priority of zero and therefore the ordering of set by Plexus is used.
Nevertheless `pom.xml` files still have the highest priority because all Tycho-mappings give way when there is a `pom.xml` (see the `AbstractTychoMapping.locatePom()`), even if they are processed first.

> 
>     2. Recursing into arbitrary (sub) directories is actually not desired (unless they contain a pom.xml or are explicitly mentioned) as it is common to move folders "out of the way" so they are not considered by Tycho pomless

Takari-polyglot only scans those directories that are (transitively) listed as modules. So nothing is scanned arbitrarily. One basically commands to scan a directory by listing it as module.

I have extended the `pomless-model` test to check that no directory is scanned unnecessarily. It is done by scanning the debug-printout and as you can see only the directories listed as modules and don't contain a other project type directly are scanned by the Aggr-Mapping. It also verifies that sub-modules are skipped.

Besides that I used a locally build version of this change (and the takari-polyglot change) and started the eclipse-platform-aggregator build and my log-line timestamp based measured showed a almost identical scan-for-projects runtime. Maybe even marginally faster.

> 
>     3. last but not least the idea is to encourage people using a structured layout instead of throwing some random folders together

While structured layouts are good, no question, I don't think that this is the right place to enforce it respectively it does not help much.
It is not even really enforced since the user can change it with the `tycho.pomless.aggregator.names` system-property. Just for aggregator builds that use sub-modules which are also build standalone, like Eclipse-Platform aggregator, it is more complicated to set up, because all the names have to be kept in sync.
And how does it improve the structure if for example one cannot place it's example-projects in an `example` folder. But e.g. has to place them in `bundles`/`plugins`? We could of course adapt the default, but other people have other constraints. And I suspect that if it is going gets tough one rather decides to create the aggregator manually than complying with the standard names. Or to adjust the names-property.


> > The remaining relations imposed by this order reflect the likelihood of a match (usually more occurring project types get a higher priority) and the cost of a match attempt (more expensive mappings get a lower priority), which results in the aggregator-mapping having the lowest priority.
> 
> By the way I don't think this is a valid measure of ordering. The ordering should purely driven from the users POV and not any technical details especially when there are _multiple_ matches, so lets say we have a bundle that contains a feature.xml, the order should reflect the precedence in this case we give 'bundle' a higher precedence than 'feature' and ignore the possible feature-nature of the project!

Because before all Mappings had priority zero, I assumed that such type-ordering was not done. But sure I'm fine with a type based ordering as well. That slight performance-optimization was just a side effect anyways and probably doesn't have a notable effect. :)

tycho-extras/tycho-pomless/pom.xml
HannesWell: Yes. We have to wait for the mentioned PR to be merged, then I will ask for a release.
I just wanted to indicate that eventually (when this PR is not a draft anymore) this is required.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoFeatureMapping.java
HannesWell: It happened in during my testing of this change, when I made the `org.sonatype.maven.polyglot.mapping.XmlMapping` the highest priority mapping, to reflect that pom.xml should always have precedence. I wanted to avoid the corresponding check in `AbstractTychoMapping.locatePom()`. But that attempt failed for the described reason.

Alternatively the XMLMapping could be adjusted to not accept `feature.xml` files. I.e. by replacing ".xml" by "pom.xml" in its `acceptLocationExtensions`.
You can test that already. Just assign the `TychoFeatureMapping` a priority below zero and run the pomless-model test build and you see that incomplete polyglot-poms are generated for features.

laeubi:
> Takari-polyglot only scans those directories that are (transitively) listed as modules. So nothing is scanned arbitrarily. One basically commands to scan a directory by listing it as module.

That is not true (and otherwise it won't work at all), as the aggregator mapping can define additional modules, and then these are also scanned and so on, so there is no "fixed" set of modules, you should keep in mind that we build the model here so takiri do nothing other than looking at **one** folder, if that folder defines modules, it will afterwards recurse in all these modules and so on.

> The new priorities just specify _reliably in which order the mappings are processed_. There is already an ordering but it is unspecified since all but the aggr-mapping have a priority of zero and therefore the ordering of set by Plexus is used.

Still it makes no sense to have a higher mapping order to prevent "XMLMapping" beeing choosen, that should simply not matter at all.

> While structured layouts are good, no question, I don't think that this is the right place to enforce it respectively it does not help much.

It is and changing this will break people currently rely on that features, the question is more what do we gain *not* enforcing this? The Aggregator mapping was specifically designed to support a structured layout.

> Besides that I used a locally build version of this change (and the takari-polyglot change) and started the eclipse-platform-aggregator build and my log-line timestamp based measured showed a almost identical scan-for-projects runtime.

Platform is not a valid measure for this, there are dozen of other people using Tycho and we can't optimize that for platform needs.

> It is not even really enforced since the user can change it with the `tycho.pomless.aggregator.names` system-property.

Exactly, everyone who thing knows a better structure can simply define its own default, as always, if you do not follow the conventions, it requires configuration.

> Because before all Mappings had priority zero, I assumed that such type-ordering was not done. But sure I'm fine with a type based ordering as well. That slight performance-optimization was just a side effect anyways and probably doesn't have a notable effect. :)

We should never "optimize" such things for performance and especially not if it is not measurable,


> I have extended the `pomless-model` test to check that no directory is scanned unnecessarily. It is done by scanning the debug-printout and as you can see only the directories listed as modules and don't contain a other project type directly are scanned by the Aggr-Mapping. It also verifies that sub-modules are skipped.

The test only covers a specific directory layout, as described this is not necessarily what users are using and it assumes that maven-cli calls the code in a specific way that is neither guaranteed, nor necessarily applicable for e.g. embedded use-cases.

tycho-its/src/test/java/org/eclipse/tycho/test/buildextension/PomlessTest.java
HannesWell: Actually polyglot-maven only searches in those directories listed as modules so it should work as intended.
But I can add an extra case for such non-nested aggregators.
Parent was referring to the module-tree parent not the pom-model parent. I try to make this more clear.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoFeatureMapping.java
laeubi: > Alternatively the XMLMapping could be adjusted to not accept `feature.xml` files. I.e. by replacing ".xml" by "pom.xml" in its `acceptLocationExtensions`.

This must be fixed then, we can't know what things has what priority and a mapping should only accept files that it really can handle.

tycho-its/src/test/java/org/eclipse/tycho/test/buildextension/PomlessTest.java
laeubi: No please try it out, create an empty folder and run maven with `mvn -f myEmptyFolder`, even though no one ever defines any modules, the mappings will be asked to locate a pom.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
HannesWell: OK, then there should be another mean to check that.
I could either create another marker file like 'started-scanning' and check if that is present in the parent or maintain a static ConcurrentHashSet of processed files can can be queried for the same.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoFeatureMapping.java
laeubi:
```
 public XmlMapping() {
    super(null);
    setPomNames("pom.xml");
    setAcceptLocationExtensions(".xml", ".pom");
    setAcceptOptionKeys("xml:4.0.0");
  }
```

I'm not sure if maybe something else is broken here, according to the code, it should only ever accept `pom.xml` and return for all other cases a null (meaning : there is no file I can handle)

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
laeubi: That you require any marker actually demonstrate that it *does* recurse into sub directories arbitrarily, as said, there should be simply no state in those mappings, and why we limit this to well known names, as otherwise it open the door to strange behaviors.

tycho-its/src/test/java/org/eclipse/tycho/test/buildextension/PomlessTest.java
HannesWell: > No please try it out, create an empty folder and run maven with `mvn -f myEmptyFolder`, even though no one ever defines any modules, the mappings will be asked to locate a pom.

OK, but in this case you tell Maven/Polyglot to look for a pom(-model) at that location. Then of course it looks it up there. But subsequently it looks only at the directories that are declared as modules in the just read pom-model. If a pom-type model is read from any of those location it looks at the directories listed as modules in that models as well and so on, until the module-aggreagation-tree is fully traversed.

But what it is NOT happening (at least from my knowledge of the polyglot-code and my observations while debugging for this change) is that starting at the build's root directory the entire file-system is scanned for models. That is what I meant when I said that polyglot doesn't scan directories arbitrarily and only those that are necessary. 

> 
> There is no "module-tree parent", it is just the default behavior that the path to the parent is ".." if you don't specify a parent path and tycho assumes you want to use that as a parent, but that do not mean it must define any modules at all.

Sorry that was unclear said from me. `Aggregator` is the more suitable term here. Although it is not accessible from the model, if you treat the aggregator->module relations as graph, then this is a directed ascylic graph (or tree) where one usually speaks of a parent. But yes in the Maven-world that's the aggregator.

HannesWell:
> > Takari-polyglot only scans those directories that are (transitively) listed as modules. So nothing is scanned arbitrarily. One basically commands to scan a directory by listing it as module.
> 
> That is not true (and otherwise it won't work at all), as the aggregator mapping can define additional modules, and then these are also scanned and so on, so there is no "fixed" set of modules, you should keep in mind that we build the model here so takiri do nothing other than looking at **one** folder, if that folder defines modules, it will afterwards recurse in all these modules and so on.

That is what I mean by not scanning arbitrary directories.

> 
> > While structured layouts are good, no question, I don't think that this is the right place to enforce it respectively it does not help much.
> 
> It is and changing this will break people currently rely on that features, the question is more what do we gain _not_ enforcing this? The Aggregator mapping was specifically designed to support a structured layout.

This change should not break any existing build. Why do you think it does? All it does is removing the restriction that only a limited set of names is supported by Tycho's automatic aggregator creation. The goal is, that if one has an aggregator directory that contains (non-aggregator) modules one simply can list it as module, regardless of its name. At the moment the names are limited to a certain set. If other names should be used for that, one has to adjust the system-property `tycho.pomless.aggregator.names`. The only thing that changes is that adjustment of `tycho.pomless.aggregator.names` is not necessary anymore. Existing builds, regardless of if they have adjusted `tycho.pomless.aggregator.names` or not, should continue to work, since the aggreagator-modules still have to be listed in another aggregator. Or it is the build's root directory (e.g. if one starts `mvn verify -f bundles`).

Technically I'm convinced it is possible, a few parts still have to be worked out. But before this is done we should first find consensus if this feature is wanted (assuming it works as expected). Otherwise it is just a waste of time for you and me.

tycho-its/src/test/java/org/eclipse/tycho/test/buildextension/PomlessTest.java
laeubi: > But what it is NOT happening (at least from my knowledge of the polyglot-code and my observations while debugging for this change) is that starting at the build's root directory the entire file-system is scanned for models. That is what I meant when I said that polyglot doesn't scan directories arbitrarily and only those that are necessary.

As said, polyglot has **nothing** to do with this, its is **maven-cli** that triggers this and tycho-pomless that supplies some kind of dynamic modules configuration.

If you run maven, then maven will query the current folder (or the file/folder given with `-f`) for a pom (that's where "locate pom" is called), if that is an existing file, it further reads this model (that's where again polyglot is queried) and if this is a "pom" type and contains modules these are also queried and so on, and that's why I said we can't have a state in this, it must be fully defined by a single file/folder and nothing else, it can be called once or several times and in any order, e.g. m2e will call this too to find out if a pomless project is still a maven project, also note, that one can also have json poms and so on and even combine this in any way

What Tycho now does is that it extends the idea of polyglot by dynamically creating a so called "connector-pom" by again query subfolder of folders that have a special name (e.g. plugins) for a pom, again here, any order, any time) and if these have a pom add them to this folders module list (as an alternative you can create a tycho.pom file and maintain the list manually).

But that also means, if you allow *any* name, that *any* folder would be assumed to be a "connector pom folder" and thus you will add *any* folder as long as it somewhere contains something that could be interpreted as a maven project and that's why your code contains special handling for if it is "the first" I guess.

You can try this out with current Tycho by creating a structure like:

`plugins/plugins/plugin/plugins/<a bundle or traditional pom.xml project>`

then create a pom in the root with just `<module>plugins</plugins>` and it will/should traverse the whole sub-directory structure, creating for each one a connector pom.

laeubi:
> This change should not break any existing build. Why do you think it does? All it does is removing the restriction that only a limited set of names is supported by Tycho's automatic aggregator creation.

Because currently it ignores certain folders while with this change it now will accept folders that are previously ignored. Just assume the following structure:

- bundles/bundle1
- bundles/bundle2
- bundles/experimental/bundle3

now I have the `<module>bundles</module>` in my `pom.xml`, it will build `bundle1` + `bundle2`, because `experimental` is not a valid name nothing will happen here. Now we allow *any* name, suddenly the build will execute `bundle1` + `bundle2` + `bundle3` because experimental will be assumed a valid choice and gets a connect pom generated that includes `bundle3` in its module list.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoFeatureMapping.java
HannesWell: The problem is the ".xml" extension.
When `PolyglotModelManager.getReaderFor()` is called it asks each mapping (in descending priority order) if it accepts the given options. And `MappingSupport.accept()` falsely returns true for a `feature.xml`:
```
  public boolean accept(final Map<String, ?> options) {
    if (options != null) {
      for (String key : getAcceptOptionKeys()) {
        if (options.containsKey(key)) {
          return true;
        }
      }
      String location = getLocation(options);
      if (location != null) {
        for (String ext : getAcceptLocationExtensions()) {
          if (location.endsWith(ext)) {
            return true;
          }
        }
      }
    }
    return false;
  }
```
The XML-model in the feature.xml file (usually) does not have the "xml:4.0.0" options-key set, but its location ends with `feature.xml`. In fact that Mapping accepts itself to provide a reader for any XML-file. That's the issue.
Until now we are (probably) lucky that the `XMLMapping` is always sorted behind the `TychoFeatureMapping` in `PolyglotModelManager.getSortedMappings()`, but I assume that this is an implementation detail of plexus.

HannesWell:
> 
> now I have the `<module>bundles</module>` in my `pom.xml`, it will build `bundle1` + `bundle2`, because `experimental` is not a valid name nothing will happen here. Now we allow _any_ name, suddenly the build will execute `bundle1` + `bundle2` + `bundle3` because `experimental` will be assumed a valid choice and gets a connect pom generated that includes `bundle3` in its module list.

OK, if the TychoAggregatorMapping would search with infinite depth this would indeed break builds. But with the current approach that it only scans with depth one (however this may be implemented), this should not occur. But this would break those builds that have  something like:
- bundles/bundle1
- bundles/bundle2
- bundles/plugins/bundle3

But is this a really common use-case? Furthermore Tycho 3.0 is a major release, so breaks would be OK.

But again. We should first find consensus if this feature is wanted at all, assuming all technical issues can be solved.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoFeatureMapping.java
laeubi: I think I fixes a while back some thing regarding the handling already, this defiantly does not looks correct and the xml mapping might need to override this.

This seems to be a very simplifed case because orignianlly polyglot was written to provide a json or yam and then the filtering by extension might work but in general this seems not sufficient, at laest the XML mapping then might better have a very low priority instead so it is used as last resort.

laeubi:
> But with the current approach that it only scans with depth one (however this may be implemented), this should not occur.

And that's wrong, you can't know the nesting in any way as you always ever only handle *one* folder, and you can't know if this is by pom, by commandline or automatically generated.

> But is this a really common use-case? Furthermore Tycho 3.0 is a major release, so breaks would be OK.

It dosen't matter if it is common or not and it wont be okay to change this this for no good reason and using other names is neither a common usecase too nor is it something we want to recommend and is already supported by config option.

akurtakov:
Will such a change help to make dependabot cooperate with tycho pomless? Right now we have dependabot enabled only for eclipse-platform-parent (https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/.github/dependabot.yml) as it chokes on pomless modules.

laeubi:
> Will such a change help to make dependabot cooperate with tycho pomless?

Nope

> as it chokes on pomless modules

can you explain or give an example? Actually dependabot should not care much as pomless modules never declare any (visible) dependencies.

akurtakov:
Sure, see this run https://github.com/eclipse/shellwax/network/updates/439045222 . ShellWax uses pomless and thus dependabot doesn't manage to run.

laeubi:
It says "page not found" for me, but if you can provide a minimal reproducer we should defiantly report this to dependabot, as this is not Tycho specific but would affect anyone using a polyglot build.

akurtakov:
The gist of it is:
```

Dependabot couldn't find a pom.xml

Dependabot couldn't find a pom.xml.

Dependabot requires a pom.xml to evaluate your Java dependencies. It had expected to find one at the path: /org.eclipse.shellwax.core/pom.xml.

If this isn't a Java project, you may wish to disable updates for it in the .github/dependabot.yml config file in this repo.
```
I agree it's general polyglot issue. How do you propose to create a reproducer ? I don't think I can call dependabot myself in anyway, can I ?

laeubi:
I think one would:

1. Create a minimal project setup with one `pom.xml` and one `polyglot module`
2. Use an outdated dependency in the `pom.xml`
3. Enable dependabot for that repository
4. Wait for Dependabot to kick in

HannesWell:
> > But with the current approach that it only scans with depth one (however this may be implemented), this should not occur.
> 
> And that's wrong, you can't know the nesting in any way as you always ever only handle _one_ folder, and you can't know if this is by pom, by commandline or automatically generated.

The limited nesting depth would of course only apply to the automatically-generated modules by Tycho. With poms one can nest as deep as wanted.
If for example the directory File that is scanned is added to a static ConcurrentHashSet before starting a scan (and removed afterwards) and if then the scan methods then checks if the parent of the directory to scan is in that set or not and acts accordingly, then I thin it should work regardless of if the module directory is declared as module in another pom or if one calls `mvn -f bundles` (and has the Tycho-extension in the correct place. Not sure if it then has to be `./.mvn` or `./bundles/.mvn`). 

> 
> > But is this a really common use-case? Furthermore Tycho 3.0 is a major release, so breaks would be OK.
> 
> It dosen't matter if it is common or not and it wont be okay to change this this for no good reason and using other names is neither a common usecase too nor is it something we want to recommend and is already supported by config option.

So if you prefer to have auto-generate aggregators for arbitrary depth like `bundle/bundle/bundle/product/` instead of the possibility to have aggregators with arbitrary names of depth one, then this can be closed. It would not have been my choice, because I think the former has less value then the latter, especially because all directories but the last must have one of the common names to make that work, so it is not even possible to find deeper nested `bundles` folders, but OK. Then so be it.

laeubi:
> then I thin it should work regardless of if the module directory is declared as module in another pom or if one calls

The problem is that it behaves differently where you call it. Just keep in mind that it is a "Tycho" Extension and not a generic "generate any module pom" Extension, so I think it is fine that it is somewhat Tycho specific, so to think even further here, It would even be better to only accept `eclipse-plugin` in a `bundles` folder, and `eclipse-feature` in a `features` folder and so on ... I hope this makes it more clear why we define fixed (default) names.

Nerveless one can think about a generic "generate-modules" extension that simply work as you proposed here, it won't need Tycho (pomless) in anyway to work and could be used in any maven setup. If you choose a slightly higher priority it could even be replace tychos current default. It could even be hosted as a separate module on the takair-polyglot project.

Beside that I think that your enhancements to documentation and getting a more semantic ordering (e.g bundle is higher than feature is higher than product is higher than updatesite ...) would still be useful.

HannesWell:
> 
> Beside that I think that your enhancements to documentation and getting a more semantic ordering (e.g bundle is higher than feature is higher than product is higher than updatesite ...) would still be useful.

Yes, I will provide a corresponding PR to implement that once the polyglot-maven enhancement is merged.
To clarify the priorities I think we need a dedicated `PomXMLMapping` with highest priority. All others should have lower onces in the mentioned order. The `XMLMapping` should have the lowest (it has zero). Considering its name it is actually valid that it accepts all xml files, which makes the dedicated Mapping necessary.


> Nerveless one can think about a generic "generate-modules" extension that simply work as you proposed here, it won't need Tycho (pomless) in anyway to work and could be used in any maven setup. If you choose a slightly higher priority it could even be replace tychos current default. It could even be hosted as a separate module on the takair-polyglot project.

OK, I will prepare a Issue to discuss that at polyglot-maven and work out a PR if they agree.

laeubi:
> To clarify the priorities I think we need a dedicated `PomXMLMapping` with highest priority.

Should a pom.xml not already disable any TychoMapping?

> Considering its name it is actually valid that it accepts all xml file

Not really if it cannot handle all XML files ...

HannesWell:
> > To clarify the priorities I think we need a dedicated `PomXMLMapping` with highest priority.
> 
> Should a pom.xml not already disable any TychoMapping?

It does it, yes. But when we introduce dedicated priorities for mappings and polyglot-maven considers them always having a dedicated `PomXMLMapping` would allow us to remove the check for a pom.xml and would avoid the interference with the `XMLMapping` and `feature.xml` files.
But lets discuss that when I have created the PR, it is the more clear and simpler with the code.

> 
> > Considering its name it is actually valid that it accepts all xml file
> 
> Not really if it cannot handle all XML files ...

It can handle it, but it doesn't handle it correctly. I just suspect that polglot-maven won't accept a PR to change the XMLMapping.
We could alternativly adjust the mapping component at runtime, but that could interfere with other mappings if one uses Tycho with other polgyglot extensions?
But lets discuss that as well with the PR once created, I will provide it after my one week vacation and hopefully then there will be some reaction in https://github.com/takari/polyglot-maven/pull/242.

PR #1265, opened 2022-08-13
Support mixed Products and product 'type' attribute

HannesWell:
With https://github.com/eclipse-pde/eclipse.pde/pull/291 PDE supports mixed Products, consisting of Features and Plug-ins.
P2 supports that already but Tycho has to be adjusted at a few locations to support it as well.

Furthermore this adds support for the `type` attribute of a product can have one of the following values: 
`bundles`, `features`, `mixed`.

I'm not sure if the change in `ProductConfiguration` is necessary, because I never hit that while debugging a product build.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,11 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## 3.0.0 (under development)
 
+### Support for Eclipse-Products with mixed Features and Plugins
+
+Tycho now supports building _mixed_ Products. In mixed Products both the listed features and listed plug-ins are installed.
+Therefore the Product attribute `type` is now supported, which can have the values `bundles`, `features` and `mixed` and takes precedence over the boolean-valued `useFeatures` attribute.
+
 ### New API Tools Mojo
 
 Tycho now provides a new API Tools Mojo, see https://github.com/eclipse/tycho/tree/master/tycho-its/projects/api-tools for an example how to use it.
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/actions/ProductDependenciesAction.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/actions/ProductDependenciesAction.java
@@ -19,6 +19,7 @@ import java.util.Set;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.equinox.internal.p2.publisher.eclipse.IProductDescriptor;
+import org.eclipse.equinox.internal.p2.publisher.eclipse.ProductContentType;
 import org.eclipse.equinox.internal.p2.publisher.eclipse.ProductFile;
 import org.eclipse.equinox.p2.metadata.IRequirement;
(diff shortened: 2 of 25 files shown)


HannesWell:
@laeubi, @akurtakov do you want to review this change?
If you are fine with it, I can rebase it to remove the commit of https://github.com/eclipse-tycho/tycho/pull/1266 so that we can have that separably.

tycho-testing-harness/src/main/java/org/eclipse/tycho/test/AbstractTychoIntegrationTest.java
```diff
@@ -143,7 +143,7 @@ protected Verifier getVerifier(String test) throws Exception {
         return getVerifier(test, true);
     }
 
-    protected String getTargetPlatform() {
+    protected static String getTargetPlatform() {
```
laeubi, line 146: It does not make much sense to have such a (protected) method when it only delegates to a public other static method as no one can ever override this....

tycho-testing-harness/src/main/java/org/eclipse/tycho/test/AbstractTychoIntegrationTest.java
```diff
@@ -174,7 +174,7 @@ protected String getMavenHome() {
         return mavenHome;
     }
 
-    protected String getTychoVersion() {
+    protected static String getTychoVersion() {
```
laeubi, line 177: see last comment...

tycho-testing-harness/src/main/java/org/eclipse/tycho/test/AbstractTychoIntegrationTest.java
HannesWell: This was part of https://github.com/eclipse-tycho/tycho/pull/1266 and has been reverted there.

tycho-testing-harness/src/main/java/org/eclipse/tycho/test/AbstractTychoIntegrationTest.java
HannesWell: Same as there.

RELEASE_NOTES.md
```diff
@@ -4,6 +4,11 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## 3.0.0 (under development)
 
+### Support for Products with mixed Features and Plugins
+
+Tycho now supports building _mixed_ Products, which have features and plug-ins listed and get both installed.
+Furthermore Product attribute 'type' is now supported, which can have the values 'bundles', 'features' and 'mixed' and supersedes the boolean-valued 'useFeatures' attribute.
```
laeubi, line 10: Do you mean `type` it takes precedence over `useFeatures`?
`supersedes` seem to indicate I cannot use `useFeatures` anymore.

HannesWell: > Do you mean `type` it takes precedence over `useFeatures`? `supersedes` seem to indicate I cannot use `useFeatures` anymore.

Yes I do. Adjusted the test accordingly.

laeubi (approved):
Looks good

HannesWell:
> Looks good

Thanks.
I stripped down the test cases a bit more so we should await the build results, but if they are green, this is ready to go.

PR #1266, opened 2022-08-14
Fix/ignore tests failing on Windows and clean-up/enhance test-code

HannesWell:
This PR aims to fix the two tests that are currently failing on Windows with errors:
- `MetaRequirementsTest`
It looks like this simply does not work on Windows, but I was not able to fix that. Locally the build fails reliably. If anybody know how to fix that, I'm happy to apply it. I don't know why it succeeds on Linux. For know I ignored it on Windows. Or should we let it keep failing?
- CustomSourceEncodingTest
Here I'm not sure about the exact purpose of the test but the applied change fixed it locally for me on Windows. But I have the impression that this change make the test worthless, but I'm also not sure how it should look like exactly.

Site-Note: When I execute the previous state of the build of that test locally from the command-line it fails like in the CI, but when I run it from my IDE it succeeds. The reasons for that is that M2E automatically adds the -Dfile.encoding=UTF-8 as VM argument when launching a build. I think that is something we should discuss in M2E.

Besides that this PR cleans up some tests and adds `AbstractTychoIntegrationTest.assertDirectoryDoesNotExist()`. I need that for https://github.com/eclipse-tycho/tycho/pull/1265, but want to keep that PR free of general test-code enhancements.

### DIFF
--- a/tycho-testing-harness/src/main/java/org/eclipse/tycho/test/AbstractTychoIntegrationTest.java
+++ b/tycho-testing-harness/src/main/java/org/eclipse/tycho/test/AbstractTychoIntegrationTest.java
@@ -189,10 +189,7 @@ public abstract class AbstractTychoIntegrationTest {
      * @return an array of matching files (will contain at least one file)
      */
     protected File[] assertFileExists(File baseDir, String pattern) {
-        DirectoryScanner ds = new DirectoryScanner();
-        ds.setBasedir(baseDir);
-        ds.setIncludes(new String[] { pattern });
-        ds.scan();
+        DirectoryScanner ds = scan(baseDir, pattern);
         File[] includedFiles = Arrays.stream(ds.getIncludedFiles()).map(file -> new File(baseDir, file))
                 .toArray(File[]::new);
         assertEquals(baseDir.getAbsolutePath() + "/" + pattern, 1, includedFiles.length);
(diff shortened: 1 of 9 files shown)


laeubi:
> For know I ignored it on Windows. Or should we let it keep failing?

I'm fine with ignoring test if they do not work reliable but maybe @akurtakov or @mickaelistria can tell what the test is about?

> * I don't know why it succeeds on Linux

I noticed that maven sometimes behaves a little bit different on windows no idea why. For one test that checks for an error in the log I had to catch the exception and then check for the error while on linux that wa snot neccesary...

> * Here I'm not sure about the exact purpose of the test

I don't know either ;-)

akurtakov:
I'm afraid that many of the tests we run are unclear what they test, test no longer relevant things or do too much to make easy to spot what's being tested. Thus any help on that level is more than welcome and removing tests when they don't make sense anymore is more than welcome. I'm concerned though about disabling test on certain OS because it fails as to me this most likely points to OS specific bugs, if the test itself is OS specific it makes perfect sense and would be nice to add such explanation to the ignore message.
Specifics:
* MetaRequirementsTest points to https://bugs.eclipse.org/bugs/show_bug.cgi?id=351487 and we no longer have different runtime for director thus test might need bigger change. But it looks like a bug if touchpoint is not run on Windows only.
* CustomSourceEncodingTest - the modification makes it irrelevant for sure. Whether the test as a whole is still relevant is another question. I personally more and more believe that support for non UTF* should be phased out slowly ;) so I'll leave that to you guys to decide.

HannesWell:
> 
>     * MetaRequirementsTest points to https://bugs.eclipse.org/bugs/show_bug.cgi?id=351487 and we no longer have different runtime for director thus test might need bigger change. But it looks like a bug if touchpoint is not run on Windows only.

I'm not familiar enough with the corresponding code to tell that immediately, but I suspected the same.

> 
>     * CustomSourceEncodingTest - the modification makes it irrelevant for sure. Whether the test as a whole is still relevant is another question. I personally more and more believe that support for non UTF* should be phased out slowly ;) so I'll leave that to you guys to decide.

I can really don't understand how that test is supposed to work, because the test explicitly checks for certain Unicode characters that are not in Latin1. So I wonder how that should ever have worked to encode the file in US_ASCII. Nevertheless it looks like the file-encoding is done separately on Windows and Linux. And as said above in the IDE build with M2E the test is fine because M2E or JDT adds the commandline flag to UTF-8 encoding before the Maven argument is added.

laeubi (approved):
As no one really seem to care much about proposing a better solution, I think we can go ahaed.

tycho-testing-harness/src/main/java/org/eclipse/tycho/test/AbstractTychoIntegrationTest.java
```diff
@@ -143,7 +143,7 @@ protected Verifier getVerifier(String test) throws Exception {
         return getVerifier(test, true);
     }
 
-    protected String getTargetPlatform() {
+    protected static String getTargetPlatform() {
```
laeubi, line 146: It does not make much sense to have such a (protected) method when it only delegates to a public other static method as no one can ever override this....

tycho-testing-harness/src/main/java/org/eclipse/tycho/test/AbstractTychoIntegrationTest.java
```diff
@@ -174,7 +174,7 @@ protected String getMavenHome() {
         return mavenHome;
     }
 
-    protected String getTychoVersion() {
+    protected static String getTychoVersion() {
```
laeubi, line 177: It does not make much sense to have such a (protected) method when it only delegates to a public other static method as no one can ever override this....

tycho-testing-harness/src/main/java/org/eclipse/tycho/test/AbstractTychoIntegrationTest.java
HannesWell: Didn't expect its purpose is to be overridden, because it was never done anywhere. But reverted it.

tycho-testing-harness/src/main/java/org/eclipse/tycho/test/AbstractTychoIntegrationTest.java
HannesWell: Reverted that as well.

Issue #1261, opened 2022-08-10
PGP signature incorrectly generated for feature if the ID & version also match a bundle


Kummallinen:
When using tycho-gpg-plugin on a repository containing a feature with the same ID & version as a bundle, for example org.eclipse.rcp, a PGP signature for the bundle is attached to the feature. This results in the p2 repository being unusable as the signature check on the feature fails on installation.

Looks like the problematic code is here: https://github.com/eclipse-tycho/tycho/blob/d9efa1acf42c74371c6115786be5951cbdf7f151/tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java#L123

I guess it should be checking the artifact type attribute rather than just checking for the presence of a matching file in "plugins"

I've only been able to test with Tycho 2.7.1 due to bugs preventing me from using later versions, but the code is unchanged so I assume it still happens.

mickaelistria:
Thanks @Kummallinen !

PR #1268, opened 2022-08-15
Bug #1261 Check artifact classifier before PGP signing

Kummallinen:
In some cases (e.g. org.eclipse.rcp) a bundle & feature can have the
same ID & version. To avoid generate an invalid sigurature in theses
cases check the artifact classifier before checking if the jar exists
in plugins/

As Platform does check signatures on feature this could be enhanced
to correctly PGP sign features in future

I couldn't see any integration tests for this feature

### DIFF
--- a/tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
+++ b/tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
@@ -120,8 +120,24 @@ public class SignRepositoryArtifactsMojo extends AbstractGpgMojoExtension {
                 // skip packed artifacts
                 continue;
             }
-            var file = new File(repository,
-                    "plugins/" + artifact.getAttribute("id") + '_' + artifact.getAttribute("version") + ".jar");
+            /*
+             * Different types of artifact have different locations in the repo. So we need to check
+             * the classifier to correctly get the location of the jar file.
+             * 
+             * Could potentially do this dynamically based on the mappings attribute with the repo.
+             */
+            String subDir = null;
+            switch (artifact.getAttribute("classifier")) {
+            case "osgi.bundle":
+                subDir = "plugins";
+                break;
+            case "org.eclipse.update.feature": //Not yet signing features
+            case "binary": //Not yet signing binaries
+            default:
+                continue; // Skip signing
+            }
+            var file = new File(repository, subDir + File.separator + artifact.getAttribute("id") + '_'
+                    + artifact.getAttribute("version") + ".jar");
             if (!file.canRead()) {
                 continue;
             }


laeubi:
> I couldn't see any integration tests for this feature

Do you like to provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue with this PR?

Kummallinen:
The only PGP related tests I could find were just using already signed artifacts, rather than signing during the test which would be required to demonstrate this issue. If someone can point me to one I've missed then I'm happy to use it as a reference for adding a test for this. Or I could add an example which could be run manually, as it is trivial to reproduce.

mickaelistria:
> To avoid generate an invalid sigurature in theses cases check the artifact classifier before checking if the jar exists in plugins/

Can you please provide detailed steps to reproduce this issue? Instead of skipping generation for features, as it might be interesting in the future, we should instead fix this case so it uses the right artifact.

Kummallinen:
> > To avoid generate an invalid sigurature in theses cases check the artifact classifier before checking if the jar exists in plugins/
> 
> Can you please provide detailed steps to reproduce this issue? Instead of skipping generation for features, as it might be interesting in the future, we should instead fix this case so it uses the right artifact.

The reproduction steps are:
1. Define a p2 repo which contains a feature & plugin with identical IDs & versions. org.eclipse.rcp is the one I discovered the issue with
2. Sign the repo with the tycho-pgp-plugin
3. Try to install the feature from the signed repo. Installation will fail due to a signature validation failure

As I noted in the initial pull request comment the feature PGP signature is validated on installation (if it didn't this bug wouldn't cause any issues). So generating a signature for a feature, or indeed a binary, does appear to be valid operation but for whatever reason the original code was written to only sign plugins.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
```diff
@@ -120,6 +120,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
                 // skip packed artifacts
                 continue;
             }
+            if(!"osgi.bundle".equals(artifact.getAttribute("classifier"))) {
+                continue; //Don't generate signatures for anything that isn't a "bundles"
+            }
```
mickaelistria, line 125: Instead of a `continue` here, as we may add support for features or other some day, it would be better to use a "positive" condition so we can easily add some `else` clause for other classifiers in the future.

Kummallinen: I was matching the surrounding code while trying to keep the change as minimal as possible.  Making the change you request would imply also modifying lines 126 & 127 so the path to the jar can change based on the artifact type.

Would it be acceptable to make that change but implement still only generating the path for bundles?

mickaelistria: yes, I think it would be ok.

PR #1273, opened 2022-08-16
Example to reproduce #1261

Kummallinen:
Example p2 repo which reproduces #1261, requested in PR #1268 

Building this repo requires an environment already set up to PGP sign artifacts, the key name & password server ID need to be passed via the test-pgp-key.name & test-pgp-key.serverId properties.

### DIFF
--- a/tycho-its/projects/issue1261/category.xml
+++ b/tycho-its/projects/issue1261/category.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<site>
+   <!--org.eclipse.rcp include a bundle with the same ID & version, which triggers #1261 -->
+   <feature id="org.eclipse.rcp"/>
+</site>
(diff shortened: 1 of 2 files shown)


laeubi:
@mickaelistria can you help converting this into a integration test?

mickaelistria:
I don't think I'll spend time to make it an automated test, particularly since it requires to setup gpg on the CI and so on, it's more work than what I'm able to spend these days. The other PR is relatively safe and this part of the code is not too fragile, so we can live without a test for the moment.
If someone feels like they can make it an automated test, please do it; if not, please close this PR.

laeubi:
@Kummallinen thanks for the test do you like to investigate if a temporary key can be generated and used for the test?

Kummallinen:
> @Kummallinen thanks for the test do you like to investigate if a temporary key can be generated and used for the test?

I can look into it but won't have the time for a few weeks. From my experiences using the gig plugin in my own builds I think getting the temporary key into a usable key ring would be the sticking point.

laeubi:
I think @merks has already done some tests with temporary PGP key, maybe the aproach can be reused/adapted for a test here?
@Kummallinen do you like to finish this PR so we can include it in the regular checks?

laeubi:
Closing as no further feedback from original author @Kummallinen if this is still interesting for you please let me know so we can reopen the PR.

PR #1287, opened 2022-08-25
Test if we need the tycho bundles product anymore

laeubi:


laeubi:
Platfrom build already suceeds with this change! @akurtakov any ide whats wrong with the failing test? It seems to fail with NPE because it can't find a properties file but as far as I can see the file itself is there...

akurtakov:
I debugged locally and it fails at DefaultEquinoxInsallation.getLauncherJar() line 51 where description.getSystemBundle(); returns null. First thought is that OSGi framework is not started at that point but the code expects it to be.

laeubi:
> I debugged locally and it fails at DefaultEquinoxInsallation.getLauncherJar() line 51 where description.getSystemBundle(); returns null. First thought is that OSGi framework is not started at that point but the code expects it to be.

Okay, I added equinox now, lets see if that helps, I somehow fear that "launching a P2 application" here mens we are not really launching a full app, but require it has a full P2 set of bundles....Actually the same is true here as for the API-Tools story, that actually P2 should provide such a product we can then reuse at Tycho...

laeubi:
It seems
- https://github.com/eclipse-tycho/tycho/issues/1290

is a perquisite for this ... so converting this to a draft for now, but if one feels to help with that any help is welcome, it seems not very profitable to manage dependencies at two places...

laeubi:
@mickaelistria @akurtakov it seems there is some hard reference when using the "standalone director application", but its a bit confusing as it is standalone why should it require the Tycho product. Any ideas? I think getting rid of the Tycho product would be good before the release is out, so we do not need to maintain it anymore for just rare use-cases... it seems all other tests already pass!

laeubi:
According to the code we need a 

https://github.com/eclipse-tycho/tycho/blob/57415c2006454138ee05b97337816c835e0a883f/tycho-p2/tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/p2/director/runtime/StandaloneDirectorRuntimeFactory.java#L69

So probably something like this:
https://github.com/eclipse-equinox/p2/blob/master/bundles/org.eclipse.equinox.p2.director.app/director.product

I think we have the following options:

1. Maintain the Tycho product just for the sake of "external director usecase" (will add additional maintainace work)
2. Use an old Tycho Release for now, and do some work to directly consume such a product from P2 project itself (will use an older release of eclipse for such purpose until we found a better soloution)
3. Drop the "external director usecase" ... (but mostly someone will complain)

laeubi:
I'm trying out "Option 2" in this PR now...

mickaelistria:
What does is this StandaloneDirectorRuntimeFactory provide? Is there any known use-case for it (seached GitHub) ? Is there a supported alternative to known user stories (eg calling the director via some other mechanism)?

laeubi:
> What does is this StandaloneDirectorRuntimeFactory provide?

As far as I understand this is sued when you build a product and want to use custom touchpoints, then Tycho installs a director application, then installs additional stuff (your touch point) and then use that to assemble the actual product.

> Is there any known use-case for it (seached GitHub)? Is there a supported alternative to known user stories (eg calling the director via some other mechanism)?

There is one integration test in Tycho, but I have never used that anywhere, the platformbuild itself seem not to use this from the result of the tycho-platform-verification-build.

A while back there where two products and they where merged: https://github.com/eclipse-tycho/tycho/commit/2dccd195091a5bfe9829bc81a4d113db2a6e30ab

the feature itself was added in 2012 by @oberlies https://github.com/eclipse-tycho/tycho/commit/7b29217072346f5d3c95201889b01d5b4e6995ef and adjusted by @nataliab here https://github.com/eclipse-tycho/tycho/commit/e22dc6f6a945074c344de1a07bbbb364ec45cba3

mickaelistria:
Reading https://www.eclipse.org/tycho/sitedocs/tycho-p2/tycho-p2-director-plugin/materialize-products-mojo.html#directorRuntime , I don't really get how the standalone director can include such "non standard touchpoints" either. ie if I develop a touchpoint, how can I inject it in the standalone director?
I believe such particular case can better be resolved by usage of the tycho-eclipserun-plugin that lets the caller configure the included bundles, so one can inject their touchpoints.
Assuming the above is true (what do you think?), then we'd better remove the standalone director entirely.

laeubi:
https://github.com/eclipse-tycho/tycho/blob/master/tycho-its/src/test/java/org/eclipse/tycho/test/product/MetaRequirementsTest.java has some explanation, and the integration test do it the following way:

1. Build the touch point and everything else
2. Then in the custom director install everything (including the touchpoint) and use that to assemble the product: https://github.com/eclipse-tycho/tycho/blob/master/tycho-its/projects/product.metaRequirements/product/pom.xml

I think this might be achievable with an eclipse-run as well but haven't tried that yet... but it might be similar inconvenient than whats currently done with the api-analysis stuff...

laeubi:
> I'm trying out "Option 2" in this PR now...

This works, so at laest would be an option to use unless we find a better solution or decide to drop this at all. From the history it might be good to let P2 produce a "director product" and then Tycho can simply consume this, install e.g. the tycho-noop security (if thats required at all) and then we don't need to care much here at tycho ...

laeubi:
@akurtakov do you have any concerns here? If not I'll merge this once the build succeeds.

akurtakov:
No concerns, I like the idea of getting rid of it.

akurtakov:
Are you looking into platform failure ?

laeubi:
> Are you looking into platform failure ?

Platform build succeeds with this change in this PR so I'm not aware of any... beside that this change was not (yet) promoted to the snapshot-repository so I suspect that if there is any problem it is not related to this change...

akurtakov:
Ah, I must have overlooked it.

PR #1299, opened 2022-08-28
Clarify priority of Tycho-mappings and update to polyglot-common 4.9

HannesWell:
As discussed in https://github.com/eclipse-tycho/tycho/pull/1264, this introduces a well defined priority for all of Tycho's polyglot mappings.
The exact order is:
- pom.xml
- bundle (i.e. `META-INF/MANIFEST.MF`)
- feature (i.e. `feature.xml`)
- repository (i.e. `category.xml` and `*.product` files)
- target (i.e. `*.target`)
- aggregator (i.e. a directory with one of the names specified via system property `tycho.pomless.aggregator.names`)

Furthermore this updates to polyglot-common 4.9 to use https://github.com/takari/polyglot-maven/pull/242 and introduces a PomXMLMapping. The latter is necessary because the default `org.sonatype.maven.polyglot.mapping.XmlMapping` erroneously claims to handle feature.xml models (because of their ending). Without a dedicated PomXMLMapping we would have to move the `XMLMapping` as second last in the list above and would have to check in each other Tycho-mapping if the project also has a `pom.xml`.


tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/PomXMLMapping.java
```diff
@@ -0,0 +1,37 @@
...
+package org.eclipse.tycho.pomless;
+
+import java.io.File;
+import java.util.Map;
+
+import org.codehaus.plexus.component.annotations.Component;
+import org.sonatype.maven.polyglot.mapping.Mapping;
+import org.sonatype.maven.polyglot.mapping.MappingSupport;
+
+@Component(role = Mapping.class, hint = "pom.xml") //TODO: adjust hint?
```
laeubi, line 23: Why do we need this? With the new priority should not TychoMappings be asked anyways before? The check for pom.xml must be done in the tycho mappings anyways...

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
```diff
@@ -71,16 +71,6 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
     private File multiModuleProjectDirectory;
     private String snapshotFormat;
 
-    @Override
-    public File locatePom(File dir) {
```
laeubi, line 75: I don't think we should remove this and instead introduce a useless pom.xml mapping....

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/PomXMLMapping.java
HannesWell: The current suggestion is to have to following mapping priority:

1. pom.xml
2. bundle
3. feature
4. repository
5. target
6. aggregator
7. XMLMapping

If we don't have a dedicated PomXMLMapping then the mapping priority would have to be
1. bundle
2. feature
3. repository
4. target
5. XMLMapping
6. aggregator

But we have to keep the check in `AbstractTychoMapping.locatePom()` if a `pom.xml` is already present in the project's root.
So while the `XMLMapping` has a low priority (in order to handle `feature.xml` files), the pom.xml files it is intended to handle have actually maximum priority. I found that odd and therefore added the PomXMLMapping so that the priority of the mappings reflects the priority of the files they handle, in order to keep that model (IMO) simpler to understand.

But of course it would also work the other way. It is more a question what we prefer.

> The check for pom.xml must be done in the tycho mappings anyways...

Why must the check be done anyways if we have a dedicated PomXMLMapping?
At least there is not test that fails with the current suggestion. But if you can think of a scenario were it is required some tests should be added.

laeubi: > At least there is not test that fails with the current suggestion.

I don't think we have much test-coverage for the pomless stuff :-\

> Why must the check be done anyways if we have a dedicated PomXMLMapping?

Because `TychoMappings` should be ignored if there is a `pom.xml`, they should not be ignored if there is any other mapping with higher priority (what is what you have implemented here), even though the visible effect is the same in most scenarios this is a different semantics.

> I found that odd and therefore added the PomXMLMapping so that the priority of the mappings reflects the priority of the files they handle,

We don't really want or should handle `pom.xml` files so adding a mapping in in Tycho is simply wrong (and maybe even more odd), the choice that `pom.xml` always takes precedence is a Tycho choice but not normative at all when it comes to polyglot in general.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
```diff
@@ -110,11 +112,6 @@ protected void initModel(Model model, Reader artifactReader, File artifactFile)
         }
     }
 
-    @Override
-    public float getPriority() {
-        return -10f;
```
laeubi, line 115: Anything wrong with negative priority? The idea was that this mapping is asked only as a last resort if nothing else matches.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoTargetMapping.java
```diff
@@ -37,13 +37,18 @@ protected String getPackaging() {
...
+    }
+
     @Override
     protected boolean isValidLocation(String location) {
         return location.endsWith(TARGET_EXTENSION);
     }
 
     @Override
-    protected File getPrimaryArtifact(File dir) {
+    public File locatePom(File dir) {
```
laeubi, line 51: Is there any reason here (and on other places) to override locatePom instead of using getPrimaryArtifact?

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/PomXMLMapping.java
HannesWell: > > At least there is not test that fails with the current suggestion.
> 
> I don't think we have much test-coverage for the pomless stuff :-\

At least with the pomless-model tests we have some more but of course there could be more.

> > I found that odd and therefore added the PomXMLMapping so that the priority of the mappings reflects the priority of the files they handle,
> 
> We don't really want or should handle `pom.xml` files so adding a mapping for it in Tycho is simply wrong (and maybe even more odd), the choice that `pom.xml` always takes precedence is a Tycho choice but not normative at all when it comes to polyglot in general.
> 
> Just assume I also want to use another mapping that do not has this rule, then Tycho can't be used without giving a higher priority than our "pom xml mapping" just because we don't handle this rule internally. You should always be aware that marking anything `@Component` it makes it visible to the whole maven world.

Yes, that's indeed a reason against the first suggestion. Although I think it wouldn't be a bad choice for other polyglot-languages as well we must not impose this rule from Tycho. I removed the PomXMLMapping again and changed the priorities to:

1. bundle
2. feature
3. repository
4. target
5. XMLMapping
6. aggregator

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
HannesWell: Was just possible with the higher prioritized PomXMLMapping, now it is reverted.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
HannesWell: Nope not at all. It was just with the initially suggested orders the XMLMapping became de-facto unused for all PDE Project-types and I therefore moved it above the XMLMapping.
With the now implemented solution it must have a lower priority than the XMLMapping so I reverted to -10.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoTargetMapping.java
HannesWell: This was because with the higher Prioritized PomXMLMapping the check for a pom.xml was not necessary in the `locatePom()` implementation in the super-class which was just calling `getPrimaryArtifact()`, so `locatePom()` could have been called directly. But since we cannot introduce a PomXMLMapping, that has to stay as it is and I reverted it.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoFeatureMapping.java
```diff
@@ -33,6 +34,16 @@ public class TychoFeatureMapping extends AbstractXMLTychoMapping {
     private static final String FEATURE_XML = "feature.xml";
     public static final String PACKAGING = "eclipse-feature";
 
+    @Override
+    protected String getPackaging() {
```
laeubi, line 38: It seems there is some reordering of code now, can you revert to the original location sow we get (hopefully) a smaller Diff?

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoFeatureMapping.java
```diff
@@ -54,7 +65,7 @@ protected void initModelFromXML(Model model, Element xml, File artifactFile) thr
 
     @Override
     protected boolean isValidLocation(String location) {
-        return location.endsWith(FEATURE_XML);
+        return Path.of(location).endsWith(FEATURE_XML);
```
laeubi, line 68: Is there any advantage of constructing a path first? Could this be a separate PR?

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoBundleMapping.java
```diff
@@ -46,6 +46,16 @@ public class TychoBundleMapping extends AbstractTychoMapping {
     private static final String BUNDLE_SYMBOLIC_NAME = "Bundle-SymbolicName";
     private static final String PACKAGING_TEST = "eclipse-test-plugin";
 
+    @Override
+    protected String getPackaging() {
```
laeubi, line 50: also reordering here...

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java
```diff
@@ -45,9 +43,18 @@ public class TychoAggregatorMapping extends AbstractTychoMapping {
...
     private static final String TYCHO_POM = "pom.tycho";
 
-    private static final Set<String> COMMON_NAMES = new HashSet<>(
-            Arrays.asList(System.getProperty(TYCHO_POMLESS_AGGREGATOR_NAMES_PROPERTY,
-                    "bundles,plugins,tests,features,sites,products,releng").split(",")));
+    private static final Set<String> COMMON_NAMES = Set.of(System.getProperty(TYCHO_POMLESS_AGGREGATOR_NAMES_PROPERTY,
+            "bundles,plugins,tests,features,sites,products,releng").split(","));
+
+    @Override
+    protected String getPackaging() {
```
laeubi, line 50: can you restore original ordering here as well so we see smaler change set?

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoFeatureMapping.java
HannesWell: This makes the test more precise. Previously that mapping would also handle a project that contains a file named:
`someOtherStringfeature.xml`

HannesWell:
I split the this change into two commits.
The first one only introduces the mapping-specific priorities (the actual purpose of this PR), the second one applies some clean-up/code re-ordering.
If you prefer I can open the second commit as separate PR, but the commits can also be reviewed one by one.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoFeatureMapping.java
laeubi: But then we need some error handling as the location not necessarily must be a java path (even though it is very likely)

HannesWell: If we do that, it should be consistent with all the other places that create Files without error handling, for example `TychoBundleMapping.isValidLocation()` and many more.
Furthermore I also wonder how errors would ideally be handled? Fallback to the String.endsWith() in a catch-block? Or just throw an Exception? Because as you said flawed paths are unlikely we maybe would better be told when they occur?

However according to the Java-doc new File() only throws an NPE, while Path.of() can throw a invalid Path exception.
So following that `new File(location).getName().equals(FEATURE_XML)`, should not throw any exception.

laeubi: Correct, `new File()` never throws any exception (except fro null) but `Path.of()` does when it think it is invalid. This can lead to strange results, but as we never know what client code passes in as a location one should better be safe here, making it fail to parse even though might simply result in false being returned.

HannesWell: To handled this consistently across all Mappings this requires changes at multiple locations and migration from the `File` to the `Path` API, to not switch back and forth.
I extracted those changes to a third commit.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
```diff
@@ -136,17 +138,17 @@ public Model read(File input, Map<String, ?> options) throws IOException {
 
     @Override
     public Model read(Reader input, Map<String, ?> options) throws IOException {
-        File file = new File(PolyglotModelUtil.getLocation(options));
+        Path file = getLocation(options).orElseThrow(() -> new IOException(""));
```
laeubi, line 141: could you add a meaningful message here?

laeubi (commented):
Just a minor comment before we can merge this, maybe you also want to squash all and use one commit message that covers all changes.

HannesWell:
> Just a minor comment before we can merge this, maybe you also want to squash all and use one commit message that covers all changes.

I would prefer to keep the changes in distinct commits. Especially the last change to NIO Path changes many places and the other changes which would make it harder to find the changes of priorities if one later goes through git.
But the specification of distinct priorities plus method reordering and Set simplifications can be squashed without making the result too complicated.

laeubi (approved):

PR #1323, opened 2022-09-10
Skip resolving if no mojo requires dependency resolution

laeubi:
This skips resolving of the project if no mojo requires actual dependency resolution.

### DIFF
--- a/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/ValidateClassPathMojo.java
+++ b/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/ValidateClassPathMojo.java
@@ -21,9 +21,11 @@ import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
 import org.apache.maven.project.MavenProject;
+import org.eclipse.tycho.ReactorProject;
 import org.eclipse.tycho.core.TychoProject;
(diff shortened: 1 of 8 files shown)


lppedd:
Hi @laeubi. Just noticed this PR. Curious to understand if it could affect - in a good way - this topic https://github.com/eclipse-tycho/tycho/discussions/1304

laeubi:
@lppedd You can try that out by specify `-Dtycho.resolver.classic=false` but be aware that 

> I have a plugin goal bound to the initialize phase, which goes through dependencies and copy some files using the Dependencies Plugin.

mostly indicates that you *are* actually requiring dependency resolution so if it would change something in your workflow depends on different aspects and your error more indicates that your local repository do not contain the required SNAPSHOT files (and as you are not running through the package phase they are not generated).

lppedd:
@laeubi Thanks! And yes, I need dependencies, but not p2 ones, so not resolved by Tycho, only standard Maven dependencies.
I don't know if it's possible to differentiate the two.

lppedd:
@laeubi I've tried with today snapshot, `-Dtycho.resolver.classic=false`, and a dumb `generate-sources` goal.
Still the same problem. Is it ok if I try to send a reproducer?

laeubi:
@lppedd [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue are always welcome, I'm just not sure if we can do much, because it does not matter how "dumb" a phase it, but if it has mojos that require dependency resolution.

PR #1327, opened 2022-09-12
Clean-up and improve documentation of baseline comparing Mojos

HannesWell:
The 'ignoredPatterns' parameter of the 'p2-metadata' and 'compare-version-with-baselines' goal is now documented.

Slightly improve performance of nested Zip-comparision and of the check if a file is ignored.


HannesWell:
> @HannesWell can you please assign a milestone this is planned for?

Yes, done it. I would like to merge this as soon as I have found out the reason for the failure.

laeubi:
Just a gentle reminder: I think we should create the 3.0.x branch and switch the master to 3.1.0-SNAPSHOT at 21.09.2022 what would mean no new features (just bugfixes), so if you want your changes to be included it would be good to have them until 20.09.2022 EOD.

HannesWell:
After the build has passed (which was the case the last time), I think this is good to go.

PR #1338, opened 2022-09-14
Bump Versions for 2022-09 release

laeubi:


akurtakov:
Overall looks good. It would be nice if deps like p2.core/repository/etc. are moved to dependencyManagement section in the parent so we are sure only single version is used.

laeubi:
> It would be nice if deps like p2.core/repository/etc. are moved to dependencyManagement section in the parent so we are sure only single version is used.

Yep that's something we should account for, I just used `mvn versions:use-latest-versions` here :-)

So we can get new platform version ASAP for now..

laeubi:
The `org.eclipse.tycho.test.apitools.ApiToolsTest` currently fails, if nothing else fails I still would merge this as we currently don't use that mojo anywhere and this is better analyzed/fixed in a separate PR...

PR #1347, opened 2022-09-16
Update Releasing with a note to sponsored items

laeubi:
@mickaelistria @akurtakov what do you think as project lead?

Beside that, I think we can move the file to the root and name it RELEASING.MD

### DIFF
--- a/devdoc/Releasing_Tycho.md
+++ b/devdoc/Releasing_Tycho.md
@@ -68,7 +68,10 @@ mvn clean deploy -Prelease -DskipTests -Dsource=8 -DjdetectJavaApiLink=false \
 git log --pretty=format:%an tycho-<previousVersion>..tycho-<newVersion> | sort | uniq
 git log --grep="Also-[bB]y:" tycho-<previousVersion>..tycho-<newVersion> | grep -i also-by | sed -e 's/.*Also-[bB]y:\s*\(.*\)/\1/' | sort | uniq
 ```
-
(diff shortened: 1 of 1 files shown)


akurtakov (commented):
I don't mind it in general. Only comment is about "special" thanks as that way it sounds to me like we thank people paying more than these contributing patches and I would like these to sound equally important.

laeubi:
@akurtakov I'm not a native speaker so its often hard to guess how it "sounds", can you maybe suggest a better wording? Maybe "we would like to also thank [...] for sponsoring contributions"?

akurtakov:
> @akurtakov I'm not a native speaker so its often hard to guess how it "sounds", can you maybe suggest a better wording? Maybe "we would like to also thank [...] for sponsoring contributions"?

I really like ^^^, go with it.
 Btw, I'm not native speaker too and probably majority of people that will read it and everyone will read it with his own mindset like I did.

mickaelistria:
I read it same way as @akurtakov and prefer the suggested rephrasing which is perfect and welcome.

laeubi:
I have updated the text, please leave an approval or suggest changes :-)

mickaelistria (approved):

akurtakov (approved):

laeubi:
What do you think about moving/renaming that file? I'll then open another PR for this...

mickaelistria:
> What do you think about moving/renaming that file?

I personally don't care as long as payload remains relatively easy to find (and in the Git repo).

PR #1351, opened 2022-09-16
Add --system when using BREE in compiler

mickaelistria:


akurtakov:
Looks like this broke Platform I-build https://ci.eclipse.org/releng/job/Builds/job/I-build-4.26/9/console  . At least this is my best guess as there is no other change in Equinox or Tycho to blame.

iloveeclipse:
Yes, I was also thinking about that. Can we please revert?

laeubi:
Instead of reverting it would be better to add an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue and fix the code.

Beside that, the platform-verification build with this change was successful. That builds the current Tycho and then builds the platform build (but without the tests), so it is not evident that this change alone is the culprit.

laeubi:
@akurtakov by the way, would it be possible to run platform build always with `-e` maven option so we get full stack traces in such case?

laeubi:
I ran teh aggregator locally and it succeeds, now retry with full test suite ...

laeubi:
Also with tests enabled it goes beyond that failure project ... @akurtakov can you adjust ibuild to use `-e` when invoking maven? It seems this is not set in the Jenkins file directly...

@iloveeclipse I still see exception like this even though we upgraded to latest EJC should this not be fixed now?
```
[INFO] --- tycho-compiler-plugin:3.0.0-SNAPSHOT:compile (default-compile) @ org.eclipse.jdt.core ---
[INFO] Compiling 1516 source files to git/eclipse-aggregator/eclipse.platform.releng.aggregator/eclipse.jdt.core/org.eclipse.jdt.core/target/classes
Failed to init Classpath for jar file git/eclipse-aggregator/eclipse.platform.releng.aggregator/eclipse.jdt.core/org.eclipse.jdt.core/target/jdtCompilerAdapter.jar-classes
java.nio.file.NoSuchFileException: git/eclipse-aggregator/eclipse.platform.releng.aggregator/eclipse.jdt.core/org.eclipse.jdt.core/target/jdtCompilerAdapter.jar-classes
```
The folder does not exits at that time and is created right after that call, but as I remeber we talked about that JDT might simply ignore non existing folder?

iloveeclipse:
I haven't changed that yet, so it is expected

Issue #1345, opened 2022-09-15
Resolve p2 dependencies against central to get the real GAV


laeubi:
P2 dependencies are not very useful in plain maven and are often confusing. Because of this the update-consumer-pom should be enhanced to make this configurable and omit p2 dependencies by default.

FYI: @hd42

hd42:
Even better would be to replace them with the corresponding Maven-Dependencies
<dependency>
      <groupId>p2.eclipse.plugin</groupId>
      <artifactId>org.eclipse.core.runtime</artifactId>
      <version>3.17.100.v20200203-0917</version>
      <type>eclipse-plugin</type>
      <scope>provided</scope>
      <optional>true</optional>
</dependency>

should become

<dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.core.runtime</artifactId>
      <version>3.17.100</version>
</dependency>

There might not be an easy way to generalize this for any bundle in the wild, but for official Eclipse bundles, this should be feasible, as they are available on Maven Central and a search for https://search.maven.org/search?q=a:org.eclipse.core.runtime%20AND%20v:3.17.100 (the version number without the .vYYYYMMDD-hhmm part) reveals exactly one artifact for current Eclipse versions.

laeubi:
Yes, but Tycho can't "search central" and guess what the right artifact is :-)

All artifacts that promote "correct" metadata are already replaced, platform sadly don't do this.

laeubi:
@mickaelistria are your enhancements regarding fixed group/artifact id already included in the 2022-09 release? If yes can you give an example of such an item?

mickaelistria:
> are your enhancements regarding fixed group/artifact id already included in the 2022-09 release? 

No, they're not, and this work is far from being complete anyway.

> Yes, but Tycho can't "search central" and guess what the right artifact is :-)

Why not? seach.maven.org allows to [search artifacts by checksum](https://central.sonatype.org/search/rest-api-guide/). It would be nice to use it.

laeubi:
> Why not? seach.maven.org allows to [search artifacts by checksum](https://central.sonatype.org/search/rest-api-guide/). It would be nice to use it.

okay if there is no concern I can add such an option, but this then would only work for central not sure if we maybe can have similar for nexus or other ...

laeubi:
The fix for this issue was gently sponsored by [Faktor Zehn GmbH](https://www.faktorzehn.com).

PR #1352, opened 2022-09-16
Resolve p2 dependencies against central to get the real GAV

laeubi:
Fixes https://github.com/eclipse-tycho/tycho/issues/1345
FYI @hd42

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,22 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## 3.0.0 (under development)
 
+### New option to transform P2 dependencies into real maven coordinates
+
+The `tycho-consumer-pom` mojo has a new option to resolve p2 introduced dependencies to 'real' maven coordinates now, when enabled it queries maven-central with the SHA1 of the file to find out what are the actual maven central coordinates
+ and place them in the generated pom consumer-pom.
+
+```
+<plugin>
+	<groupId>org.eclipse.tycho</groupId>
+	<artifactId>tycho-packaging-plugin</artifactId>
+	<version>${tycho-version}</version>
+	<configuration>
+		<mapP2Dependencies>true</mapP2Dependencies>
+	</configuration>
+</plugin>
+```
+
 ### New tycho-p2-plugin:dependency-tree mojo
 
 Sometimes it is useful to find out how dependencies of a project are actually pulled in. Tycho now supports a new `tycho-p2-plugin:dependency-tree` mojo that outputs a tree view of the P2 dependecies of a tycho project.
--- a/tycho-its/projects/packaging.consumer.pom/bundle/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/packaging.consumer.pom/bundle/META-INF/MANIFEST.MF
@@ -1,0 +1,8 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Plugin with API
+Bundle-SymbolicName: api-bundle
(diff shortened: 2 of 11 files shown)


RELEASE_NOTES.md
```diff
@@ -4,6 +4,22 @@ This page describes the noteworthy improvements provided by each release of Ecli
...
+The `tycho-consumer-pom` mojo has a new option to resolve p2 introduced dependencies to 'real' maven coordinates now, when enabled it queries maven-central with the SHA1 of the file to find out what are the actual maven central coordinates
+ and place them in the generated pom consumer-pom.
+
+```
+<plugin>
+	<groupId>org.eclipse.tycho</groupId>
+	<artifactId>tycho-packaging-plugin</artifactId>
+	<version>${tycho-version}</version>
+	<configuration>
+		<resolveP2Dependencies>true</resolveP2Dependencies>
```
mickaelistria, line 18: IMO, this should be `true` by default.

laeubi: If it is great and everyone loves it maybe... but I better have it disabled by default to not get complains, e.g because this do not currently uses any configured mirrors, proxies, authentication, offline mode and so on ...

And it only works with access to maven-central at the moment...

laeubi:
Having a bundle that requires

```
Require-Bundle: org.eclipse.core.runtime
```
and the current 2022-09 target results in a consumer pom with the following content:
 ```
 <dependencies>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.preferences</artifactId>
      <version>3.10.100</version>
      <type>eclipse-plugin</type>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.registry</artifactId>
      <version>3.11.200</version>
      <type>eclipse-plugin</type>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.common</artifactId>
      <version>3.16.200</version>
      <type>eclipse-plugin</type>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.eclipse.platform</groupId>
      <artifactId>org.eclipse.equinox.app</artifactId>
      <version>1.6.200</version>
      <type>eclipse-plugin</type>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
    <dependency>
      <groupId>org.osgi</groupId>
      <artifactId>org.osgi.service.prefs</artifactId>
      <version>1.1.2</version>
      <scope>compile</scope>
      <optional>false</optional>
    </dependency>
  </dependencies>
```

RELEASE_NOTES.md
mickaelistria: You're too shy ;)
I think the name isn't explicit about what it does. Basically `resolve` is not something we can use in Tycho and hope it's clear enough, as Tycho resolves tons of things in tons of contexts. Maybe something like `preferRemoteArtifactGAV` ?

laeubi: actually it is some kind of reverse-resolve but this also sounds strange :-)

laeubi:
I also added support for resolving items that carry (valid) maven `pom.properties` infos and basic offline support.. so if the build succeeds I think we can use this as a basic starting point ...

RELEASE_NOTES.md
laeubi: I think I will name this `mapP2Dependencies` that seems a better name here...

PR #1359, opened 2022-09-18
Use --system but disable --release in such a case

laeubi:
Lets see if that works better...


laeubi:
@mickaelistria this should fix the issue with platform build (verification already succeeded) and thus if the Jenkins build is done this can be merged.

PR #1361, opened 2022-09-18
Adjust and clean-up/simplify sisu-OSGi-connect

HannesWell:
This PR is split in two commits.

1. A few minor clean-ups/simplifications for the code of `sisu-osgi-connect` that I noticed while working on https://github.com/eclipse-tycho/tycho/pull/1328.

2. The path for the `connect.bundles` file is fixed in the release-notes and the path of the `connect.properties` is changed from `META-INF/sisu-connect.properties` to `META-INF/sisu/connect.properties`, to correspond to `META-INF/sisu/connect.bundles`.
@laeubi I assume that was the actual intention, wasn't it?

### DIFF
--- a/sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusFrameworkConnectServiceFactory.java
+++ b/sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusFrameworkConnectServiceFactory.java
@@ -15,10 +15,8 @@ package org.eclipse.sisu.osgi.connect;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.URL;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
@@ -26,9 +24,8 @@ import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
-import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
+import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.ServiceLoader;
@@ -44,6 +41,7 @@ import org.codehaus.plexus.logging.Logger;
 import org.codehaus.plexus.personality.plexus.lifecycle.phase.Disposable;
 import org.codehaus.plexus.personality.plexus.lifecycle.phase.Initializable;
 import org.codehaus.plexus.personality.plexus.lifecycle.phase.InitializationException;
+import org.codehaus.plexus.util.StringUtils;
 import org.eclipse.sisu.equinox.EquinoxServiceFactory;
 import org.eclipse.sisu.equinox.embedder.EquinoxLifecycleListener;
 import org.osgi.framework.Bundle;
@@ -73,7 +71,7 @@ public class PlexusFrameworkConnectServiceFactory implements Initializable, Disp
 	private Logger log;
 
 	private static final Map<ClassRealm, PlexusConnectFramework> frameworkMap = new HashMap<>();
-	
+
 	private static final Map<ClassLoader, ClassRealm> loaderMap = new HashMap<>();
 
 	@Requirement(role = EquinoxLifecycleListener.class)
@@ -109,15 +107,15 @@ public class PlexusFrameworkConnectServiceFactory implements Initializable, Disp
 		}
 		Framework foreignFramework = getForeignFramework(realm);
 		if (foreignFramework != null) {
-			PlexusConnectFramework connectFramework = new PlexusConnectFramework(foreignFramework, log, this, realm,
-					true);
-			frameworkMap.put(realm, connectFramework);
-			return connectFramework;
+			PlexusConnectFramework connectFwk = new PlexusConnectFramework(foreignFramework, log, this, realm, true);
+			frameworkMap.put(realm, connectFwk);
+			return connectFwk;
 		}
 		if (log.isDebugEnabled()) {
-			printRealm(realm, 0, new HashSet<ClassRealm>());
+			printRealm(realm, 0, new HashSet<>());
 		}
-		Collection<ClassRealm> realms = collectRealms(realm, new LinkedHashSet<ClassRealm>());
+		Collection<ClassRealm> realms = collectRealms(realm, new LinkedHashSet<>());
+
 		log.debug("Create framework for " + this + " with realm " + realm);
 		Logger fwLogger = new PlexusConnectFramework(null, log, this, realm, false);
 		Map<String, String> p = readProperties(realm, fwLogger);
@@ -129,23 +127,25 @@ public class PlexusFrameworkConnectServiceFactory implements Initializable, Disp
 		p.put(Constants.FRAMEWORK_STORAGE, storagePath + File.separator + "storage");
 		p.put(Constants.FRAMEWORK_BEGINNING_STARTLEVEL, "6");
 		p.put("osgi.instance.area", storagePath + File.separator + "instance");
-		ServiceLoader<ConnectFrameworkFactory> sl = ServiceLoader.load(ConnectFrameworkFactory.class,
-				getClass().getClassLoader());
-		ConnectFrameworkFactory factory = sl.iterator().next();
+
+		var loader = ServiceLoader.load(ConnectFrameworkFactory.class, getClass().getClassLoader());
+		ConnectFrameworkFactory factory = loader.findFirst()
+				.orElseThrow(() -> new NoSuchElementException("No ConnectFrameworkFactory found"));
+
 		PlexusModuleConnector connector = new PlexusModuleConnector(factory);
 		Framework osgiFramework = factory.newFramework(p, connector);
 		PlexusConnectFramework connectFramework = new PlexusConnectFramework(osgiFramework, log, this, realm, false);
 		PlexusFrameworkUtilHelper.registerHelper(connectFramework);
 		osgiFramework.init(connectFramework);
 		frameworkMap.put(realm, connectFramework);
+
 		connectFramework.start(osgiFramework.getBundleContext());
 		for (ClassRealm r : realms) {
 			connector.installRealm(r, osgiFramework.getBundleContext(), connectFramework);
 		}
 		osgiFramework.start();
 
-		Collection<EquinoxLifecycleListener> values = lifecycleListeners.values();
-		for (EquinoxLifecycleListener listener : values) {
+		for (EquinoxLifecycleListener listener : lifecycleListeners.values()) {
 			connectFramework.debug("Calling " + listener + "...");
 			listener.afterFrameworkStarted(connectFramework);
 		}
@@ -163,17 +163,11 @@ public class PlexusFrameworkConnectServiceFactory implements Initializable, Disp
 				printRealm(parentRealm, indent, printed);
 				indent += 2;
 			}
-			for (int j = 0; j < indent; j++) {
-				System.out.print(' ');
-			}
-			System.out.println("> " + realm.getId() + " (parent = " + realm.getParentRealm() + ")");
+			String indentation = StringUtils.repeat(" ", indent);
+			System.out.println(indentation + "> " + realm.getId() + " (parent = " + realm.getParentRealm() + ")");
 			Enumeration<URL> resources = realm.loadResourcesFromSelf(PlexusModuleConnector.MAVEN_EXTENSION_DESCRIPTOR);
-			while (resources != null && resources.hasMoreElements()) {
-				URL url = resources.nextElement();
-				for (int j = 0; j < indent; j++) {
-					System.out.print(' ');
-				}
-				System.out.println("  " + url);
+			if (resources != null) {
+				resources.asIterator().forEachRemaining(url -> System.out.println(indentation + "  " + url));
 			}
 			realm.display();
 			for (ClassRealm imports : realm.getImportRealms()) {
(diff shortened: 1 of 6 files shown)


sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusFrameworkConnectServiceFactory.java
```diff
@@ -102,22 +100,26 @@ protected String getName(ClassLoader classLoader) {
 	 * @return get (or creates) the Framework that is made of the given classloader
 	 * @throws BundleException if creation of the framework failed
 	 */
-	synchronized PlexusConnectFramework getFramework(ClassRealm realm) throws BundleException {
-		PlexusConnectFramework cachedFramework = frameworkMap.get(realm);
-		if (cachedFramework != null) {
-			return cachedFramework;
-		}
+	PlexusConnectFramework getFramework(ClassRealm realm) {
+		return frameworkMap.computeIfAbsent(realm, r -> {
```
laeubi, line 104: I don't think that computeIfabset brings any value here beside that we are forced to wrap exceptions in another exception...

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusFrameworkConnectServiceFactory.java
```diff
@@ -129,23 +131,22 @@ synchronized PlexusConnectFramework getFramework(ClassRealm realm) throws Bundle
 		p.put(Constants.FRAMEWORK_STORAGE, storagePath + File.separator + "storage");
 		p.put(Constants.FRAMEWORK_BEGINNING_STARTLEVEL, "6");
 		p.put("osgi.instance.area", storagePath + File.separator + "instance");
-		ServiceLoader<ConnectFrameworkFactory> sl = ServiceLoader.load(ConnectFrameworkFactory.class,
-				getClass().getClassLoader());
-		ConnectFrameworkFactory factory = sl.iterator().next();
+
+		var loader = ServiceLoader.load(ConnectFrameworkFactory.class, getClass().getClassLoader());
+		ConnectFrameworkFactory factory = loader.findFirst().orElseThrow();
```
laeubi, line 136: Please add a meaningful exception (with message) to be thrown here

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusFrameworkConnectServiceFactory.java
```diff
@@ -238,11 +227,11 @@ private static void printFrameworkState(Framework framework, Logger log) {
 		Comparator<Bundle> bySymbolicName = Comparator.comparing(Bundle::getSymbolicName,
 				String.CASE_INSENSITIVE_ORDER);
 		Comparator<Bundle> byState = Comparator.comparingInt(Bundle::getState);
-		Arrays.stream(bundles).sorted(byState.thenComparing(bySymbolicName)).forEachOrdered(bundle -> {
-			log.info(toBundleState(bundle.getState()) + " | " + bundle.getSymbolicName() + " (" + bundle.getVersion()
-					+ ")");
+		Arrays.stream(bundles).sorted(byState.thenComparing(bySymbolicName)).forEachOrdered(b -> {
```
laeubi, line 230: can we keep `bundle` instead of `b` as a name here?

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusModuleConnector.java
```diff
@@ -141,15 +140,14 @@ public synchronized void installRealm(ClassRealm realm, BundleContext bundleCont
 		if (installBundle(bundleContext, realmBundleName, logger) != null) {
 			installed.add(realmBundleName);
 		}
-		boolean isExtensionRealm = !realmExports.artifacts.isEmpty() || !realmExports.artifacts.isEmpty()
-				|| !realmExports.bundleStartMap.isEmpty();
+		boolean isExtensionRealm = !realmExports.artifacts.isEmpty() || !realmExports.bundleStartMap.isEmpty();
```
laeubi, line 143: Why was `!realmExports.bundleStartMap.isEmpty()` removed?

laeubi:
Yeah the path part was changed at the development phase a bit, an I'm not completely happy with the way to configure this but I think we need to get some experience with that :-)

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusModuleConnector.java
HannesWell: `!realmExports.artifacts.isEmpty()` was duplicated and therefore redundant in an or statement, because `a || a == a`.

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusFrameworkConnectServiceFactory.java
HannesWell: Sure, done.

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusFrameworkConnectServiceFactory.java
HannesWell: Done.

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusFrameworkConnectServiceFactory.java
HannesWell: In combination with using a ConcurrentHashMap it has the advantage that the method itself does not have to be synchronized and therefore a higher level of concurrency (theoretically) can be achieved.
And because the exception was wrapped into another exception in the calling method anyways we don't loose anything in this regard. The wrapping now just happens earlier.

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusModuleConnector.java
laeubi: Seems the diff has tricked me... I wonder if this is/was really a duplication or it should actually check something else...

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusFrameworkConnectServiceFactory.java
laeubi: I think the key point here is "theoretically" ;-)

Given that the startup of a framework is many decades slower than any synchronization overhead We should not optimize for no good reason here. And just because the caller (might) wrap the exception we should not wrap it here as it makes debugging and tracing harder for no gain.

HannesWell:
> Yeah the path part was changed at the development phase a bit, an I'm not completely happy with the way to configure this but I think we need to get some experience with that :-)

No problem, its a new thing and it is really hard to know from the beginning how to use it/configure it in all cases. :)

Two other things I noticed (that maybe should be addressed separately if we know we need it):
- A EquinoxServiceFactory with `hint="connect"` cannot be injected into Mojo but into a Plexus-Component, but a factory with hint `"tycho-core"` can be injected into a Mojo. I have no clue why, but at least I was not able to do that in #1328 and therefore created a separated component.
- `EquinoxServiceFactory` only launches a framework when querying a service, but in some cases one only wants to launch the framework without querying a service, because the desired class is not available as OSGi-Service and has to be instantiated directly. At least that is the case in #1328 (given that it uses/works with osgi-connect finally).
But we can discuss that in #1328.

laeubi:
> * A EquinoxServiceFactory with `hint="connect"` cannot be injected into Mojo but into a Plexus-Component, but a factory with hint `"tycho-core"` can be injected into a Mojo.

Please be carefull that Mojo + Plexus Component use different annotations (`@Component` versus `@Requirement`) and one can't mix both!

> * `EquinoxServiceFactory` only launches a framework when querying a service, but in some cases one only wants to launch the framework without querying a service, because the desired class is not available as OSGi-Service and has to be instantiated directly.

Yes, OSGi is all about services, so you should use them. Anyways you can query *any* service if you think you can do better without it even `getService(Object.class)`

HannesWell:
> > * A EquinoxServiceFactory with `hint="connect"` cannot be injected into Mojo but into a Plexus-Component, but a factory with hint `"tycho-core"` can be injected into a Mojo.
> 
> Please be carefull that Mojo + Plexus Component use different annotations (`@Component` versus `@Requirement`) and one can't mix both!

🤦🏽‍♂️ looks like I indeed blindly copied it and didn't notice `@Component` versus `@Requirement`. Thanks for the hint! That should be it.

> 
> > * `EquinoxServiceFactory` only launches a framework when querying a service, but in some cases one only wants to launch the framework without querying a service, because the desired class is not available as OSGi-Service and has to be instantiated directly.
> 
> Yes, OSGi is all about services, so you should use them. Anyways you can query _any_ service if you think you can do better without it even `getService(Object.class)`

In the mentioned case there was simply no useful service that I could query, but yes I just queried some unused service to launch the osgi-instance. But using Object probably indicates the intend of not using it better.

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusModuleConnector.java
HannesWell: I asked myself the same question as well, but I think only you can answer that. :)

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusFrameworkConnectServiceFactory.java
HannesWell: > I think the key point here is "theoretically" ;-)
> 
> Given that the startup of a framework is many decades slower than any synchronization overhead We should not optimize for no good reason here. And just because the caller (might) wrap the exception we should not wrap it here as it makes debugging and tracing harder for no gain.

It was not about the synchronization overhead itself, it was about that in a parallel build, again "theoretically" :D , multiple framework instances could be launched in parallel if different plug-ins request that.
But if you think this is too theoreticall and would better be done with a 'old'-style get-in-case-of-null-compute-and-put I can revert that.

laeubi:
> In the mentioned case there was simply no useful service that I could query

Then probably there should be one!

Just keep in mind that the whole purpose of connect is to interact with the OSGi framework (otherwise you might better instantiate the classes directly without any OSGi), so you can ave you a lot of trouble if there is a service you can call and that do the dirty bits without you knowing. I also used something similar anyways but more as a legacy way, so you should be able to directly use any class up to a point where one service will be required. And yes Eclipse is nasty in its usage of static activator access... a good opportunity to fix that and using real service/ds instead :-)

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusFrameworkConnectServiceFactory.java
laeubi: I have done it here for different reasons:

1. checked exceptions
2. easier to debug
3. better lock the object (e.g. for maybe other methods as well)
4. code might "call back" while the init state and with `frameworkMap.put(realm, connectFramework);` I can control when this framework will become visible (in this case before `start` was called), while with a concurrentmap is is only visible after the method returns.

so yes I think we better revert this without a strong usecase / analysis, beside that I always use concurrent map if possible :-)

HannesWell: >     4. code might "call back" while the init state and with `frameworkMap.put(realm, connectFramework);` I can control when this framework will become visible (in this case before `start` was called), while with a concurrentmap is is only visible after the method returns.

I wonder why one wants to see/use the framework instance before it is completely set up?
Nevertheless I reverted that part.

HannesWell:
> > In the mentioned case there was simply no useful service that I could query
> 
> Then probably there should be one!
> 
> Just keep in mind that the whole purpose of connect is to interact with the OSGi framework (otherwise you might better instantiate the classes directly without any OSGi), so you can ave you a lot of trouble if there is a service you can call and that do the dirty bits without you knowing. I also used something similar anyways but more as a legacy way, so you should be able to directly use any class up to a point where one service will be required. And yes Eclipse is nasty in its usage of static activator access... a good opportunity to fix that and using real service/ds instead :-)

Fully agree, but that is a huge task since that pattern is used very often in the Eclipse eco-system.

sisu-osgi/sisu-osgi-connect/src/main/java/org/eclipse/sisu/osgi/connect/PlexusFrameworkConnectServiceFactory.java
laeubi: There are two cases:

1. Code using FrameworkUtil.getBundle(class) that can run as we start the framework
2. EquinoxFrameworklisteners calling methods

It might be possible to call start in a lazy way to prevent this but I think we don't need to complicate this even more :-)

laeubi:
> Fully agree, but that is a huge task since that pattern is used very often in the Eclipse eco-system.

Every little piece counts :-) As DS now days even require less setup I don't see a reason to not use it ...

HannesWell:
> > Fully agree, but that is a huge task since that pattern is used very often in the Eclipse eco-system.
> 
> Every little piece counts :-) As DS now days even require less setup I don't see a reason to not use it ...

That's right. I just meant it is too much to wait for it now. :)

Besides that, do you have any more remarks? If not I think this is ready.

laeubi (approved):
Sounds reasonable!

PR #1362, opened 2022-09-18
More clean-ups and fixes of baseline comparing Mojos

HannesWell:
After https://github.com/eclipse-tycho/tycho/pull/1327 has been merged, the results of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/564 are not as expected.
This PR adds some debug output for testing and tries to build the platform with it.

### DIFF
--- a/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/ManifestComparator.java
+++ b/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/ManifestComparator.java
@@ -79,7 +79,7 @@ public class ManifestComparator implements ContentsComparator {
         return !result.isEmpty() ? new CompoundArtifactDelta("different", result) : null;
     }
 
-    protected void addDelta(TreeMap<String, ArtifactDelta> result, Name key, String message) {
+    private void addDelta(TreeMap<String, ArtifactDelta> result, Name key, String message) {
         result.put(key.toString(), new SimpleArtifactDelta(message));
     }
 
--- a/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/NestedZipComparator.java
+++ b/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/NestedZipComparator.java
@@ -16,6 +16,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 
 import org.codehaus.plexus.component.annotations.Component;
 import org.codehaus.plexus.component.annotations.Requirement;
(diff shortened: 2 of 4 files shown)


.github/workflows/verify-platform.yml
```diff
@@ -24,7 +24,8 @@ jobs:
        path: 'platform'
        submodules: 'recursive'
        lfs: true
-       repository: 'eclipse-platform/eclipse.platform.releng.aggregator'
+       repository: 'HannesWell/eclipse.platform.releng.aggregator'
```
laeubi, line 27: Is this intended?

.github/workflows/verify-platform.yml
```diff
@@ -61,7 +62,8 @@ jobs:
        JAVA_HOME: ${{ env.JAVA_HOME_17_X64 }}
       working-directory: 'platform'
       run: >- 
-        mvn -U -V -e -B -ntp
+        mvn -U -V -e -B -ntp -X
```
laeubi, line 65: Will this be reverted or is this by intention? The verification here should really not fail for any baseline comparison as we never change any code.

.github/workflows/verify-platform.yml
HannesWell: It is an intended change for testing this PR, but I will revert that before merging this.

.github/workflows/verify-platform.yml
HannesWell: Same as above.

HannesWell:
There was indeed a regression in #1327.
In contrast to `FileUtils.copyStreamToFile()` the method `Files.copy()` fails if the target file already exists, unless REPLACE_EXISTING is passed. And `Files.createTempFile()` returns the path of an existing file. So without passing `REPLACE_EXISTING` the comparison of nested zip cannot work.

HannesWell:
Besides that I think if the zip comparison failed with an exception a warning should always be printed displaying the exception and not only a simple delta without context should be presented.
Without that it is hard to tell why there is a difference and an IOexception is actually always not acceptable.

.github/workflows/verify-platform.yml
HannesWell: > It is an intended change for testing this PR, but I will revert that before merging this.

Done.

mickaelistria:
> the comparison of nested zip cannot work.

I believe Eclipse Platform build is badly affected, see https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-183/2/console . Can we have a remediation ASAP?

HannesWell:
> > the comparison of nested zip cannot work.
> 
> I believe Eclipse Platform build is badly affected, see https://ci.eclipse.org/platform/job/eclipse.platform/job/PR-183/2/console . Can we have a remediation ASAP?

Yep merged it, which was the plan anyway. :)

mickaelistria:
thank you!

laeubi:
@akurtakov @mickaelistria given all these "can we get ASAP" demands, I really like to ask if we can get at least more **COMPUTER POWER** resources for Tycho because that currently stalls our development a lot!

I have opened https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1955

But also @merks I really would appreciate to take this into account for any "funding" discussions, its really a pitty that we are so limited here and have static assignments of build nodes really seems ancient given today "cloud computing" capabilities.

I have even sponsored an own build node already https://ci.eclipse.org/tycho/computer/laeubisoft%2Dbuild%2Dnode/ but still if a lot of changes "rush in" that's not enough and feels a bit odd that one has to contribute development AND processing power. But without that, we have had a backlogs that take half a day to be processed in the past, this does not fit the "we need fixes ASAP from Tycho" demands...

PR #1366, opened 2022-09-19
Bump

mickaelistria:


laeubi:
I think we should first bump the target (that could be merged anyways) and then let dependabot rebase it ... but as mentioned in the other PR this is a regression in equinox... so I see no real way to upgrade unless it is fixed.

mickaelistria:
Yes, I'll change that; it was more a desperate attempt, I didn't place much hope there.

PR #1370, opened 2022-09-20
Instead of directly use the implementation make methods available through the interface

laeubi:


akurtakov:
I like that one a lot. There are so many things like that that just go unnoticed for years.

laeubi:
This has really bugged me several times while refactoring so I finally like to get rid of this ;-)

Beside that, it would make it possible to  now convert this into a reusable plexus component.

PR #1378, opened 2022-09-20
Remove the ECF Transport option

laeubi:
The Tycho Transport is the default for a while now and we got no complains so far, this change now removes the ECF transport option completely.


akurtakov:
Does it mean we can remove ecf from target platform and everywhere else used?

laeubi:
> Does it mean we can remove ecf from target platform and everywhere else used?

If this build succeeds and got merged we can try to do so I'm just not sure if the transport is already used everywhere (what will be one of the next steps to move this out of OSGi into Maven part).

akurtakov:
> > Does it mean we can remove ecf from target platform and everywhere else used?
> 
> If this build succeeds and got merged we can try to do so I'm just not sure if the transport is already used everywhere (what will be one of the next steps to move this out of OSGi into Maven part).

OK, thanks for the info.

Issue #1388, opened 2022-09-21
Cannot invoke "org.apache.maven.project.MavenProject.getRemoteArtifactRepositories()" because "project" is null


pzi-dsa:
This problem showed up after updating our project to use the Eclipse 4.25 and is causing our builds to fail.
For some reason, Maven attempts to resolve some of our 3rd party artifacts against artifacts of the current reactor build, failing with a NPE. The project in question seems to be `p2.eclipse.plugin:org.bouncycastle.bcprov:eclipse-plugin:1.70.0.v20220507-1208:provided`.

```
Caused by: java.lang.NullPointerException: Cannot invoke "org.apache.maven.project.MavenProject.getRemoteArtifactRepositories()" because "project" is null
    at org.apache.maven.project.artifact.MavenMetadataSource.retrieve (MavenMetadataSource.java:200)
    at org.apache.maven.repository.legacy.resolver.DefaultLegacyArtifactCollector.recurse (DefaultLegacyArtifactCollector.java:551)
    at org.apache.maven.repository.legacy.resolver.DefaultLegacyArtifactCollector.collect (DefaultLegacyArtifactCollector.java:148)
    at org.apache.maven.artifact.resolver.DefaultArtifactResolver.resolve (DefaultArtifactResolver.java:504)
    at org.apache.maven.repository.legacy.LegacyRepositorySystem.resolve (LegacyRepositorySystem.java:367)
    at org.eclipse.tycho.osgi.configuration.MavenDependenciesResolverConfigurer.resolve (MavenDependenciesResolverConfigurer.java:106)
    at org.eclipse.tycho.core.shared.MavenDependenciesResolver.resolve (MavenDependenciesResolver.java:58)
    at org.eclipse.tycho.p2.resolver.MavenTargetDefinitionContent.<init> (MavenTargetDefinitionContent.java:135)
    at org.eclipse.tycho.p2.target.TargetDefinitionResolver.resolveContentWithExceptions (TargetDefinitionResolver.java:186)
    at org.eclipse.tycho.p2.target.TargetDefinitionResolver.resolveContent (TargetDefinitionResolver.java:110)
    at org.eclipse.tycho.p2.target.TargetDefinitionResolverService.resolveFromArguments (TargetDefinitionResolverService.java:86)
    at java.util.concurrent.ConcurrentHashMap.computeIfAbsent (ConcurrentHashMap.java:1708)
[...]

PR #1389, opened 2022-09-21
Add reproducer for issue #1388.

pzi-dsa:
The issue here seems to be caused by the 'provided' dependencies of the org.eclipse.egit.core [pom](https://repo.eclipse.org/content/groups/releases/org/eclipse/egit/org.eclipse.egit.core/6.3.0.202209071007-r/org.eclipse.egit.core-6.3.0.202209071007-r.pom).


laeubi:
Thanks for the example. This is actually a bug in maven, where it assumes that for each model obtained from an `MavenWorkspaceReader` there is a corresponding reactor project in the map. @pzi-dsa do you mind to report this to maven? I'll see if I can find a workaround (at worst needs a try/catch) here...

laeubi:
This is now included here: https://github.com/eclipse-tycho/tycho/pull/1392

pzi-dsa:
Thank you for further looking into this :)

PR #1398, opened 2022-09-21
Pin bcprov

laeubi:


mickaelistria (changes requested):
IMO, it's much saner to wait for approval (hopefully it wont last too long) than to start incorporating hacks of that form.

laeubi:
> IMO, it's much saner to wait for approval (hopefully it wont last too long) than to start incorporating hacks of that form.

This is not a hack, I just choose the wrong base branch... my fault.

laeubi:
By the way the review is already due, I asked two days ago, no reaction, so this is simply to late for 3.0.0

https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/3286

laeubi:
I would be fine delay this until Monday, but still this PR can prove that it works just in case we don't get feedback. Actually we do not have any functional requirement on that version.

laeubi:
@akurtakov @mickaelistria so its monday an no response .. so I think we should merge this to get IP warnings fixed and hope for 3.1.x?

akurtakov:
I'm fine with that.

mickaelistria:
OK.

Issue #1377, opened 2022-09-20
[regression] tycho-buildtimestamp-jgit impact with pomDependencies=consider


briandealwis:
I've hit a subtle bug in Tycho that was introduced in 2.7.0 and is present in 3.0.0-SNAPSHOT.

Basically a project that:
- uses `pomDependencies=consider`,
- uses the tycho-buildtimestamp-jgit provider
- includes a Maven dependency with OSGi metadata, like Guava,
- and has a feature that includes this Maven dependency
will fail with a resolving error on the Maven dependency.

I've created an demonstration project in https://github.com/briandealwis/tycho/tree/jgit-buildtimestamp-bug/tycho-its/projects/pomDependencyConsider.buildtimestamp.jgit

- the demonstration works fine with Tycho 2.6.0
- the demonstration fails using Tycho 2.7.5 with tycho-buildtimestamp-jgit 2.6.0

```
$ mvn package -e
[INFO] Error stacktraces are turned on.
[INFO] Scanning for projects...
[WARNING] No explicit target runtime environment configuration. Build is platform dependent.
[INFO] ### Using TychoRepositoryTransport for remote P2 access (You can disable this with -Dtycho.p2.transport=ecf) ###
[INFO]     Cache location:         /usr/local/google/home/bdealwis/.m2/repository/.cache/tycho
[INFO]     Transport mode:         online
[INFO]     Update mode:            cache first
[INFO]     Minimum cache duration: 60 minutes
[INFO]       (you can configure this with -Dtycho.p2.transport.min-cache-minutes=<desired minimum cache duration>)
[INFO] Resolving dependencies of MavenProject: pomDependencyConsider.buildtimestamp.jgit:feature:0.0.1-SNAPSHOT @ /usr/local/google/home/bdealwis/Projects/CTE/repo-tycho/tycho-its/projects/pomDependencyConsider.buildtimestamp.jgit/feature/pom.xml
[INFO] The following requirements are not satisfied yet and must be provided through pom dependencies:
[INFO]    - org.eclipse.equinox.p2.iu; com.google.guava 0.0.0
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Build Order:
[INFO] 
[INFO] parent                                                             [pom]
[INFO] feature                                                [eclipse-feature]
[INFO] 
[INFO] ----------< pomDependencyConsider.buildtimestamp.jgit:parent >----------
[...]

PR #1400, opened 2022-09-21
Add integration test to for jgit buildtimestamp provider bug

briandealwis:
Re: #1377


laeubi:
@briandealwis thanks for the test-case, this is now included in https://github.com/eclipse-tycho/tycho/pull/1405

PR #1406, opened 2022-09-22
Use some instanceof patterns

mickaelistria:


laeubi:
@mickaelistria can we hold back such changes until 3.0.0 release (Fr next week) to get smaller change sets in case of regression fixes?

mickaelistria:
Sure, I mostly submitted it to see how CI likes it; it's not meant for 3.0 at all.

laeubi:
@mickaelistria as no one has reported major regression until now I think its fine to merge these when the test/ci is fine ...

mickaelistria:
Merged as part of #1425

PR #1407, opened 2022-09-22
Reproducer for generate-sources issue

lppedd:
Command from root project directory: `mvn clean generate-sources`
(**do not** `install` before)

Expected issue output:
```
[INFO] Reactor Summary for repro-parent 1.0.0-SNAPSHOT:
[INFO]
[INFO] repro-parent ....................................... SUCCESS [  0.059 s]
[INFO] a .................................................. SUCCESS [  3.354 s]
[INFO] b .................................................. FAILURE [  0.005 s]
[INFO] BUILD FAILURE
[INFO] Total time:  58.727 s
[INFO] Finished at: 2022-09-22T11:20:47+02:00
[ERROR] Failed to execute goal on project b: Could not resolve dependencies for project com.lppedd.tycho.repro:b:eclipse-plugin:1.0.0-SNAPSHOT: Could not find artifact com.lppedd.tycho.repro:a:jar:1.0.0-SNAPSHOT -> [Help 1]    
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException
[ERROR]
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :b
```

This is a follow up for [this comment](https://github.com/eclipse-tycho/tycho/pull/1323#issuecomment-1254142927).

### DIFF
--- a/tycho-its/projects/resolver.generateSources/a/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/resolver.generateSources/a/META-INF/MANIFEST.MF
@@ -1,0 +1,9 @@
+Manifest-Version: 1.0
+Automatic-Module-Name: a
+Bundle-ManifestVersion: 2
+Bundle-Name: a
+Bundle-Vendor: My Company
+Bundle-Version: 1.0.0.qualifier
+Bundle-SymbolicName: a; singleton:=true
+Bundle-ActivationPolicy: lazy
+Bundle-RequiredExecutionEnvironment: JavaSE-11
--- a/tycho-its/projects/resolver.generateSources/a/build.properties
+++ b/tycho-its/projects/resolver.generateSources/a/build.properties
@@ -1,0 +1,3 @@
+source.. = src/
+bin.includes = .,\
+               META-INF/
--- a/tycho-its/projects/resolver.generateSources/a/pom.xml
+++ b/tycho-its/projects/resolver.generateSources/a/pom.xml
@@ -1,0 +1,13 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>com.lppedd.tycho.repro</groupId>
+		<artifactId>repro-parent</artifactId>
+		<version>1.0.0-SNAPSHOT</version>
+	</parent>
+
+	<artifactId>a</artifactId>
+	<packaging>eclipse-plugin</packaging>
+</project>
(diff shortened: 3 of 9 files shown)


lppedd:
Btw @laeubi this seems to be independent of any plugin. You can remove `gmaven` and it will error out anyway.

laeubi:
@lppedd can you check the ECA problem? Please make sure to push your commit with the email you used to sign the ECA!

lppedd:
@laeubi I will, in the meantime I'm using a Maven extension to forcefully resolve dependencies.
See https://stackoverflow.com/questions/73840829/maven-extension-to-forcibly-resolve-a-dependency-how-dangerous-is-it
Maybe that's the only way?

laeubi:
So this is more a generic maven problem than a Tycho one?

laeubi:
@lppedd you probably want to push forward the following tickets at maven what most probably would allow to resolve this maven issue: https://issues.apache.org/jira/browse/MNG-7538 because that would allow to tell maven that there is something but the final name is (not yet) known.

lppedd:
@laeubi yes, I think this is a Maven core issue. Probably we can discard this PR, but if you find 5 free minutes just test the reproducer yourself, you'll notice this is a pretty important issue (at least that's my thought). The issue you linked seems related but I read

> that is used while resolving an Artifact against a remote repository

In the reproducer the failure is not related to a remote artifact, but to a local one only.

laeubi:
> In the reproducer the failure is not related to a remote artifact, but to a local one only.

Yeah its a bit unspecific, actually, as you have noted, when maven tries to resolves an artifact it do so by asking the WSR, if that can't give the file it asks the remote (that probably download the file then), but in both cases (either download or not (yet) generated local item) it would simply help to say: Hey I know here is something, and if you need it I can try to gather it then.

> Probably we can discard this PR

If you like, we can `@Ignore` the testcase and give in the documentation the links to maven issue that need to be addressed so we can enable it once the maven issues are fixed (I think WSR also needs a new method that returns a `Supplier<File>`) we can check with Tycho again...

laeubi:
@lppedd just noted Tycho already contains some hacks in that area:

https://github.com/eclipse-tycho/tycho/blob/0f790c7c2498f6de0344d5ad1f7c0cc685c2ef2f/tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java#L398

lppedd:
@laeubi 

> If you like, we can @Ignore the testcase

Will make this change ASAP, and check the legal stuff (it's strange, I remember submitting it through the Eclipse site).

> just noted Tycho already contains some hacks in that area

How does that trick compare to the WSR one?

laeubi:
It creates a temp file to trick maven that it things the artifact is resolved...

lppedd:
This one can be closed as you pushed it alongside #2035

Issue #1411, opened 2022-09-23
New Regression in 3.0.0


cdietrich:
https://ci.eclipse.org/mwe/job/mwe2/job/cd_tycho30/234/console

```
07:04:36  [ERROR] Failed to execute goal org.eclipse.tycho:tycho-packaging-plugin:3.0.0-SNAPSHOT:package-feature (default-package-feature) on project org.eclipse.emf.mwe.core.feature: Execution default-package-feature of goal org.eclipse.tycho:tycho-packaging-plugin:3.0.0-SNAPSHOT:package-feature failed: eclipse-plugin artifact with ID "org.eclipse.emf.mwe.core" and version matching "0.0.0" was not found in the target platform -> [Help 1]
```

any idea what has changed from yesterday where it still was working?
https://ci.eclipse.org/mwe/job/mwe2/job/cd_tycho30/232/console

laeubi:
Can you run with `-e` option to get a full stack trace?

All recent changes on 3.0.0 here:
https://github.com/eclipse-tycho/tycho/commits/tycho-3.0.x

as always provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue would help alot!

cdietrich:
05:46:57  [INFO] org.eclipse.emf.mwe2.parent 2.14.0-SNAPSHOT ........ SUCCESS [  1.997 s]
05:46:57  [INFO] org.eclipse.emf.mwe2.target 2.14.0-SNAPSHOT ........ SUCCESS [  0.255 s]
05:46:57  [INFO] org.eclipse.emf.mwe2.plugins-parent 2.14.0-SNAPSHOT  SUCCESS [  2.888 s]
05:46:57  [INFO] org.eclipse.emf.mwe.activities 1.8.0-SNAPSHOT ...... SUCCESS [  4.146 s]
05:46:57  [INFO] org.eclipse.emf.mwe2.runtime 2.14.0-SNAPSHOT ....... SUCCESS [  3.282 s]
05:46:57  [INFO] org.eclipse.emf.mwe.core 1.8.0-SNAPSHOT ............ SUCCESS [  3.536 s]
05:46:57  [INFO] org.eclipse.emf.mwe.ui 1.8.0-SNAPSHOT .............. SUCCESS [  3.735 s]
[...]

PR #1413, opened 2022-09-23
Add a verification build for mwe

laeubi:
FYI @cdietrich maybe you can help here?


laeubi:
> please note: this one is mwe. xtext main build with tycho3 are

Alright, I renamed this, maybe this can still serve as an inspiration :-)

Would the used commandlien be suitable to reproduce https://github.com/eclipse-tycho/tycho/issues/1411 ?

laeubi:
@cdietrich I'll close this for now, maybe you can think about if it would be usefull if we add xtext/mwe to the verification suite.

PR #1419, opened 2022-09-26
Move the MetadataSerializable from OSGi > Maven

laeubi:


laeubi:
@akurtakov I try to move a test from "OSGi World" to "Maven World" and would like to have the tycho maven components here, but it seems I need to adjust something as the test already fails with:

>  No implementation for org.apache.maven.repository.RepositorySystem was bound.

before my actual test code even starts. Any hints?

akurtakov:
So you say that the tests are failing even before your change?

laeubi:
No the test are failing here:

https://github.com/eclipse-tycho/tycho/blob/916c7ad23beef52ffeb3debcc0480bb616994f30/tycho-p2/tycho-p2-facade/src/test/java/org/eclipse/tycho/p2/facade/test/MetadataSerializableImplTest.java#L44

so before I can do anything, so I suspect my setup of running a `AbstractMojoTestCase` is missing something (maybe a dependency to maven-???)

akurtakov:
Ah, I got it. AbstractMojoTestCase needs maven-compat as a test dependency to work.

laeubi:
> AbstractMojoTestCase needs maven-compat as a test dependency to work

Is there an "non compat" alternative? I must confess I have assumed writing a "maven test" would be rather standard and easy but it seems its not :-\

akurtakov:
Not that I know.

laeubi:
> Not that I know.

Alright I'll ask on the mailinglist then, it seems the test is working now!

PR #1397, opened 2022-09-21
EclipseRunMojo: Respect <executionEnvironment> configuration

sratz:
This got lost in 0eeb95d1e58f3c93ced88dfb0f6a76da33d9074d.

Fixes #1396.

PR #1421, opened 2022-09-26
EclipseRunMojo: Add unit tests for <executionEnvironment>

sratz:
Follow-up on #1397.


laeubi:
@sratz many thanks for the testcase!

sratz:
I just noticed a small issue in the test which could let some issue slip through unnoticed. See #1426

Issue #1437, opened 2022-09-29
Tycho Setup seems outdated


laeubi:
I tried to use the Tycho setup, but Oomph refuses to install, some settings seem outdated. @HannesWell can you take a look? do we need to updates the setup e.g. for new m2e 2.x? The setupa also mention "gerrit" on some places what seems wrong.

HannesWell:
There are two connectors that are incompatible with M2E-2.0 in the setup. I removed them and their p2-repos (and one orphan) via #1439.
I also searched for "Gerrit" but did not find that string in the setup?

PR #1439, opened 2022-09-29
Remove Connectors incompatible with M2E-2.0 from Tycho's Oomph-setup

HannesWell:
Fixes https://github.com/eclipse-tycho/tycho/issues/1437


HannesWell:
Canceled the builds since the Oomph setup is not checked.

PR #1441, opened 2022-09-30
Using Tycho 3.0.0 as the bootstrap version.

laeubi:


laeubi:
FYI @akurtakov what recently came into my mind, once we are done with completely migrating to the connect-framework and moving Tycho stuff out of OSGi we actually won't need any Tycho "bootstrap" stuff anymore to build Tycho itself :-)

akurtakov:
Wow, that be awesome achievement.

PR #1446, opened 2022-10-01
Move the publisher tools to maven

laeubi:


laeubi:
Needs more smaller steps...

Issue #24, opened 2021-04-09
Create a installer button to instantly start coding


laeubi:
Like with eclipse-m2e/m2e-core#120 but as we already have a setup file maybe just some additional wiring is required?

laeubi:
@HannesWell if I remeber correctly you have done this for m2e already, do you think you can propose such a button for Tycho as well?

Here is a "clone into eclipse" button but it points to @mickaelistria repository (can we adjust this to use the tycho gh pages now?):

https://github.com/eclipse-tycho/tycho/blob/master/CONTRIBUTING.md

but m2e seems to have d dedicated button for the setup as well.

mickaelistria:
> Here is a "clone into eclipse" button but it points to @mickaelistria repository

You'll need a proxy as GitHub doesn't allow non-https links. My repo is the proxy that just does a redirect. You can put this proxy someplace else, but I don't know what you'd gain from it.

HannesWell:
> @HannesWell if I remeber correctly you have done this for m2e already, do you think you can propose such a button for Tycho as well?

Here we go: https://github.com/eclipse-tycho/tycho/pull/1459 :)

PR #1459, opened 2022-10-03
[Setup] Add Oomph-configuration setup for Tycho

HannesWell:
Fixes https://github.com/eclipse-tycho/tycho/issues/24

Based on the Oomph Config for M2E I created one for Tycho and added a Button  to use it to the Contribution guide.
The button image will be created at https://download.eclipse.org/oomph/www/setups/svg/ when the setup exists, which is checked by a Oomph build job once a day (IIRC), but I can check that again.

Furthermore we can add an icon for Tycho that is displayed with the description if one opens the config. For M2E I reused the IDE icon, but for Tycho I didn't found any suitable icon. So if somebody can suggest one, I can reference it (it has to be somewhere on the internet) but we can probably also leaf that blank.

@laeubi what do you think about the descriptions?

We could also add the Button more prominent in the main READ.ME, like M2E does it?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,10 @@ and the minimal reproducer project to Tycho's [issue tracker](https://github.com
 
 ## Development environment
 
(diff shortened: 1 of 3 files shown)


laeubi:
I would first add the stuff and then we can maybe move stuff around, on the other hand the CONTRIBUTING.MD is recommended when you create issues ... Tycho has no dedicated icon.

PR #1461, opened 2022-10-04
Fix broken equals in StandardEEResolutionHints

laeubi:
Regression from e261bbe1889964c5a0e60083952521e6a42f2567

This causes the target platform to be resolved over and over again, FYI @mickaelistria 

By the way this shows that less lines do not imply less errors / better code, this was really hard to debug as the method looks so innocent and obvious here...


mickaelistria:
Thanks for catching!

> By the way this shows that less lines do not imply less errors / better code

It's almost always true that less code implies less errors and often better code. The issue is not the result, it's the path towards it: every change is a risk of introducing an error, even changes that leads to expected improvements.

PR #1470, opened 2022-10-05
Improve javadoc of DeclarativeServicesMojo

### LINKED ISSUE #1468: Clarify meaning of https://tycho.eclipseprojects.io/doc/3.0.0/tycho-ds-plugin/declarative-services-mojo.html#enabled
According to https://tycho.eclipseprojects.io/doc/3.0.0/tycho-ds-plugin/declarative-services-mojo.html#enabled there is a plugin parameter called `enabled`. But the description is very vague and in fact the according field doesn't seem to be evaluated: https://github.com/eclipse-tycho/tycho/blob/0f790c7c2498f6de0344d5ad1f7c0cc685c2ef2f/tycho-ds-plugin/src/main/java/org/eclipse/tycho/ds/DeclarativeServicesMojo.java#L65.

Is this used anywhere via reflection or can this private field just be removed?
It seems redundant as there is already https://github.com/eclipse-tycho/tycho/blob/0f790c7c2498f6de0344d5ad1f7c0cc685c2ef2f/tycho-ds-plugin/src/main/java/org/eclipse/tycho/ds/DeclarativeServicesMojo.java#L71.

kwin:
This closes #1468

### DIFF
--- a/tycho-ds-plugin/src/main/java/org/eclipse/tycho/ds/DeclarativeServicesMojo.java
+++ b/tycho-ds-plugin/src/main/java/org/eclipse/tycho/ds/DeclarativeServicesMojo.java
@@ -39,8 +39,12 @@ import aQute.bnd.osgi.Jar;
 import aQute.bnd.osgi.Resource;
 
 /**
- * This mojo could be added to a build if validation of the classpath is desired
- * before the compile-phase.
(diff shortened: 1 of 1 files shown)


laeubi:
Thanks for the improvements!

PR #1472, opened 2022-10-06
Add option to replace packa-type by extension for the consumer pom

laeubi:


akurtakov:
How would it differ between eclipse-plugin and eclipse-test-plugin?

laeubi:
> How would it differ between eclipse-plugin and eclipse-test-plugin?

Technically `eclipse-test-plugin` and `eclipse-plugin` are just the same, both produce a jar-file... so for the case of using it as a maven dependency that simply don't matter at all.

mickaelistria:
I agree with @laeubi , the consumer pom doesn't care about eclipse-plugin or eclipse-test-plugin; in practice they're not really *artifact* types but more lifecycle types; in the end resulting in a regular jar artifact from consumer POV.

PR #1474, opened 2022-10-06
Clarify allowed values for parameter "dsVersion"

### LINKED ISSUE #1473: Clarify values for "declarative-services" mojo parameter "dsVersion"
It seems that the parameter https://github.com/eclipse-tycho/tycho/blob/663a1e16aaa77854823c3f825373b01b5d1fd25c/tycho-ds-plugin/src/main/java/org/eclipse/tycho/ds/DeclarativeServicesMojo.java#L62 is not passed to Bnd at all. On the other hand the version may be useful to set in the context of https://github.com/eclipse-m2e/m2e-core/pull/967. PDE only supports the following values though: https://github.com/eclipse-pde/eclipse.pde/blob/8312046eb6c566df0846f4125c2a5d41dcf0d0a3/ds/org.eclipse.pde.ds.annotations/src/org/eclipse/pde/ds/internal/annotations/DSAnnotationVersion.java.

Please clarify the intent of this parameter and outline possible values.

kwin:
Obey version configuration in Bnd

This closes #1473

### DIFF
--- a/tycho-ds-plugin/src/main/java/org/eclipse/tycho/ds/DeclarativeServicesMojo.java
+++ b/tycho-ds-plugin/src/main/java/org/eclipse/tycho/ds/DeclarativeServicesMojo.java
@@ -35,6 +35,7 @@ import org.eclipse.tycho.core.osgitools.OsgiBundleProject;
 
 import aQute.bnd.component.DSAnnotations;
 import aQute.bnd.osgi.Analyzer;
+import aQute.bnd.osgi.Constants;
 import aQute.bnd.osgi.Jar;
 import aQute.bnd.osgi.Resource;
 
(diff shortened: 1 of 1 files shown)


laeubi:
@kwin do you think you can enhance  org.eclipse.tycho.test.compiler.CompilerClasspathEntryTest for the new support e.g. that it fails when using 1.3 options but setting is 1.2? You can use CompilerClasspathEntryTest.testDSComponents() as an example

kwin:
I am not sure what this IT is supposed to test. I added a commit now to make sure errors in Bnd lead to build errors and all warnings all logged. Please take it up from here and add more ITs.

laeubi:
@kwin an integration test could be a small tycho project where you would expect it fail now with your change where previously Tycho would just compile (I suppose this is the main goal here?)

Another option would be to inspect the resulting xml that it uses the configured DS version.

PR #1477, opened 2022-10-06
Delete corrupted jars and download them again

laeubi:


laeubi:
If you look at these PR: https://github.com/eclipse-tycho/tycho/pull/1476 there are two "failing" test because a jar can't be read (I assume it is corrupted and now cached somewhere), so the goal is in such a case delete the file and trigger download again. I already fixed something similar somewhere else and seems here is another such case where it can happen.

mickaelistria:
Please add such information in the commit message.

mickaelistria (approved):
Thanks

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
```diff
@@ -204,6 +205,31 @@ private MavenProject getMavenProject(ReactorProject reactorProject) {
...
+            logger.warn("First attempt failed, deleting file " + location + " :: " + e);
+            //something is wrong, delete the file and try to fetch it again!
+            if (!location.delete()) {
+                logger.warn("File can't be deleted!");
+            }
+            location = descriptor.getLocation(true);
+            if (location == null) {
+                return Optional.empty();
+            }
+            if (!location.exists()) {
```
kwin, line 225: Rather https://docs.oracle.com/javase/7/docs/api/java/io/File.html#isFile() and the message should probably explain the most common cause:
- Path is a directory or
- File is not accessible to the current user

laeubi:
Okay debug output is:
```
[WARNING] First attempt failed, deleting file C:\Users\runneradmin\.m2\repository\p2\osgi\bundle\org.eclipse.equinox.launcher.gtk.linux.x86\1.1.551.v20171108-1834\org.eclipse.equinox.launcher.gtk.linux.x86-1.1.551.v20171108-1834.jar :: org.eclipse.tycho.core.osgitools.OsgiManifestParserException: Exception parsing OSGi MANIFEST C:\Users\runneradmin\.m2\repository\p2\osgi\bundle\org.eclipse.equinox.launcher.gtk.linux.x86\1.1.551.v20171108-1834\org.eclipse.equinox.launcher.gtk.linux.x86-1.1.551.v20171108-1834.jar: zip file is empty
[WARNING] FIle is not null but do not exits! C:\Users\runneradmin\.m2\repository\p2\osgi\bundle\org.eclipse.equinox.launcher.gtk.linux.x86\1.1.551.v20171108-1834\org.eclipse.equinox.launcher.gtk.linux.x86-1.1.551.v20171108-1834.jar
```
So it seems for some reason Tycho thinks that the file is already downloaded and just returns a file pointer to a non existing file ...

laeubi:
It seems somehow related to the windows execution only, all attempts to verify an re-download seem to result in the same problem (beside that I was able to fail the checksum validation on Jenkins...)

PR #1488, opened 2022-10-07
ApiFileGenerationMojo: Add missing parameters to sync with ant task

sratz:
Offer the same parameters that the ant task [1] does.

[1] https://help.eclipse.org/latest/topic/org.eclipse.pde.doc.user/reference/api-tooling/ant-tasks/filegeneration-ant-task.htm?cp=4_3_0_0_11


laeubi:
@sratz thanks for this, are you using the mojo in your builds?

sratz:
We are using the ant task to generate .api-description and I wanted to migrate to tycho-apitools-plugin of Tycho 3.0.0, but couldn't, because we are using some of those additional parameters.

Issue #436, opened 2021-12-14
Support maven-p2-site generation as part of regular tycho build


laeubi:
A [recent discussion](https://github.com/eclipse/tycho/discussions/430) has shown that there are demands to have pure p2-sites as part of the tycho build as well.

The proposal is to add a new packaging type p2-maven-repository that has the [tycho-p2-repository:assemble-maven-repository](https://www.eclipse.org/tycho/sitedocs/tycho-p2/tycho-p2-repository-plugin/assemble-maven-repository-mojo.html) bind to the `package` of its life-cycle and automatically depend on al bundle-project of the reactor.

This would make the usage of the site mojo inside a regular tycho build more convenient.

mickaelistria:
I don't really get what such a new packaging type would make noticeable simpler? Can you please provide some examples of before/after?

laeubi:
The main goal would be to have a simple pom.xml in the form of

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<groupId>tycho.example</groupId>
	<artifactId>p2-site</artifactId>
	<version>1.0.0-SNAPSHOT</version>
	<packaging>maven-repository</packaging>
</project>
```

that simply builds a maven-repository site based on all reactor projects. The problem is that currently (tom ake it right) one has to declare all projects as dependencies, or use the reactor strategy and find a way to build the site project as the last item inside a tycho build.

My idea would be that the TychoMavenLifeCycleParticipant could handle such packaging special and simply add all  bundle-projects as dependencies.

mickaelistria:
What about transitive deps or extra deps? I'm afraid that in practice, for a p2 repo to be useful, one would always need some tricks to add some extra deps, and that can already be covered by category.xml. So overall, I think it could be a regular `eclipse-repository` with an extra customatization of the https://www.eclipse.org/tycho/sitedocs/tycho-p2/tycho-p2-repository-plugin/assemble-repository-mojo.html to "includeAllReactorArtifacts" (and which would be able to run with or without a category.xml).

laeubi:
[...]

PR #1489, opened 2022-10-07
Add a packaging for maven site

laeubi:
Fix https://github.com/eclipse-tycho/tycho/issues/436


tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/P2SiteProject.java
```diff
@@ -0,0 +1,57 @@
+package org.eclipse.tycho.core.osgitools;
```
akurtakov, line 1: Please put license header.

Issue #1483, opened 2022-10-07
Promote checksums from the maven repository for maven-target locations


laeubi:
P2 supports checksum and maven repositories have checksum as well. We should (if available) promote those. 
Similar to
- https://github.com/eclipse-tycho/tycho/issues/661

it would be good to make this configurable.

mickaelistria:
> it would be good to make this configurable.

Why? Can anything bad happen with extra checksums?

laeubi:
> > it would be good to make this configurable.
> 
> Why? Can anything bad happen with extra checksums?

At least they might require extra I/O to be fetched and there is always a case where someone complains ;-)

mickaelistria:
> there is always a case where someone complains ;-)

I think it's not really true. Often people don't complain about what they don't know is happening; also I think this is very hypothetical someone complains, so better not introduce more complexity and more user options until we have 1 serious complaint.

PR #1490, opened 2022-10-07
Promote maven artifact signatures to p2 metadata

laeubi:
Fix https://github.com/eclipse-tycho/tycho/issues/1483

FYI @mickaelistria


laeubi:
Seems P2 Advice processing is strange... needs further work here ...

laeubi:
This should finally promote the checksums from maven artifacts to the meta-data and get rid of the warning for maven artifacts from target locations.

PR #1493, opened 2022-10-09
Document packaging types

kwin:


src/site/markdown/PackagingTypes.md
```diff
@@ -0,0 +1,35 @@
+# Packaging Types
+
+Tycho defines the following custom Maven packaging types targeted for Eclipse Plug-in development.
+
+1. `eclipse-plugin` corresponds to [Eclipse Plug-in and Plug-in Fragment projects](https://wiki.eclipse.org/PDE/User_Guide#Plug-in)
+1. `eclipse-test-plugin` is similar to `eclipse-plugin` but only contains ITs to be executed inside an Eclipse application. There is a noticable difference between `eclipse-plugin` and `eclipse-test-plugin` with respect to the goal executed during `integration-test`. The former uses `tycho-surefire-plugin:integration-test` while the latter uses `tycho-surefire-plugin:test`.
```
laeubi, line 6: The numbering seems odd (maybe an unordered list is sufficient here? beside that, we should not really talk about IT but "Plugin Tests" because thats the term used in Eclipse/PDE.
If @mickaelistria is fine with that, I would like to note here that today it is recommended to use `eclipse-plugin` in most cases.

mickaelistria: > If @mickaelistria is fine with that, I would like to note here that today it is recommended to use eclipse-plugin in most cases.

Yes, I think it's OK to recommend eclipse-plugin in most cases and to mention `eclipse-test-plugin` is suitable for test-only bundles and maps to PDE "Run As > JUnit Plugin Test" action.

kwin: > unordered list is sufficient here

done in https://github.com/eclipse-tycho/tycho/pull/1493/commits/5095c0c5474614fb5a78173ec75f80265c317aa2 Also replaced IT by "Plugin Tests".

@laeubi Can you extend this PR with the recommended layout for `eclipse-plugin`?

PR #1496, opened 2022-10-10
Clarify CONTRIBUTING.md about mojo changes

laeubi:
@mickaelistria @akurtakov We have already recently talked about that we do not want require a majo-version for mojo changes that are compatible (e.g. just moving a mojo or replace it by something other), but this was not reflected in the contributing guide.

Therefore I'd like to propose the following change.


akurtakov:
Fine with me. "Drop-in replacement" is important promise.

PR #1511, opened 2022-10-11
Move the MirrorApplicationService to tycho

laeubi:


akurtakov:
@laeubi You're on a spree! I'm really excited about what I see in Tycho.

laeubi:
Hm... in the IDE the test runs but with maven surefire it fails, seem there is something different in the classpath...

laeubi:
Strange:

- Jenkins CI succeeds
- Github fails on both win/linux
- my local build also fails in maven but succeeds in Eclipse ...

laeubi:
The problem is when many tests run, if one executes only the test alone everything works fine...

laeubi:
Even running *all* tests from tycho-core suceeds but running it from the root tree fails... so it seems there is something confused when running in the reactor (maybe a build order dependency?)

laeubi:
@akurtakov any idea?

if I run 

```
tycho$ mvn clean install  -Dtycho.localArtifacts=ignore -DskipTests
tycho/tycho-core$ mvn clean install -Dtycho.localArtifacts=ignore
```

tests pass.

if I run

```
tycho$ mvn clean install  -Dtycho.localArtifacts=ignore
```
test fail in `tycho-core` module...

so it seems some kind of dependency makes a problem when it is build together with tycho-core...

akurtakov:
To eliminate possibly ignoring local artifact when you shouldn't. Do you see same behavior in both cases when using verify instead of install?

laeubi:
> To eliminate possibly ignoring local artifact when you shouldn't

As tycho-core is only maven, this is/was actually not required.

> Do you see same behavior in both cases when using verify instead of install?

I used the install because tycho-core depends on some snapshot artifacts so I think one needs to first install them?

laeubi:
It seems the test uses the jared file, so this is not the culprit here...

laeubi:
I now did a 1:1 comparison and it seems the only difference is that if one executes it directly only maven-deüendencies are used, while otherwise the p2 dependencies are used at some places... I'll try to convert tycho-core-shared to a plain maven-jar ...

laeubi:
> I'll try to convert tycho-core-shared to a plain maven-jar ...

Yep that makes it run, but now I need to find a way to convert the project to a plain jar but make the old bundles happy... what is hard because of the wrecked platform maven meta-data and highly dependency coupling of Tycho & Platform... :-\

laeubi:
So back to the drawing board, minimal reproducer is:

`mvn -pl tycho-core,tycho-bundles/org.eclipse.tycho.core.shared clean verify -Dtest=org.eclipse.tycho.p2tools.MirrorApplicationServiceTest#testMirrorFeatureWithContent -DfailIfNoTests=false`

next idea is to filter the dummy realm ...

laeubi:
At least now CI and GitHub both fail consistently with the same errors. Just filtering every P2 dependency from the dummy real already works to a certain extend but then some lookup of Agent Services seem not working...

laeubi:
I got a little step further running 

`mvn -pl tycho-core -am clean verify  -Dtycho.localArtifacts=ignore`

now succeeds ... but running the full build still shows problems.

laeubi:
Finally!

PR #1523, opened 2022-10-13
Add a surrogate implementation of UI dependencies

laeubi:
There is an optional import for org.eclipse.ui.workbench to support UI test. If the import can not be resolved, there is no fallback in two scenarios:

1. we only want compile but do not pull in half of platform 
2. the bundle can not be resolved

for those cases the minimal necessary API is copied into the bundle that will be shaded by the imports if available in the running system.

FYI @akurtakov @mickaelistria this is an attempt to reduce the compile time dependencies to platform-ui what pulls in a large dependency chain and has many bad maven meta-data that causes trouble...

### DIFF
--- a/tycho-surefire/org.eclipse.tycho.surefire.osgibooter/pom.xml
+++ b/tycho-surefire/org.eclipse.tycho.surefire.osgibooter/pom.xml
@@ -45,7 +45,7 @@
 				<executions>
 					<execution>
 						<id>copy</id>
-						<phase>process-resources</phase>
+						<phase>validate</phase>
 						<goals>
 							<goal>copy</goal>
 						</goals>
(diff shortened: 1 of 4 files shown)


laeubi:
@waynebeaton would it be problematic to embeds the two classes from platform into Tycho?

laeubi:
This is obsolete if we use:

- https://github.com/eclipse-tycho/tycho/pull/1544

still it would be useful to know if it is "safe" to copy code from other Eclipse projects or if this requires a CQ...

mickaelistria:
> This is obsolete if we use:

So can we close the issue? #1544 looks much more sane.

laeubi:
This is now oboslete

waynebeaton:
> @waynebeaton would it be problematic to embeds the two classes from platform into Tycho?

This looks fine to me.

> still it would be useful to know if it is "safe" to copy code from other Eclipse projects or if this requires a CQ...

I get a lot of email notifications from GitHub. Adding me in copy to some random issue will have mixed results. If you have an IP question, you'll get more consistent results when you send them it to emo-ip-team@eclipse.org.

laeubi:
Thanks for the feedback an no problem about the delay its all fine :+1:

PR #1524, opened 2022-10-13
Get rid of P2 defined items and use instead maven deps

laeubi:
This replaces the target platform by maven dependencies so we are sure all modules consume the same artifacts.


laeubi:
@akurtakov any idea why " project org.eclipse.tycho.p2.resolver.impl.test: Could not determine test framework provider"? Should the import of junit not be enough? Or is there probably a requirement of the provider itself not fulfilled?

akurtakov:
Probably it's not ready to handle import-packages only. The providers work with bundle names e.g. https://github.com/eclipse-tycho/tycho/blob/273b20d1b2e6ebab063b28efd3d1c90801151a1a/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/provider/impl/JUnit47Provider.java#L75 and I have no idea how/when resolving package to bundle name happens and whether it's not too late .

laeubi:
I'll check this, at least here it worked with import package already: https://github.com/eclipse-tycho/tycho/pull/1522 so it must be something missing here...

akurtakov:
> I'll check this, at least here it worked with import package already: #1522 so it must be something missing here...

Rebasing this PR on master will make it smaller and hopefully easier to spot what's wrong.

laeubi:
> Rebasing this PR on master will make it smaller and hopefully easier to spot what's wrong.

Done, I also changed the maven cache key to see if the helps with the strange build error on GH Actions

laeubi:
> The providers work with bundle names e.g.

I think your are right so will try to configure this explicitly...

laeubi:
Lets try a different approach: https://github.com/eclipse-tycho/tycho/pull/1529

PR #1529, opened 2022-10-13
Use pom dependencies for core.shared and embedder.shared

laeubi:


laeubi:
Now it show the same error in the OSGi-Tests as seen in the converted maven tests, so it seems there is a dependency that break things and is different between P2 defined target and maven defined pom dependencies :-\

@akurtakov if you might want to take a look, what I did here is to let all "true" OSGi bundles use the tycho-bundles-target, while the "shared" using pom dependencies = consider and a pom defined "target".

Now if as the test combine both worlds inside an `eclipse-plugin-test` something in the intersection of both makes things fail as the extension registry is not found....

akurtakov:
I actually have an idea. IIRC JUnit 4.x jars in maven central don't have OSGi metadata so this is probably the issue. If you verify that the issue happens with JUnit 4.x but not with JUnit 5.x tests - that's it.

laeubi:
I think all OSGi test use Junit4, but these use the target platform (orbit) P2 bundles, aslo the test start, but loading some stuff inside the test then fails.

laeubi:
Even more strange, the **Jenkins** build **passes** the tests: https://ci.eclipse.org/tycho/job/tycho-github/job/PR-1529/2/console ... it is just Github and my local machine where it fails with such strange errors...

laeubi:
Next assumption is: There are now two bundles of the same name but only one is choose what unluckily is not the one the test uses later on ... so it actually depends on the order of jars read from the filesystem.... I'll add some debug output.

laeubi:
This was a good idea but also don't work out well, need to think about another way out of this dependency chaos...

PR #1531, opened 2022-10-13
Update maven target locations

laeubi:


laeubi:
@akurtakov in the target file there is also the following TODO:

```
<!-- Currently still using Orbit variants instead of upstream one in Platform
				because Tycho 2.7.x which is used to build current Tycho 3.0.0-SNAPSHOT
doesn't support new/official JUnit bundles -->
```

as we are now using Tycho 3.0 should this location be updated/deleted/...?

akurtakov:
> @akurtakov in the target file there is also the following TODO:
> 
> ```
> 
> ```
> 
> as we are now using Tycho 3.0 should this location be updated/deleted/...?

I think so. @mickaelistria did this work IIRC but good catch on yet another thing we would better get fixed.

laeubi:
GH failures are not related...

laeubi:
@mickaelistria do you like to create a follow-up regarding the junit bunldes updates?

PR #1540, opened 2022-10-16
Fix junit59withvintage project data

laeubi:


laeubi:
Only metadata changes ...

PR #1543, opened 2022-10-16
Fix surefire java version to be built on 1.8

laeubi:


laeubi:
@akurtakov any idea what could cause the test-failure? I can't reproduce this with a local build and the code-part seems unrelated to surefire or any java compile version...

laeubi:
At laest GH do not shows the failure, local build is also fine...

laeubi:
I added a new PR without changes here ant it fails with the same tests

https://github.com/eclipse-tycho/tycho/pull/1547

but master succeeds:

https://ci.eclipse.org/tycho/job/tycho-github/job/master/

to make progress here I want to disable the test for now to see if anything else fails on Jenkins @akurtakov @mickaelistria any objections?

akurtakov:
I don't see why you have to even ask us for temp disablement of test for experiment in this PR.

laeubi:
> I don't see why you have to even ask us for temp disablement of test for experiment in this PR.

Question is more how long such temp disabling could be used, e.g. I don't see any chance to debug this if it is not reproducible locally, as the test assumes a failure where non seem to happen... so if the test only fails on Jenkins CI but not for anyone else, I fear we need to disable it unless someone finds out how to reproduce this so it can be debugged ... Or we conclude that the test is not that important and remove it permanently.

tycho-surefire/pom.xml
```diff
@@ -14,9 +14,8 @@
 	<modelVersion>4.0.0</modelVersion>
 	<parent>
 		<groupId>org.eclipse.tycho</groupId>
-		<artifactId>tycho-bundles</artifactId>
```
akurtakov, line 17: Could this be the issue? I don't see why this change is in this PR and tycho-bundles parent pom overrides target-platform-configuration.

laeubi: Actually tycho-bundles as a parent is not required here as all submodules using non tycho-build (jar packaging) or directly use tycho-bundles as their parent....

Beside that I created a PR without any (code) changes and it also fails with this test :-\ 

See
- https://github.com/eclipse-tycho/tycho/pull/1547
- https://ci.eclipse.org/tycho/job/tycho-github/job/PR-1547/1/

so it is a bit strange ...

laeubi:
I have created https://github.com/eclipse-tycho/tycho/issues/1550 to track enabling the test again.

PR #1571, opened 2022-10-21
Move remaining service from OSGi > Tycho core

laeubi:


laeubi:
Platformbuild already suceeds with this change... so now its just a matter of converting the tests.

laeubi:
All test are migrated now and everything compiles fine, but some tests still need some adjustments, any help would be appreciated.

If this PR is merged, we finally can transform the remaining Tycho build bundles into maven build ones and then Tycho do not needs Tycho to be build, all services are available as plexus components and we can use everything everywhere in Tycho without the need of an facades, proxies or extra services.

laeubi:
Jenkins succeed, but the Jenkinsfile needs to be adjusted as the osgi test results no longer there:

> ‘*/*/target/work/data/.metadata/.log’ doesn’t match anything: ‘*’ exists but not ‘*/*/target/work/data/.metadata/.log’

PR #1578, opened 2022-10-23
Remove no longer required service registrations

laeubi:


laeubi:
Seems this requires more smaller steps...

PR #1580, opened 2022-10-24
Move the transport to the p2-plugin

laeubi:


laeubi:
FYI @akurtakov this should finally remove references to ECF from Tycho. Lets see

laeubi:
This looks promising:

1. Platform build passes successfully
2. Only the reactor make integration test fail (what might be just a configuration issue due to the move and now requires some more bundles to be started).

laeubi:
The problem seems that the eventbus is a SCR component and we exclude scr from the tycho build extension so it is not available and causes problems...

So I included it again, if that works, we can think of simply register it also as plexus component for such cases where SCR is not used...

laeubi:
Obsolete now

PR #1581, opened 2022-10-24
remove target platform from Tycho setup

Bananeweizen:
There is no PDE and target anymore.


akurtakov:
@laeubi @merks Can you pleaase push this one if the change is fine? It's not needed but I'm not sure of Oomph mechanics.

Bananeweizen:
@akurtakov The change is not urgent. The "Set target platform action" will just silently fail since a target with that name is not available. So no need to push this in a hurry, the Oomph setup is not broken.

akurtakov:
Let's merge it. I trust you :) .

laeubi:
> Let's merge it. I trust you :) .

Yes thats fine, thanks @akurtakov for taking care, I forgott that when dleting the target file with recent commit.

PR #1593, opened 2022-10-25
automatically mark target folders derived, fixes #1591

### LINKED ISSUE #1591: automatically mark target folders derived in eclipse
I would like to add the https://nodj.github.io/AutoDeriv/ plugin to the Tycho Oomph setup. It allows marking workspace files as derived via a simple gitignore-like config file. That way, stuff in target folders is no longer included in search results etc. I'm using this since years in company projects.
Any objections? Otherwise I would just create a PR.

Bananeweizen:
That avoids contents from target folders to appear in file search and other derived-files-aware tooling. One preference has be to disabled to avoid AutoDeriv to generate an info warning.

Also disable the license confirmation generally in the Tycho setup.

Looks like this afterwards, but that can be further customized via preference page: 
image

### DIFF
--- a/.derived
+++ b/.derived
@@ -1,0 +1,5 @@
+# AutoDeriv plugin configuration.
+# To make this work, you need to install the plugin from https://nodj.github.io/AutoDeriv/
+
+# target folders of Maven
+target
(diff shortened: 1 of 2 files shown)


laeubi:
@Bananeweizen thanks for helping improve the tycho project setup!

PR #1599, opened 2022-10-27
Enable RemoteAgentCompositeLoadingTest again

### LINKED ISSUE #1550: Investigate and enable RemoteAgentCompositeLoadingTest again
As discussed below, the `RemoteAgentCompositeLoadingTest` is currently disabled (ad2b1b2184f7318bff03ba2da8f5c9da6731c97b) because it only fails on Jenkins. We should enable it again if we find a way to reproduce the failure locally...

I added a new PR without changes here ant it fails with the same tests

https://github.com/eclipse-tycho/tycho/pull/1547

but master succeeds:

https://ci.eclipse.org/tycho/job/tycho-github/job/master/

to make progress here I want to disable the test for now to see if anything else fails on Jenkins @akurtakov @mickaelistria any objections?

_Originally posted by @laeubi in https://github.com/eclipse-tycho/tycho/issues/1543#issuecomment-1280403053_

laeubi:
Lets se if recent refactoring have made things better here...

Fix #1550


laeubi:
https://ci.eclipse.org/tycho/job/tycho-github/view/change-requests/job/PR-1599/1/ succeeds (but status is not reported) so I think we are fine to merge this

PR #1600, opened 2022-10-28
Fix build path errors

Bananeweizen:
Just rerun "Update Maven projects" on all existing projects, no manual changes. The removed entries all caused build path errors complaining about non existing source/resource folders. With this change, I have only warnings remaining, but no errors.

image

If you remember the discussion from EclipseCon: As long as the build path still contained errors, I had to manually fix them (like setting the classpath entries optional). Once all build path errors were fixed manually, the "Update Maven projects" action was suddenly able to fix the errors on its own (i.e. overwriting my "optional" and just removing the entries).

### DIFF
--- a/sisu-osgi/sisu-osgi-connect/.classpath
+++ b/sisu-osgi/sisu-osgi-connect/.classpath
@@ -28,11 +28,5 @@
 			<attribute name="test" value="true"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry excluding="**" kind="src" output="target/test-classes" path="src/test/resources">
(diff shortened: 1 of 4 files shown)


laeubi:
@Bananeweizen very strange, do you like to open an issue for m2eclipse? This should really be fixed there as well! I assume the issue arise like this:

1. You have a maven project with java+resource folder and some content and m2e adds the entries (but without optional)
2. Then you delete everything from the resources folder and check it in

Because the folder exits locally, m2e do not remove the entries but then if you checkout this on another computer the errors appear.

PR #1602, opened 2022-10-28
Various small code modernizations.

akurtakov:
These just pile in my workspace while investigating different things thanks to sonarlint.


Bananeweizen:
I'm always happy about such cleanups, therefore please continue. :)

Bananeweizen:
@laeubi: Do you have an idea how this sometimes goes down the rabbit hole without ends? So for actually doing some warning fixes in https://github.com/eclipse-tycho/tycho/pull/1609 I try building, which gets me sudden git changes for the launch configs: https://github.com/eclipse-tycho/tycho/pull/1608. So I switch to the console for building, but notice other warnings that should not be there: https://github.com/eclipse-tycho/tycho/pull/1610. Running another build and closely monitoring the console to verify the warnings are gone, I see a typo in plexus: https://github.com/codehaus-plexus/plexus-containers/pull/65. And after I have submitted that one, I notice it is meanwhile deprecated and search for similar issues in Eclipse Sisu instead (without knowing what these projects even do). And I haven't even looked at the inconsistent jdt.core.prefs that I also noticed during fixing the warnings... 😕

laeubi:
@merks once asked me why I'm so concerned if a lot of scattering changes all around the eclipse (or more general open-source world) touching several concerns from documentation to code and building / releng stuff and I think you found the answer and can feel a little the pain I see every day 😁

Its always like you feel that just **this** PR needs to be merged/fixed and you could sit back and just enjoy coding new and cool feature but then suddenly something pop up in that way and you just need to **that** last PR until ...

akurtakov:
What you describe guys is my world for the last decade :) . I no longer even try to put these things aside - just fix what I see and can fix so next time I get to the project there will hopefully be less such distractions.

PR #1609, opened 2022-10-30
Fix some eclipse warnings

Bananeweizen:
No functional changes, rather just going after different warnings sorted by type.


laeubi:
@Bananeweizen it seems this breaks something do you like to take a look? or probably trying some smaller batches?

Bananeweizen:
removed the try-with-resource refactoring, as it seems to have changed some logic in the exception handling there.

akurtakov:
I feel better about the future of a project when I see such changes compared to trying to get a new feature in quick and dirty way :)

laeubi:
Thanks @Bananeweizen for helping cleaning things up with Tycho codebase!

PR #1610, opened 2022-10-30
Avoid BND warning about missing developer id

Bananeweizen:
BND complains: Cannot consider developer in line 'foo' of file 'bar/pom.xml' for bundle header 'Bundle-Developers' as it does not contain the mandatory id.

To avoid this developers must either have an ID or be removed. Maven itself considers the list of developers the list of people to be contacted for the project. Since the currently listed developers are no longer active, the error can be avoided best (and with little future maintenance effort) by having an organizational entry as developer, pointing the Tycho project.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,19 +42,11 @@
 	</scm>
 	<developers>
 		<developer>
-			<name>Igor Fedorenko</name>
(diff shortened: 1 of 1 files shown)


laeubi:
Good idea!

PR #1611, opened 2022-10-30
fix github organization links

Bananeweizen:
While there is currently some magic in place to redirect github.com/eclipse/tycho we should probably rather use the full url containing the eclipse-tycho organization instead.

Also update the developerConnection entry to point to github.


laeubi:
Thanks, this was really heavily outdated...

PR #1612, opened 2022-10-30
Avoid superfluous info markers

Bananeweizen:
As long as the POM editor is opened, it will add an info marker for every comment starting with "NOTE". Looks like this, and is rather confusing since it depends on the editor lifecycle, not the workspace resources (i.e. you have no such marker if you don't open the editor).
image


mickaelistria:
I don't think it's m2e; but TM4E that creates some marker for patterns in comments. A different marker type could probably be used (task could be used instead of problems).
That said, I don't see a particular value to keep a NOTE prefix here. So I think it's good to merge it.

Issue #1621, opened 2022-11-02
Do not overwrite MANIFEST.MF and feature.xml files during packaging


lppedd:
Currently, I'm trying to implement multi-branding for an application.
This means I process resources via filtering, `MANIFEST.MF` and `feature.xml` files too to rename bundle symbolic names at build time.

The Tycho Packaging Plugin however always overwrite those two files when packaging the product, even if they're already present under `${project.build.directory}`.

https://github.com/eclipse-tycho/tycho/blob/77243cebc93d0bdbaf31e0e1b1055e29182958d8/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackagePluginMojo.java#L267-L269

And

https://github.com/eclipse-tycho/tycho/blob/77243cebc93d0bdbaf31e0e1b1055e29182958d8/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackageFeatureMojo.java#L124-L130

The proposed change is checking for `!file.exists()` before writing.  
If it's ok I have a PR ready, as I haven't found any workaround.

laeubi:
@lppedd I think this is "by intend" and I'm not sure if it would work nice to change the defaults, e.g. if one do not perform clean then maybe outdated content might be used.

The usual solution to this in maven is to allow to configure the **source** of the file to use instead, so you could write your feature to `target/my_enhanced_feature.xml` and then configure the mojo read its from there instead.

mickaelistria:
I agree `!file.exists()` is not a good check here as we want the file in `target/` to be overwritten if necesary between different invocations of `mvn`. I believe what you want to try is to customize your MANIFEST.MF/feature.xml with a mojo executed during the `process-resources` phase, directly modifying the file in the `target/` folder.

laeubi:
> I believe what you want to try is to customize your MANIFEST.MF/feature.xml with a mojo executed during the `process-resources` phase, directly modifying the file in the `target/` folder.

The file is only copied in the package phase, thus I think a config parameter where the package-mojo reads the file from is more suitable and flexible.

lppedd:
@laeubi done as best as I could. I'm using a custom `MavenResourcesFiltering` so this complicates it a bit.
I'll provide a test for `feature.xml` too as soon as I understand why `maven-resources-plugin` is not running for the `eclipse-feature` packaging.

lppedd:
Nevermind, I've simplified the test.

lppedd:
[...]

PR #1623, opened 2022-11-03
Allow specifying custom MANIFEST, feature, category and product files

lppedd:
See issue #1621.

### DIFF
--- a/tycho-maven-plugin/src/main/resources/META-INF/plexus/components.xml
+++ b/tycho-maven-plugin/src/main/resources/META-INF/plexus/components.xml
@@ -167,8 +167,9 @@
               <initialize>
                 org.eclipse.tycho:target-platform-configuration:${project.version}:target-platform
               </initialize>
-              <generate-resources>
-              </generate-resources>
+              <process-resources>
+                org.apache.maven.plugins:maven-resources-plugin:${resources-plugin.version}:resources
+              </process-resources>
               <prepare-package>
                org.eclipse.tycho:tycho-packaging-plugin:${project.version}:update-consumer-pom
               </prepare-package>
--- a/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/ValidateIdMojo.java
+++ b/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/ValidateIdMojo.java
@@ -17,6 +17,7 @@ import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
 import org.eclipse.tycho.PackagingType;
 
 /**
@@ -24,10 +25,15 @@ import org.eclipse.tycho.PackagingType;
  */
 @Mojo(name = "validate-id", defaultPhase = LifecyclePhase.VALIDATE, threadSafe = true)
 public class ValidateIdMojo extends AbstractVersionMojo {
+    /**
+     * Whether to skip the project's artifact ID validation against the OSGi ID.
+     */
+    @Parameter
+    private boolean skip;
 
     @Override
     public void execute() throws MojoExecutionException, MojoFailureException {
-        if (!project.getArtifactId().equals(getOSGiId())) {
+        if (!skip && !project.getArtifactId().equals(getOSGiId())) {
             failBuildDueToIdMismatch();
         }
     }
--- a/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackagePluginMojo.java
+++ b/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackagePluginMojo.java
@@ -311,8 +311,15 @@ public class PackagePluginMojo extends AbstractTychoPackagingMojo {
 	}
 
 	protected Manifest getManifest() throws IOException, MojoExecutionException {
+		// If the user explicitly set a MANIFEST file in the archive configuration,
+		// read and update that one instead of looking under the source files
+		final File archiveManifestFile = archive.getManifestFile();
+		final File manifestFile = archiveManifestFile != null
+				? archiveManifestFile
+				: new File(project.getBasedir(), "META-INF/MANIFEST.MF");
+
 		Manifest mf;
-		try (InputStream is = new FileInputStream(new File(project.getBasedir(), "META-INF/MANIFEST.MF"))) {
+		try (InputStream is = new FileInputStream(manifestFile)) {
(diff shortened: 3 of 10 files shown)


tycho-maven-plugin/src/main/resources/META-INF/plexus/components.xml
```diff
@@ -167,8 +167,9 @@
               <initialize>
                 org.eclipse.tycho:target-platform-configuration:${project.version}:target-platform
               </initialize>
-              <generate-resources>
-              </generate-resources>
+              <process-resources>
+                org.apache.maven.plugins:maven-resources-plugin:${resources-plugin.version}:resources
+              </process-resources>
```
lppedd, line 172: As discussed, I've added the resource processing phase.

tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/ValidateIdMojo.java
```diff
@@ -17,17 +17,23 @@
...
 /**
  * Validates that project Maven and OSGi ids match.
  */
 @Mojo(name = "validate-id", defaultPhase = LifecyclePhase.VALIDATE, threadSafe = true)
 public class ValidateIdMojo extends AbstractVersionMojo {
+    /**
+     * Whether to skip the project's artifact ID validation against the OSGi ID.
+     */
+    @Parameter
+    private boolean skip;
```
lppedd, line 32: This is not really required, but a way to skip the validation to apply some "hacks" might be useful.

tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackagePluginMojo.java
```diff
@@ -311,8 +311,15 @@ protected void writeManifest(File output, Manifest mf) throws IOException {
 	}
 
 	protected Manifest getManifest() throws IOException, MojoExecutionException {
+		// If the user explicitly set a MANIFEST file in the archive configuration,
+		// read and update that one instead of looking under the source files
+		final File archiveManifestFile = archive.getManifestFile();
+		final File manifestFile = archiveManifestFile != null
+				? archiveManifestFile
+				: new File(project.getBasedir(), "META-INF/MANIFEST.MF");
```
lppedd, line 319: No need to add another parameter, we just reuse the archive configuration.

laeubi:
I think this are useful additions, an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate basic usage and a note at the release notes would be good.

lppedd:
@laeubi for sure. I just want to make it work with building the p2 repository. Other changes might be required on these mojos.

lppedd:
I have pinned down the actual issue while debugging. I have to get approval to share the info first and then I'll post here.

lppedd:
@laeubi better be safe than sorry lol

Here we are inside the resolution process spawned by `AssembleRepositoryMojo#getVisibleRepositories()`.  
You can see the requirements for the `p2updatesite` repository are still referring to old feature IDs.

old_requires_c

While the installable units have been updated to the new names, to the renamed feature IDs.

renamed_units_c

This is why the the build fails, as capabilities cannot be satisfied. Any idea on how this works?

lppedd:
This is where the requirements for `p2updatesite` are originally set.
See the call stack.

image

lppedd:
I've debugged more and I've seen the `category.xml` file that is considered is always the one in `${project.basedir}`.
You can actually see requirements are recomputed during the `assemble-repository` mojo execution.

image

lppedd:
@laeubi last question: can I somehow override `P2DependencyResolver`?
I would like to provide my own which extends it via a separate plugin.

Anyway I'll try to add an example/IT soon so we can merge.

laeubi:
>  can I somehow override `P2DependencyResolver`?

With Tycho 4: yes (it is jsut a regular maven plexus component), with Tycho 3 no.

lppedd:
@laeubi not even if I use a custom `<resolver>`?
https://tycho.eclipseprojects.io/doc/latest/target-platform-configuration/target-platform-configuration-mojo.html#resolver

laeubi:
That should work too. Bu you then need to adjust the config of the target.

lppedd:
The last commit completes the work done here, at least for the `eclipse-repository` packaging.  
What happens is, we actually specify our custom base path where to find the modified configuration file. Now filtering can be used to its full potential.

The same concept,

image

can be applied to other types of configurations, e.g. `MANIFEST.MF` files or `.product` files, in other Mojos, if needed.  
This seemingly "dumb" solution does not break anything.

lppedd:
Now, if this is ok, I'll prepare the examples and ITs, but there is one thing to clarify.
You can see that I'm targeting the 2.7.x branch, why? Because I cannot move to 3.0.0 for our product currently. Would it be possible to push a new update for the 2.7 line in case?

laeubi:
In general you should first implement it for the current master. if that works one might backport it to a branch.

About 2.7.x we usually do not publish bugfix/maintainance releases for Tycho. But in the past some companies gracefully sponsored such releases because they require "old" releases, if that's the case here as well you can contact me for details.

Beside that, of course we can accept "community-driven" releases as well, this usually includes:

- Update the version to the next release on the corresponding branch (you can just open a PR for that)
- check that the build succeeds and snapshots are published
- identify bugfixes/feature to be backported and have a PR for them including updates to the release notes
- Wait for feedback from the community if there are more bugfixes and/or regressions
- Then one of the commiters has to push the release to central

That said, you probably just want to update to the latest tycho-release, actually there should be not much holding you back from that beside havin a Java 17 JVM for build and probably older JVM in the toolchains for compile (if desired).

lppedd:
@laeubi sounds good! Sorry for the strange workflow. Wanted to be sure I could get it working with our current version.

So what I'll do now:
1. cherry pick to my local master
2. test the master snapshot on my build
3. open a PR targeting master
4. ask for a backport to the nearest planned release 

Is it ok?

laeubi:
> Is it ok?

Yep, just keep in mind that as of now, the next planned release is Tycho 4.0

lppedd:
@laeubi got it. How much time do I have to get it merged?  
Btw, on master the the target platform seems disappeared, was it removed/moved?

laeubi:
Things has changed much at 3.0/4.0 the release of 4.0 will most likley not happen before next year so there is enough time here.

lppedd:
@laeubi oh ok!
That could be a "problem".  I need those changes to ship a prototyped product by the end of December.  
I wonder what's the best path to take than. Maybe I can self host Tycho on our build machine until it's out? Not sure if licensing allows that tbh.

laeubi:
Tycho publishes [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) builds that are usually available after an hour after a change was merged.

PR #1643, opened 2022-11-09
Support considering maven dependencies for compilation

### LINKED ISSUE #1556: Support considering maven dependencies for compilation
Currently if I declare a dependency in the pom (e.g. to use some annotations with class or source retention scope), Tycho ignores them unless I somewhere in the manifest reference them and they are OSGi bundles and thus compilation fails even if these are irrelevant at runtime.

There should be an option that any item declared in the pom with the appropriate scope are considered for compilation.

FYI @hd42

laeubi:
Previously dependencies declared in the pom are ignored by Tycho completely and even though one could enable these to be considered in target platform
this still requires them to be imported in the bundle manifest to finally be usable for compilation.

Now each pom defined dependency is always considered for compilation as this matches the expectation of most maven users and finally allows to have 'compile only' dependencies to be used,
for example with annotations that are only retained at source or class level.

Fix #1556

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,28 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ### Migration guide 3.x > 4.x
 
+#### pom declared dependencies are considered for compile
+
+Previously dependencies declared in the pom are ignored by Tycho completely and even though one could enable these to be considered in target platform
+this still requires them to be imported in the bundle manifest to finally be usable for compilation.
+
+Now each pom defined dependency is always considered for compilation as this matches the expectation of most maven users and finally allows to have 'compile only' dependencies to be used,
+for example with annotations that are only retained at source or class level.
+
+One example that uses [API-Guardian](https://github.com/apiguardian-team/apiguardian) annotations can be found here: https://github.com/eclipse/tycho/tree/master/tycho-its/projects/compiler-pomdependencies
+
+You can disable this feature through the `tycho-compiler-plugin` configuration:
+```
+<plugin>
+	<groupId>org.eclipse.tycho</groupId>
+	<artifactId>tycho-compiler-plugin</artifactId>
+	<version>${tycho-version}</version>
+	<configuration>
+		<pomOnlyDependencies>ignore</pomOnlyDependencies>
+	</configuration>
+</plugin>
+```
+
 #### Properties for tycho-surefire-plugin's 'useUIThread' and 'useUIHarness' parameters
 
 The configuration parameters `useUIThread` and `useUIHarness` parameter of the `tycho-surefire-plugin` can now be set via the properties `tycho.surefire.useUIHarness` respectively `tycho.surefire.useUIThread`.
--- a/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
+++ b/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
@@ -40,6 +40,7 @@ import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.artifact.resolver.ArtifactResolutionRequest;
 import org.apache.maven.artifact.resolver.ArtifactResolutionResult;
+import org.apache.maven.artifact.resolver.filter.ScopeArtifactFilter;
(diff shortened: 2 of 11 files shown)


lppedd:
Couldn't the fact `pomOnlyDependencies` is on by default be dangerous?
Someone might have dependencies declared by accident or whatever, and suddenly they take part in the classpath.

If you want to keep that on by default, wouldn't the correct scope to consider be `provided` in Maven?
This is how I currently use JetBrains nullability annotations.
```
<dependency>
  <groupId>org.jetbrains</groupId>
  <artifactId>annotations</artifactId>
  <version>23.0.0</version>
  <scope>provided</scope>
</dependency>
```

laeubi:
> Couldn't the fact `pomOnlyDependencies` is on by default be dangerous?

I can't think of any danger beside that maybe some want to disable it because of whatever reason... all Tycho integration tests pass so all officially supported features are good with this change.

> Someone might have dependencies declared by accident or whatever, and suddenly they take part in the classpath.

Then it should be fixed and its better Tycho reveals such accidental configurations.

> If you want to keep that on by default, wouldn't the correct scope to consider be `provided` in Maven?

You can choose whatever scope you like, so there is nothing Tycho requires here.

lppedd:
> You can choose whatever scope you like, so there is nothing Tycho requires here

Sounds good! Thanks.

Last question. Currently, to avoid having to manually download JAR files, I've setup the maven-dependency-plugin's `copy-dependencies` goal.
Dependencies are declared as `compile` and they get automatically copied under a directory, which is then included via `bin.includes`

```
bin.includes = META-INF/,\
               dependencies/
```

Could this interfere in some way or dependencies are just merged?

lppedd:
> I assume you list the dependencies then already in the Library Path of the bundle

Yes, dependencies are also declared in the bundle's classpath, e.g.:

```xml
<classpath>
  ...
  <classpathentry exported="true" kind="lib" path="dependencies/derbyshared.jar"/>
  <classpathentry exported="true" kind="lib" path="dependencies/derby.jar"/>
  ...
</classpath>

```

laeubi:
> I'm reasoning about this... Is this a reproducible thing or sometimes it's a way sometimes another way?

It depends ;-)

But if you really **copy** it doesn't really matter if the one from maven repo or the one from folder is used.

I'm currently looking in the platform failures and if thats also fine we already cover a large area of use-cases (probably also yours), still try out the current [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) build if this is merged is the best way to verify and give feedback on this for your case as well.

lppedd:
> if you really copy

That's how it's supposed to work, yep. I'm thinking about what edge/strange cases could happen. I'm just worried about all the facets that Tycho introduces to manage dependencies, version 4 seems to be a big deal compared to 2.7.x.

A bit OT (or not?): could the same thing I'm doing with `copy-dependencies` be done using straight Tycho somehow?
That would avoid dealing with such thing.

> still try out the current tycho snapshot build if this is merged

Already testing on the SNAPSHOT stream! Thanks!

laeubi:
> I'm thinking about what edge/strange cases could happen.

If there is something edge/strange the best is to provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the use-case, that's the very first thing that is checked and what we try to support as much as possible, I think the only times we removed an itest was with the pack200 removal.

> A bit OT (or not?): could the same thing I'm doing with `copy-dependencies` be done using straight Tycho somehow?

Don't think so, the more interesting question would be why you embedd dependencies at all (whats bad from OSGi POV) so the best would be to get that fixed up :-)

laeubi:
You can use maven-target support to consume stuff directly in your target including adding missing manifests:

https://läubisoft.gmbh/en/articles/using-maven-artifacts-in-pde-rcp-and-tycho-builds/

lppedd:
From the article.

> This was also recently enhanced by us to even generate basic meta-data for non-OSGi artifacts. This works well with Maven builds, but fails to give a good user experience in the IDE

I think this is related to that `wrapAsBundle` thing that I was mentioning. That's why I did not end up using it.

> if you already using a target file for your project...

I'm not, but actually I quite like the idea, might try to experiment with it. Thanks!  
Currently our "target" is simply represented as a Maven repository with a p2 layout.

laeubi:
> wrapAsBundle

is really basic and does not allow customization, but target files does!

The great thing about target support is that it works in Eclipse+Maven seamlessly.

laeubi:
Platform build ok, Jenkins as well, just the usual windows failures ...

PR #1663, opened 2022-11-14
Fix URIs pointing the old tycho location to use the new location

merks:
https://github.com/eclipse/tycho ->
https://github.com/eclipse-tycho/tycho


merks:
That was very fast!  🚀

PR #1668, opened 2022-11-15
Fix the base path for the setup's clone location

merks:


merks:
@akurtakov I missed this one during the previous round of URI fixes.

laeubi (approved):

laeubi:
We don't need to wait for the build here I think...

merks:
@laeubi Thanks.  Indeed this obviously has no impact on the build.  I often cancel such builds for the sake of climate change...

laeubi:
Actually one can even configure what files should trigger builds, but I didn't get to this from my long backlog yet :-\

akurtakov:
That's how alive project looks and works - by the time you end up seeing someone - someone already acted on and it's merged :) . "Given enough eyeballs, all bugs are shallow"

PR #1676, opened 2022-11-16
Remove the incomplete resolver feature

laeubi:
The incomplete resolving feature was introduced to allow mixed reactor builds. They way it worked has several implications:

- pom-dependencies=consider was required what has a lot of other implications
- the resolver does not really allow incomplete resolving, because of this we have tried to generate missing requirements
- this required several rounds of resolving and is slow and fragile

This feature is now removed and will be replaced by the new resolver approach.

**This PR is just to check what tests fail without this feature** (expectation: one test fails) **so it can be used as a base-commit to implements the new approach on top of it** (that fixes the test again).


laeubi:
So it seems we already (implicitly) relay on this working completely, so the new feature has to be enabled regardless of the resolver mode...

PR #1679, opened 2022-11-16
Update BuildProperties.md

LorenzoBettini:
Document additional.bundles supported since Tycho 2.4.0


laeubi:
I wish you find a way out the jungle there :-)

PR #1680, opened 2022-11-16
Add a special builder to tycho-build-extension

laeubi:
Work in progress...


laeubi:
Platform build succeeds and Jenkins also, so adding this dependencies at least to not seem to harm at the first place.

lppedd:
Was the performance gain really that big that it was worth making this the default?

laeubi:
There is not really any disadvantage... from my testing, depending on the system configuration, performance gains up to 50% and more are possible. Beside that `mvnd` uses the smartbuilder as well and people seem to like that :-)

lppedd:
@laeubi I like the idea in general. The only problem is the confusing log (as with standard Maven).  Plus the project seems to be on the inactive side (probably because the effort goes into mvnd now).

Is https://github.com/takari/concurrent-build-logger a viable option for logging? The docs are a bit confusing and I don't get if an entirely custom version of Maven is needed.

laeubi:
If it is possible to get better log-output with an extension we can include/enable this, probably you can get in touch with the `mvnd` people if/how they do it, probabbly it can be incooperated into https://github.com/takari/concurrent-build-logger
As far as I heared the smartbuilder changes should be contributed back to the original and it should be revived but I'm not sure how the status is, at laest I have added a PR to improve smartbuilder logging:
- https://github.com/takari/takari-smart-builder/pulls

and there are some attempts to modernize it (e.g. get rid of guava), so if you are interested in improving things it would be great to participate in these areas.

lppedd:
@laeubi or, we can try to solve `mvnd` issues, e.g., https://github.com/eclipse-tycho/tycho/issues/2042
So we don't have to pull in other extensions.

laeubi:
> @laeubi or, we can try to solve `mvnd` issues, e.g., #2042 So we don't have to pull in other extensions.

No plans for that... Tycho is a maven extension and works well with maven, so no support for 3rd party tools that do something different. `mvnd` is just poorly supported in IDEs and CIs and it is the choice of their maintainer to fire up an own tool instead of improving existing maven, but Tycho can't afford being tested / build for different eco-systems.

PR #1683, opened 2022-11-17
Join the former tycho.embedder.shared and embedder api into tycho-api

laeubi:


laeubi:
I'll leave this open as a reference @akurtakov you probably like to take a look here, actually everything compiles and most tests run, but due to the restructuring there is "something" messed up with the maven dependencies and I have not really an idea what is missing here and even if I think this will make the structure of tycho more clear I currently have to much on the table to dig deeper here :-\

PR #1695, opened 2022-11-20
Remove the deprecated LocalDependencyResolver

laeubi:
It is actually only used in tests, now we have the option to move this to tycho-core (see https://github.com/eclipse-tycho/tycho/pull/1694 ) we can finally get rid of it.

Fixes https://github.com/eclipse-tycho/tycho/issues/486


laeubi:
@akurtakov maybe also something you want to look into? It seems there are still some codes rely on the no longer used Localresolver.

akurtakov:
> @akurtakov maybe also something you want to look into? It seems there are still some codes rely on the no longer used Localresolver.

Sorry for not being really responsive lately, I'm quite busy with other tasks but will try to come back to tycho when things calm down a bit.

laeubi:
No problem, this is lingering around for such a long time we are not in hurry here..

laeubi:
I think first step is to remove the P2TestStubResolver! this currently seem to hide some (possible) regressions.

Issue #1660, opened 2022-11-13
Target platform resolved multiple times


Bananeweizen:
I've just converted Eclipse-CS from a target platform module to a plain target file reference. Doing that I noticed the target is resolved as many times as modules refer to it. Shouldn't we resolve it only once?
configuration is `${basedir}/../somedir/somefile.target`, which resolves to the same absolute path each time. This is 2.7.5, I still have to verify on 3.0.

See first and last line of this snippet:
```
[INFO] Resolving target definition file:/C:/dev/eclipse-cs/git/eclipse-cs/net.sf.eclipsecs.branding/../net.sf.eclipsecs.target/net.sf.eclipsecs.target.target for environments=[win32/win32/x86_64, linux/gtk/x86_64, macosx/cocoa/x86_64], include source mode=honor, execution environment=StandardEEResolutionHints [executionEnvironment=OSGi profile 'JavaSE-11' { source level: 11, target level: 11}], remote p2 repository options=org.eclipse.tycho.p2.remote.RemoteAgent@2006fdaa...
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
[INFO] Resolving dependencies of MavenProject: net.sf.eclipsecs:net.sf.eclipsecs.branding:10.3.3-SNAPSHOT @ C:\dev\eclipse-cs\git\eclipse-cs\net.sf.eclipsecs.branding\pom.xml
[INFO] Resolving target definition file:/C:/dev/eclipse-cs/git/eclipse-cs/net.sf.eclipsecs.checkstyle/../net.sf.eclipsecs.target/net.sf.eclipsecs.target.target for environments=[win32/win32/x86_64, linux/gtk/x86_64, macosx/cocoa/x86_64], include source mode=honor, execution environment=StandardEEResolutionHints [executionEnvironment=OSGi profile 'JavaSE-11' { source level: 11, target level: 11}], remote p2 repository options=org.eclipse.tycho.p2.remote.RemoteAgent@2006fdaa...
...
```

laeubi:
@Bananeweizen do I get ti correctly that you have different targets that point to the same target file via includes? 
Can you provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue with a minimal setup?

[...]

PR #1698, opened 2022-11-20
integration test for #1660: same target resolved multiple times

Bananeweizen:


Bananeweizen:
As expected the integration test fails because it resolves the same target twice. Locally the relevant log output is:
```
[INFO] Resolving target definition file:/C:/dev/tycho/git/tycho/tycho-its/projects/sameAbsoluteTarget/bundle1/../target-platform/target-platform.target
...
[INFO] Resolving target definition file:/C:/dev/tycho/git/tycho/tycho-its/projects/sameAbsoluteTarget/bundle2/../target-platform/target-platform.target
```
Those are 2 different paths, but just 1 unique target file, so it should not resolved a second time.

laeubi:
@Bananeweizen I think that's actually the issue that we need to normalize the URI here so tycho do not think it is two different files.

Bananeweizen:
Closing this one, since the commit is contained in your fix at #1702 now.

PR #1699, opened 2022-11-20
disable failing VerifierServiceImplTest

Bananeweizen:
Every PR fails with that one test on Windows, and everyone must scan the logs to see whether it is that "known failure". Better disable it, if there is no fix on the horizon.

### DIFF
--- a/tycho-p2/tycho-p2-repository-plugin/src/test/java/org/eclipse/tycho/p2/tools/verfier/VerifierServiceImplTest.java
+++ b/tycho-p2/tycho-p2-repository-plugin/src/test/java/org/eclipse/tycho/p2/tools/verfier/VerifierServiceImplTest.java
@@ -30,6 +30,7 @@ import org.eclipse.tycho.p2.tools.FacadeException;
 import org.eclipse.tycho.p2.tools.RepositoryReferences;
 import org.eclipse.tycho.testing.TychoPlexusTestCase;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -73,6 +74,7 @@ public class VerifierServiceImplTest extends TychoPlexusTestCase {
     }
 
     @Test
+    @Ignore("fails on Windows")
     public void testFileRepositoryWithTamperedArtifact() throws Exception {
         final RepositoryReferences repositories = sourceRepos("tampered_file");
         assertEquals(false, verify(repositories));


laeubi:
Actually it would be better to find out why this fails, I think @HannesWell wanted to take a look at it.

Bananeweizen:
Even if Hannes would take a look at it tomorrow, disabling it today would still be the right thing to do. First rule of handling flaky tests: Quarantine them, to avoid further test runs being flaky. Only _then_ fix them.

laeubi:
The probelem is that even if we disable this test, there are still some (random) failures on Windows in the integration tests. I just see them more as a gentle reminder we need better Windows (testing) support :-)

Bananeweizen:
If you need a reminder, please create a task. I'm aware that you probably know exactly what happens when looking at a PR. All the people contributing just every now and then don't, as we saw by the discussion in #1671. By having a (generally) green master build it becomes easier for other contributors.

HannesWell:
> Actually it would be better to find out why this fails, I think @HannesWell wanted to take a look at it.

That's right. Sorry I didn't have the time to do it yet. But I'll try to have a look today.


> Even if Hannes would take a look at it tomorrow, disabling it today would still be the right thing to do. First rule of handling flaky tests: Quarantine them, to avoid further test runs being flaky. Only _then_ fix them.

In my opinion the number one priority should be to fix the failing test. It may be a bug in the main code or, if one is lucky, the test itself is only poorly written. Just disabling the test maybe just hides a bug. And having a failing test is a stronger motivation to fix it than a ignored test. IMHO tests should only be ignored if one is certain, that the test itself is wrong but it should not (yet) be deleted.

Bananeweizen:
In my world the CI system is the ultimate source of truth. And the green master build is the highest priority of all. If that assertion is broken, every other change cannot be trusted anymore. That's why in my company projects, no changes can be merged onto a broken master by bitbucket policy. Since the next priority is having the brokenness repaired as soon as possible, we revert the last change without even looking for a bug fix. It's much better to have time for the real bug fix without everyone else waiting for you. That strict policy both gives everyone trust, and avoids a high level of stress in case the master ever breaks.

I would really like to see a similarly strict policy here. Black and white, no gray "yeah that one test is okayish, even if it fails". And the discussion already showed, that @HannesWell is under stress now, because everyone expects him to fix this fast now. That's just not useful. Disable the test, fix it with no pressure, enable it with the fix.

laeubi:
Disabling the test does not make the world better or the truth more true you just don't see it anymore ;-)

Also it is not black and white ... We have had for years a Jenkins CI (and also today) that only runs on linux. Then we moved on to Github and because that CI was done because of infra-crash I have added a github action. Still we sometimes have had failures there we don't see else where but over time this was stabilized as well.

Then some time back someone complained that there is a bug with windows path and windwos was added as an additional verification step, but showed that some test seem not to run under windows (most of them is fixed now), still we are certain that where Tycho is used most (windows) we are always green and we do not accidently introduce new problems on windows. Beside that I also added now the verification build of the platform build that gives an extra check we cover lots of use-cases.

So we have a lot of additional checks that give extra checks even if we could have only used the Jenkins CI ... but imposing a policy that we can not merge because of some failures on this extra checks seems not very suitable unless we have someone that is dedicated in analyze (and fix!) windows specific issues and even maybe GitHub windows runner specific issues.

Issue #1631, opened 2022-11-05
Tycho 3.0.0 download speed is much slower as 2.7.5 if mirrors are used


Phillipus:
A Tycho 2.7.5 build of our product takes just 3 and a half minutes compared with about 8 minutes on Tycho 3.0.0. Each build was timed with an empty `.m2` directory. The extra time is spent downloading the artefacts.

Is there more going on with 3.0.0, or perhaps some new settings that might improve speed?

laeubi:
> Is there more going on with 3.0.0, or perhaps some new settings that might improve speed?

We generally enable all settings that deliver best performance, but of course it depends on the actual setup.

What would be more interesting is how it performs after the inital download, because otherwise it even depends more on the setup and used servers.

Phillipus:
Hi, after the initial download it's a about the same time to compile and assemble in both versions. But for some reason, the download of all the dependencies is slower on 3.0.0.

Phillipus:
I ran it with the `-X` debug option and noticed that the slow down in 3.0.0 is at these points:

`[DEBUG] Writing P2 metadata for osgi.bundle,com.ibm.icu,67.1.0 to .m2\repository\p2\osgi\bundle\com.ibm.icu\67.1.0\com.ibm.icu-67.1.0-p2artifacts.xml...`

lppedd:
> the download of all the dependencies is slower on 3.0.0

Noticed the same. Also in my case the download progress report has disappeared in 3.0.0 >

lppedd:
@Phillipus have you tried with `-D=tycho.p2.transport=ecf`? Can't right now but I'll try that too.

laeubi:
> @Phillipus have you tried with `-D=tycho.p2.transport=ecf`? Can't right now but I'll try that too.

ECF is no longer supported. Neither maven nor Tycho is optimized for an empty repository, actually a lot of effort is taken into reusing stuff from the local repository to speed up things.

Phillipus:
More testing shows that it's not actually the download being slower but some other processing at the point of `Writing P2 metadata for....` only for p2 bundles. All the other support bundles that Tycho/Maven downloads are fast.

laeubi:
> only for p2 bundles.

What is a P2 bundle?

[...]

PR #1703, opened 2022-11-21
[draft] Add alternative http transport providers

laeubi:
To make the transport pluggable (for example to better analyze https://github.com/eclipse-tycho/tycho/issues/1631) this adds a simple http-transport abstraction so the core operation of fetching a file or performing a head operation.


laeubi:
Most of the code (beside the actual implementation) was now already integrated in Tycho, keep this as a reminder to possibly migrate the implementations as well...

laeubi:
Obsolete now

Issue #1689, opened 2022-11-18
Concurrency Issue with p2.artifacts.properties


laeubi:
@laeubi  does that help?

```
[2022-11-17T20:31:01.559Z] [INFO] ------------------------------------------------------------------------
[2022-11-17T20:31:01.559Z] [INFO] BUILD FAILURE
[2022-11-17T20:31:01.559Z] [INFO] ------------------------------------------------------------------------
[2022-11-17T20:31:01.559Z] [INFO] Total time:  02:19 min
[2022-11-17T20:31:01.559Z] [INFO] Finished at: 2022-11-17T21:31:01+01:00
[2022-11-17T20:31:01.559Z] [INFO] ------------------------------------------------------------------------
[2022-11-17T20:31:01.559Z] [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:3.0.0:validate-classpath (default-validate-classpath) on project com.godyo.p5.dashboard: Execution default-validate-classpath of goal org.eclipse.tycho:tycho-compiler-plugin:3.0.0:validate-classpath failed: java.io.FileNotFoundException: /p4/.m2/repository/.meta/p2-artifacts.properties (No such file or directory) -> [Help 1]
[2022-11-17T20:31:01.559Z] org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:3.0.0:validate-classpath (default-validate-classpath) on project com.godyo.p5.dashboard: Execution default-validate-classpath of goal org.eclipse.tycho:tycho-compiler-plugin:3.0.0:validate-classpath failed: java.io.FileNotFoundException: /p4/.m2/repository/.meta/p2-artifacts.properties (No such file or directory)
[2022-11-17T20:31:01.559Z]     at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:306)
[2022-11-17T20:31:01.559Z]     at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:211)
[2022-11-17T20:31:01.559Z]     at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:165)
[2022-11-17T20:31:01.559Z]     at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:157)
[...]

PR #1708, opened 2022-11-21
issue #1689: avoid file writing when no changes

Bananeweizen:
I don't really have an idea whether that is actually possible in practice or not. However, given that this method will update the file on disk in every case, we should avoid that if the old and new file content would be the same.
And even if this is theoretically not possible, the code should be merged to my mind.

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.core.shared/src/main/java/org/eclipse/tycho/p2/repository/FileBasedTychoRepositoryIndex.java
+++ b/tycho-bundles/org.eclipse.tycho.core.shared/src/main/java/org/eclipse/tycho/p2/repository/FileBasedTychoRepositoryIndex.java
@@ -110,6 +110,10 @@ public class FileBasedTychoRepositoryIndex implements TychoRepositoryIndex {
 
     @Override
     public synchronized void save() throws IOException {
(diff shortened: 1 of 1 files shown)


laeubi (approved):
Sounds reasonable!

laeubi:
There is only one sideffect: `save()` previously have called `reconcile()` that effectively reloads the files if it might be modified by another process.

Bananeweizen:
> There is only one sideffect: `save()` previously have called `reconcile()` that effectively reloads the files if it might be modified by another process.

Looking at that piece of code, I think the newly read file content (parsed into the "gavs" member) is actually never used anywhere, except when writing the file the next time. Therefore it should be okay to defer the reconcile method until the next time we really need to write something. It's basically a lazy reconcile, and that lazy is early enough for the write() method to be consistent.

laeubi:
@Bananeweizen I agree with your analysis, but it seems it still [breaks one test](https://ci.eclipse.org/tycho/job/tycho-github/job/PR-1708/1/testReport/junit/org.eclipse.tycho.p2resolver/LocalArtifactRepositoryFactoryTest/testLoad/) can you take a look?

Bananeweizen:
@laeubi fixed

laeubi:
Great!

PR #1714, opened 2022-11-22
Streamline "addPublic*" parameters

kwin:
Extend Javadoc

### DIFF
--- a/tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
+++ b/tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
@@ -46,8 +46,12 @@ import org.eclipse.equinox.p2.repository.artifact.IArtifactDescriptor;
 /**
  * Modifies the p2 metadata (<code>artifacts.xml</code>) to add PGP signatures for each included
  * artifact. Signatures are added as <code>pgp.signatures</code> property on the artifact metadata,
- * in armored form; and public keys of the signers are added as <code>pgp.publicKeys</code> property
- * on the repository metadata, in armored form.
+ * in armored form; and public keys of the signers are optionally added as
+ * <code>pgp.publicKeys</code> property on the repository metadata, in armored form.
+ * 
+ * @see <a href=
+ *      "https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2Fp2_pgp.html">Using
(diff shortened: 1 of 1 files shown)


mickaelistria (approved):

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
```diff
@@ -68,14 +72,22 @@ public class SignRepositoryArtifactsMojo extends AbstractGpgMojoExtension {
...
+     * If set to {@code true} adds the public key belonging to the signature to the repository
+     * metadata.
+     */
     @Parameter(defaultValue = "true")
     private boolean addPublicKeyToRepo;
 
-    @Parameter(defaultValue = "true")
-    private boolean addPublicKeysToArtifacts;
+    /**
+     * If set to {@code true} adds the public key belonging to the signature to each signed
```
lppedd, line 83: Is `{@code ...}` handled correctly in the generated docs? @kwin

laeubi:
I don't think adding a link to the eclipse-platform(!) documentation is the right choice here. This describes a very specific usecase that is most probably not applicable to most projects outside the eclipse world.

mickaelistria:
> This describes a very specific usecase that is most probably not applicable to most projects outside the eclipse world

It's p2 documentation, and PGP signatures are part of p2 metadata. Tycho has no use case with PGP signature in p2 metadata that makes sense outside of p2 world, so it seems very fine to reference the upstream/target documentation.

laeubi:
> It's p2 documentation, and PGP signatures are part of p2 metadata.

The page says

> **Eclipse** documentation - Current Release **Eclipse IDE 2022-09**

Further on

> **In Eclipse applications**, any plug-in can define some PGP public keys as trusted using the [org.eclipse.equinox.p2.engine.pgp](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/extension-points/org_eclipse_equinox_p2_engine_pgp.html) extension-point. Users can review and manage trusted keys in the **Opens the Trust preference pageInstall/Update > Trust preference page**.

So this all indicates (to me) that this is very specific on how Eclipse works and other application might behave differently, especially in terms of UI, key management and alike.

mickaelistria:
I'm inclined to merge it as it and replace the link to another one when a better one is available.

laeubi:
Just wanted to mention that the link is more how Eclipse (IDE) uses this currently than a generic one, but if you think its fine I won't veto it.

PR #1718, opened 2022-11-22
Report download speed to the caller

laeubi:
Currently Tycho do not report download statistics, but this can badly affect download performance because the P2 MirrorSelector heavily depends on this information to choose the next mirror to use for downloads.

This adds the necessary pieces to report proper download status.

Fix https://github.com/eclipse-tycho/tycho/issues/1631

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/DefaultHttpCacheConfig.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/DefaultHttpCacheConfig.java
@@ -27,6 +27,7 @@ public class DefaultHttpCacheConfig implements HttpCacheConfig, Initializable {
 
 	private boolean offline;
 	private boolean update;
+	private boolean interactive;
 
 	@Requirement
 	private LegacySupport legacySupport;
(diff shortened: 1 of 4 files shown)


laeubi:
You should see download messages again now, and yes we need a different implementation that especially supports superior caching performance for repeated builds.

Most of the time, one uses "static" update-sites that actually never change, still P2 queries those pages all the time and ask "is something changed?" ... even worse if something changed this of course requires another request to the server... especially for small and non existing files this can add considerable overhead (before even downloading a single artifact).

The new Tycho transport records HTTP caching headers and even uses eager caching for a certain amount of time so large target platforms resolve much faster. Still the user can force a "recheck" by passing `-U` option to maven.

lppedd:
@laeubi oh ok, this sounds good. I miss the "real time" download feedback tho. Maybe I can enhance `DownloadStatusOutputStream` to log every n seconds or n bytes? Not sure if we can decently bring back KB/s too).

laeubi:
Actually Tycho has previously contained some code to suppress some of the useless download status because they are filling the logs and actually don't give much valuable information to the user. Tycho now even respects the `-B` flag to suppress download messages at all...

lppedd:
> respect -B

@laeubi that is totally fine. With 2.7.5 I looked at log often in interactive mode tho, I think it can be useful (think inside of an IDE for example).

Issue #1466, opened 2022-10-04
PGP signing mojo re-adds new signature properties for already signed artifacts


mickaelistria:
If you're including in your p2 repo an artifact that already has some PGP signature (eg slf4j-api from Platform I-Build), and call the sign-p2-artifacts goal on it, then some new properties with the new signature and new key are added, leading to duplicated properties.
The mojo should provide a way to handle that case of already PGP-signed bundles:
* skip (keep original signature only)
* overwrite
* merge

merks:
Yes, I can confirm that this happens.   I've worked around various shortcomings and problems in the aggregator:

https://github.com/eclipse-cbi/p2repo-aggregator/issues/12

Here I add a fingerprint of the signing certificate to the artifact metadata, if there is one, and I record the original PGP signature/key:

https://github.com/eclipse-cbi/p2repo-aggregator/blob/ec3412a980ab15c3041e6b2792b3b82c04fa9dc9/plugins/org.eclipse.cbi.p2repo.aggregator.engine/src/org/eclipse/cbi/p2repo/aggregator/engine/MirrorGenerator.java#L446-L492

Then, after Tycho PGP signs all bundles and features, even ones that are jar-signed, I post process the metadata.  That let's me implement what you describe above:

https://github.com/eclipse-cbi/p2repo-aggregator/blob/ec3412a980ab15c3041e6b2792b3b82c04fa9dc9/plugins/org.eclipse.cbi.p2repo.aggregator.engine/src/org/eclipse/cbi/p2repo/aggregator/engine/SignatureCleaner.java#L71-L73

Some of this might be reusable (copy it and improve it):

https://github.com/eclipse-cbi/p2repo-aggregator/blob/ec3412a980ab15c3041e6b2792b3b82c04fa9dc9/plugins/org.eclipse.cbi.p2repo.aggregator.engine/src/org/eclipse/cbi/p2repo/aggregator/engine/SignatureCleaner.java#L170-L297

The processing here also lets me "reject" certain certificates, ones Tycho doesn't recognize as being treated as unsigned (even when running with a 17.0.5+8 JRE where jarsigner should detect that) and then I retain the new PGP signatures for those artifacts.

But my biggest problem now, one that I cannot work around, is that PGP signed features don't validate. 

- https://download.eclipse.org/staging/2022-12/buildInfo/archive/download.eclipse.org/staging/2022-12/index/org.eclipse.uml2.feature.jar_5.5.2.v20210228-1829.html
[...]

PR #1720, opened 2022-11-23
Enhance the PGP signing support.

merks:
Improve support for how already-PGP-signed artifacts are handled to support skip, replace, or merge.

Use SignedContentFactory for checking jar signatures which allows for skipping jars that are only signed by certificates anchored in Java's cacerts.

Support PGP signing features, and optionally also binary artifacts.

Support using Bouncy Castle for signing to improve performance and to allow signing to be done in parallel.  This also better support providing integration tests for the various sign-p2-artifacts mojo's options.

Ensure that only keys actually used by signatures are added to the repository and/or artifact properties.  Determining the default key is needed only when signing with Bouncy Castle, in which case it's determined by signing a document and checking which key is used; that's because the default key can be specified in the gpg.config so just listing the secret key fingerprints will not always correctly identify the correct default.

For testing purposes, support generating key information and loading it. Use that to provide integration tests for the new options as well as for existing options.

https://github.com/eclipse-tycho/tycho/issues/1466


laeubi:
@merks this looks like very good addition!

> Support using Bouncy Castle for signing to improve performance and to allow signing to be done in parallel. This also better support providing integration tests for the various sign-p2-artifacts mojo's options.

Maybe this can be made configurable? I think @mickaelistria has had some concerns that we might need "native support" here (from my side BC seems enough especially for testing...).

merks:
It is configurable and it is not the default.  Note too the comments about the performance.  It's generally 5 to 10 times faster and can be done in parallel...

> @merks this looks like very good addition!
> 
> > Support using Bouncy Castle for signing to improve performance and to allow signing to be done in parallel. This also better support providing integration tests for the various sign-p2-artifacts mojo's options.
> 
> Maybe this can be made configurable? I think @mickaelistria has had some concerns that we might need "native support" here (from my side BC seems enough especially for testing...).

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
```diff
@@ -30,213 +26,269 @@
...
+    @Parameter(defaultValue = "skip")
+    private PGPKeyBehavior pgpKeyBehavior;
+
+    /**
+     * Configure to {@code false} to sign using Bouncy Castle libraries rather than launching
+     * {@code gpg} for that purpose. This is much faster and it runs in parallel, so is very much
+     * faster.
+     */
+    @Parameter(defaultValue = "true")
+    private boolean signWithGnuPG;
```
laeubi, line 127: This should become a String parameter (e.g. `signer`)

Then you can have
```
@Component(role = SignerFactory.class)
private Map<String, SignerFactory>  factories;
```

and have a one or more impls e.g.
```
@Component(role = SignerFactory.class, hint="my")
MySignerFactory implements SignerFactory {

... 

}
```
The even users can supply their own implementation if desired later on...

merks:
@laeubi 

Note that GitGuardian:

image

doesn't like this file, which yes, is private key information but not that I or anyone cares about:

https://github.com/eclipse-tycho/tycho/pull/1720/files#diff-07625bf05da7c7fc0b04285c752c7f00a0f58e2f75164a530ed18ec6cb85d0ff

Do you think that might cause a problem?  I suppose I could rework the test/framework so that this is generated on the fly.  What do you think?

BTW, I'm really impressed by the integration test framework!!

laeubi:
> Do you think that might cause a problem? I suppose I could rework the test/framework so that this is generated on the fly. What do you think?

Generating a key on the fly seems suitable if e.g. BC offers such a function and then using https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createTempFile(java.lang.String,%20java.lang.String,%20java.nio.file.attribute.FileAttribute...) it should be quite safe.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
laeubi: Yes this is a bit bad (we probably ask maven-pgp-plugin to make this more open).

Maybe we still can use a string field with the values "bc" or "gpg" so if in the future we want to make it open for more implementations?

merks: Oh, that's a good idea!  Plan ahead for potential future flexibility.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
```diff
@@ -30,213 +26,269 @@
...
-            switch (artifact.getAttribute("classifier")) {
-            case "osgi.bundle":
-                subDir = "plugins";
-                break;
-            case "org.eclipse.update.feature": //Not yet signing features
-            case "binary": //Not yet signing binaries
-            default:
-                continue; // Skip signing
+
+            artifactRepository.getClass().getMethod("save").invoke(artifactRepository);
```
laeubi, line 192: Can you open a issue at P2 so we probably get a way to invoke save without reflection?

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
```diff
@@ -30,213 +26,269 @@
...
-                zipUnArchiver.extract();
-                artifactsJar.delete();
+        var signedContentFactory = serviceFactory.getService(SignedContentFactory.class);
+
+        var signer = newSigner(project);
+        var keys = KeyStore.create();
+
+        var artifactRepositoryManager = agent.getService(IArtifactRepositoryManager.class);
+        try {
+            var artifactRepository = (IFileArtifactRepository) artifactRepositoryManager
```
laeubi, line 168: If it helps her you probably can use `SimpleArtifactRepositoryIO` directly instead of using the manager to load/save repository.

laeubi (commented):
Some small remarks, beside that it would be good to add a note to the [RELEASE_NOTES.MD](https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md) about the new options.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
merks: The problem here is that if I cast to something with a save method, I get an error about class loader problems. There is already a public save method.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
merks: The useful methods for getting the artifact location using whatever mapping rules are defined is important so no, that doesn't help.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
```diff
@@ -30,213 +26,269 @@
...
+    @Parameter(defaultValue = "skip")
+    private PGPKeyBehavior pgpKeyBehavior;
+
+    /**
+     * Configure the signer used for PGP signing. Currently supported are {@code gpg} for launching
+     * the native {@code gpg} executable, and {@code bc} for using Bouncy Castle libraries. The
+     * latter is much faster and it can sign in parallel, so is very much faster.
+     */
+    @Parameter(defaultValue = "pgp")
+    private String signer;
```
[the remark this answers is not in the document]
merks: Are you sure.  I think that's risky because a great many options are ignored in this case and also note that even the bouncy castle thing only works because we can get the gpg keys, so without gpg one just cannot get secret keys from anywhere.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/BouncyCastleSigner.java
```diff
@@ -0,0 +1,271 @@
...
+
+    /**
+     * This creates a new key ring for this user ID with this passphrase.
+     */
+    public BouncyCastleSigner(String passphrase, String... userIDs) {
+        this.passphrase = passphrase;
+        try {
+            init(userIDs);
+        } catch (PGPException | IOException e) {
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
merks: okay

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/BouncyCastleSigner.java
```diff
@@ -0,0 +1,271 @@
...
+
+    public BouncyCastleSigner(String keyname, String passphrase, String publicKeys, String secretKeys) {
+        this.passphrase = passphrase;
+        this.publicKeys = publicKeys;
+        this.secretKeys = secretKeys;
+        this.keyname = keyname.toLowerCase();
+        try {
+            initPrivateKey();
+        } catch (PGPException | IOException e) {
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
merks: okay

tycho-gpg-plugin/pom.xml
```diff
@@ -55,11 +55,24 @@
 			<version>2.6.200</version>
 		</dependency>
 		<dependency>
-			<groupId>org.junit.jupiter</groupId>
-			<artifactId>junit-jupiter</artifactId>
-			<version>5.9.1</version>
-			<scope>test</scope>
+			<groupId>org.eclipse.tycho</groupId>
```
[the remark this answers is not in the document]
merks: Yes, that's nicer.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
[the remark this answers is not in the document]
merks: The property thing is a good idea though.  What name do you suggest for it?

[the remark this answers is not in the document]
merks: I changed it to "gpg" until you are really sure.  I added the property pgp.signer.

merks:
The latest push addresses all your constructive suggestions except I didn't change the default to "bc" yet, pending "are you really sure"?  😨  😱

mickaelistria:
Thanks. I'm not following much, but I see only goodness. As to BC vc GPG by default, I'd rather stick to GPG. Moving to BC is very likely to cause integration issues that would break existing jobs on Jenkins.

merks:
@mickaelistria Thanks, I know we're all very busy.  I have the same concern about changing defaults being potentially disruptive, but that being said, it's also **much** faster and likely no one will discover the possibility.   I'll leave it to @laeubi to make the final decision on that front.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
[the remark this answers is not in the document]
laeubi: We usually use something prefixed with "tycho" so probably `tycho.pgp.signer`.

As long as users can "switch back" with a configuration option it is fine, and i think BC is the future and should be improved/promoted as much as possible, so what do you think is missing? Without checking it in details I would expect not much more than specify a secret key (maybe a passphrase?) on the commandline or the maven config and then everything is signed with that key. There seem [some examples on how to accomplish it with BC](https://github.com/rodbate/bouncycastle-examples/blob/master/src/main/java/bcfipsin100/pgp/Sign.java), then it could be enhanced to probably read some of the "magic locations" where usually stuff is stored to automate this more, and we probabbly want to reuse the maven credentials encryption later on, but thats of course all optional for a first step.

So if you could use that in your test-setup I think it should be usable in "real code" as well?

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
laeubi: Can you share the error you get? It might be that we need to export a package from the `tycho-maven-plugin` that is `tycho-maven-plugin/src/main/resources/META-INF/maven/extension.xml` the maven way to share classes is really ugly :-\

laeubi:
@mickaelistria @merks I think if one can switch to "old" GPG by simply adding a configuration parameter the risk is quite manageable (and do not require much effort). And if we see issues with BC we should fix them, having operation system agnostic (and faster!) support will likley push acceptance of that feature!

mickaelistria:
What about trying to switch to BC after 2022-12 to keep things safe until then?

laeubi:
> What about trying to switch to BC after 2022-12 to keep things safe until then?

Tycho 4.0 will already only be released after 2022-12 most likely even after 2023-03 I assume.
Platform could even add the configuration option to disable BC beforehand if required as maven simply ignores unknown configuration options.

laeubi:
BTW license-check seems broken so this can be ignored...

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
merks: If I change it to this:

image

The following is what happens, with the key message being `org.eclipse.equinox.internal.p2.artifact.repository.simple.SimpleArtifactRepository cannot be cast to class org.eclipse.equinox.internal.p2.artifact.repository.simple.SimpleArtifactRepository (org.eclipse.equinox.internal.p2.artifact.repository.simple.SimpleArtifactRepository is in unnamed module of loader org.codehaus.plexus.classworlds.realm.ClassRealm @7b451bf4; org.eclipse.equinox.internal.p2.artifact.repository.simple.SimpleArtifactRepository is in unnamed module of loader org.codehaus.plexus.classworlds.realm.ClassRealm @4987383f)`:
```
08:57:53 [ERROR] Failed to execute goal org.eclipse.tycho:tycho-gpg-plugin:4.0.0-SNAPSHOT:sign-p2-artifacts (default) on project org.eclipse.tycho.maven.site: A type incompatibility occurred while executing org.eclipse.tycho:tycho-gpg-plugin:4.0.0-SNAPSHOT:sign-p2-artifacts: class org.eclipse.equinox.internal.p2.artifact.repository.simple.SimpleArtifactRepository cannot be cast to class org.eclipse.equinox.internal.p2.artifact.repository.simple.SimpleArtifactRepository (org.eclipse.equinox.internal.p2.artifact.repository.simple.SimpleArtifactRepository is in unnamed module of loader org.codehaus.plexus.classworlds.realm.ClassRealm @7b451bf4; org.eclipse.equinox.internal.p2.artifact.repository.simple.SimpleArtifactRepository is in unnamed module of loader org.codehaus.plexus.classworlds.realm.ClassRealm @4987383f)
08:57:53 [ERROR] -----------------------------------------------------
08:57:53 [ERROR] realm =    plugin>org.eclipse.tycho:tycho-gpg-plugin:4.0.0-SNAPSHOT
08:57:53 [ERROR] strategy = org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy
08:57:53 [ERROR] urls[0] = file:/D:/Users/merks/.m2/repository/org/eclipse/tycho/tycho-gpg-plugin/4.0.0-SNAPSHOT/tycho-gpg-plugin-4.0.0-SNAPSHOT.jar
08:57:53 [ERROR] urls[1] = file:/D:/Users/merks/.m2/repository/org/codehaus/plexus/plexus-utils/3.5.0/plexus-utils-3.5.0.jar
08:57:53 [ERROR] urls[2] = file:/D:/Users/merks/.m2/repository/org/codehaus/plexus/plexus-archiver/4.6.0/plexus-archiver-4.6.0.jar
08:57:53 [ERROR] urls[3] = file:/D:/Users/merks/.m2/repository/org/codehaus/plexus/plexus-io/3.4.0/plexus-io-3.4.0.jar
08:57:53 [ERROR] urls[4] = file:/D:/Users/merks/.m2/repository/commons-io/commons-io/2.11.0/commons-io-2.11.0.jar
08:57:53 [ERROR] urls[5] = file:/D:/Users/merks/.m2/repository/org/apache/commons/commons-compress/1.22/commons-compress-1.22.jar
08:57:53 [ERROR] urls[6] = file:/D:/Users/merks/.m2/repository/org/iq80/snappy/snappy/0.4/snappy-0.4.jar
08:57:53 [ERROR] urls[7] = file:/D:/Users/merks/.m2/repository/org/tukaani/xz/1.9/xz-1.9.jar
08:57:53 [ERROR] urls[8] = file:/D:/Users/merks/.m2/repository/com/github/luben/zstd-jni/1.5.2-5/zstd-jni-1.5.2-5.jar
08:57:53 [ERROR] urls[9] = file:/D:/Users/merks/.m2/repository/org/apache/maven/plugins/maven-gpg-plugin/3.0.1/maven-gpg-plugin-3.0.1.jar
08:57:53 [ERROR] urls[10] = file:/D:/Users/merks/.m2/repository/org/apache/maven/shared/maven-shared-utils/3.3.4/maven-shared-utils-3.3.4.jar
08:57:53 [ERROR] urls[11] = file:/D:/Users/merks/.m2/repository/org/codehaus/plexus/plexus-component-annotations/2.1.1/plexus-component-annotations-2.1.1.jar
08:57:53 [ERROR] urls[12] = file:/D:/Users/merks/.m2/repository/org/apache/maven/shared/maven-artifact-transfer/0.12.0/maven-artifact-transfer-0.12.0.jar
08:57:53 [ERROR] urls[13] = file:/D:/Users/merks/.m2/repository/org/apache/maven/shared/maven-common-artifact-filters/3.0.1/maven-common-artifact-filters-3.0.1.jar
08:57:53 [ERROR] urls[14] = file:/D:/Users/merks/.m2/repository/org/sonatype/sisu/sisu-inject-bean/1.4.2/sisu-inject-bean-1.4.2.jar
08:57:53 [ERROR] urls[15] = file:/D:/Users/merks/.m2/repository/org/sonatype/sisu/sisu-guice/2.1.7/sisu-guice-2.1.7-noaop.jar
08:57:53 [ERROR] urls[16] = file:/D:/Users/merks/.m2/repository/org/sonatype/plexus/plexus-sec-dispatcher/1.4/plexus-sec-dispatcher-1.4.jar
08:57:53 [ERROR] urls[17] = file:/D:/Users/merks/.m2/repository/org/sonatype/plexus/plexus-cipher/1.4/plexus-cipher-1.4.jar
08:57:53 [ERROR] urls[18] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.artifact.repository/1.4.500/org.eclipse.equinox.p2.artifact.repository-1.4.500.jar
08:57:53 [ERROR] urls[19] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.common/3.16.200/org.eclipse.equinox.common-3.16.200.jar
08:57:53 [ERROR] urls[20] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.registry/3.11.200/org.eclipse.equinox.registry-3.11.200.jar
08:57:53 [ERROR] urls[21] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.metadata/2.6.300/org.eclipse.equinox.p2.metadata-2.6.300.jar
08:57:53 [ERROR] urls[22] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.repository/2.6.200/org.eclipse.equinox.p2.repository-2.6.200.jar
08:57:53 [ERROR] urls[23] = file:/D:/Users/merks/.m2/repository/org/bouncycastle/bcprov-jdk18on/1.72/bcprov-jdk18on-1.72.jar
08:57:53 [ERROR] urls[24] = file:/D:/Users/merks/.m2/repository/org/bouncycastle/bcpg-jdk18on/1.72.2/bcpg-jdk18on-1.72.2.jar
08:57:53 [ERROR] urls[25] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.core.jobs/3.13.100/org.eclipse.core.jobs-3.13.100.jar
08:57:53 [ERROR] urls[26] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.preferences/3.10.100/org.eclipse.equinox.preferences-3.10.100.jar
08:57:53 [ERROR] urls[27] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.core/2.9.200/org.eclipse.equinox.p2.core-2.9.200.jar
08:57:53 [ERROR] urls[28] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.security/1.3.1000/org.eclipse.equinox.security-1.3.1000.jar
08:57:53 [ERROR] urls[29] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.osgi/3.18.100/org.eclipse.osgi-3.18.100.jar
08:57:53 [ERROR] urls[30] = file:/D:/Users/merks/.m2/repository/org/osgi/org.osgi.service.prefs/1.1.2/org.osgi.service.prefs-1.1.2.jar
08:57:53 [ERROR] urls[31] = file:/D:/Users/merks/.m2/repository/org/osgi/osgi.annotation/8.0.1/osgi.annotation-8.0.1.jar
08:57:53 [ERROR] urls[32] = file:/D:/Users/merks/.m2/repository/org/apache/maven/resolver/maven-resolver-util/1.9.1/maven-resolver-util-1.9.1.jar
08:57:53 [ERROR] urls[33] = file:/D:/Users/merks/.m2/repository/org/codehaus/plexus/plexus-interpolation/1.26/plexus-interpolation-1.26.jar
08:57:53 [ERROR] urls[34] = file:/D:/Users/merks/.m2/repository/org/codehaus/plexus/plexus-cipher/2.0/plexus-cipher-2.0.jar
08:57:53 [ERROR] urls[35] = file:/D:/Users/merks/.m2/repository/org/eclipse/tycho/sisu-equinox-launching/4.0.0-SNAPSHOT/sisu-equinox-launching-4.0.0-SNAPSHOT.jar
08:57:53 [ERROR] urls[36] = file:/D:/Users/merks/.m2/repository/org/apache/commons/commons-exec/1.3/commons-exec-1.3.jar
08:57:53 [ERROR] urls[37] = file:/D:/Users/merks/.m2/repository/org/eclipse/tycho/sisu-osgi-api/4.0.0-SNAPSHOT/sisu-osgi-api-4.0.0-SNAPSHOT.jar
08:57:53 [ERROR] urls[38] = file:/D:/Users/merks/.m2/repository/org/eclipse/tycho/tycho-targetplatform/4.0.0-SNAPSHOT/tycho-targetplatform-4.0.0-SNAPSHOT.jar
08:57:53 [ERROR] urls[39] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.frameworkadmin/2.2.100/org.eclipse.equinox.frameworkadmin-2.2.100.jar
08:57:53 [ERROR] urls[40] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.metadata.repository/1.4.100/org.eclipse.equinox.p2.metadata.repository-1.4.100.jar
08:57:53 [ERROR] urls[41] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.jarprocessor/1.2.300/org.eclipse.equinox.p2.jarprocessor-1.2.300.jar
08:57:53 [ERROR] urls[42] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.publisher/1.7.200/org.eclipse.equinox.p2.publisher-1.7.200.jar
08:57:53 [ERROR] urls[43] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.director/2.5.400/org.eclipse.equinox.p2.director-2.5.400.jar
08:57:53 [ERROR] urls[44] = file:/D:/Users/merks/.m2/repository/org/ow2/sat4j/org.ow2.sat4j.core/2.3.6/org.ow2.sat4j.core-2.3.6.jar
08:57:53 [ERROR] urls[45] = file:/D:/Users/merks/.m2/repository/org/ow2/sat4j/org.ow2.sat4j.pb/2.3.6/org.ow2.sat4j.pb-2.3.6.jar
08:57:53 [ERROR] urls[46] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.updatesite/1.2.300/org.eclipse.equinox.p2.updatesite-1.2.300.jar
08:57:53 [ERROR] urls[47] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.publisher.eclipse/1.4.100/org.eclipse.equinox.p2.publisher.eclipse-1.4.100.jar
08:57:53 [ERROR] urls[48] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.frameworkadmin.equinox/1.2.200/org.eclipse.equinox.frameworkadmin.equinox-1.2.200.jar
08:57:53 [ERROR] urls[49] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.simpleconfigurator.manipulator/2.2.0/org.eclipse.equinox.simpleconfigurator.manipulator-2.2.0.jar
08:57:53 [ERROR] urls[50] = file:/D:/Users/merks/.m2/repository/org/osgi/org.osgi.service.component/1.5.0/org.osgi.service.component-1.5.0.jar
08:57:53 [ERROR] urls[51] = file:/D:/Users/merks/.m2/repository/org/osgi/org.osgi.namespace.extender/1.0.1/org.osgi.namespace.extender-1.0.1.jar
08:57:53 [ERROR] urls[52] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.concurrent/1.2.100/org.eclipse.equinox.concurrent-1.2.100.jar
08:57:53 [ERROR] urls[53] = file:/D:/Users/merks/.m2/repository/org/eclipse/tycho/p2-maven-plugin/4.0.0-SNAPSHOT/p2-maven-plugin-4.0.0-SNAPSHOT.jar
08:57:53 [ERROR] urls[54] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.engine/2.7.500/org.eclipse.equinox.p2.engine-2.7.500.jar
08:57:53 [ERROR] urls[55] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.osgi.compatibility.state/1.2.700/org.eclipse.osgi.compatibility.state-1.2.700.jar
08:57:53 [ERROR] urls[56] = file:/D:/Users/merks/.m2/repository/org/eclipse/tycho/sisu-osgi-connect/4.0.0-SNAPSHOT/sisu-osgi-connect-4.0.0-SNAPSHOT.jar
08:57:53 [ERROR] urls[57] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.touchpoint.natives/1.4.400/org.eclipse.equinox.p2.touchpoint.natives-1.4.400.jar
08:57:53 [ERROR] urls[58] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.touchpoint.eclipse/2.3.300/org.eclipse.equinox.p2.touchpoint.eclipse-2.3.300.jar
08:57:53 [ERROR] urls[59] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.garbagecollector/1.2.0/org.eclipse.equinox.p2.garbagecollector-1.2.0.jar
08:57:53 [ERROR] urls[60] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.simpleconfigurator/1.4.100/org.eclipse.equinox.simpleconfigurator-1.4.100.jar
08:57:53 [ERROR] urls[61] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.app/1.6.200/org.eclipse.equinox.app-1.6.200.jar
08:57:53 [ERROR] urls[62] = file:/D:/Users/merks/.m2/repository/org/apache/felix/org.apache.felix.scr/2.2.4/org.apache.felix.scr-2.2.4.jar
08:57:53 [ERROR] urls[63] = file:/D:/Users/merks/.m2/repository/org/codehaus/mojo/animal-sniffer-annotations/1.9/animal-sniffer-annotations-1.9.jar
08:57:53 [ERROR] urls[64] = file:/D:/Users/merks/.m2/repository/org/bouncycastle/bcpg-jdk15on/1.70/bcpg-jdk15on-1.70.jar
08:57:53 [ERROR] urls[65] = file:/D:/Users/merks/.m2/repository/org/bouncycastle/bcprov-jdk15on/1.70/bcprov-jdk15on-1.70.jar
08:57:53 [ERROR] urls[66] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.core.runtime/3.26.0/org.eclipse.core.runtime-3.26.0.jar
08:57:53 [ERROR] urls[67] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.core.contenttype/3.8.200/org.eclipse.core.contenttype-3.8.200.jar
08:57:53 [ERROR] urls[68] = file:/D:/Users/merks/.m2/repository/biz/aQute/bnd/biz.aQute.bndlib/6.3.1/biz.aQute.bndlib-6.3.1.jar
08:57:53 [ERROR] urls[69] = file:/D:/Users/merks/.m2/repository/biz/aQute/bnd/biz.aQute.bnd.util/6.3.1/biz.aQute.bnd.util-6.3.1.jar
08:57:53 [ERROR] urls[70] = file:/D:/Users/merks/.m2/repository/org/osgi/org.osgi.service.repository/1.1.0/org.osgi.service.repository-1.1.0.jar
08:57:53 [ERROR] urls[71] = file:/D:/Users/merks/.m2/repository/org/osgi/org.osgi.util.function/1.2.0/org.osgi.util.function-1.2.0.jar
08:57:53 [ERROR] urls[72] = file:/D:/Users/merks/.m2/repository/org/osgi/org.osgi.util.promise/1.2.0/org.osgi.util.promise-1.2.0.jar
08:57:53 [ERROR] urls[73] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.director.app/1.2.200/org.eclipse.equinox.p2.director.app-1.2.200.jar
08:57:53 [ERROR] urls[74] = file:/D:/Users/merks/.m2/repository/org/eclipse/platform/org.eclipse.equinox.p2.repository.tools/2.3.100/org.eclipse.equinox.p2.repository.tools-2.3.100.jar
08:57:53 [ERROR] urls[75] = file:/D:/Users/merks/.m2/repository/commons-codec/commons-codec/1.15/commons-codec-1.15.jar
08:57:53 [ERROR] Number of foreign imports: 1
08:57:53 [ERROR] import: Entry[import  from realm ClassRealm[project>org.eclipse.tycho.maven:org.eclipse.tycho.maven.parent:1.0.0-SNAPSHOT, parent: ClassRealm[maven.api, parent: null]]]
08:57:53 [ERROR] 
```
I've seen various forms of this problem over the last week.

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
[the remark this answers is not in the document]
merks: I will change it to tycho.pgp.signer for the next force push.

I will further investigate if the one use of gpg can be eliminated by reading the underlying secrete keys and the public keystore directly; I already have code for the latter.  This would have the advantage that one doesn't have to configure or start the gpg agent making it significantly easier to set up a build.  But let's do that as a separate step and if that's successful then we can change the default to bc.   I.e., I'd very much like to get past this PR because I expect I'm going to need it (the ability to sign features) to deal with the SimRel 2022-12 signing problems.  

Okay?

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
laeubi: Can you try adding `org.eclipse.equinox.internal.p2.artifact.repository.simple` to the exported packages of https://github.com/eclipse-tycho/tycho/blob/master/tycho-maven-plugin/src/main/resources/META-INF/maven/extension.xml ?

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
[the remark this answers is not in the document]
laeubi: Agree, lets go for it!

laeubi (approved):
Looks good

tycho-gpg-plugin/src/main/java/org/eclipse/tycho/gpg/SignRepositoryArtifactsMojo.java
merks: You're awesome! 🥇

laeubi:
Thanks @merks for improving this!

merks:
@laeubi Thank you for the fast and constructive feedback!

Issue #1713, opened 2022-11-22
Support standard PGP signing workflow


laeubi:
Currently we have the [tycho-gpg:sign-p2-artifacts](https://tycho.eclipseprojects.io/doc/latest/tycho-gpg-plugin/sign-p2-artifacts-mojo.html) to sign artifacts after-the-fact. The more usual approach would be that one uses default [maven-pgp-plugin](https://maven.apache.org/plugins/maven-gpg-plugin/) to sign an artifact as part of the build.

Similar to 
- https://github.com/eclipse-tycho/tycho/issues/1482

this means we need to update the p2 metadata (currently this does not include neither PGP signatures nor checksums), [the usual configuration](https://maven.apache.org/plugins/maven-gpg-plugin/usage.html) is that one binds the pgp-maven-plugin to the verify phase so it happens before deploy but after package, so in this case we might need to use an own mojo than a default lifecycle-binding.

PR #1722, opened 2022-11-24
Propagate log to ProxySignerWithPublicKeyAccess

akurtakov:
Fixes nasty:
```
[ERROR] Failed to execute goal
org.eclipse.tycho:tycho-gpg-plugin:4.0.0-SNAPSHOT:sign-p2-artifacts
(pgpsigner) on project org.eclipse.linuxtools.releng-site: Execution
pgpsigner of goal
org.eclipse.tycho:tycho-gpg-plugin:4.0.0-SNAPSHOT:sign-p2-artifacts
failed: Cannot invoke
"org.apache.maven.plugin.logging.Log.info(java.lang.CharSequence)"
because the return value of
"org.apache.maven.plugins.gpg.ProxySignerWithPublicKeyAccess.getLog()"
is null -> [Help 1]
```

### DIFF
--- a/tycho-gpg-plugin/src/main/java/org/apache/maven/plugins/gpg/AbstractGpgMojoExtension.java
+++ b/tycho-gpg-plugin/src/main/java/org/apache/maven/plugins/gpg/AbstractGpgMojoExtension.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2021 Red Hat Inc. and others.
+ * Copyright (c) 2021, 2022 Red Hat Inc. and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -18,7 +18,9 @@ public abstract class AbstractGpgMojoExtension extends AbstractGpgMojo {
     @Override
     protected ProxySignerWithPublicKeyAccess newSigner(MavenProject project)
             throws MojoExecutionException, MojoFailureException {
-        return new ProxySignerWithPublicKeyAccess(super.newSigner(project));
+        ProxySignerWithPublicKeyAccess signer = new ProxySignerWithPublicKeyAccess(super.newSigner(project));
+        signer.setLog(getLog());
+        return signer;
     }
 
 }


akurtakov:
Is @merks patch supposed to go in soon? I'm trying to push gpg signing in more projects and this is a blocker for now https://github.com/eclipse-linuxtools/org.eclipse.linuxtools/pull/137 . If that would not delay more than few days that should be fine.

akurtakov:
If I read the other PR comments proper - I can rebase my patch on top of master when the other PR is in.

merks:
Yes I hope it's done very very soon.  Note that I would change it here so that it's passed to the bouncy castle signer has well.

image

This way there will be nothing left to do.  Okay?

merks:
The change is here now:

https://github.com/eclipse-tycho/tycho/pull/1720/files#diff-9885bd65123412650a0560d9ec1fc4fbb4557a0598ba91160f12d00159ff33c2

akurtakov:
Thanks. Closing in favor of https://github.com/eclipse-tycho/tycho/pull/1720

laeubi:
@akurtakov I wonder if you plan to replace code-signing completely?
Then 
- https://github.com/eclipse-tycho/tycho/issues/1713

might be interesting for you, and linuxtools could be a good incubator for that feature. Or are you mostly want to sign third party artifacts?

akurtakov:
I have no plans to replce code signing completely. The goal is to get rid of [ancient Orbit used](https://github.com/eclipse-linuxtools/org.eclipse.linuxtools/blob/master/releng/org.eclipse.linuxtools.target/linuxtools-e4.26.target#L46) which is signed with expired artifacts. Once this is done we can think for other enhancements. Small steps :)

merks:
@akurtakov  FYI, one of the new options is `skipIfJarsignedAndAnchored`  if you set this to true and `skipIfJarsigned` to false, and you make sure you run the build with the latest Java 11 or 17, the logic will automatically determine that the JRE doesn't consider certain bundles signed and will PGP sign then.  This way you can avoid manually "forcing" the signatures via a long manually maintained list.

akurtakov:
Interesting, I would definetely consider this one. Although in this case both javassist and cyberneko are up for update so I have to get the pgp signing proper to update.

merks:
I'm not sure I follow.  Note by way of example, that bcpg is jar signed, but it is not by a rooted certificate in the cacerts, it's a JCA certificate,  so it's automatically PGP signed with this new option.  Maybe that applies for these things you mention too...

laeubi:
> but it is not by a rooted certificate in the cacerts

while one need to take into consideration that the `cacerts` might variy among user installations.

merks:
Yes, that's why I mention using the latest JDK because these things evolved and old roots are removed while new roots are added.  Also now, particular algorithms can be selectively blocked for specific date ranges.   Given we ship EPP with Temurin 17 latest, that's a good one to use...

Issue #1724, opened 2022-11-24
Support easier configuration for PGP signing


merks:
In a free style job, PGP signing is configured like this:

![image](https://user-images.githubusercontent.com/208716/203795115-7ea258b3-a680-4146-a641-146409b8b419.png)

Then there is a bunch of boilerplate stuff like this
```
gpg --batch --import "${KEYRING}"
for fpr in $(gpg --list-keys  --with-colons  | awk -F: '/fpr:/ {print $10}' | sort -u);
do
  echo -e "5\ny\n" |  gpg --batch --command-fd 0 --expert --edit-key $fpr trust;
done

# https://bugs.eclipse.org/bugs/show_bug.cgi?id=573866
mkdir -p ~/.gnupg
cat >> ~/.gnupg/gpg.conf <<EOL
use-agent 
pinentry-mode loopback
EOL
cat >> ~/.gnupg/gpg-agent.conf <<EOL
allow-loopback-pinentry
EOL
echo RELOADAGENT | gpg-connect-agent
echo "test" | gpg --clearsign --passphrase "${KEYRING_PASSPHRASE}"
```

But it's really easy to read the *.asc containing the secret keys with BouncyCastle directly.  I expect that we can even just use the default key because I think there is only one.  So with a command line option like -Dtycho.pgp.signer.bc.secretKeys=${KEYRING_PASSPHRASE} we should be able to do PGP signing without using gpg at all.  The certainly works in my tests locally...

So in a pipeline, maybe we would only need to do this:
```
      steps {
        withCredentials([file(credentialsId: 'secret-subkeys.asc', variable: 'KEYRING')]) {
          withCredentials([string(credentialsId: 'gpg-passphrase', variable: 'KEYRING_PASSPHRASE')]) {
            sh '''
            mvn clean verify \
              -Dmaven.repo.local=$WORKSPACE/.m2/repository \
              -Psign -Pgpg-sign \
              -Dgpg.passphrase="${KEYRING_PASSPHRASE}" \
              -Dtycho.pgp.signer.bc.secretKeys=${KEYRING_PASSPHRASE}
              -Dproject.build.sourceEncoding=UTF-8 
            '''
          }
        }
      }
```

I'm not really sure we can confirm if this type of approach works on Eclipse infrastructure without first implementing it and trying it.  But, as I said, this works for me locally....

laeubi:
[...]

PR #1726, opened 2022-11-24
Support initializing bouncy castle signer by directly loaded secret keys

merks:
This avoids needing the native gpg keybox to be initialized by importing the secret keys file and then later fetching that information from gpg, simplifying configuration on a build server.

https://github.com/eclipse-tycho/tycho/issues/1724


laeubi (approved):
Looks good.

laeubi:
Github seems wrecked at the moment but Jenkins is fine.

PR #1730, opened 2022-11-25
Disable VerifierServiceImplTest

laeubi:
FYI @Bananeweizen hopefully we can get some idea why this fails on github...

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/VerifierServiceImpl.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/VerifierServiceImpl.java
@@ -13,10 +13,11 @@
  *******************************************************************************/
 package org.eclipse.tycho.core;
 
-import java.io.ByteArrayOutputStream;
+import java.io.OutputStream;
 import java.net.URI;
 import java.util.Collection;
-import java.util.Iterator;
+import java.util.Set;
+import java.util.function.BiConsumer;
 
 import org.codehaus.plexus.component.annotations.Component;
 import org.codehaus.plexus.component.annotations.Requirement;
@@ -79,11 +80,14 @@ public class VerifierServiceImpl implements VerifierService {
             final IArtifactRepository artifactRepository, Logger logger) {
         final IQueryResult<IInstallableUnit> collector = metadata.query(QueryUtil.ALL_UNITS, monitor);
         boolean valid = true;
-        for (Iterator<IInstallableUnit> iterator = collector.iterator(); iterator.hasNext();) {
-            IInstallableUnit iu = iterator.next();
+        Set<IInstallableUnit> set = collector.toSet();
+        logger.debug("Verify content of " + set.size() + " units...");
+        for (IInstallableUnit iu : set) {
             final Collection<IArtifactKey> artifacts = iu.getArtifacts();
             for (IArtifactKey key : artifacts) {
-                valid &= verifyArtifactExists(key, artifactRepository, logger);
+                boolean verifyArtifactExists = verifyArtifactExists(key, artifactRepository, logger);
+                logger.debug("Verify " + key + " exits: " + verifyArtifactExists);
+                valid &= verifyArtifactExists;
             }
         }
         return valid;
@@ -103,12 +107,14 @@ public class VerifierServiceImpl implements VerifierService {
 
         IQueryResult<IArtifactKey> allKeys = repository
                 .query(new ExpressionMatchQuery<>(IArtifactKey.class, ExpressionUtil.TRUE_EXPRESSION), null);
-        for (Iterator<IArtifactKey> keyIt = allKeys.iterator(); keyIt.hasNext();) {
-            IArtifactKey key = keyIt.next();
-
+        Set<IArtifactKey> set = allKeys.toSet();
+        logger.debug("Verify content of " + set.size() + " artifacts...");
+        for (IArtifactKey key : set) {
             IArtifactDescriptor[] descriptors = repository.getArtifactDescriptors(key);
             for (IArtifactDescriptor descriptor : descriptors) {
-                valid &= verifyArtifactContent(repository, logger, descriptor);
+                boolean verifyArtifactContent = verifyArtifactContent(repository, logger, descriptor);
+                logger.debug("Verify artifact content " + descriptor + ": " + verifyArtifactContent);
+                valid &= verifyArtifactContent;
             }
         }
         return valid;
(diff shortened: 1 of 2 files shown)


Bananeweizen:
This doesn't fail on github. It fails on _Windows_. That's exactly why I wanted it disabled. I have to temporarily disable it for every local build, and I've therefore stopped looking into bugfixes.

laeubi:
See this PR it now fails on Ubuntu as well (but not on the Jenkins) ... so it seem to depend on some system state. I can't reproduce it locally, if you can it would be great to attach a debugger to the test:
https://maven.apache.org/surefire/maven-surefire-plugin/examples/debugging.html

and see why it does validate while it should not.

laeubi:
And can you maybe try passing  `-Dtycho.localArtifacts=ignore` and/or building with an empty m2 rep (`-Dmaven.repo.local=testme`) makes any difference?

laeubi:
In the good case the output should look like this:

```
Checking metadata from 'tycho-p2/tycho-p2-repository-plugin/src/test/resources/repositories/tampered_file/' and artifacts from 'tycho-p2/tycho-p2-repository-plugin/src/test/resources/repositories/tampered_file/'
Verify content of 11 units...
Verify osgi.bundle,jarsigning,0.0.1.201109191414 exits: true
Verify org.eclipse.update.feature,jarsigning.feature,0.0.1.201109191414 exits: true
Verify content of 2 artifacts...
OK
Verify artifact content canonical: org.eclipse.update.feature,jarsigning.feature,0.0.1.201109191414: true
Problems downloading artifact: osgi.bundle,jarsigning,0.0.1.201109191414.
  File has invalid content:/tmp/signatureFile4781508675350802476.jar
    Invalid content:Dummy.class: Dummy.class
```

When the test fails on Github it show:
```
 Checking metadata from 'file:/home/runner/work/tycho/tycho/tycho-p2/tycho-p2-repository-plugin/src/test/resources/repositories/tampered_file/' and artifacts from 'file:/home/runner/work/tycho/tycho/tycho-p2/tycho-p2-repository-plugin/src/test/resources/repositories/tampered_file/'
Verify content of 11 units...
Verify osgi.bundle,jarsigning,0.0.1.201109191414 exits: true
Verify org.eclipse.update.feature,jarsigning.feature,0.0.1.201109191414 exits: true
Verify content of 2 artifacts...
Verify artifact content canonical: org.eclipse.update.feature,jarsigning.feature,0.0.1.201109191414: true
Verify artifact content canonical: osgi.bundle,jarsigning,0.0.1.201109191414: true
The integrity of the metadata repository 'file:/home/runner/work/tycho/tycho/tycho-p2/tycho-p2-repository-plugin/src/test/resources/repositories/tampered_file/' and artifact repository 'file:/home/runner/work/tycho/tycho/tycho-p2/tycho-p2-repository-plugin/src/test/resources/repositories/tampered_file/' has been verified successfully) expected:<false> but was:<true>
```

laeubi:
@merks do you see the failure on windows as well? Any idea why the signature seems not to be checked here (sometimes)?

laeubi:
By the way this case shows that it is _somehow_ possible to bypass code-sign verification in P2 what itself is much more scary than a failing test... even though it should be detected at runtime somehow.

laeubi:
FYI I have now disabled the Github Workflow until we find out whats causing this.

merks:
@laeubi It looks ready to merge.

laeubi:
@merks this code jsut adds some debug infos but the problem remains that `VerifierServiceImplTest` fails in Github (and @Bananeweizen reposrte it also fails for him on windows), So if you probably can cehck if it fails for you on windows as well or maybe have an idea what can cause P2 to ignore sigantures that would be great.

merks:
Sorry I posted on the wrong PR.  I meant that comment for here:

https://github.com/eclipse-tycho/tycho/pull/1734

laeubi:
@akurtakov @mickaelistria any idea what might causing this?

merks:
This test works for me on Windows locally:

image

laeubi:
@Bananeweizen do the test also fails for you locally? If not I suspect this to be some caching issue in Github action (if I remember correctly artifacts are only checked on first download?),

Bananeweizen:
Nope. This branch fails similar like before, just the new debug output is there AFAIKS:
```
[ERROR] Failures:
[ERROR]   VerifierServiceImplTest.testFileRepositoryWithTamperedArtifact:81 The repository should not verify! (messages below)
Checking metadata from 'file:/C:/dev/tycho/git/tycho/tycho-p2/tycho-p2-repository-plugin/src/test/resources/repositories/tampered_file/' and artifacts from 'file:/C:/dev/tycho/git/tycho/tycho-p2/tycho-p2-repository-plugin/src/test/resources/repositories/tampered_file/'
Verify content of 11 units...
Verify osgi.bundle,jarsigning,0.0.1.201109191414 exits: true
Verify org.eclipse.update.feature,jarsigning.feature,0.0.1.201109191414 exits: true
Verify content of 2 artifacts...
OK
Verify artifact content canonical: org.eclipse.update.feature,jarsigning.feature,0.0.1.201109191414: true
OK
Verify artifact content canonical: osgi.bundle,jarsigning,0.0.1.201109191414: true
The integrity of the metadata repository 'file:/C:/dev/tycho/git/tycho/tycho-p2/tycho-p2-repository-plugin/src/test/resources/repositories/tampered_file/' and artifact repository 'file:/C:/dev/tycho/git/tycho/tycho-p2/tycho-p2-repository-plugin/src/test/resources/repositories/tampered_file/' has been verified successfully) expected:<false> but was:<true>
[INFO]
[ERROR] Tests run: 4, Failures: 1, Errors: 0, Skipped: 0
```

mickaelistria:
Are those using different Java versions? jarsigning relies on some algorithms and it might have been assumed that some algorithms are provided and verified by the JRE. However, the policiy in the JRE has been to remove algorithms as they become considered unsafe. So maybe it's the case here that from a version of Java to another, the available crypto/signature algorithms have been removed and that jarsigner check default to "safe" if it cannot verify anymore.

merks:
@mickaelistria  You hit the nail on the head!  Switching from 17.0.4 to 17.0.5 it fails:

image

So indeed, it treats that jar as unsigned and therefore does not detected tampering.  The world is a safer place?

laeubi:
@mickaelistria @merks @Bananeweizen thanks for giving the right hints and test things, this for me proves the test has (or will) become useless and disable this one unless someone steps up to update it to whatever has to be used today.

> So indeed, it treats that jar as unsigned and therefore does not detected tampering. The world is a safer place?

Even though it first might sound strange it actually is! 
Just recap what is detected here: One file in that jar is/was tampered. Now assume you think the algorithm for this detection is no longer safe, it is better to tell the user "I can't verify this" (and for example in a restricted environment prevent the code from being used) than to verify it and tell the user everything is fine, because the tampering would probably never detected because the algorithm was broken and a bad file is in there.

The rational is, that there actually should only be two cases:

1. In your environment only signed (and trusted) jars are allowed, this will effectively prevent an tampered or insecure jars to being used
2. Signing is optional, in which case an attacker can simply **strip all signatures** to get the same effect (the item will be treated as unsigned), so the knowledge of one file tapered is not really adding much.

Issue #1728, opened 2022-11-24
gpg signing works for the first p2 repo build but not for the second one


akurtakov:
Look at https://ci.eclipse.org/linuxtools/job/linuxtools/job/PR-137/15/console :

* [INFO] --- tycho-gpg-plugin:4.0.0-SNAPSHOT:sign-p2-artifacts (pgpsigner) @ org.eclipse.linuxtools.releng-site --- completes
* [INFO] --- tycho-gpg-plugin:4.0.0-SNAPSHOT:sign-p2-artifacts (pgpsigner) @ org.eclipse.linuxtools.docker-site ---
gpg: signing failed: Bad passphrase
gpg: signing failed: Bad passphrase
fails

I've tried locally and both succeeds but not on Jenkins so if someone has ideas what could be the reason or how to debug it?
The relevant part from the errorlog :
```
Caused by: org.apache.maven.plugin.MojoExecutionException: Exit code: 2
    at org.apache.maven.plugins.gpg.ProxySignerWithPublicKeyAccess.generateSignature (ProxySignerWithPublicKeyAccess.java:109)
    at org.eclipse.tycho.gpg.SignRepositoryArtifactsMojo.handle (SignRepositoryArtifactsMojo.java:274)
    at org.eclipse.tycho.gpg.SignRepositoryArtifactsMojo.lambda$execute$0 (SignRepositoryArtifactsMojo.java:186)
    at java.util.stream.ForEachOps$ForEachOp$OfRef.accept (ForEachOps.java:183)
```

akurtakov:
@merks considering your latest work maybe you have some interest in this one?

merks:
I'll have a look tomorrow. 

Have you done all the magical things described here?

https://github.com/eclipse-tycho/tycho/issues/1724

E.g., maybe `-Dgpg.passphrase="${KEYRING_PASSPHRASE}"` is missing though I see it there, so the message suggest it's not actually correct though...

merks:
The only thing I see that's missing/different that I see in other places where it works is specifying the `<keyname>b6d3ab9bcc641282</keyname>` like here:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/c6cc4ae1713b3023b97dfecf0cae20b2cdac575b/eclipse.platform.releng.tychoeclipsebuilder/pom.xml#L46

In your build log I see this:
```
[...]

PR #1734, opened 2022-11-26
For PGP signing, get the IProvisioningAgent from the service factory

merks:
Injecting it as an @Component may cause multiple Equinox frameworks to be instantiated.

https://github.com/eclipse-tycho/tycho/issues/1728


laeubi (approved):

merks:
Thanks. 👍

merks:
@laeubi FYI, this did eliminate the stack traces I saw previously.  Woo hoo.

PR #1740, opened 2022-11-26
Remove obsolete Eclipse metadata files

HannesWell:
With the recent migration of PDE Plugin projects and recent versions of M2E all the Eclipse metadata seems to be generated automatically and therefore are not necessary in git anymore.
This should also prevent unstaged changes in these metadata that occasionally occur and are especially annoying on the first project setup.

I tested it and the same projects are imported and checking a few projects also showed that the generated metadata are the same. Also I get a error free workspace when I start with an empty workspace and start Tycho's Oomph setup on a cleaned Tycho git repo.

@laeubi what do you think?


merks:
Won't there be warnings about missing project encoding preferences?

HannesWell:
> Won't there be warnings about missing project encoding preferences?

I think the resources preferences are then generated automatically, but I can check that.

HannesWell:
> Won't there be warnings about missing project encoding preferences?

Just checked it and there are non because `org.eclipse.core.resources.prefs` are generated with content like:
```
eclipse.preferences.version=1
encoding//src/main/java=UTF-8
encoding//src/main/resources=UTF-8
encoding/<project>=UTF-8
```

So I think this is fine. Does anybody has more concerns? If not I think this is ready to go.

merks:
Thanks for checking.  👍

laeubi:
I would like to @mickaelistria to gechk this becuase last time he was the one adding the metadata back. The only thing I like to ensure is that there is no back-and-forth because that was quite annoying last time when switching branches...

HannesWell:
> I would like to @mickaelistria to gechk this becuase last time he was the one adding the metadata back. The only thing I like to ensure is that there is no back-and-forth because that was quite annoying last time when switching branches...

Agree. @mickaelistria could you please test this PR?

mickaelistria:
As long as "File > Open Projects from Filesystem" works without need for extra connector beyond what's shipped by m2e in the related EPP products, I don't mind seeing some some metadata files removed.

HannesWell:
> As long as "File > Open Projects from Filesystem" works without need for extra connector beyond what's shipped by m2e in the related EPP products, I don't mind seeing some some metadata files removed.

Checked that as well and that works. Nevertheless there is a still a pop-up about missing connectors for some other plugins, but that should be addressed by #1747.
Personally I would use `Import ... > Maven >Existing Maven Projects` or the Tycho Oomph setup because then I don't have to deselect the test projects.

So then I think this is ready after the build has succeeded?

HannesWell:
The failing tests on Windows also fail on the master, so they are likely unrelated. Submitting.

Bananeweizen:
This change has an unfortunate side effect. All projects with zero .prefs files in git now have their settings directory as "unversioned" folder in the package explorer, because that directory is only created on first run. In other projects I typically check in a .gitignore which would prohibit adding further files, but makes the directory itself being versioned. Any thoughts on that?
grafik

laeubi:
Can we simply ignore `.settings` folder?

Bananeweizen:
> Can we simply ignore `.settings` folder?

BIG NO. At least not generally. You can ignore it exactly in the projects I mentioned above (where no .prefs exist). In all JTD projects you _must_ have at least JDT settings under version control, otherwise people see totally different errors/warnings/Java EE levels, as most of those are just not contained in a POM, and therefore cannot be managed by m2e. Those settings then depend on Eclipse versions, workspace preferences, installed JREs etc. I committed quite a lot of new specific preferences to the JDT settings of Tycho some weeks ago just because we already had exactly that broken situation.

HannesWell:
Some Tycho projects like `sisu-equinox-embedder` already have a dedicated .gitignore with content like the following to 'override' the global rule although no 'global' rule to ignore settings exists yet:
```
# override global rule
!/.settings/

# only include specified files
/.settings/*
!/.settings/org.eclipse.jdt.core.prefs
!/.settings/org.eclipse.jdt.ui.prefs
```
If we exclude the .settings folder 'globally' it should only be done for the first three level so that one can easaly add .settings for test cases if it is necessary (I assume that one strips down the test cases as far as possible anyway before submission and that if settings are present, they are required):
```
/.settings
/*/.settings
/*/*/.settings
```

On the other hand, the "unversioned" .settings folder is a visual disturbance yes, but technically I see no problem or do you see one?

PR #1747, opened 2022-11-27
Fix some M2E 'Maven Project Build Lifefycle Mapping Problem' warnings

HannesWell:
Fix some M2E 'Maven Project Build Lifefycle Mapping Problem' warnings and replace the `maven-bundle-plugin` by the `bnd-maven-plugin`.
The former uses the latter under the hood so we should better use the 'original' to get enhancements quicker.


laeubi:
@HannesWell I think we can even remove the whole osgi manifest generation all together I think.

- sisu-osgi/sisu-equinox-launching/pom.xml
- sisu-osgi/sisu-osgi-api/pom.xml

it does not really make sense for sisu to have OSGi headers as it does not really support OSGi for our use-cases.

- tycho-embedder-api/pom.xml

with OSGi-connect, we should not need this anymore and just could remove any imports from the static MANIFEST.MF files on tycho-core-shared and tycho-embedder-shared parts

HannesWell:
Sounds good. Removed all manifest generating executions and the therefore necessary `maven-jar-plugin` executions from
- sisu-equinox-launching
- sisu-osgi-api
- tycho-embedder-api

lets see if the tests still pass.

> with OSGi-connect, we should not need this anymore and just could remove any imports from the static MANIFEST.MF files on tycho-core-shared and tycho-embedder-shared parts

Which projects are you referring to exactly? I did not find them?
Furthermore I wonder if we should do the removal in a dedicated PR?

laeubi:
> Which projects are you referring to exactly? I did not find them?



> Which projects are you referring to exactly? I did not find them?

https://github.com/eclipse-tycho/tycho/blob/master/tycho-bundles/org.eclipse.tycho.core.shared/src/main/resources/META-INF/MANIFEST.MF
https://github.com/eclipse-tycho/tycho/blob/master/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/resources/META-INF/MANIFEST.MF

We just have to keep this one, but it might be generated automatically:
https://github.com/eclipse-tycho/tycho/blob/master/tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/resources/META-INF/MANIFEST.MF

HannesWell:
Ah looked for the wrong names. Thanks.

> We just have to keep this one, but it might be generated automatically: https://github.com/eclipse-tycho/tycho/blob/master/tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/resources/META-INF/MANIFEST.MF

Yes, it looks doable to generate them. Will do that.

tycho-core/pom.xml
```diff
@@ -67,22 +67,22 @@
 			</plugin>
 
 			<plugin>
-				<groupId>org.apache.felix</groupId>
-				<artifactId>maven-bundle-plugin</artifactId>
+				<groupId>biz.aQute.bnd</groupId>
```
HannesWell, line 70: @laeubi is this still required? Or can I just remove it too?

tycho-surefire/org.eclipse.tycho.surefire.osgibooter/pom.xml
```diff
@@ -141,32 +142,50 @@
...
 				<configuration>
-					<source>${java.version}</source>
-					<target>${java.version}</target>
+					<bnd><![CDATA[
+						Import-Package: !*
+						Bundle-SymbolicName: org.eclipse.tycho.surefire.osgibooter;singleton:=true
+						Automatic-Module-Name: ${bsn}
+						Bundle-Activator: org.eclipse.tycho.surefire.osgibooter.Activator
+						Bundle-ActivationPolicy: lazy
+						Bundle-Vendor: %providerName
```
HannesWell, line 170: Do we really have to internationalize the Bundle-Vendor?
If not the `plugin.properties` can be deleted too.

HannesWell (commented):
> We just have to keep this one, but it might be generated automatically: https://github.com/eclipse-tycho/tycho/blob/master/tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/resources/META-INF/MANIFEST.MF

Done. The generated Manifest looks like the following.
This also considers the removal of `common-junit4-3.0.0-M5.jar` (or was that unintentional?)

```
Manifest-Version: 1.0
Automatic-Module-Name: org.eclipse.tycho.surefire.osgibooter
Bundle-ActivationPolicy: lazy
Bundle-Activator: org.eclipse.tycho.surefire.osgibooter.Activator
Bundle-ClassPath: .,jars/common-java5-3.0.0-M5.jar,jars/maven-failsafe
 -plugin-3.0.0-M5.jar,jars/maven-surefire-common-3.0.0-M5.jar,jars/sur
 efire-api-3.0.0-M5.jar,jars/surefire-booter-3.0.0-M5.jar,jars/surefir
 e-extensions-api-3.0.0-M5.jar,jars/surefire-logger-api-3.0.0-M5.jar,j
 ars/surefire-shared-utils-3.0.0-M5.jar
Bundle-Description: Tycho integrates Maven with Eclipse and OSGi
Bundle-ManifestVersion: 2
Bundle-Name: Tycho Surefire OSGi Booter Eclipse Application
Bundle-SymbolicName: org.eclipse.tycho.surefire.osgibooter;singleton:=
 true
Bundle-Vendor: %providerName
Bundle-Version: 4.0.0
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
```

tycho-core/pom.xml
laeubi: If the build works without it can be removed... :-)

tycho-surefire/org.eclipse.tycho.surefire.osgibooter/pom.xml
laeubi: Don't think so, this is really nothing visible to the user.

tycho-core/pom.xml
HannesWell: It looks like the generated manifest was not included in the build jars for a while (at least `tycho-core-3.0.0.jar` and the current 4.0.0-SNAPSHOT) only have a trivial, non-OSGi compliant, MANIFEST.MF.
So I'm pretty sure it can be removed, but lets see what the build says now (I first wanted to fix the other failures).

HannesWell:
Just checked that there are no more executions of the maven-bundle-plugin at all and no more bnd-maven-plugin executions that can be removed (trivially).  Also I verified that there are no more unncessary MANIFEST.MF file and also inlined+removed more plugin.properties (which were probably not working anyway because of the missing corresponding `Bundle-Localization` header.

So after the build has succeeded, I think this is ready for submission.

PR #1754, opened 2022-11-28
avoid duplicate dependencies

Bananeweizen:
Since it's not the first time that Maven throws warnings because of duplicate dependencies in Tycho, we need more strict checks.


laeubi (approved):
This is a good idea!

PR #1755, opened 2022-11-28
Remove invalid classpath entries

Bananeweizen:
On a fresh clone of Tycho all these directories just don't exist.
grafik


laeubi:
@HannesWell can you check that this works as well with
- #1740

HannesWell:
> @HannesWell can you check that this works as well with
> 
>     * [Remove obsolete Eclipse metadata files #1740](https://github.com/eclipse-tycho/tycho/pull/1740)

All those `.classpath` files have been removed from git with #1740 and M2E should generate them properly then. So this should be obsolete now.

Issue #115, opened 2021-05-03
Create a 'tycho-baseline' plugin


laeubi:
Getting semantic versioning right is not an easy task. BND allows for [comparing a baseline versus an actual version](https://bnd.bndtools.org/commands/baseline.html), it would be good if tycho could offer a mojo that helps in doing this in an automated way.

mickaelistria:
Such mojo would better be provided by BND directly, and then maybe used by Tycho.
Tycho already has ways to include PDE API Tools analysis during build, which helps with doing semver in.some extent by reporting errors against baseline. This is for example enabled as part of Platform Build and allows to report wrong versioning of packages/bundles as errors during CI/Gerrit builds. So nothing more is needed from Platform POV.

laeubi:
BND already offers a maven plugin for this but this would require to build my projects by BND and the goal would be to have this available in Tycho itself.

As you wrote PDE API Tools ' helps with doing semver in.some extent' but I really like full support without binding to eclipse specific tools.

> So nothing more is needed from Platform POV.

Tycho is more than platform needs :-)

mickaelistria:
> Tycho is more than platform needs :-)

Sure, but just beware that in practice, most of its current contributors (at least @akurtakov and I) are only caring about Platform and won't provide much support for cases that are not part of Platform.

laeubi:
> Sure, but just beware that in practice, most of its current contributors

Looking at the [monthly pulse](https://github.com/eclipse/tycho/pulse/monthly) I won't see this to pessimistic.

HannesWell:
> Tycho already has ways to include PDE API Tools analysis during build, which helps with doing semver in.some extent by reporting errors against baseline. This is for example enabled as part of Platform Build and allows to report wrong versioning of packages/bundles as errors during CI/Gerrit builds.

As written on some mailing-list in the past (I think it was m2e), the setup currently used in Platform could be moved into a dedicated API analysis plug-in to simply its use in other projects (yes one could copy past it, but that's IMHO not very elegant).
This would also have the advantage, that the plug-in could be automatically enabled/disabled depending if a plug-in has the APIAnalysis builder listed in its `.project` build-commands. Which would avoid the `skipAPIAnalysis=true` properties in many test projects.
[...]

PR #1767, opened 2022-11-30
Add a tycho-baseline-mojo

laeubi:
Fix https://github.com/eclipse-tycho/tycho/issues/115


laeubi:
@HannesWell I think m2e would be a good incubator for this feature, so if you like to give it a try and find any issue let me know!

PR #1768, opened 2022-11-30
Lock version ranges for platform dependency

laeubi:
- org.eclipse.equinox.common
- org.eclipse.core.jobs

FYI @merks


laeubi:
This is to clumsy right now, disable the filtering all together and hope for the next release:
- https://github.com/eclipse-tycho/tycho/pull/1770

PR #1773, opened 2022-11-30
Reproducer for "JupiterTestEngine not a subtype" when using JUnit 5

lppedd:
I have no idea what I'm doing wrong, so I prepared a reproducer.  
The target platform is in its own reactor artifact, and it imports some features of Eclipse 4.23, plus JUnit 5 Maven dependencies.

The full error stack trace is:
```
org.apache.maven.surefire.api.util.SurefireReflectionException: java.util.ServiceConfigurationError: org.junit.platform.engine.TestEngine: org.junit.jupiter.engine.JupiterTestEngine not a subtype
	at org.apache.maven.surefire.api.util.ReflectionUtils.instantiateOneArg(ReflectionUtils.java:127)
	at org.apache.maven.surefire.booter.SurefireReflector.instantiateProvider(SurefireReflector.java:211)
	at org.apache.maven.surefire.booter.ProviderFactory.createProvider(ProviderFactory.java:118)
	at org.apache.maven.surefire.booter.ProviderFactory.invokeProvider(ProviderFactory.java:83)
	at org.eclipse.tycho.surefire.osgibooter.OsgiSurefireBooter.run(OsgiSurefireBooter.java:142)
	at org.eclipse.tycho.surefire.osgibooter.HeadlessTestApplication.start(HeadlessTestApplication.java:29)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:203)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:136)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:104)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:401)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:255)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:659)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:596)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1467)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1440)
Caused by: java.util.ServiceConfigurationError: org.junit.platform.engine.TestEngine: org.junit.jupiter.engine.JupiterTestEngine not a subtype
	at java.base/java.util.ServiceLoader.fail(ServiceLoader.java:593)
	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNextService(ServiceLoader.java:1244)
	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNext(ServiceLoader.java:1273)
	at java.base/java.util.ServiceLoader$2.hasNext(ServiceLoader.java:1309)
	at java.base/java.util.ServiceLoader$3.hasNext(ServiceLoader.java:1393)
	at java.base/java.lang.Iterable.forEach(Iterable.java:74)
	at org.junit.platform.launcher.core.LauncherFactory.collectTestEngines(LauncherFactory.java:145)
	at org.junit.platform.launcher.core.LauncherFactory.createDefaultLauncher(LauncherFactory.java:131)
	at org.junit.platform.launcher.core.LauncherFactory.create(LauncherFactory.java:125)
	at org.junit.platform.launcher.core.LauncherFactory.create(LauncherFactory.java:109)
	at org.apache.maven.surefire.junitplatform.JUnitPlatformProvider.<init>(JUnitPlatformProvider.java:90)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
	at org.apache.maven.surefire.api.util.ReflectionUtils.instantiateOneArg(ReflectionUtils.java:123)
```

Any idea?

### DIFF
--- a/tycho-its/projects/target.artifact.junit/bundle-example/src/test/java/com/tycho/repro/MyExampleTestIT.java
+++ b/tycho-its/projects/target.artifact.junit/bundle-example/src/test/java/com/tycho/repro/MyExampleTestIT.java
@@ -1,0 +1,10 @@
+package com.tycho.repro;
+
+import org.junit.Test;
+
+public class MyExampleTestIT {
+    @Test
+    public void exampleTest() {
+        // Noop
+    }
+}
--- a/tycho-its/projects/target.artifact.junit/target-platform/target-platform.target
+++ b/tycho-its/projects/target.artifact.junit/target-platform/target-platform.target
@@ -1,0 +1,99 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<?pde version="3.8"?>
+<target includeMode="feature" name="target-platform">
+  <locations>
+    <location includeDependencyDepth="none" includeSource="true" missingManifest="generate" type="Maven">
+      <dependencies>
+        <dependency>
+          <groupId>org.junit.jupiter</groupId>
+          <artifactId>junit-jupiter-api</artifactId>
+          <version>5.9.1</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.junit.jupiter</groupId>
+          <artifactId>junit-jupiter-engine</artifactId>
+          <version>5.9.1</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.junit.jupiter</groupId>
+          <artifactId>junit-jupiter-migrationsupport</artifactId>
+          <version>5.9.1</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.junit.jupiter</groupId>
+          <artifactId>junit-jupiter-params</artifactId>
+          <version>5.9.1</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.junit.platform</groupId>
+          <artifactId>junit-platform-commons</artifactId>
+          <version>1.9.1</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.junit.platform</groupId>
+          <artifactId>junit-platform-engine</artifactId>
+          <version>1.9.1</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.junit.platform</groupId>
+          <artifactId>junit-platform-launcher</artifactId>
+          <version>1.9.1</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.junit.platform</groupId>
+          <artifactId>junit-platform-runner</artifactId>
+          <version>1.9.1</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.junit.platform</groupId>
+          <artifactId>junit-platform-suite-api</artifactId>
+          <version>1.9.1</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.junit.platform</groupId>
+          <artifactId>junit-platform-suite-commons</artifactId>
+          <version>1.9.1</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.junit.vintage</groupId>
+          <artifactId>junit-vintage-engine</artifactId>
+          <version>5.9.1</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.apiguardian</groupId>
+          <artifactId>apiguardian-api</artifactId>
+          <version>1.1.2</version>
+          <type>jar</type>
+        </dependency>
+        <dependency>
+          <groupId>org.opentest4j</groupId>
+          <artifactId>opentest4j</artifactId>
+          <version>1.2.0</version>
+          <type>jar</type>
+        </dependency>
+      </dependencies>
+    </location>
+    <location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
+      <repository location="https://download.eclipse.org/eclipse/updates/4.23"/>
+      <unit id="org.eclipse.platform.sdk" />
+      <unit id="org.eclipse.rcp.feature.group" />
+      <unit id="org.eclipse.platform.ide" />
+      <unit id="org.eclipse.test.feature.group" />
+    </location>
+  </locations>
+  <environment>
+    <nl>en_US</nl>
+  </environment>
+  <targetJRE path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
+</target>
--- a/tycho-its/projects/target.artifact.junit/bundle-example/.classpath
+++ b/tycho-its/projects/target.artifact.junit/bundle-example/.classpath
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/5"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src/main/java"/>
+	<classpathentry kind="src" output="test" path="src/test/java">
+        <attributes>
+    		<attribute name="test" value="true"/>
+    	</attributes>
+    </classpathentry>
+	<classpathentry kind="output" path="target/classes"/>
+</classpath>
--- a/tycho-its/projects/target.artifact.junit/bundle-example/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/target.artifact.junit/bundle-example/META-INF/MANIFEST.MF
@@ -1,0 +1,11 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Example Bundle
+Bundle-SymbolicName: bundle-example
+Bundle-Version: 1.0.0.qualifier
+Bundle-Vendor: Example Vendor
+Require-Bundle: org.eclipse.ui,
+ org.eclipse.core.runtime
+Bundle-RequiredExecutionEnvironment: JavaSE-11
+Bundle-ActivationPolicy: lazy
+Bundle-ClassPath: .
--- a/tycho-its/projects/target.artifact.junit/bundle-example/build.properties
+++ b/tycho-its/projects/target.artifact.junit/bundle-example/build.properties
@@ -1,0 +1,4 @@
+source.. = src/main/java/
+output.. = target/classes/
+bin.includes = META-INF/,\
+               .
--- a/tycho-its/projects/target.artifact.junit/bundle-example/pom.xml
+++ b/tycho-its/projects/target.artifact.junit/bundle-example/pom.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.tycho.repro</groupId>
+    <artifactId>tycho-repro-junit</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>bundle-example</artifactId>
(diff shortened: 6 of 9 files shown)


laeubi:
Most likely surefire sees more than one test-engine and one of them is incompatible (wrong classlaoder for example).

lppedd:
@laeubi it is a bit strange tho. I thought something like this had been done in #1457.

I have debugged a lot but I wasn't able to understand where to introduce changes.

lppedd:
@laeubi I've noticed you pushed some stuff related to JUnit. I've rebased to check if this issue is still present.

tycho-its/projects/target.artifact.junit/target-platform/target-platform.target
```diff
@@ -0,0 +1,99 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<?pde version="3.8"?>
+<target includeMode="feature" name="target-platform">
+  <locations>
+    <location includeDependencyDepth="none" includeSource="true" missingManifest="generate" type="Maven">
+      <dependencies>
```
laeubi, line 6: with recent tycho, adding thigs manually to the target togehter with a Junit5 container is superflous.

lppedd: @laeubi so starting with Tycho 4, I can omit JUnit 5 from the target platform, that's the meaning?
Basically JUnit 5 is going to be resolved automatically if it's in the `.classpath` file?

laeubi: Yes it was also previously resolved (but only from the target) now it is automatically resolved from maven if not already in the target.

lppedd: @laeubi I'll try removing it from the target then! Thanks.
Although by how you're describing it, is seems keeping it as it is - in the target - should work.

laeubi: There are two demo projects here:
https://github.com/eclipse-tycho/tycho/tree/master/demo/testing/surefire

so in general junit 5 _should_ work, so I just assumed that probably something with the target setup is wrong.

tycho-its/projects/target.artifact.junit/bundle-example/src/test/java/com/tycho/repro/MyExampleTestIT.java
```diff
@@ -0,0 +1,10 @@
+package com.tycho.repro;
+
+import org.junit.Test;
```
laeubi, line 3: Why using Junit4 here?

lppedd: @laeubi it's on purpose. On my project I have existing JUnit 4 tests that I don't want to touch

joeshannon:
I am also experiencing this while attempting to update Tycho (testing with 4.0.0-SNAPSHOT) together with Eclipse release p2 as the target (2022-09).

With debugging I found that indeed there are two provider bundles of `TestEngine`:
```
org.eclipse.tycho.surefire.osgibooter:4.0.0(id=5)
junit-platform-engine:1.9.0(id=276)
```

Which of these is expected to be the provider?

As far as I can see, there are no projects with a JUnit 5 container or an explicit dependency on JUnit 5 in the target platform.
Will attempt to investigate further.

laeubi:
As said please start with the demo project, as these already work and show a state that is expected to work. Also make sure to use 2022-12 release as this is the primary target Tycho is tested with.

lppedd:
> Also make sure to use 2022-12

Unfortunately this is not possible. We depend on a custom base platform derived from 2022-03.

tycho-its/projects/target.artifact.junit/target-platform/target-platform.target
joeshannon: Thanks - those demos are only using maven-surefire-plugin I think so I've been looking at https://github.com/eclipse-tycho/tycho/tree/master/demo/testing/tycho/standalone

joeshannon:
I have spent a bit of time debugging this and have learnt a few things.

The JUnit Platform is provided by one of the Tycho surefire fragments (e.g. `org.eclipse.tycho.surefire.junit59`) which are fragments of `org.eclipse.tycho.surefire.osgibooter`.

It is also added to the runtime via various plugins which are added (e.g `junit-jupiter-engine`, `junit-platform-commons`, `junit-platform-engine`, `junit-platform-launcher`). Should all of these be added for tests running with Tycho Surefire? It does lead to multiple copies being in the runtime. Looking at an older setup (2022-03 TP with Tycho 2.7.5) there don't seem to be as many bundles added. Could this be related to other changes to classpath containers and support for `maven-surefire-plugin`?

The `OsgiSurefireBooter` supplies a custom class loader to invoke the Surefire framework with. This is a `CompositeClassLoader` with three loaders in an array which are searched in order.

The first loader wraps the test bundle, the second is the bundle loader for `org.eclipse.tycho.surefire.osgibooter` and the third is a modified `ContextFinder`.

The particular failure I was experiencing was caused by the test-plugin/fragment-host's Eclipse Buddy Policy which was set to `global`. Therefore when The test engine is loaded, the first class loader in `CompositeClassLoader`, with this global policy is able to load `JupiterTestEngine` from `junit-jupiter-engine` whereas it should be the version from `org.eclipse.tycho.surefire.osgibooter`.

I'm not entirely sure how to improve this situation. This is reproducible in the current integration tests by setting this buddy policy.

@lppedd I'm not sure if Buddying is relevant to your issue but might be worth checking. Otherwise there might be some other dependency causing the mismatched engine class to be loaded.

tycho-its/projects/target.artifact.junit/target-platform/target-platform.target
laeubi: It all depends on your use-case, setup and what you want to archive see:
https://tycho.eclipseprojects.io/doc/master/TestingBundles.html

laeubi:
> The particular failure I was experiencing was caused by the test-plugin/fragment-host's Eclipse Buddy Policy which was set to global.

You can prevent many hassle by stay away from any header starting with `Eclipse-<whatever>` ... today there is always ever a better, specification compliant alternative (e.g.  [DynamicImport-Package](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#d0e2063)).

lppedd:
Hi @joeshannon, thanks for the investigation.  
I will have a look again to understand if I'm importing stuff that I should not import.  
One question: where did you set breakpoints in the Maven or Tycho code?

joeshannon:
@laeubi Thanks - yes I have been making progress to slowly remove these from our products. Unfortunately `DynamicImport-Package` doesn't solve this issue.

Have been debugging Tycho code with a breakpoint in `OsgiSurefireBooter#run`. This is where Tycho invokes the Surefire framework.

JUnit then eventually uses a Java `ServiceLoader` to load engine classes, so also been debugging which bundle classes from `CompositeClassLoader` are loaded from.

I've just debugged the test PR. At some point, the JUnit dependencies to the bundle and also configuring it to use `DynamicImport-Package` for *.

```
BundleLoader[bundle-example_1.0.0.202301311640].findClass(org.junit.jupiter.engine.JupiterTestEngine)
BundleLoader[bundle-example_1.0.0.202301311640].findLocalClass(org.junit.jupiter.engine.JupiterTestEngine)
ModuleClassLoader[bundle-example_1.0.0.202301311640 - /scratch/30day_tmp/delete27jan/tycho-ide/workspace_git/tycho.git/tycho-its/projects/target.artifact.junit/bundle-example/target/classes].findClassImpl(org.junit.jupiter.engine.JupiterTestEngine)
ModuleClassLoader[bundle-example_1.0.0.202301311640 - /scratch/30day_tmp/delete27jan/tycho-ide/workspace_git/tycho.git/tycho-its/projects/target.artifact.junit/bundle-example/target/test-classes].findClassImpl(org.junit.jupiter.engine.JupiterTestEngine)
ModuleClassLoader[bundle-example_1.0.0.202301311640 - /scratch/30day_tmp/delete27jan/tycho-ide/workspace_git/tycho.git/tycho-its/projects/target.artifact.junit/bundle-example/target/bundle-example-1.0.0-SNAPSHOT.jar].findClassImpl(org.junit.jupiter.engine.JupiterTestEngine)
ModuleClassLoader[bundle-example_1.0.0.202301311640 - /scratch/30day_tmp/delete27jan/tycho-ide/workspace_git/tycho.git/tycho-its/projects/target.artifact.junit/bundle-example/target/bundle-example-1.0.0-SNAPSHOT-test-d05d81e7-0852-4a5d-abc4-a23aef0ad247.jar].findClassImpl(org.junit.jupiter.engine.JupiterTestEngine)
BundleLoader[bundle-example_1.0.0.202301311640] attempting to resolve dynamic package: org.junit.jupiter.engine
BundleLoader[bundle-example_1.0.0.202301311640] using dynamic import source: org.junit.jupiter.engine -> [junit-jupiter-engine_5.9.1]
BundleLoader[junit-jupiter-engine_5.9.1].findLocalClass(org.junit.jupiter.engine.JupiterTestEngine)
```

This is added by Tycho (I think this only applies to `eclipse-plugin` i.e. `TychoIntegrationTestMojo` and `BndTestMojo`): https://github.com/eclipse-tycho/tycho/blob/39a042926a05f67fcaa6b311e78dada3eea59e24/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java#L378 

One fix might be to change the order of the loaders in the instantiated  `CompositeClassLoader`, to ensure that the booter's classes are prioritised, but this might have consequences.

laeubi:
@joeshannon You should start with one of the examples and enhance them in a way that fails, that would be much easier to diagnose, as currently it is quite hard to guess what might be wrong in an unknown setup even using old dependencies.

joeshannon:
[This change](https://github.com/joeshannon/tycho/commit/55d5c46722005097c1a4b96eb231aeb362ef80f6) is probably one of the simplest reproducers, using an existing integration test.

This adds `DynamicImport-Package` to the test bundle. It looks quite different to the original reproducer here but I believe they have the same underlying cause, as described in my comment above.

laeubi:
I now created a PR to remove the dynamic import as it seem to cause problems and there is currently no use-case that require it:
- https://github.com/eclipse-tycho/tycho/pull/2049

lppedd:
`DynamicImport` seems like a recipe for issues anyway. Good that you got rid of it.  
I've rebased, let's see!

lppedd:
Nope, not luckly this time 😢

joeshannon:
I can submit a draft PR for the classloading change? I would be interested to see if it solves your issue, and also to run all the tests with it to see if it is compatible.

lppedd:
@joeshannon definitely! I will merge/rebase your changes on this PR afterwards.

joeshannon:
The change is #2057 . I'm not entirely convinced this is the best fix but the tests do seem to pass still (I don't think the single failure is related, after a quick glance).

laeubi:
Yes the failure is unrelated, sadly the test-setup is a bit brittle (it should work better with the new bnd-test-mojo), maybe @lppedd can check this with his example and real world use-case, one thing that came into my mind: Maybe we can have a new property (tycho.test.classloaderOrder) with the values `testProbeFirst` (old behaviour) and `booterFirst` (changed behavior)), and then have a configuration option in the mojo for this. I know it is a bit more work to pass this along (you can take a look how the "failsafe" property is passed along) but then it would be quite safe to include this as if it causes issues one can easily switch between both variants.

lppedd:
One step I plan on doing is standardizing the target platform we use for all the environments/IDEs.
Currently, we have devs that use Eclipse with a versioned target file, and devs that simply use Tycho as a way to inject dependencies into the Maven model through declared repositories (e.g., in IntelliJ IDEA).

The reproducer that I shared in this PR is very very similar to a possible end result of the single `.target` file that will be used.
I'm just waiting for this to work, to be able to say to the management "I need to allocate time to try out various ways of testing".

lppedd:
@joeshannon your change seems to be working!

Tho, I have to be honest, I still do not fully understand the implications of such a change.

lppedd:
This is now passing. Thanks again for the investigation work @joeshannon.  
@laeubi can we merge this too once it finishes building?

PR #1781, opened 2022-12-01
Use plexus components instead of direct lookup of service in pgp-plugin

laeubi:
@merks lets see if this works out well, do you maybe like to test this with some real world examples or are you fine with it given the tests succeed?


merks:
If the tests work, I think it should be fine.

laeubi:
At laes the PGP test do not fail but jenkins has one failure in an unrelated par ... strange.

PR #1784, opened 2022-12-01
Construct the URI directly instead of first creating an URL

laeubi:


HannesWell:
The problem in the `FeatureWithRestrictionsTest` is that the way the URL is created is wrong and leads to an invalid URL.

Currently it is like
```
File repo = ResourceUtil.resolveTestResource("projects/feature.restrictions/repository").getAbsoluteFile();
String url ="file:" + repo;
```
But correct would be:
`repo.toURI().toString()`
For the first (wrong) way the result is something like `file:C:\dev\git\org.eclipse.tycho\tycho-its\projects\feature.restrictions\repository` and for the second way the result is `file:/C:/dev/git/org.eclipse.tycho/tycho-its/projects/feature.restrictions/repository/`
The main problem is the missing slash after the schema. The backward-slashes are no big deal (IIRC).

The test case can be adjusted, but the question is if users expect to be able to use such invalid URLs for convenience?
If we want to still allow the usage of invalid URIs we could consider to use `org.eclipse.core.runtime.URIUtil.fromString()`.
But in general the 'support' for invalid URLs is somehow hindering the adjustments of Eclipse-platforms to use proper File->URI/URL conversion and is therefore hindering full support for path's that e.g. contain spaces (IIRC it is still problematic in some cases if you have a space in your workspace-path, at least on windows).
So I'm tempted to say that users should learn to use proper URLs.

laeubi:
Thanks for the analysis... We already have org.eclipse.tycho.p2resolver.TargetDefinitionResolver.convertRawToUri(String) that I think could be reused here, the rationale was not really convenience, but that if you use variables like `${project.basedir}/some/path` in `pom.xml` there is not really a way for the user so convert this to an URI compatible way. I think we should reuse this here and then maybe even the test makes sense as-is so it covers that case :-)

PR #1785, opened 2022-12-01
Normalize the host bundle symbolic name

### LINKED ISSUE #1769: Test plugin JAR generation creates invalid Bundle-SymbolicName
Method call [`TychoIntegrationTestMojo#createTestPluginJar`](https://github.com/eclipse-tycho/tycho/blob/1e3c6cd22154bb3499528cf2b88d2aef1ed79b05/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/TychoIntegrationTestMojo.java#L208).

image

The resulting `Bundle-SymbolicName` is `com.ibm.pdtools.common.component.core;singleton:=true.90fd8144-0691-4773-a78b-00a93545e4f6`, which seems to be wrong.

If you confirm this indeed looks wrong I can submit a PR.

lppedd:
Fix #1769.

The first commit does the real thing.  
The second commit reformats the entire source file.

Kept them separate to ease reviewing.

### DIFF
--- a/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/TychoIntegrationTestMojo.java
+++ b/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/TychoIntegrationTestMojo.java
@@ -220,7 +220,7 @@ public class TychoIntegrationTestMojo extends AbstractTestMojo {
                     Analyzer analyzer = new Analyzer(jar)) {
                 Manifest bundleManifest = mainArtifact.getManifest();
                 String hostVersion = bundleManifest.getMainAttributes().getValue(Constants.BUNDLE_VERSION);
-                String hostSymbolicName = bundleManifest.getMainAttributes().getValue(Constants.BUNDLE_SYMBOLICNAME);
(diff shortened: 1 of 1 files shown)


laeubi:
@lppedd what Eclipse formatter profile do you use for this reformating? I just tried to reformat the code from within my Eclipse but don't get the any changes here..

lppedd:
@laeubi I actually use IntelliJ. The formatting profile is basically "adapt to what you find" (e.g. 4 spaces indentation).
Formatting of chained calls has been done manually.

laeubi:
If the reformating is not essential I think it would be better to just fix the bug as this otherwise causes to much noise with different formatters and manual formatting in conjunction.

lppedd:
@laeubi I can format it with the Eclipse formatter settings (`org.eclipse.jdt.core.prefs`), not an issue if you say it's ok.

lppedd:
@laeubi reformatted again using the Eclipse formatter settings for the project (`tycho-surefire-plugin/.settings/org.eclipse.jdt.core.prefs`).

lppedd:
@laeubi btw, would be cool to have a unified formatter (.xml, .epf) valid for the entire workspace. I could then simply import that formatter and work like if I was in Eclipse.

PR #1786, opened 2022-12-01
Improve the Baseline-Mojo

laeubi:
- add integration tests
- include the current version in the baseline search
- ...


laeubi:
This later will produce a nice report if any differences are found
grafik

PR #1791, opened 2022-12-03
Print download statistics similar to maven when running interactive

laeubi:


lppedd:
That's awesome. I was really missing this one, thanks!

laeubi:
Great you like it, this should makes p2 downloads appears as "regular" maven downloads, 
if you like my work on Tycho you might consider say [thank you](https://github.com/sponsors/laeubi?frequency=one-time) or even [sponsor further development ](https://github.com/sponsors/laeubi?frequency=recurring).

PR #1802, opened 2022-12-05
Add support for isolating classloadings for some bundles

laeubi:
Currently all bundles are shared, but some of them are only on the classpath for technical reasons (e.g. DS) this adds support for marking a bundle as using an isolated bundle loader. Beside that, lazy bundles are no longer started by default, this allows to more selectively control what is actually exposed.


laeubi:
This requires
- https://github.com/eclipse-tycho/tycho/pull/1805

to be effective...

PR #1805, opened 2022-12-06
remove org.eclipse.equinox.security dependency

laeubi:
With the new Tycho transport the binding to org.eclipse.equinox.security is obsolete now.


laeubi:
`equinox.security` will still be pulled in transitive because of

- https://github.com/eclipse-equinox/p2/issues/200
- https://github.com/eclipse-equinox/p2/issues/201

PR #1808, opened 2022-12-06
Add m2e integration for build-qualifier and validate-id/version

laeubi:
This adds support for errors detected in the validate phase of tycho if version/id missmatch happens.

This will show validation errors in m2e when using Tycho:
grafik
grafik

FYI @HannesWell


HannesWell:
That's great 🚀

HannesWell:
Since you started this, do you want to do the same for the `org.eclipse.tycho.extras:tycho-p2-extras-plugin:compare-version-with-baselines` goal?

I think it would be highly appreciated by all Eclipse-Platform devs because one would already be notified in the IDE about a necessary service-version bump. In the past I suggested something similar as an extension of the PDE-API-tools where M2E could provide a implementation of a service that computes the qualifier of a Plug-in and therefpre then suggest service version updates (I cannot find the discussion anymore).
But doing it similar like you did here, would actually be much more straight forward and simple and would require no new APIs in PDE and no extra implementation in M2E. :)
If you don't want to do it I can do it as well.

laeubi:
It depends a bit how the mojos work, but my goal is to integrate as much of them as possible but currently this is blocked by:
- https://github.com/eclipse-m2e/m2e-core/issues/1126

because you would need here the final jar as well.

HannesWell:
> It depends a bit how the mojos work, but my goal is to integrate as much of them as possible but currently this is blocked by:
> 
>     * [Tycho configuration is bound to package-plugin instead of the package-type eclipse-m2e/m2e-core#1126](https://github.com/eclipse-m2e/m2e-core/issues/1126)
> 
> 
> because you would need here the final jar as well.

I'm not sure if you need the final jar for this. From my understanding (but I have not yet looked into the details), the mojo only checks if the qualifier of a artifact is different compared to the one in the [baseline](https://tycho.eclipseprojects.io/doc/master/tycho-extras/tycho-p2-extras-plugin/compare-version-with-baselines-mojo.html#baselines) and if it is different and there is not at least a service-version bump, it fails. From that I think all you need is the  full-version(major, minor, service and qualifier) from the baseline and the full-version in the workspace. For the qualifier the [build-qualifier-mojo](https://tycho.eclipseprojects.io/doc/master/tycho-packaging-plugin/build-qualifier-mojo.html) has to run before (at least in the usual setup), but that should be it. I don't think that the final jar is necessary for that.

In general I'm not sure if Tycho's Plugins should run so far that the jars are build, I'm a concerned that this could have a noticeable performance impact, but that would have to be tested to be sure.

PR #1813, opened 2022-12-06
Clean-up/Modernize code in tycho-buildtimestamp-jgit plugin

HannesWell:
At the moment one can only use new-lines as separator for the jgit.ignore property read by the `tycho-buildtimestamp-jgit` plugin.

This often leads to lengthy configurations like in M2E:
https://github.com/eclipse-m2e/m2e-core/blob/f9551f4de2433b99ed4d93b2036485ec2c38a2fa/m2e-parent/pom.xml#L148-L156

Supporting commas as separator would allow to use more compact configurations.
Since one usually does not use commas in file-names I think that should be OK. It should be documented in the release notes, which is currently missing.
Or are there reasonable cases where one wants to ignore files with commas in their name?


HannesWell:
> I think this is not a good idea as this one tries to be similar to how gitignore files work. So maybe it would be better to support an external file, or even just add this to the git-ignore itself, Tycho should respect this file as well.

Understand, then just keep it as it is. Moving it to a separate file does not make it more compact and therefore does not makes sense IMHO.

.gitigore files are implicitly considered since these files don't contribute to the time-stamp since git itself already ignores them.

PR #1825, opened 2022-12-07
Add Maven wrapper

henryju:
Contrary to what the documentation was [saying](https://github.com/eclipse-tycho/tycho/blob/a52b5e749721223bea9b04455e1431a89ce3cf20/CONTRIBUTING.md?plain=1#L34), building Tycho [requires](https://github.com/eclipse-tycho/tycho/blob/a52b5e749721223bea9b04455e1431a89ce3cf20/pom.xml#L51) Maven 3.8.6

I think it is simpler to use the Maven wrapper to ensure everyone build using the correct (and the same) version of Maven.


laeubi:
@henryju I really don't like to introduce maven wrappe, please update the documentation instead that should be sufficent.

Beside that github actions/jenkins file are most often more up-to date.

lppedd:
I think this can be useful to always have the correct Maven version and not depend on what the contributor/CI machine has.
I use it almost everywhere if possible.

akurtakov:
It's already outdated and AFAICT it will add even one more thing to update on new maven release. Ad dependabot seems to not support automating maven-wrapper https://github.com/dependabot/dependabot-core/issues/485 I don't like the idea of one more manual thing to update. Closing for now. If/when updating the wrapper is automated we can have the discussion again.

laeubi:
One interesting danger when using wrapper jars:
https://blog.gradle.org/wrapper-attack-report

SteveBombardier (approved):

PR #1830, opened 2022-12-07
Dependecy updates

laeubi:


laeubi:
Still blocked by BND lib review, I have added a comment to [maven/mavencentral/biz.aQute.bnd/biz.aQute.bndlib/6.4.0](https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/5269)

akurtakov:
maybe exclude bnd for now and let it be handled via https://github.com/eclipse-tycho/tycho/pull/1779

laeubi:
Failures probably because of https://issues.apache.org/jira/browse/MSHARED-1125 @akurtakov do you think its worth take a look or should I rollback? I'm not sure how much time I'll find to adjust the test as I'm a bit busy these days ...

akurtakov:
In general I have found that doing so many updates at once always fails so its better to go smaller

laeubi:
Problem is that the verifier update requires some more changes that can not be split... lets see if my fix works..

laeubi:
Lets see if this works better then: https://github.com/eclipse-tycho/tycho/pull/1842

PR #1832, opened 2022-12-07
[API-Generation] Improve handling of extraManifests/extraSourceLocations

HannesWell:
The META-INF/MANIFEST.MF file may not be rooted in the project's basedir but for example in target/classes if it is generated. The APIFileGenerator can be told to use that using the extraManifests property. Therefore those files have to be considered too, when deciding to execute the mojo.

Additionally convert the 'extraManifests' and 'extraSourceLocations' to fields of `List<File>`. This has the advantage that, if a specified element is relative, it is automatically resolved against the project's basedir by Maven when the mojo is configured. Furthermore it allows to abstract away the platform-specific path separation that is required to prepare the `APIFileGenerator` properly. Therefore without this change this Mojo is effectively platform-dependend.

The `APIFileGenerator` should be adjusted to accept List of files, but for now we have to do it this way.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -38,6 +38,11 @@ Any baseline problems will then be reported to the build:
 
 ![grafik](https://user-images.githubusercontent.com/1331477/205283998-484c6a13-0a66-4b34-9386-599a27bff53e.png)
 
+### Parameter enhancements for tycho-apitools-plugin:generate goal
+
+The parameters of the `tycho-apitools-plugin:generate` goal have been completed and improved.
+
+
 ### Migration guide 3.x > 4.x
 
 #### Choosable HTTP transports
--- a/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiFileGenerationMojo.java
+++ b/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiFileGenerationMojo.java
@@ -13,7 +13,9 @@
 package org.eclipse.tycho.apitools;
 
 import java.io.File;
+import java.util.List;
 import java.util.jar.JarFile;
+import java.util.stream.Collectors;
 
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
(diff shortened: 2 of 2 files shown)


laeubi:
I think @sratz is using this Mojo and has added extra parameters in the past, so I think it would best if he can take a look here, as long as the platform check works its all fine for me ...

sratz:
> Sure. :) But in general I assume these changes are all for the better.

Yes, this looks good!

> Has it been released in this form yet?

It was first added in 3.0.0, see https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#new-api-tools-mojo, but was missing some of the parameters.

Interestingly, its documentation appears to be missing completely on https://tycho.eclipseprojects.io/doc/latest/index.html for some reason.

>  If yes I think I should add a corresponding release note.

That would probably be helpful. Maybe just some generic '_parameters of tycho-apitools-plugin:generate have been completed and improved_', that would also cover the changes I made in https://github.com/eclipse-tycho/tycho/pull/1488 ;)

HannesWell:
> 
> Interestingly, its documentation appears to be missing completely on https://tycho.eclipseprojects.io/doc/latest/index.html for some reason.

I have no clue and must admit that I'm not fully aware of how Tycho builds its site. Probably the site goal/plugin is used. I also have problems to navigate to the goals of plug-ins that are not pinned on the left.

> 
> > If yes I think I should add a corresponding release note.
> 
> That would probably be helpful. Maybe just some generic '_parameters of tycho-apitools-plugin:generate have been completed and improved_', that would also cover the changes I made in #1488 ;)

Sounds good. Added a corresponding entry based on your suggestion.

laeubi:
@akurtakov @mickaelistria @kwin any idea at the site-docs issue? I also noticed that my https://github.com/eclipse-tycho/tycho/blob/master/src/site/markdown/StructuredBuild.md is missing in the output there? I'll try to take a look if the upload has an issue but according to:
https://github.com/eclipse-tycho/eclipse-tycho.github.io/commit/e8656e7d5786f970941d7a0bbb40db9b8e930fda

I can see something **is** updated ...

kwin:
I see it in https://tycho.eclipseprojects.io/doc/master/staging/index.html, but IMHO you should not upload the staging folder. There is an old version of the site in https://github.com/eclipse-tycho/eclipse-tycho.github.io/tree/main/doc/master and a new one in https://github.com/eclipse-tycho/eclipse-tycho.github.io/tree/main/doc/master/staging, so something is wrong with the GitHub Action generating the site...

kwin:
@laeubi It looks like the path in https://github.com/eclipse-tycho/tycho/blob/3c169535a451481f0a75a2d9d0de79a63f2b23c8/.github/workflows/sitedocs.yml#L64 is wrong. You only want to copy the files/folders below `staging` but not the `staging` directory itself.

laeubi:
@kwin that explains it ... 🤦

kwin:
https://github.com/eclipse-tycho/tycho/pull/1850

lppedd:
I don't really understand what the apitools mojo does. Where can I find more documentation on this one?

laeubi:
> I don't really understand what the apitools mojo does. Where can I find more documentation on this one?

Then you probably do not need to use it ;-)

This is quite Eclipse-Specific: https://www.eclipse.org/pde/pde-api-tools/

lppedd:
@laeubi that's quite interesting tho. Seems like a cool additional safety net for build time. Thanks!

PR #1842, opened 2022-12-08
Update maven-verifier to 2.x

laeubi:


laeubi:
I got the verifier update successful now with only a few exceptions, so if anyone like to help with this it would be welcome, at the moment I have no idea what can cause this maybe something with passing the arguments.

laeubi:
@akurtakov maybe something you can help with as well after M1?

I got most things fixed now but there are 5 failing test left after the upgrade to Maven Verifier 2.x I just don't understand what might cause this change in behavior .. but probably we just see a bug here that was not discovered previously?

The most notable changes for Verifier 2.x is that method names changed and arguments are no longer parsed and passed "as is" but I can't see why it should make a difference here, still a second pair of eyes might reveal something.

Next thing is that Verifier 2 executes in the same process by default instead of forking, I have disabled that so we always fork because some test don't work well with it, so it should not be an issue here.

akurtakov:
I have the will but no promises :)

laeubi:
Rebased...

laeubi:
Maybe one should first finish this one as a smaller increment:
- https://github.com/eclipse-tycho/tycho/pull/1257

akurtakov:
@dependabot rebase

laeubi:
> @dependabot rebase

This is not a dependabot PR ;-)

I can try to rebase this and resolve conflicts but last time there where still some testfailures du to the upgrade so someone must be look into this if we want to upgrade.

akurtakov:
I would probably not find time to look into it as it's just M1 but I'll try to handle some of the deprecations done in this PR separately so it becomes less of a churn.

laeubi:
In the meanwhile [maven-verifier is deprecated](https://github.com/apache/maven-verifier/blob/master/MIGRATION.md) - therefore it might make more sense to go to the new maven-executor directly.

I keep this PR as a reminder and for future work.

PR #1850, opened 2022-12-09
Remove "staging" directory from GH pages

kwin:


laeubi:
Thanks, lets see if that works!

PR #1852, opened 2022-12-09
For EclipseRunMojo, resurrect argLine and appArgLine options

merks:
https://github.com/eclipse-tycho/tycho/discussions/1824


merks:
@laeubi If this is okay, I think it would be good to include this in 3.0.x as well.  Probably the release notes need attention that too...

laeubi (approved):
Looks good, regarding 3.x please see https://github.com/eclipse-tycho/tycho/blob/master/CONTRIBUTING.md#backporting

PR #1853, opened 2022-12-09
For EclipseRunMojo, resurrect argLine and appArgLine options

merks:
https://github.com/eclipse-tycho/tycho/discussions/1824

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,15 @@
 
 This page describes the noteworthy improvements provided by each release of Eclipse Tycho.
 
+## 3.0.2
+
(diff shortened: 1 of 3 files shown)


merks:
@laeubi These are the same changes as in the master branch.  I've updated the release notes appropriately I hope...

merks:
Yes, I was kind of torn by that issue.   But I was concerned that leaving the other part unchanged, and then someone reading that part out of context would also be misinformed...   But yes, if someone already read the notes then they wouldn't really expect them to change after the fact.  Let's see that the tests pass and then I'll force push just changes to the release notes for your review. 

Thanks for such timely responses!!

merks:
@laeubi This is better I think.

PR #1855, opened 2022-12-09
Add test case for the improved generate-pde-source-header mojo

HannesWell:
This PR adds tests for the enhancements added in https://github.com/eclipse-tycho/tycho/pull/1841.


HannesWell:
At least one of the two failing tests also fails on the master and I don't think to other failure is related to the addition of a test-case, so this is ready to go.

Issue #1648, opened 2022-11-09
A generated source feature does not actually include the source bundles


merks:
When I run the Maven/Tycho build in the attached project:

[generated-source-feature.zip](https://github.com/eclipse-tycho/tycho/files/9973077/generated-source-feature.zip)

The generated repository doesn't contain source bundles.  I see that the p2 metadata for the source feature includes "real" bundles and not the "source" bundles, e.g., `org.objectweb.asm.tree.analysis` not `org.objectweb.asm.tree.analysis.source`:

![image](https://user-images.githubusercontent.com/208716/200885835-ae469c4e-14e7-4bd6-944b-9b5fd9c8b440.png)

This automatic feature thing is super cool!   Unfortunately it doesn't help build a repo with all the sources with this bug...

laeubi:
Thanks for reporting this, I'll take a look at is! In the meanwhile you maybe want to transform this into an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue? 

For example https://github.com/eclipse-tycho/tycho/blob/master/tycho-its/src/test/java/org/eclipse/tycho/test/target/TargetPlatformLocationsTest.java seems a good place to enhance for testing that source bundles/features are actually produced correctly.

laeubi:
Fixed by https://github.com/eclipse-tycho/tycho/pull/1693

PR #1857, opened 2022-12-10
[backport] A generated source feature should include the source bundles

merks:
https://github.com/eclipse-tycho/tycho/issues/1648


merks:
@laeubi This is just back porting the simple fix...

PR #1863, opened 2022-12-11
Add support for validate feature version increments

laeubi:


laeubi:
This then might looks like this
grafik

Issue #1864, opened 2022-12-12
Platform build fails with OOM


akurtakov:
Since yesterday (Dec 11th) Platform build OOMs as it can be seen at https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/master/833/console and https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/PR-757/3/console . As there doesn't look like any changes in platform could cause it I'm suspecting Tycho issue.

akurtakov:
@laeubi Any ideas what could happen here?

laeubi:
Nothing obvious ...  it might be related to https://github.com/eclipse-tycho/tycho/commit/484a44fefe36278dd939a82a43c70205a1645dca but this is just guessing here...

laeubi:
@akurtakov can we change the platform builds to pass `-e` maven option? This will really help in such cases as one at least gets an exception trace then.

laeubi:
I'm now running this locally, I just wonder: How many ram is used for the maven build?

akurtakov:
No definition from project side https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/Jenkinsfile so whatever JIPPs default it. @fredg02 can you answer the question?

laeubi:
I run the aggregator build (without test) and at least this does succeeds...  I have not set any special options here so it might depend of course... the stack trace is really weird I'll try to enhance this to get a better message but can't really guess this should cause OOM...

akurtakov:
So it's definetely xmlunit . https://ci.eclipse.org/platform/job/eclipse.platform.releng.aggregator/job/PR-760/1/console :
```
java.lang.OutOfMemoryError: Java heap space
11:14:02      at com.sun.org.apache.xerces.internal.dom.DeferredDocumentImpl.getNodeObject (DeferredDocumentImpl.java:940)
11:14:02      at com.sun.org.apache.xerces.internal.dom.DeferredElementNSImpl.synchronizeData (DeferredElementNSImpl.java:127)
11:14:02      at com.sun.org.apache.xerces.internal.dom.ElementNSImpl.getLocalName (ElementNSImpl.java:307)
11:14:02      at com.sun.org.apache.xerces.internal.dom.CoreDocumentImpl.importNode (CoreDocumentImpl.java:1545)
[...]

PR #1870, opened 2022-12-12
Handle errors in artifact comparator more graceful

laeubi:
- only record full diff in XML if requested by the user
- if comparing zips fails, fall back to byte compare

FYI @akurtakov @HannesWell 

Closes https://github.com/eclipse-tycho/tycho/issues/1864


akurtakov:
This looks good to me and platfomr build succeeded thus merging.

PR #1872, opened 2022-12-12
Mark Mojos threadsafe

laeubi:
actually maven do not do any special action for mojos not marked thread safe... if we find problem we should fix them but marking them as not threadsafe only emits useless warning messages.


laeubi:
As always it "depends", `synchronized` only works in the same vm and only for the given instance, or you need to sync on a static object. I general one should make sure that each instance works on its own working directory.

PR #1878, opened 2022-12-14
Use ArtifactDelta.NO_DIFFERENCE instead of null

hd42:


laeubi:
@hd42 thanks for this cleanup :+1:

PR #1879, opened 2022-12-14
Remove the LocalDependencyResolver from tests

laeubi:
@akurtakov there is still test-code using that and this causes issues do you think you can help in migrating these tests?


akurtakov:
I'm quite busy with Platform changes as M1 is right after new year and this is my last work week for the year. I'll try to find some time after vacation though.

akurtakov:
There seem to be some conflicting files? Code deleted was adjusted so is removal still wanted?

laeubi:
> There seem to be some conflicting files? Code deleted was adjusted so is removal still wanted?

Yes removal is still wanted, I just deleted some tests in the meanwhile that reference this part of the code but it is still used internally, will try to update...

laeubi:
The PR is now up-to-date.

tycho-core/src/main/java/org/eclipse/tycho/core/resolver/DefaultDependencyResolverFactory.java
```diff
@@ -47,18 +46,11 @@ public DependencyResolver lookupDependencyResolver(ReactorProject reactorProject
...
         }
         if (properties.getProperty("tycho.test.targetPlatform") != null) {
-            try {
-                @SuppressWarnings("deprecation")
-                String hint = org.eclipse.tycho.core.osgitools.targetplatform.LocalDependencyResolver.ROLE_HINT;
-                return container.lookup(DependencyResolver.class, hint);
-            } catch (ComponentLookupException e) {
-                throw new RuntimeException("Could not instantiate required component", e);
-            }
+            throw new RuntimeException("don't use that");
```
laeubi, line 53: @akurtakov Each test that currently fails with this message uses the old (or ancient) way of setting the TP and needs adjustments, so if you have an idea that would be good if you can help adjust them, or if they are no longer useful, delete them.

laeubi:
The `org.eclipse.tycho.core.osgitools.EquinoxResolverTest.testBundleNativeCode_usingInvalidAliases()` seems an unnoticed regression as the tests now using the real Tycho resolver, as no one have complained, I will disable the test for now.

laeubi:
@akurtakov I have rebased (again) with the ltest master, now only tests that use the old way to set a targetplatform should fail we can:

1. adjust them
2. delete them
3. convert to regular itest

akurtakov:
If not easily adjustable  delete them as these are not testing anything that's usable now.

laeubi:
Currently failing tests:
```
Error:  Errors: 
Error:    DependencyComputerTest.testStrictBootClasspathAccessRules:142->AbstractTychoMojoTestCase.getSortedProjects:166 » CompoundRuntime
Error:    MavenDependencyCollectorTest.testNestedJars:39->AbstractTychoMojoTestCase.getSortedProjects:143->AbstractTychoMojoTestCase.getSortedProjects:166 » CompoundRuntime
Error:    TychoTest.testBundleExtraClasspath:169->AbstractTychoMojoTestCase.getSortedProjects:143->AbstractTychoMojoTestCase.getSortedProjects:166 » CompoundRuntime
Error:    TychoTest.testMNGECLIPSE942:134->AbstractTychoMojoTestCase.getSortedProjects:143->AbstractTychoMojoTestCase.getSortedProjects:166 » CompoundRuntime
Error:    TychoTest.testMissingClasspathEntries:146->AbstractTychoMojoTestCase.getSortedProjects:143->AbstractTychoMojoTestCase.getSortedProjects:166 » CompoundRuntime
Error:    TychoTest.testProjectPriority:121->AbstractTychoMojoTestCase.getSortedProjects:143->AbstractTychoMojoTestCase.getSortedProjects:166 » CompoundRuntime
```

laeubi:
Given that Tycho support "file" reference for target files it *might* be just to put the stuff in a target and reference that in the pom... @mickaelistria do you maybe want to take a look here?

akurtakov:
Do you think it makes sense to add system property so one can set target platform via -D https://tycho.eclipseprojects.io/doc/master/target-platform-configuration/target-platform-configuration-mojo.html#target  ? 
In that case fixing the tests should become trivial and I can even see how it can help people trying to build against e.g. 4.25 and 4.27 .

laeubi:
It should be already possible like this:
```
<plugin>
	<artifactId>target-platform-configuration</artifactId>
	<configuration>
		<target>
			<file>${younameit}</file>
		</target>
	</configuration>
</plugin>
```

but one can of course also add a property that just do that, just take into account that the mojo is just a placeholder for config assistance, so you need to evaluate the property/value in `DefaultTargetPlatformConfigurationReader` and then just add `result.addTarget(target)`, so if you think its useful I think it could be added without a problem.

laeubi:
@akurtakov actually once there was a property `tycho.targetPlatform` but it was deprecated / removed do you think we should add support for it again?

akurtakov:
If it helps. IIRC it was to point to smth like p2 repo directly aka resolved target only.

laeubi:
> IIRC it was to point to smth like p2 repo directly aka resolved target only.

I think these are "directory" targets, I'll check if one could probably use that ...

laeubi:
I was now able to fix the tests in tycho-core using a target platform file ...

laeubi:
I hope I have now found all places, this also uncovered two smaller issues in Tycho that are also fixed with this PR.

PR #1884, opened 2022-12-15
Make includeAllSources=true by default

mickaelistria:
Encourage OSS by default; facilitate best practice for Eclipse projects to publish their sources.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,12 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## 4.0.0 (under development)
 
+### ⚠️ Source bundles now included by default in p2 repositories
+
+The `includeAllSources` parameter for `tycho-p2-repository-plugin:assemble-repository` now has `true` as default value which means that source bundles -when available- will be included in every p2 repository you're building. This is meant to facilitate adoption of best practices for Eclipse projects and to encourage OSS as 1st class citizen.
+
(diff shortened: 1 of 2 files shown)


laeubi:
I don't think this should be enabled as default, there are many commercial products build with Tycho and not only OSS ones.

mickaelistria:
And why should we favor commercial products over OSS ones in default settings?

laeubi:
I just don't think that enforcing this as a default will help anything but can create big problems if is undergoing unnoticed. There a several reasons not to publish the sources with the p2 repsoitory, just think about that most of the time one want to deliver something to the end user to install into eclipse, they have no need for the sources and it actually wasted space and transfer times.

I'm also always reserved to promote "Eclipse best practice", there are many thing that are claimed "best practice" but are actually "we always did it that way" practice ...

Saying it is best practice to always include all sources in a P2 repository from my point of view is only valid if you assume that this p2 repository is consumed for developing software on top of it, something that is not true for all cases.

lppedd:
I would not do this. This requires some time so that everyone know it's going to be changed at least.
I would give it a couple of versions within a "Future invasive changes" sections or alike.

lppedd:
Just as a precaution I've set it to `false` explicitly in all our projects. I can't say how careful other people are tho...

akurtakov:
Closing this one as including all sources is not unconditionally wanted and it has the potential to break build in hard to recover way as pointed by https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/770 .

PR #1886, opened 2022-12-17
Allow specifying additional filesets for feature's packaging

### LINKED ISSUE #1869: Support additionalFileSets for package-feature Mojo
I've noticed the `additionalFileSets` parameter of `tycho-packaging-plugin` is supported only for plugins, not for features.
I'm trying to filter `feature*.properties` files, and would like to include them instead of what's used via `build.properties`.

I can give it a go myself in case.

lppedd:
Closes #1869.

I have also aligned the plugin/feature `additionalFileSets` javadoc.

### DIFF
--- a/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackageFeatureMojo.java
+++ b/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackageFeatureMojo.java
@@ -168,6 +168,18 @@ public class PackageFeatureMojo extends AbstractTychoPackagingMojo {
             jarArchiver.setDestFile(outputJar);
 
             try {
+                // Additional file sets win over bin.includes ones, so we add them first
+                if (additionalFileSets != null) {
+                    for (final var fileSet : additionalFileSets) {
+                        final var directory = fileSet.getDirectory();
+
+                        // noinspection ConstantConditions
+                        if (directory != null && directory.isDirectory()) {
+                            archiver.getArchiver().addFileSet(fileSet);
+                        }
+                    }
+                }
+
                 archiver.getArchiver().addFileSet(getManuallyIncludedFiles(buildProperties));
                 if (licenseFeature != null) {
                     archiver.getArchiver()
--- a/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/AbstractTychoPackagingMojo.java
+++ b/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/AbstractTychoPackagingMojo.java
@@ -69,6 +69,29 @@ public abstract class AbstractTychoPackagingMojo extends AbstractMojo {
     @Parameter(defaultValue = "true")
     protected boolean strictBinIncludes;
 
+    /**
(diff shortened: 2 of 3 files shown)


tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackageFeatureMojo.java
```diff
@@ -168,6 +191,18 @@ public void execute() throws MojoExecutionException, MojoFailureException {
             jarArchiver.setDestFile(outputJar);
 
             try {
+                // Additional filesets win over bin.includes ones, so we add them first
+                if (additionalFileSets != null) {
+                    for (final var fileSet : additionalFileSets) {
+                        final var directory = fileSet.getDirectory();
+
+                        // noinspection ConstantConditions
+                        if (directory != null && directory.isDirectory()) {
```
lppedd, line 200: I'm not sure why these checks are present in `package-plugin`. There is a `@Nonnull` annotation, and the method explicitly states it's a directory. Can we get rid of them?

tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/PackageFeatureMojo.java
```diff
@@ -96,6 +97,28 @@ public class PackageFeatureMojo extends AbstractTychoPackagingMojo {
...
+     *   &lt;directory&gt;${project.build.directory}/mytool-gen/&lt;/directory&gt;
+     *   &lt;includes&gt;
+     *    &lt;include&gt;&#42;&#42;/*&lt;/include&gt;
+     *   &lt;/includes&gt;
+     *  &lt;/fileSet&gt;
+     * &lt;/additionalFileSets&gt;
+     * </pre>
+     */
+    @Parameter
+    private DefaultFileSet[] additionalFileSets;
```
laeubi, line 120: should this not beeing inherited from the Abstract mojo?

lppedd: @laeubi definitely a good idea since . However the `package-iu` mojo would not use it, while still having it in its configuration.

lppedd: (If what you meant was moving it to `AbstractTychoPackagingMojo`).

laeubi: I must confess I have no idea for what one needs `package-iu` ... but think it would not be a big deal having one useless paramter there.

lppedd: @laeubi agree, at one point we might want to support addition file sets in that ui mojo too maybe. I'll move it if that's ok.

lppedd: @laeubi done. I've re-touched to documentation. Let me know if it's ok.

PR #1891, opened 2022-12-19
Add Text-File comparator to baseline compare to ignore lineending-diffs

HannesWell:
This changes adds a text-file comparator that ignores differences in line-endings for text files.
Comparing text-files with a baseline is problematic when the text-files are build into the baseline jars with another (usually platform depended newline) than used on the current computer.
This often happens when projects are build on Jenkins and one wants to make a local build on Windows and git is configured to check-out text files Windows-style (but commit them Linux style).

This improves the experience with the `p2-metadata`, `compare-version-with-baselines` and `tycho-baseline` mojos in this regard.

Some tests would probably be good as well.

### DIFF
--- a/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/TextContentsComparator.java
+++ b/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/TextContentsComparator.java
@@ -1,0 +1,67 @@
+/*******************************************************************************
+ * Copyright (c) 2022, 2022 Hannes Wellmann and others.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Hannes Wellmann - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.tycho.zipcomparator.internal;
+
+import java.io.IOException;
+import java.util.Arrays;
+
+import org.codehaus.plexus.component.annotations.Component;
+import org.eclipse.tycho.artifactcomparator.ArtifactComparator.ComparisonData;
+import org.eclipse.tycho.artifactcomparator.ArtifactDelta;
+import org.eclipse.tycho.artifactcomparator.ComparatorInputStream;
+
+@Component(role = ContentsComparator.class, hint = TextContentsComparator.TYPE)
+public class TextContentsComparator implements ContentsComparator {
+
+    public static final String DEFAULT_TEXT_FILE_EXTENSIONS = "txt,java,html,xml,exsd";
+
+    public static final String TYPE = "text";
+
+    private static final char CR = '\r';
+    private static final char LF = '\n';
+
+    @Override
+    public ArtifactDelta getDelta(ComparatorInputStream baselineStream, ComparatorInputStream reactorStream,
+            ComparisonData data) throws IOException {
+        byte[] baseline = baselineStream.asBytes();
+        byte[] reactor = reactorStream.asBytes();
+
+        int bI = 0;
+        int rI = 0;
+        int mismatch = Arrays.mismatch(baseline, reactor);
+        while (mismatch >= 0) {
+            int baselineNewLine = newLineLength(baseline, bI + mismatch);
+            int reactorNewLine = newLineLength(reactor, rI + mismatch);
+            if (baselineNewLine < 0 || reactorNewLine < 0) {
+                return ArtifactDelta.DEFAULT;
+            }
+            bI += baselineNewLine;
+            rI += reactorNewLine;
+
+            mismatch = Arrays.mismatch(baseline, bI, baseline.length, reactor, rI, reactor.length);
+        }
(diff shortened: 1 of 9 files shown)


tycho-baseline-plugin/src/main/java/org/eclipse/tycho/baseline/BaselineContext.java
```diff
@@ -26,6 +26,8 @@ public interface BaselineContext {
 	void reportBaselineProblem(String message) throws MojoFailureException;
 
 	List<String> getIgnores();
+	
+	List<String> getTextFileExtensions();
```
laeubi, line 30: is this really something we like to configure?

tycho-baseline-plugin/src/main/java/org/eclipse/tycho/baseline/FeatureBaselineComparator.java
```diff
@@ -334,7 +334,7 @@ private List<Diff> computeJarDelta(MavenProject project, BaselineContext context
 			ignores.addAll(context.getIgnores());
 			ArtifactDelta artifactDelta = zipComparator.getDelta(getStream(baselineJarUnit, context),
 					new ComparatorInputStream(reactor, file.getAbsolutePath()),
-					new ComparisonData(ignores, false));
+					new ComparisonData(ignores, List.of(".xml"), false));
```
laeubi, line 337: I don't think we should ignore any xml here this does not seem correct.

laeubi:
Some integration test are here:

https://github.com/eclipse-tycho/tycho/tree/master/tycho-its/projects/baseline

tycho-baseline-plugin/src/main/java/org/eclipse/tycho/baseline/FeatureBaselineComparator.java
HannesWell: It is not about ignoring, it is about defining if a (feature)XML-file is compared as text-file or not.
And I would say that a feature.xml should be compared as text-file, if not compared per XML-element.

tycho-baseline-plugin/src/main/java/org/eclipse/tycho/baseline/BaselineContext.java
HannesWell: If you know a reliable way to detect if a file is a text-file then it does not need to be configurable. But since I don't know any my solution currently is to provide a good default set of text-file-extensions (it can probably be enhanced) and let the user adjust it if that does not match that users reality.

tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/TextContentsComparator.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Override
+    public ArtifactDelta getDelta(ComparatorInputStream baselineStream, ComparatorInputStream reactorStream,
+            ComparisonData data) throws IOException {
+        byte[] baseline = baselineStream.asBytes();
+        byte[] reactor = reactorStream.asBytes();
+
+        int bI = 0;
+        int rI = 0;
+        int mismatch = Arrays.mismatch(baseline, reactor);
```
laeubi, line 41: This here seems a bit hard to understand if/how it works, why not having one for loop and simply advance the pointer if a newline is found? I'm also not sure if all cases are handled correctly I only see `\r\n` and `\n` handled but what about `\r` or even `\n\r`

and then I think this can be extracted into a public static method so it might be reused by other comparators.

tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/ZipComparatorImpl.java
```diff
@@ -146,6 +146,9 @@ private String getContentType(String name) {
         if (name.endsWith(".xml")) {
             return XmlComparator.XML;
         }
+        if (data.isTextFile(name)) {
```
laeubi, line 149: As mentioned we should only enable this for specific types (most likely `txt`)

laeubi (changes requested):

tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/TextContentsComparator.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+
+import org.codehaus.plexus.component.annotations.Component;
+import org.eclipse.tycho.artifactcomparator.ArtifactComparator.ComparisonData;
+import org.eclipse.tycho.artifactcomparator.ArtifactDelta;
+import org.eclipse.tycho.artifactcomparator.ComparatorInputStream;
+
+@Component(role = ContentsComparator.class, hint = TextContentsComparator.TYPE)
+public class TextContentsComparator implements ContentsComparator {
```
laeubi, line 24: You might want to move the code from org.eclipse.tycho.zipcomparator.internal.XmlComparator.createDelta(String, ComparatorInputStream, ComparatorInputStream) into this calss as a static method that can be used to generate unified diff for different text formats

laeubi:
This should now be fixed by:
- https://github.com/eclipse-tycho/tycho/pull/1934

Issue #1900, opened 2022-12-22
tycho-surefire-plugin:test does not resolve conditional bundles bound to non-primary environment


kwin:
The order of `environment` entries in https://tycho.eclipseprojects.io/doc/latest/target-platform-configuration/target-platform-configuration-mojo.html#environments seems to matter, as only the conditional bundles belonging to the *first* environment is properly resolved, e.g. when used in in a dependency given in https://tycho.eclipseprojects.io/doc/latest/tycho-surefire-plugin/test-mojo.html#dependencies.

Compare with the discussion at https://github.com/eclipse-m2e/m2e-core/discussions/1153

Expected behaviour:
The dependencies would be resolved from any(!) of the given environments matching conditional bundles.

kwin:
The issue seems to be in https://github.com/eclipse-tycho/tycho/blob/b12caef50b268837a0c23a7d9ce24655006427e4/tycho-core/src/main/java/org/eclipse/tycho/p2resolver/P2ResolverImpl.java#L153 which iterates over all environments. Unfortunately the first environment might already throw an exception (which is not caught), therefore the subsequent environment are not called.

kwin:
I provided a failing IT in https://github.com/eclipse-tycho/tycho/pull/1901 (only for Mac OS for now)

laeubi:
I think this is correct, i can't see that m2e is actually a "conditional bundle" that is one that defines a  eclipse-platform-filter.

kwin:
I don't get what is considered correct here. @laeubi Would you mind elaborating a bit? The bundle with filters in the example is not an m2e bundle but an Eclipse JDT one.

laeubi:
It is correct that the resolution fails. You example bundle does not define a platform-filter thus it is not platform dependent and need to resolve on all declared environments.

kwin:
How is this supposed to work with `eclipse-test-bundle` which are Bundle Fragments? I don't think that they are allowed to add a platform filter. In any case the error message is really poor and should explicitly state that in fact the bundle can only be resolved for one environment and not for others.

laeubi:
It is not a requirement that a `plugin-test` is a fragment, there is no such thing as "test fragments" anywhere in the spec.

> Also I don't think it is reasonable that all extra dependencies of a bundle need to have the same platform filter.

Extra requirements are nothing special, they are just there as a way to add items that are not directly required but the same rules apply to them.

kwin:
[...]

PR #1901, opened 2022-12-22
Add failing IT

kwin:
This relates to #1900

### DIFF
--- a/tycho-surefire/tycho-surefire-plugin/pom.xml
+++ b/tycho-surefire/tycho-surefire-plugin/pom.xml
@@ -27,6 +27,42 @@
 	<name>Tycho Surefire Plugin</name>
 
 	<build>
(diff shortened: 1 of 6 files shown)


kwin:
I couldn't find conditional bundles for other OS's in any of the features provided from https://download.eclipse.org/eclipse/updates/4.17/ (or a newer version of Eclipse), but I can also extend to other OS if someone can point me to a feature with conditional bundles for other operating systems.

laeubi:
In general for testing purpose it is often better to not rely on any other update site provinding a very special item but to create them as part of the test.

laeubi (changes requested):
Please add this as a regular [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) not as a maven-invoker one.

PR #1920, opened 2023-01-04
Move InstallableUnitProvider to the SPI

laeubi:


laeubi:
The idea using a `SessionScoped` is currently blocked by 
- https://github.com/apache/maven/pull/948

lppedd:
A question: hypothetically, let's say I want to contribute my implementation of a SPI interface.  
My custom plugin project imports Tycho SPI 4.0, while the project that is built with Tycho uses 4.1.
Would this version difference work (assuming no signature changes) or is there a check for the version?

laeubi:
Because of how maven works here, your extension will always get the version Tycho uses, as it export it and there can only every be one exported version. So if there are compatible changes it would work, but as with every tight integration with a product it is always good to keep up with the latest versions when they are available.

For such an extender it would most probably make sense to use the same version as Tycho, so users can use one version property for both and have always matching items.

PR #1933, opened 2023-01-07
Inject Junit Dependencies into the maven-model

laeubi:


laeubi:
While this works, doing it that way do make maven download a lot of stuff. I'll adjust the change so it keeps the info about the maven coordinates, not inject them directly. Then another change could build up on this by do the following:

1. convert current impl in compiler-plugin by the new classpath contributor using the Targetplatform and fallback to maven approach
2. only inject what was actually resolved into the model
3. probably mark some items as "not interesting" for injection as they are transitively required already.

PR #1950, opened 2023-01-10
Add includeAllSources to target-platform-configuration

mickaelistria:
For includeAllSources to work as expected during assemble-repository, the sources must first be included in the resolution. An includeAllSources flag is added to target-platform-configuration to instruct Tycho to add into resulting resolved TP all available source bundles it's aware of from the dependency sources.

### DIFF
--- a/tycho-embedder-api/META-INF/MANIFEST.MF
+++ b/tycho-embedder-api/META-INF/MANIFEST.MF
@@ -1,0 +1,23 @@
+Manifest-Version: 1.0
+Bnd-LastModified: 1664886812981
+Build-Jdk-Spec: 18
+Bundle-Description: Tycho integrates Maven with Eclipse and OSGi
+Bundle-DocURL: https://www.eclipse.org/
+Bundle-License: https://www.eclipse.org/legal/epl-v20.html
+Bundle-ManifestVersion: 2
+Bundle-Name: Tycho Embedder API
+Bundle-SymbolicName: org.eclipse.tycho.embedder-api
+Bundle-Vendor: Eclipse Foundation
+Bundle-Version: 3.1.0.SNAPSHOT
+Created-By: Apache Maven Bundle Plugin 5.1.8
+Export-Package: org.eclipse.tycho.build;version="0.11.0";uses:="org.apac
+ he.maven.execution,org.apache.maven.plugin,org.apache.maven.project",or
+ g.eclipse.tycho.classpath;version="0.11.0";uses:="org.apache.maven.plug
+ in,org.eclipse.tycho",org.eclipse.tycho.resolver;version="0.11.0";uses:
+ ="org.apache.maven.execution,org.apache.maven.project,org.eclipse.tycho
+ ",org.eclipse.tycho.runtime;version="0.11.0"
+Import-Package: java.io,java.lang,java.util,org.apache.maven.execution,o
+ rg.apache.maven.plugin,org.apache.maven.project,org.eclipse.tycho;versi
+ on="[3.1,4)"
+Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=17))"
+Tool: Bnd-6.3.1.202206071316
--- a/tycho-its/projects/p2Repository.includeAllSources/Maven.target
+++ b/tycho-its/projects/p2Repository.includeAllSources/Maven.target
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<?pde version="3.8"?>
+<target name="Maven" sequenceNumber="1">
+	<locations>
+		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="https://download.eclipse.org/eclipse/updates/4.25/R-4.25-202208311800"/>
+			<unit id="org.apache.batik.css" version="0.0.0"/>
+		</location>
+	</locations>
+</target>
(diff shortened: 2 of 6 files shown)


mickaelistria:
@laeubi Do you think this is a good approach? Or maybe a better one would be to recompute the target platform adding extra requirements as part of assemble-repository mojo? (I would actually prefer it but am not sure it's a safe thing to do there)

target-platform-configuration/src/main/java/org/eclipse/tycho/target/TargetPlatformConfigurationMojo.java
```diff
@@ -192,6 +192,13 @@ public class TargetPlatformConfigurationMojo extends AbstractMojo {
     @Parameter(name = DefaultTargetPlatformConfigurationReader.TARGET_DEFINITION_INCLUDE_SOURCE)
     private IncludeSourceMode targetDefinionIncludeSource;
 
+    /**
+     * When <code>true</code>, the resolved target platform is augmented with matching source
+     * bundles that can be found in the dependencies sources.
+     */
+    @Parameter(name = DefaultTargetPlatformConfigurationReader.INCLUDE_ALL_SOURCES)
+    private boolean includeAllSources;
```
laeubi, line 200: Should not
https://tycho.eclipseprojects.io/doc/latest/target-platform-configuration/target-platform-configuration-mojo.html#targetDefinitionIncludeSource 

with `targetDefinitionIncludeSource=force`

be something you like to archive here? In such case all locations are instructed to include the source bundles regardless of configuration in the target platform.

mickaelistria:
> I think targetDefinitionIncludeSource=force should be used to force all sources regardless of configuration (if the user do not want to control it by their TP), then all sources should be available in the target.

That's not the purpose of this flag and this flag is not really in the scope of the current issue. The same issue exists without a target definition (eg direct p2 repos).

> Then we should fetch the sources from the target of the project and not from the computed target requirements so P2Resolver must not be changed here but the mirror call, that should then solve the issue, I think this should be org.eclipse.tycho.TargetPlatform.resolveArtifact(String, String, String)

OK, I'll try using that instead of growing the resolver.

laeubi:
> That's not the purpose of this flag and this flag is not really in the scope of the current issue. The same issue exists without a target definition (eg direct p2 repos).

I think we should not need it then, if the resolver can access the units, other parts of Tycho can do that as well. Even though the term "TargetPlatform" is a bit overloaded, the *resolved* artifacts are actually the `DependencyArtifacts` an `TargetPlatform` contains everything (pom dependencies, target file content and repo defined content).

As we know there is just one source with one specific ID we do not really need the resolver, as we do not want to really "resolve" the source.

`org.eclipse.tycho.source.SourceFeatureMojo` uses `org.eclipse.tycho.core.resolver.P2ResolverFactory` but I think today `org.eclipse.tycho.TargetPlatform` should be more lightweight approach.

mickaelistria:
> So I just tried it, and without changing the resolver, then TargetPlatform doesn't contain the extra source bundles that are not explicitly required. 

Sorry, the TargetPlatform object seems to contain all visible artifacts; but the `targetPlatformRepository` basically contains the DependencyArtifacts. Indeed, feeding the mirror application with the TargetPlatform should allow to get all available sources.

laeubi:
> So I just tried it, and without changing the resolver, then TargetPlatform doesn't contain the extra source bundles

Can you say what `TargetPlatform` you specifically refer to?

> The content.xml in target/targetPlatformRepository/ which is used as input of the mirror app in AssembleRepositoryMojo only contains the resolved artifacts, not everything that's declared available.

`MirrorApplication` is today "just" a plain java  code and you should be able to pass it any Tycho objects ... This could probably simplified today with less indirection I was just not able to come to that point with my cleanups yet :-)

laeubi:
Probably `org.eclipse.tycho.p2.tools.RepositoryReferences` could just carry the full Target platform I think we mostly used URLs here because of the former restrictions of Tycho with a non-connected Framework ...

laeubi:
One more thing that came into my mind, we have

- org.eclipse.tycho:tycho-p2-publisher-plugin:${project.version}:publish-osgi-ee,
- org.eclipse.tycho:tycho-p2-publisher-plugin:${project.version}:publish-products,
- org.eclipse.tycho:tycho-p2-publisher-plugin:${project.version}:publish-categories,

that publish stuff into this "target repository", maybe one can use similar approach to publish the sources so assemble-repository can find them there?

mickaelistria:
Back then, I initially thought about making it an extra mojo rather than a flag. it looks like you last comment kind of adds credit to this idea. Do you agree a `tycho-p2-repository:add-sources` mojo would be actually simpler/better?

laeubi:
Thinking more about it, I think we actually need it at the MirrorApp or do it afterwards... otherwise we can't really know what sources are needed ... I'll check your example and make some experiments today.

laeubi:
This should be merged as a prerequisite, then passing the TP to the mirror mojo and then the Slicer can use the TP to find the source units (I'll prepare another commit on top of this then):
- https://github.com/eclipse-tycho/tycho/pull/1957

mickaelistria:
> Thinking more about it, I think we actually need it at the MirrorApp or do it afterwards... otherwise we can't really know what sources are needed ... 

FWIW, the current proposal is "good enough" for the use-case of Platform. It's fine if this can be achieved some other way, but functionally, I don't think we need something "better" here; reading the DependencyArtifacts to try to decide of source bundles to add works OK.
But if you want to try another approach which works at least equally fine; feel free. If you don't find time to work on it soon, maybe the most productive approach would be to merge this PR and consider improvements later.

laeubi:
I'll push a suggestion soon :-)

mickaelistria:
Closing in favor of https://github.com/eclipse-tycho/tycho/pull/1960

PR #1951, opened 2023-01-10
Add a demo how to use bnd-testing-plugin with Tycho

laeubi:
Just work in progress...


laeubi:
Due to different BND constraints, this does not work nicely and a better integration can be archived with a own mojo here:
- https://github.com/eclipse-tycho/tycho/pull/1961

PR #1960, opened 2023-01-11
Pr 1950

laeubi:
WDYT @mickaelistria ? If tests passing we might want to squash this before the merge ...


mickaelistria:
Implementation looks OK to me. If the test passes, then it's good to be (squashed and) merged

mickaelistria:
looks like it working; and the PGP tests need to be updated after this fix.

laeubi:
Is it expected that the pgp include the source? I'll try to fix the test then or check why sources now always included...

mickaelistria:
I can't tell the details of what's expected for sure, as @merks wrote this test. But I believe the assertion are incorrect as `includeAllSources` was not complete until then. It seems now more correct, and the assertions need to be checked to properly include -where they fit- the newly added sources.

merks:
I'm not sure how to parse the question "Is it expected that the pgp include the source?"  Are you asking whether it's expect that source bundles should be PGP signed, then yes, if the bundle for which it is the source is PGP signed then the source bundle should be PGP signed.  (And similarly if the bundle is jar signed the source bundle should be jar signed.)

laeubi:
The question is: 
Is the test written in a way that it expects that source bundles are added, but did not previously asserted that --> we nee d to fix the test

Or  is the test written without source bundles in mind, so they should never be part of the result --> code must probably be fixed

mickaelistria:
I think the test was written accepting the limitations/bugs of `includeAllSources` that we intend to fix right here; so the test must be fixed to add the source bundles (as not PGP signed ones because they are jarsigned already).

merks:
The category.xml is like this:
```
<site>
   <feature id="org.eclipse.tycho.maven.all"/>
   <feature id="org.eclipse.tycho.maven.all.source"/>
   <bundle id="org.eclipse.equinox.common"/>
   <bundle id="org.eclipse.platform_root"/>
</site>
```
but the pom specifies this:
```
			<plugin>
				<configuration>
					<repositoryName>Tycho Maven</repositoryName>
					<includeAllSources>true</includeAllSources>
				</configuration>
			</plugin>
```
So I think that `includeAllSources` was working differently when the expectations where captured than they are now.  So I think the expectations should be updated to reflect the additional sources...

merks:
Or maybe easier if  `<includeAllSources>true</includeAllSources>` is changed to false?

laeubi:
> Or maybe easier if `<includeAllSources>true</includeAllSources>` is changed to false?
> 
> It's this file: /tycho-its/projects/gpg.sign.p2.basic/site/pom.xml

You can decide if you think this is important for test test or not :-)
I'll then try to adjust the test to what ever is decided here.

mickaelistria:
I think the current combination if includeAllSources=true + PGP signing is more interesting to cover (as it's what Platform uses) and is also more complex, so we can hope that if this case works, there are good chances things work with includeAllSources=false, more than the other way round.

merks:
I agree with @mickaelistria that keeping includeAllSources=true is more interesting and useful to test so adjusting the tests based on the new/fixed/improved implementation for that case would be good.

PR #1970, opened 2023-01-13
Pr 1964

laeubi:


laeubi:
FYI @iloveeclipse this update should hopefully fix the memory problem we see with previous surefire updates.

Just in case you have some setup to test/prove this without running the full itest-build you can do the following:

1. Fetch this PR
2. run `mvn clean install -DskipTests -T1C -Dtycho.localArtifacts=ignore`
3. run your test with local snapshot build

laeubi:
Still some license not vetted and two tests are failing...

Issue #1969, opened 2023-01-12
Maven Loockup can become really slow and should cache previous requests


mickaelistria:
https://ci.eclipse.org/tycho/job/tycho-github/job/master/38/console

```
ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 435.657 s <<< FAILURE! - in org.eclipse.tycho.test.packaging.ConsumerPomTest
[ERROR] org.eclipse.tycho.test.packaging.ConsumerPomTest.testReplaceP2  Time elapsed: 435.652 s  <<< FAILURE!
java.lang.AssertionError: dependency with groupId=org.eclipse.platform and artifactId=org.eclipse.equinox.preferences not found in [Dependency {groupId=org.osgi, artifactId=org.osgi.service.prefs, version=1.1.2, type=jar}]
	at org.junit.Assert.fail(Assert.java:89)
	at org.eclipse.tycho.test.packaging.ConsumerPomTest.assertHasDependency(ConsumerPomTest.java:50)
	at org.eclipse.tycho.test.packaging.ConsumerPomTest.testReplaceP2(ConsumerPomTest.java:39)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
	at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
[...]

PR #1972, opened 2023-01-13
Backport of #1969

laeubi:
FYI @hd42


laeubi:
Only the current timeouts from maven central AFAIK...

PR #1958, opened 2023-01-11
Issue 1943 reproducer

glatuske:
Fix for #1943

I have cherry-picked the missing change, decremented Tycho version for view modules and added a test case to existing JUnit 5.9 integration tests for the non working parametrized tests with method source.

PR #1974, opened 2023-01-13
add test case for parameterized tests with method source for JUnit 5.9

glatuske:
Cherry-picked integration test case for JUnit 5.9 with parametermized tests and method source as request in https://github.com/eclipse-tycho/tycho/pull/1958#issuecomment-1381439800


laeubi:
Thanks for the test-case!

PR #1983, opened 2023-01-15
Improve Oomph setup

Bananeweizen:
The working set wrongly matched everything that would be in "dev\tycho\git\foo.git", as I noticed when I created an additional clone of the Maven sources.


Bananeweizen:
If it is not clear from the XML alone, the new predicate means "in same repository as project tycho-core".

Issue #1981, opened 2023-01-15
Tycho doesn't respect --no-transfer-progress


Bananeweizen:
Building m2e with "mvn verify -ntp" all the non-Tycho downloads are silenced, but Tycho still shows progress messages, although it shouldn't:
![grafik](https://user-images.githubusercontent.com/406876/212534423-d5f6fae5-765b-4a20-ad49-4ac913ddcaaa.png)

laeubi:
Tycho currently only support `-B`  option.

Bananeweizen:
Surprisingly, I did not even find a related getter in MavenExecutionRequest, while I would have expected that in the version of the maven-core.jar already. Am I looking into the wrong part of the code?

laeubi:
Usual sequence is:

1. Checkout maven code
2. Search for `"-ntp"` (Should be somewhere in `MavenCli` or near to it...
3. Check how this is then passed and where this set the value, some values are even not accessible from the plugin at all
4. now use that in Tycho e.g. `org.eclipse.tycho.p2maven.transport.TychoRepositoryTransport` wher currently only interactive is evaluated (from `session.getRequest().isInteractiveMode()`)

laeubi:
@Bananeweizen I think this is fixed now, do you like to propose a patch for maven to add an explicit accessor here:

https://github.com/apache/maven

I think this would be useful for others as well.

PR #1984, opened 2023-01-15
Respect -ntp argument of Maven

Bananeweizen:
If the -ntp argument is present, MavenCli creates a new instance of QuietMavenTransferListener, which is just a concrete version of the abstract super class, with no code added.

Since there is no explicit flag in MavenExecutionRequest, we can only check for the class name as workaround.


Bananeweizen:
I have not been able to test this locally. I'm further away from a build than ever before. I even get JDT compiler exceptions every some seconds in the IDE.
Fixes #1981

laeubi:
> I even get JDT compiler exceptions every some seconds in the IDE.

You mean the  error popup? You probably want to share your feelings about its usefulness here:

- https://github.com/eclipse-platform/eclipse.platform.ui/issues/509

Beside that I always use `mvn clean install -DskipTests -T1C -Dtycho.localArtifacts=ignore` to perform a full build and test it with an example project.

Bananeweizen:
Great tip. Together with the temporary directory for maven, I got it working that way. And m2e then builds without the previous download output and rather "waits" several seconds there. I'd consider this working now.

PR #1998, opened 2023-01-18
Stronger typing in target-platform-configuration mojo

mickaelistria:
That should improve a bit the assistance in pom editors.


laeubi:
I'm not sure if this works as you can have any number of URl, File, GAV... not only one item.

mickaelistria:
Indeed, it's not perfect not complete, but at least, it should allow that completion works for the first children; so it's a bit better.

laeubi:
I'm just not sure if maven would complain / fail if there are multiple items?

mickaelistria:
> I'm just not sure if maven would complain / fail if there are multiple items?

Maven wouldn't as it doesn't validate the mojo parameters; but maybe lemminx-maven would... I didn't yet manage to try it for real, but if it shows a warning or an error, it's probably not worth it. Let's wait a bit for actual testing then.

laeubi:
Oka,y then it might be enough to just add a javadoc that says the elemtn can orrour one or more times

lppedd:
Getting rid of XML-related types is probably also better for the polyglot UX.

laeubi:
In the maven channel it was mentioned that the actual correct type for such a case would be `org.codehaus.plexus.configuration.PlexusConfiguration`

mickaelistria:
> In the maven channel it was mentioned that the actual correct type for such a case would be org.codehaus.plexus.configuration.PlexusConfiguration

You mean instead of Xpp3Dom ?

laeubi:
> You mean instead of Xpp3Dom ?

Yes and in one place we even use Object, both seems to be invalid (but works here as the mojo is actually never executed).

target-platform-configuration/src/main/java/org/eclipse/tycho/target/TargetPlatformConfigurationMojo.java
```diff
@@ -183,7 +184,7 @@ public class TargetPlatformConfigurationMojo extends AbstractMojo {
      * </pre>
      */
     @Parameter(name = DefaultTargetPlatformConfigurationReader.DEPENDENCY_RESOLUTION)
-    private Object dependencyResolution; // TODO stronger typing
+    private PlexusConfiguration dependencyResolution; // TODO stronger typing
```
laeubi, line 187: I wonder what (beside `extraRequirements`) is actually allowed here? The doc just says "for example" ...

laeubi:
@mickaelistria one thing that recently came into my mind:

probably we should simply fade out the support for specify multiple targets in the configuration here:

1. With the target reference one can now emulate this with a standard technique that also works in the PDE
2. PDE itself never has supported this, and I admit I think never will support that given point (1)
3. I have never seen any project that is actually using it, so it seems a very rare use-case
4. [With the new support of specify a target as a repository](https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#define-targets-in-repository-section) one might better use that approach if an additional (Tycho build specific) target is desired and might be more descriptive then.

This would simplify this whole thing a lot as we then can specify as you suggested, and can simply assume there is always one and given an error otherwise.

mickaelistria:
> probably we should simply fade out the support for specify multiple targets in the configuration here

I agree with that and find the reasons you mention are very good. So let's do as you suggest and design the mojo so it's assumed only 1 target is allowed. We can put a note in the documentaiton about multiple of them being allowed, that should be enough.

Issue #1482, opened 2022-10-06
Lots of "No digest algorithm is available to verify download of " during assemble-repository


mickaelistria:
From some random projects (known examples so far are TM4E and LinuxTools but I suspect it's widespread) and build it with `mvn verify -DskipTests`; during the assemble-repository step, you get
```
[INFO] --- tycho-p2-repository-plugin:3.0.0-SNAPSHOT:assemble-repository (default-assemble-repository) @ org.eclipse.tm4e.repository ---
[WARNING] [c7c8a64d-0775-468f-a89c-2d83d6f5d371][extension>org.eclipse.tycho:tycho-maven-plugin:3.0.0-SNAPSHOT] No digest algorithm is available to verify download of org.eclipse.update.feature,org.eclipse.tm4e.language_pack.feature,0.1.0.202206131926.
[WARNING] [c7c8a64d-0775-468f-a89c-2d83d6f5d371][extension>org.eclipse.tycho:tycho-maven-plugin:3.0.0-SNAPSHOT] No digest algorithm is available to verify download of osgi.bundle,org.eclipse.tm4e.ui,0.6.0.202206241118.
[WARNING] [c7c8a64d-0775-468f-a89c-2d83d6f5d371][extension>org.eclipse.tycho:tycho-maven-plugin:3.0.0-SNAPSHOT] No digest algorithm is available to verify download of osgi.bundle,org.eclipse.tm4e.registry.source,0.6.1.202208131654.
[WARNING] [c7c8a64d-0775-468f-a89c-2d83d6f5d371][extension>org.eclipse.tycho:tycho-maven-plugin:3.0.0-SNAPSHOT] No digest algorithm is available to verify download of osgi.bundle,org.jcodings,1.0.57.
[WARNING] [c7c8a64d-0775-468f-a89c-2d83d6f5d371][extension>org.eclipse.tycho:tycho-maven-plugin:3.0.0-SNAPSHOT] No digest algorithm is available to verify download of osgi.bundle,org.eclipse.tm4e.markdown.source,0.5.0.202206241118.
[WARNING] [c7c8a64d-0775-468f-a89c-2d83d6f5d371][extension>org.eclipse.tycho:tycho-maven-plugin:3.0.0-SNAPSHOT] No digest algorithm is available to verify download of osgi.bundle,org.eclipse.tm4e.core.source,0.5.0.202206232130.
[...]

PR #2003, opened 2023-01-18
No digest algorithm

mickaelistria:


laeubi:
At thir point of time we only generate inital metadata (what fails in earlier stages as OSGi is not ready there), the error message is ommited for the final produced artifact (that is attatch-p2-metadata).

laeubi:
I'm a bit confused as you linked 
- https://github.com/eclipse-tycho/tycho/issues/1482
what is much older as the mentioned commit, so can you explain what exactly message you can reproduce/causes issues in the test case I then can take a look at this.

mickaelistria:
Ineed, my analysis was incomplete. I found 1 commit that added the NO_MD5 option (which is poorly named as it's "no signature"). But there are other and older places where this is set too which are more involved in the issue.

laeubi:
I'll try to take a look at the "No signature available" stuff, so I assume the new test-case shows this message already?

mickaelistria:
> I'll try to take a look at the "No signature available" stuff, so I assume the new test-case shows this message already?

Yes.
Note that https://github.com/eclipse-tycho/tycho/commit/231d495aeccf88586faf6e9d9b16c5d2e1ff3933 was a decent functional fix, making the test pass successfully (not showing the message); I am just trying to find something better if possible.

laeubi:
I have a fix for this locally that do work with standard P2 (so no extra work by Tycho) i'll push this tomorrow after enhancing my testcase a bit ...

laeubi:
See:
- https://github.com/eclipse-tycho/tycho/pull/2008

laeubi:
I think this one is not required anymore?

mickaelistria:
> I think this one is not required anymore?


Indeed

PR #2008, opened 2023-01-20
Add an option to generate checksums by the P2MetadataMojo

laeubi:
FYI @mickaelistria this reproduces the message for me with a very small and isolated case (no target platform required just plain Tycho and some empty artifacts) ... I'll check how to mitigate the message.


mickaelistria (approved):
Thanks!

Issue #2009, opened 2023-01-20
Issue getting sources


mickaelistria:
See attached project [jdom-sources.zip](https://github.com/eclipse-tycho/tycho/files/10469777/jdom-sources.zip), and build it with a clean repo.
Expected: Build works
Got:
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-p2-repository-plugin:4.0.0-SNAPSHOT:assemble-repository (default-assemble-repository) on project include-jdom-sources: Could not assemble p2 repository: Mirroring failed: Messages while mirroring artifact descriptors.: [Failed to transfer artifact packed: osgi.bundle,org.jdom.source,1.1.1.v201101151400.: [Download of osgi.bundle,org.jdom.source,1.1.1.v201101151400 failed on repository file:/resolution-context-artifacts@%252Fhome%252Fmistria%252Fgit%252Feclipse.platform.releng.aggregator-GitHub%252Feclipse.platform.releng.tychoeclipsebuilder%252Feclipse.platform.repository-test. Retrying. : [Artifact packed: osgi.bundle,org.jdom.source,1.1.1.v201101151400 is not available in the repository file:/home/mistria/sandbox/lolrepo/]; Artifact not found: packed: osgi.bundle,org.jdom.source,1.1.1.v201101151400.]] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
```

mickaelistria:
Note that on the 2nd build, when artifact is already in the local repository, then the error isn't happening...

mickaelistria:
I'm trying to write an automated test, but it's made trickier as we need to use a dedicated Maven repo (which makes the test considerably slower). If there a known good practice to mimic for that case? If not, should I just set the local repo to a temp directory to run this test?

mickaelistria:
As I'm debugging, I can see that the line causing the issues are https://github.com/eclipse-tycho/tycho/blob/c2025b31b8d88cd0042d3555de63f4df32997e03/tycho-bundles/org.eclipse.tycho.core.shared/src/main/java/org/eclipse/tycho/p2/repository/MirroringArtifactProvider.java#L176..L181 which were introduce with #1980 .
[...]

PR #2010, opened 2023-01-20
Exclude packed artifacts

mickaelistria:
Those aren't supported anymore, but some older p2 repo tend to still list them with high priority. Explicitly exclude those.

Fixes https://github.com/eclipse-tycho/tycho/issues/2009#issuecomment-1398980830

### DIFF
--- a/tycho-its/projects/p2Repository.includeAllSources.oldOrbit/category.xml
+++ b/tycho-its/projects/p2Repository.includeAllSources.oldOrbit/category.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<site>
+   <bundle id="org.jdom"/>
+</site>
--- a/tycho-its/projects/p2Repository.includeAllSources.oldOrbit/jdom-from-old-orbit.target
+++ b/tycho-its/projects/p2Repository.includeAllSources.oldOrbit/jdom-from-old-orbit.target
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<?pde version="3.8"?>
+<target name="eclipse-sdk-prereqs">
+  <locations>
+    <location includeAllPlatforms="true" includeMode="slicer" type="InstallableUnit">
+      <!-- Transitive dependency of jxpath, not available as OSGi bundle on Maven Central -->
+      <unit id="org.jdom" version="1.1.1.v201101151400"/>
+      <unit id="org.jdom.source" version="1.1.1.v201101151400"/>
+      <repository location="https://download.eclipse.org/tools/orbit/downloads/drops/R20201118194144/repository/"/>
+    </location>
+  </locations>
+  <targetJRE path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-17"/>
+</target>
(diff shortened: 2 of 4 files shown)


tycho-bundles/org.eclipse.tycho.core.shared/src/main/java/org/eclipse/tycho/p2/repository/MirroringArtifactProvider.java
```diff
@@ -174,9 +175,12 @@ public final IArtifactDescriptor[] getArtifactDescriptors(IArtifactKey key) thro
             return localArtifactRepository.getArtifactDescriptors(key);
         }
         if (remoteProviders.contains(key)) {
-        	//we can use it if the remote contains it, but want to shadow the repository returned by the descriptor
-        	//just in case someone uses this to download an artifact we are asked again...
+            //we can use it if the remote contains it, but want to shadow the repository returned by the descriptor
+            //just in case someone uses this to download an artifact we are asked again...
             return Arrays.stream(remoteProviders.getArtifactDescriptors(key))
+                    .filter(desc -> Arrays.stream(desc.getProcessingSteps())
```
laeubi, line 181: You should be able to use `ArtifactTransferPolicy.isCanonicalFormat(descriptor)`

mickaelistria: Thanks for the tip. Applied.

laeubi:
@mickaelistria is this a test-case for the recent platform failures, e.g. could be used to debug / check if 
- https://github.com/eclipse-tycho/tycho/pull/2013

could affect this too?

mickaelistria:
The test covers the issue reported in comments for https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/773 or on the mailing-list https://www.eclipse.org/lists/platform-releng-dev/msg39053.html

mickaelistria:
And no I don't think this test covers the issue described in #2013 . This seems to be another bug.

laeubi:
@mickaelistria I have now tried my patch from https://github.com/eclipse-tycho/tycho/pull/2013 and your test-case and it also "fixes" the build, so we now have a very good test-case for this, would you mind submit the testcase alone in a PR so we see it fails the current build and can base possible fixes/further investigations on it?

laeubi:
I think the root-cause is this place:
```
org.eclipse.equinox.p2.internal.repository.mirroring.Mirroring
org.eclipse.equinox.p2.internal.repository.mirroring.Mirroring.run(boolean, boolean)
org.eclipse.equinox.p2.internal.repository.mirroring.Mirroring.mirror(IArtifactDescriptor, boolean)
```
there a flag is checked (`includePacked`) and if that flag is `true` so it tires to actually mirror packed items! So the fix should be not to skip packed descriptors here (what we can't always control like here) but to set that flag to `false`. Nerveless I think your approach of filtering packed descriptors should be applied as the `MirroringArtifactProvider` always only using non-packed items.

mickaelistria:
This PR only contains the test now. Let's hope this test fail.
The fix that was initially proposed with it is https://github.com/eclipse-tycho/tycho/commit/b679a060b69edd5fd0da711de6f2e826fa06dcf4

mickaelistria:
Replaced with https://github.com/eclipse-tycho/tycho/pull/2015

Issue #2005, opened 2023-01-19
Consumer-POM should use packaging-type jar instead of eclipse-plugin


hd42:
When I create a consumer-pom for my plugin project, I want that project to be usable as a plain old maven dependency. But Maven seams to read the POM (at least for transitive dependencies) and decides to look for a file with the extension .eclipse-plugin instead of .jar.

This can be avoided by adding 
```
		<extensions>
			<extension>
				<groupId>org.eclipse.tycho</groupId>
				<artifactId>tycho-build</artifactId>
				<version>3.0.2-SNAPSHOT</version>
			</extension>
		</extensions>
```
to the POM referencing the Plugin-JAR, but that is something counterintuitive for the JARs consumers, who are not using Tycho - else they wouldn't be using the JAR as a maven dependency.

It would be nice if Tycho could (optionally?) replace the packing type when creating the consumer-pom.

laeubi:
The problem is that "plain old maven" sadly do only know a fixed list of "maven managed" artifact types, everything else requires an extension.

I opened this here at Maven:
- https://github.com/apache/maven/pull/964

but will take care to put an option to "lying" about the actual packaging type to Tycho consumer pom generator.

laeubi:
@hd42 can you share a minimal demo project for further analysis why maven actually try to use the wrong type?

laeubi:
The fix for this issue was gently sponsored by [Faktor Zehn GmbH](https://www.faktorzehn.com).

PR #2012, opened 2023-01-21
Backport of #2005

laeubi:
FYI @hd42


laeubi:
@hd42 the new snapshot should be available in a few hours do you like to check if this works like you expected (just keep in mind it must be activated first, see changed test-case).

PR #2013, opened 2023-01-21
Restore the sideeffect of download and the check that the local repos really contain the artifact.

laeubi:
@mickaelistria this is just an idea that came into my mind, that probably some code in Tycho relies on the previous sideefect of "contains" downloading (and checking) the artifact, so probably this is a fix for the Platform issues?

FYI @iloveeclipse

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.core.shared/src/main/java/org/eclipse/tycho/p2/repository/MirroringArtifactProvider.java
+++ b/tycho-bundles/org.eclipse.tycho.core.shared/src/main/java/org/eclipse/tycho/p2/repository/MirroringArtifactProvider.java
@@ -174,8 +174,8 @@ public class MirroringArtifactProvider implements IRawArtifactFileProvider {
             return localArtifactRepository.getArtifactDescriptors(key);
         }
         if (remoteProviders.contains(key)) {
-        	//we can use it if the remote contains it, but want to shadow the repository returned by the descriptor
-        	//just in case someone uses this to download an artifact we are asked again...
+            //we can use it if the remote contains it, but want to shadow the repository returned by the descriptor
+            //just in case someone uses this to download an artifact we are asked again...
             return Arrays.stream(remoteProviders.getArtifactDescriptors(key))
                     .map(base -> new MirrorArtifactDescriptor(base, this)).toArray(IArtifactDescriptor[]::new);
         }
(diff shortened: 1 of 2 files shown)


mickaelistria:
> this is just an idea that came into my mind, that probably some code in Tycho relies on the previous sideefect of "contains" downloading (and checking) the artifact, so probably this is a fix for the Platform issues?

It could be, yet the current state makes sense from Tycho POV. I hope we can better isolate all the current Platform build issues (we already did for a few of them), and just make things work without restoring the early download. This would be beneficial for Tycho and all its consumers.
So I wouldn't advocate restoring the previous behavior "blindly" until it's made clear that it is the cause of some issue; and if we can have a good example proving it can be a cause, then we can improve the current state to fix it.

laeubi:
It only restores the part on "contains" (the getdescriptors is still lazy), because *some* code seem to use

```
if (contains(...)) {
   return getFile();
}
```
where it is assumed that the file exits after the call... I thought I have found all places, but just in case wanted to prepare this PR, it should not make things wore at least...

mickaelistria:
OK, if you're confident that this change makes Tycho behave better in general, then feel free to merge it.

iloveeclipse:
Something is broken in recent tycho.
See this new fail in JDT UI jenkins:
https://ci.eclipse.org/jdt/job/eclipse.jdt.ui-github/job/PR-406/1/console
```
13:02:21  [ERROR] Number of foreign imports: 1
13:02:21  [ERROR] import: Entry[import  from realm ClassRealm[project>eclipse.jdt.ui:eclipse.jdt.ui:4.27.0-SNAPSHOT, parent: ClassRealm[maven.api, parent: null]]]
```

If we could fix something with this PR, please merge.

laeubi:
This failures seems not related, can you restart the build (Using the replay option of Jenkins) and probably add `-U` to the maven arguments to force updates of snapshots?

laeubi:
> OK, if you're confident that this change makes Tycho behave better in general, then feel free to merge it.

This is currently just a guess, is there any reproducing testcase/build I can use to probably verify (or debug) this?

iloveeclipse:
@laeubi : I guess the problem is in https://github.com/eclipse-tycho/tycho/issues/2014

laeubi:
I'll set this to draft for a while if we still need it but it seems not be part of the (reported) problem.

mickaelistria:
Isn't it covered by https://github.com/eclipse-tycho/tycho/pull/2015 ?

laeubi:
I hope so it seems this was not an issue with that problem at laest.

PR #2015, opened 2023-01-22
Pr 2010 testcase source

laeubi:
Rebased and also setting `setIncludePacked=false`
- https://github.com/eclipse-tycho/tycho/pull/2010


laeubi:
FYI @akurtakov maybe we should either remove this option in P2 at all or at least set it to false by default, I'll leave that up to you :-)

laeubi:
Hm... seems one test now fails, have to investigate, @mickaelistria if you have an idea if this is to be expected let me know.

laeubi:
@mickaelistria I found the problem with failing test, so from my side when Jenkins succeeds now this can be merged (feel free to push the button if I'm not around).

PR #2027, opened 2023-01-26
Add the AbstractApplication / MirrorApplication to Tycho code base

laeubi:
I don't know if it is strictly required but would be good to get a +1 as PL from @mickaelistria or @akurtakov here.

I hope the commit message and readme is sufficent, but I copy the relevant parts to this PR as well just in case:

> Tycho already customizes the P2 applications but they are not really used as a real application and instead called directly. In the past especially the MirrorApplication required some adjustments to fit Tycho needs, but it was not written with extensibility in mind.
> 
> Even though sometimes we we able to work around that limitations or apply patches to P2 this really has slowed down the development and ability of Tycho to deliver new value to their users. On the other hand none of these changes has had added any value to pure P2 users and there are no real innovation in these areas over the years.
> 
> Because of this, Tycho now includes a copy of these classes (but remains the extension pattern to allow apply of P2 patches to Tycho) to fasten development and decouple Tycho form P2 in the area of tools.

I have split the PR in two commits, one that adds the readme + original code (with just imports adjusted) as a base, and a second one that links Tycho with the copied code and apply some fixes due to changed API in P2 and unnecessary static Activator references.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/p2tools/MirrorApplication.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/p2tools/MirrorApplication.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2011 SAP SE and others.
+ * Copyright (c) 2010, 2023 SAP SE and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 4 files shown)


akurtakov:
As much I dislike coping code here and there I also know the pain having to wait 3 months for a change in p2 to be released so tycho can make use of it. Thus if you are 100% sure this is the path forward - go for it.

PR #2028, opened 2023-01-26
Filter provided units and artifacts

laeubi:
A work in progress (test is required and depend on #2027) but something that came into my mind when reading 
https://github.com/eclipse-m2e/m2e-core/issues/1214#issuecomment-1404244327

FYI @HannesWell  As m2e references some 3rd party sites, we might can use this option to include everything but those already available from the referenced repos without mention all of them in features.


HannesWell:
That sounds like a very handy feature und should further improve the quality of p2-repos 👍🏽

laeubi:
@HannesWell test succeed so this could be investigated at m2e after the next snapshot deployment.

PR #2035, opened 2023-01-30
Add a TychoReactorReader that mimics maven behavior for Tycho packagings

laeubi:
FYI @lppedd lets see if this makes anything worse ...

This includes 
- https://github.com/eclipse-tycho/tycho/pull/1407

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoReactorReader.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoReactorReader.java
@@ -1,0 +1,103 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Christoph Läubrich and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *    Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.tycho.core.maven;
+
+import static java.util.function.Function.identity;
+import static java.util.stream.Collectors.toMap;
+
+import java.io.File;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import javax.inject.Inject;
+
+import org.apache.maven.SessionScoped;
+import org.apache.maven.artifact.ArtifactUtils;
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.model.Model;
+import org.apache.maven.project.MavenProject;
+import org.apache.maven.repository.internal.MavenWorkspaceReader;
+import org.codehaus.plexus.component.annotations.Component;
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.repository.WorkspaceReader;
+import org.eclipse.aether.repository.WorkspaceRepository;
+import org.eclipse.tycho.PackagingType;
+
+/**
+ * An implementation of a workspace reader similar to what maven does but for the Tycho packaging
+ * types
+ */
+@SessionScoped
+@Component(role = WorkspaceReader.class, hint = TychoReactorReader.HINT)
+public class TychoReactorReader implements MavenWorkspaceReader {
+    static final String HINT = "tycho-reactor";
+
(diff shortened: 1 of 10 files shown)


tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoReactorReader.java
```diff
@@ -0,0 +1,103 @@
...
+    private boolean isTychoReactorArtifact(Artifact artifact) {
+        if (artifact.getClassifier() == null || artifact.getClassifier().isBlank()) {
+            return PackagingType.TYCHO_PACKAGING_TYPES.contains(artifact.getProperty("type", ""));
+        }
+        return false;
+    }
+
+    @Override
+    public List<String> findVersions(Artifact artifact) {
+        return getTychoReactorProject(artifact).map(project -> List.of(artifact.getVersion())).orElse(List.of());
```
lppedd, line 93: In my custom extension I always return an empty list here. 
If I recall correctly this is basically unused.

laeubi: `DefaultVersionRangeResolver` calls this method.

tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoReactorReader.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public File findArtifact(Artifact artifact) {
+        return getTychoReactorProject(artifact).map(project -> {
+            org.apache.maven.artifact.Artifact mainArtifact = project.getArtifact();
+            if (mainArtifact != null && mainArtifact.getFile() != null && mainArtifact.getFile().exists()) {
+                return mainArtifact.getFile();
+            }
+            return project.getBasedir();
```
lppedd, line 66: Not sure if it's more correct to use `project.getBasedir()` or `project.getBuild().getOutputDirectory()` (latter is what I currently return).

`ReactorReader` does this:
```java
if (isTestArtifact( artifact )) {
    if (project.hasLifecyclePhase("test-compile")) {
        return new File(project.getBuild().getTestOutputDirectory());
    }
} else {
    String type = artifact.getProperty("type", "");
    if (project.hasLifecyclePhase("compile") && COMPILE_PHASE_TYPES.contains(type)) {
        return new File(project.getBuild().getOutputDirectory());
    }
}
```
And that's why I've picked `project.getBuild().getOutputDirectory()`.

lppedd:
I can definitely say this is going to work, as I already use this solution!

tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoReactorReader.java
lppedd: @laeubi ahh yes, you're right, not sure why the search didn't bring up any reference.

tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoReactorReader.java
laeubi: Tycho uses the basedir in such situations already so this is consistent (even though a bit bad and we might later want to get rid of it).

laeubi:
> I can definitely say this is going to work, as I already use this solution!

It seems to pass the testsuite and your new test is also recognized, but lets wait for the Jenkins build ... the problem is that there are sometimes quite special cases in Tycho so even if it works most of the times one can be surprised some times :-)

tycho-its/src/test/java/org/eclipse/tycho/test/resolver/TychoGenerateSourcesIssueTest.java
```diff
@@ -0,0 +1,17 @@
...
+
+import org.apache.maven.it.Verifier;
+import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
+import org.junit.Test;
+
+public class TychoGenerateSourcesIssueTest extends AbstractTychoIntegrationTest {
+    @Test
+    public void testTychoGenerateSourcesIssue() throws Exception {
+        Verifier verifier = getVerifier("resolver.generateSources", false, false);
+        verifier.addCliOption("-Dtycho.resolver.classic=false");
```
lppedd, line 13: Just noticed this. Maybe it's better to remove this line?

PR #2039, opened 2023-01-31
Do not resolve using Maven coords when the requested artifact is from p2

lppedd:
I've reorganized `TychoWorkspaceReader` in a way it's more readable and maintainable.

`findMavenArtifact` has been left blank with indication on how to implement it. 

An alternative to the `findP2Artifact`/`findMavenArtifact` separation is to have a single method which knows if we want to resolve as p2 or Maven, e.g., `findArtifact(artifact, isP2Artifact)`.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoWorkspaceReader.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoWorkspaceReader.java
@@ -31,20 +31,13 @@ import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.aether.repository.LocalRepository;
 import org.eclipse.aether.repository.WorkspaceReader;
 import org.eclipse.aether.repository.WorkspaceRepository;
-import org.eclipse.tycho.ArtifactDescriptor;
-import org.eclipse.tycho.ArtifactKey;
-import org.eclipse.tycho.DependencyArtifacts;
-import org.eclipse.tycho.MavenDependencyDescriptor;
-import org.eclipse.tycho.ReactorProject;
-import org.eclipse.tycho.TychoConstants;
+import org.eclipse.tycho.*;
 import org.eclipse.tycho.core.osgitools.DefaultReactorProject;
-import org.eclipse.tycho.core.resolver.P2ResolverFactory;
 import org.eclipse.tycho.core.utils.TychoProjectUtils;
 
 @Component(role = WorkspaceReader.class, hint = "TychoWorkspaceReader")
 public class TychoWorkspaceReader implements MavenWorkspaceReader {
-
-    private WorkspaceRepository repository;
+    private final WorkspaceRepository repository;
 
     @Requirement
     private LegacySupport legacySupport;
@@ -54,9 +47,6 @@ public class TychoWorkspaceReader implements MavenWorkspaceReader {
 
     @Requirement
     private ModelWriter modelWriter;
-
-    @Requirement
-    P2ResolverFactory factory;
 
     public TychoWorkspaceReader() {
         repository = new WorkspaceRepository("tycho", null);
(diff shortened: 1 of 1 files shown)


tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoWorkspaceReader.java
```diff
@@ -68,110 +53,149 @@ public WorkspaceRepository getRepository() {
     }
 
     @Override
-    public File findArtifact(Artifact artifact) {
+    public File findArtifact(final Artifact artifact) {
+        final var isP2Artifact = artifact.getGroupId().startsWith(TychoConstants.P2_GROUPID_PREFIX);
```
laeubi, line 57: please don't use `var` for trivial local variables.

tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoWorkspaceReader.java
```diff
@@ -68,110 +53,149 @@ public WorkspaceRepository getRepository() {
...
+        } catch (final IOException e) {
+            logger.debug("Cannot write the POM model", e);
         }
+
+        return null;
+    }
+
+    private File findP2Artifact(final Artifact artifact) {
+        // Attempt a fast lookup first
+        final var cachedFile = getFileForArtifact(artifact);
```
laeubi, line 125: Same here, please use regular types

tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoWorkspaceReader.java
```diff
@@ -68,110 +53,149 @@ public WorkspaceRepository getRepository() {
...
+
+    private File findPomArtifact(final Artifact artifact) {
+        logger.debug("Find the POM file for " + artifact);
+        final var pomFile = getFileForArtifact(artifact);
+
+        if (pomFile.isFile()) {
+            return pomFile;
+        }
+
+        // noinspection ResultOfMethodCallIgnored
```
laeubi, line 110: avoid vendor specific comments in code

tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoWorkspaceReader.java
```diff
@@ -68,110 +53,149 @@ public WorkspaceRepository getRepository() {
...
+
+    private File findP2Artifact(final Artifact artifact) {
+        // Attempt a fast lookup first
+        final var cachedFile = getFileForArtifact(artifact);
+
+        if (cachedFile.isFile()) {
+            return cachedFile;
+        }
+
+        final var session = legacySupport.getSession();
```
laeubi, line 131: Same here...

tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoWorkspaceReader.java
```diff
@@ -68,110 +53,149 @@ public WorkspaceRepository getRepository() {
...
+
+        if (session != null) {
+            final var currentProject = session.getCurrentProject();
```
laeubi, line 134: and all of those ...

tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoWorkspaceReader.java
```diff
@@ -68,110 +53,149 @@ public WorkspaceRepository getRepository() {
...
+        //  compared to the inputted artifact information.
+        //  The returned Maven information should also be validated.
+        //  For that, see MavenDependencyInjector#isValidMavenDescriptor
         return null;
     }
 
-    private Model getP2Model(Artifact artifact) {
-        Model model = new Model();
+    private File getFileForArtifact(final Artifact artifact) {
+        final var repositorySession = legacySupport.getRepositorySession();
```
laeubi, line 168: please avoid var here as well.. this is not JavaScript ;-)

laeubi (changes requested):
Looks fine in general but please avoid erasing types with `var` this makes the code very hard to read and seems unnecessary here.

tycho-core/src/main/java/org/eclipse/tycho/core/maven/TychoWorkspaceReader.java
```diff
@@ -68,110 +58,149 @@ public WorkspaceRepository getRepository() {
...
+
+            if (dependencyMetadata.isPresent()) {
+                logger.debug("Attempting to resolve " + artifact + " for project " + currentProject + "...");
+
+                for (final ArtifactDescriptor descriptor : dependencyMetadata.get().getArtifacts()) {
+                    final ArtifactKey artifactKey = descriptor.getKey();
+                    final String artifactId = artifactKey.getId();
+                    final String version = artifactKey.getVersion();
+
+                    if (artifactId.equals(artifact.getArtifactId()) && version.equals(artifact.getVersion())) {
```
lppedd, line 151: Should we check if the artifact type matches?  
Examples:
```
p2.eclipse.plugin  = eclipse-plugin
p2.eclipse.feature = eclipse-feature
```

laeubi: Features are usually nothing one can "resolve" in maven (at least at the moment), but technically it might be good to check the artifact type.

lppedd: @laeubi 

> are usually nothing one can "resolve" in maven

Guess I'm the first one resolving (I need the JAR to be on disk) feature bundles then 😄   
Ok, I'll add the check.

lppedd: Done.

lppedd:
Do you think it's possible to merge this before the end of next week?  
We need to run builds on our CI machine and I can't build my own version of Tycho there.

laeubi:
I'll try to review this the next days.

laeubi (approved):

laeubi:
Thanks for improving this @lppedd

PR #2046, opened 2023-02-02
Add a new handler for the FTP protocol

lppedd:
An FTP connection for an host is cached up to the end of the build.


p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
```diff
@@ -0,0 +1,198 @@
...
+        if (localFile.isFile()) {
+            FileUtils.forceDelete(localFile);
+        }
+
+        FileUtils.moveFile(tempFile, localFile);
+        localFile.setLastModified(remoteFile.getTimestampInstant().toEpochMilli());
+        return localFile;
+    }
+
+    @PreDestroy
```
laeubi, line 107: Is `PreDestroy` actually called? I think Plexus uses `Initlizable` and `Disposable`?

lppedd: @laeubi it is definitely called, at least in Maven 3.8.6+. Tested via debug, and I use it in multiple plugins in our env.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
```diff
@@ -0,0 +1,198 @@
...
+        final Credentials credentials = authenticator.getServerCredentials(uri);
+
+        if (credentials != null) {
+            client.login(credentials.getUserName(), credentials.getPassword());
+        } else {
+            client.login("anonymous", "");
+        }
+
+        client.enterLocalPassiveMode();
+        client.setFileType(FTP.BINARY_FILE_TYPE);
```
lppedd, line 152: I've now explicitly set the transfer type as binary.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
laeubi: Good to know :+1:

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
```diff
@@ -0,0 +1,198 @@
...
+            } catch (final IOException e) {
+                logger.error("Error disconnecting from host " + entry.getKey(), e);
+            }
+        }
+    }
+
+    protected FTPClient createFtpClient() {
+        final FTPClient client = new FTPClient();
+        client.setControlKeepAliveTimeout(Duration.ofSeconds(15));
+        client.setAutodetectUTF8(true);
```
laeubi, line 124: Usually it is good to use passive mode (as some firewalls blocking FTP traffic otherwise). 
Probably one can use a system property to control this (with defaulting to passive mode).

laeubi (approved):

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
lppedd: @laeubi a question: why are we still using the old Plexus annotations and not the Javax ones? Do we need to keep some sort of compatibility?

laeubi:
@lppedd do you know of any public ftp site we can use for an integration test? Or maybe something like https://mockftpserver.sourceforge.net/fakeftpserver-getting-started.html ?

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
laeubi: 1. As long as it works it doesn't matter and there is no advantage in using "new" annotations
2. the JSR annotation do not work with Java 17 in Maven 3.8.x

lppedd:
> do you know of any public ftp site we can use for an integration test? Or maybe something like [mockftpserver.sourceforge.net/fakeftpserver-getting-started.html](https://mockftpserver.sourceforge.net/fakeftpserver-getting-started.html) ?

I think mocking would be the best way.   
Don't merge until I've at least come up with a very simple test.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
lppedd: @laeubi yup, I now set explicitly:
```
client.enterLocalPassiveMode();
client.setFileType(FTP.BINARY_FILE_TYPE);
```

laeubi:
You can probbably take a look at `org.eclipse.tycho.test.TYCHO279HttpProxy.ProxySupportTest` where we start an embedded jetty server to test the proxy support, something similar for FTP would be good.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
laeubi: This seems a good way.

lppedd:
@laeubi should this FTP handler support a proxy already?  
I see there is a `ProxyHelper` class in case.

Or should I just test it with a very basic setup?

laeubi:
Basic setup is enough, if someone later on wants to add proxy support it can be added later on.

lppedd:
@laeubi I wrote a new `AbstractTychoIntegrationTest`, however it hangs indefinitely in `getVerifier(...)` inside the `@Before` method, has it ever happened to you?

lppedd:
My god this FTP mocking is a PITA... Took me hours to get it working and had to implement a couple commands.

laeubi:
> My god this FTP mocking is a PITA... Took me hours to get it working and had to implement a couple commands.

Writing a good test-case or finding a small reproducing test-case is often the hardest part at all :-)
BTW If you think another [Embedded FTP Server](https://www.google.com/search?client=firefox-b-e&q=Embedded+FTP+Server) is better suited, feel free to use what ever works here!

lppedd:
@laeubi no no, that's the best one can have. Other solutions are more intricate and - if necessary - do no allow the same level of customization.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
```diff
@@ -0,0 +1,200 @@
...
+        final FTPFile remoteFile = getRemoteFile(client, remotePath);
+        final boolean isRemoteMissing = remoteFile == null;
+
+        if (localFile.isFile() && (isRemoteMissing || !mustRefresh(localFile, remoteFile))) {
+            return localFile;
+        }
+
+        if (isRemoteMissing) {
+            logger.warn("Could not find file: " + uri);
+            return null;
```
laeubi, line 79: You could throw `FIleNotFoundException` here.

lppedd: @laeubi do I keep the `logger.warn` there?

lppedd: To be honest I recall not throwing an exception because it was erroring out my build, while it was ok to keep going.
Mmm, not sure

laeubi: Then we should find out whats going on here, because a `null` file will mean it the gets downloaded with the standard URL handle and then should anyways fail with FileNotFound just that the request is made two times. But it is important that it is of the correct Type:
- `FileNotFoundException` the file is not available but everything else is ok
- `AuthenticationFailedException` username/password was wrong
- `IOException` generic failure (e.g. host not reachable at all)

Logging is not required.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
```diff
@@ -0,0 +1,200 @@
...
+
+        if (!client.hasFeature(FTPCmd.MDTM)) {
+            logger.debug("Could not retrieve the last modification timestamp for: " + uri);
+            return -1;
+        }
+
+        final FTPFile file = client.mdtmFile(uri.getPath());
+
+        if (file == null) {
+            throw new IOException("Could not find file: " + uri);
```
laeubi, line 50: This should be FileNotFoundException then

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
```diff
@@ -0,0 +1,200 @@
...
+            return localFile;
+        }
+
+        if (isRemoteMissing) {
+            logger.warn("Could not find file: " + uri);
+            return null;
+        }
+
+        final File parent = FileUtils.createParentDirectories(localFile);
+        final File tempFile = File.createTempFile("download", ".tmp", parent);
```
laeubi, line 83: Can you use `Files` (from NIO API) here as it is generally more safe see 
https://owasp.org/www-community/vulnerabilities/Insecure_Temporary_File

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
```diff
@@ -0,0 +1,200 @@
...
+
+        try (final OutputStream os = new FileOutputStream(tempFile)) {
+            if (!client.retrieveFile(remotePath, os)) {
+                final var message = client.getReplyString();
+                throw new IOException(String.format("Error retrieving file: %s. Message: %s", remotePath, message));
+            }
+        } catch (final IOException e) {
+            tempFile.delete();
```
laeubi, line 91: temp files should also always marked as delete on exit

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
```diff
@@ -0,0 +1,200 @@
...
+        final FTPClient client = new FTPClient();
+        client.setControlKeepAliveTimeout(Duration.ofSeconds(15));
+        client.setAutodetectUTF8(true);
+        client.setBufferSize(8192);
+        return client;
+    }
+
+    private FTPClient getClient(final URI uri) throws IOException {
+        final String host = uri.getHost();
+        final int port = uri.getPort() < 0 ? 21 : uri.getPort();
```
laeubi, line 127: Can we extract the number 21 into a constant, probably even with a systemproperty using Integer.getInteger("...", 21)?

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
```diff
@@ -0,0 +1,200 @@
...
+
+        if (client.isConnected()) {
+            return client;
+        }
+
+        client.connect(host, port);
+
+        if (!FTPReply.isPositiveCompletion(client.getReplyCode())) {
+            client.disconnect();
+            throw new IOException("Could not connect to host: " + key);
```
laeubi, line 139: It might be a good ide to give the user as much information as possible, e.g the replycode + reply message if possible.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
lppedd: @laeubi but if I use NIO, I cannot mark it as delete on close.
I can use NIO everywhere at this point, but this will have to be `Files.delete(...)`

laeubi: Keep it simply as`Path.toFile().deleteOnExit()`  :-)
And if you can use NIO more, thats would be good.

lppedd: @laeubi or we can do just `Files.createTempFile(parent.toPath(), "download", ".tmp").toFile()`

My "problem" with using NIO in Maven stuff is that Maven itself always use the old File API. So every time I question its usage.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
lppedd: Updated with the returned error message (includes the error code).

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
lppedd: The `tycho.p2.transport.ftp.port` property can now change the default FTP port.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
lppedd: Changed.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
lppedd: Couldn't reproduce anymore. So changed to a `FIleNotFoundException`.

lppedd:
I've added a check to avoid socket exceptions escaping at disposal time, e.g., if the connection has been already dropped and we call logout > error.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
```diff
@@ -160,9 +159,8 @@ private FTPClient getClient(final URI uri) throws IOException {
         }
 
         if (!FTPReply.isPositiveCompletion(client.getReplyCode())) {
```
laeubi, line 161: Maybe one can have method (that possibly throws IOException) checkResult(client, message_with_placeholder)

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
```diff
@@ -0,0 +1,217 @@
...
+                // We check if the connection is still active, as it might
+                // have been dropped earlier because of a timeout
+                if (client.isAvailable()) {
+                    client.logout();
+                    client.disconnect();
+                }
+            } catch (final FTPConnectionClosedException e) {
+                // Connection already closed by the host
+            } catch (final IOException e) {
+                logger.error("Error disconnecting from host: " + entry.getKey(), e);
```
laeubi, line 122: I think debug is enough, as there is not much to do anyways...

laeubi:
@lppedd I think this is quite complete now, if you can squash all changes and write one descriptive commit message we can merge this and probably improve later on if issues are found.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
lppedd: I always prefer to see the actual error + stacktrace. Sometimes FTP servers have issues and errors can be of help.
Timeouts are not shown, so other errors might be of relevance.

laeubi: You can always enable -X if it is of any interest, but errors when closing a connection only pollute the usual logs.

lppedd: @laeubi not convinced by that, mostly because debug logging is rarely on, so you'd never know in the first place, plus it's a single stack trace.
What I might do is display a message only as error, and ask for debug to display the full stack trace, sounds reasonable?

laeubi: Not really, when closing fails, and the maven process end, there is literally no need to scare the user, what should happen here? The process will exit and any socket will forcefully be closed anyways, so there is nothing to worry about.

lppedd: @laeubi the socket is closed but as a user I want to know if there has been an issue anyway. From a tool perspective it's important the user is always aware of errors, even if the message is concise it can be expanded with debug.

lppedd: I can downgrade to warning, that's acceptable

laeubi: But what error? From user POV there is no error, the build has succeed and we are just cleaning up resources, so this "problem" will never influence the build outcome and thus must not be shown if not debug is requested.

lppedd: Well imho a connection error should be notified anyway, as warning at least. FTP repos are going to be used mostly internally and the people that manage the build might manage the server too (I do devops and I manage the full stack), and might be interested in being notified of connection issues. Depending on the type of error the client or sever might be tweaked or at least the server log can be inspected. If we show that only in debug mode it's never going to be seen, or very rarely

laeubi: If no one notice a problem, then there is no problem, Tycho is not an FTP-Server-Debugging-Tool.

lppedd:
@laeubi squashed in a single commit and changed the dispose exception logging to debug.

laeubi:
Thanks for your hard work in improving FTP support for Tycho :+1:

Issue #2051, opened 2023-02-05
Rebundled jars should keep their original GAV


ptziegler:
When building a project with `addMavenDescriptor=true`, the pom file is included in the assembled artifact. This file lists all rebundled Maven dependencies using their synthetic GAV, instead of their original GAV.

For example, a dependency to
```
<dependency>
	<groupId>com.squareup.okhttp3</groupId>
	<artifactId>okhttp</artifactId>
	<version>4.10.0</version>
</dependency>
```
is listed as:
```
<dependency>
	<groupId>p2.eclipse.plugin</groupId>
	<artifactId>wrapped.com.squareup.okhttp3.okhttp</artifactId>
	<version>4.10.0</version>
	<type>eclipse-plugin</type>
	<scope>provided</scope>
	<optional>true</optional>
</dependency>
```

If this artifact is used in a pure Maven environment, the dependency resolver computes an incomplete dependency tree and the user has to explicitly add the dependency to OkHttp to their project.

This behavior is caused by the following line in the MavenTargetDefinitionContent class
```
unit = publish(BundlesAction.createBundleDescription(tempFile), tempFile,
        /*
         * do not propagate maven artifacts info for wrapped bundles
         */ null);
```
where Tycho intentionally forgets about the relationship between the original and rebundled jar.

Is this still necessary?

laeubi:
A wrapped bundle can't retain the original dependency information because it is different, but actually the original dependency should remain (and the generated one should be removed), what version of Tycho are you using? If it is 
the current [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) build (4.0.0-SNAPSHOT), then please provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue so it could be fixed.

laeubi:
@ptziegler By the way you might want to help okhttp to add osgi headers in the fisrt place what would make wrapping unnecessary here.

ptziegler:
> If it is the current [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) build (4.0.0-SNAPSHOT), then please provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue so it could be fixed.

Given that the project is still using Java 11, I'm at 2.7.5. However, the same problem also shows up with the latest nightly build. I'll see if I can cook up a simple example project.

[...]

PR #2052, opened 2023-02-05
Create integration tests for issue #2051.

ptziegler:
The first test checks whether the rebundled jars are listed as dependencies in the consumer pom.
The second test checks whether the correct coordinates are used as project dependencies when running dependency:tree to e.g. perform a static dependency analysis.

### DIFF
--- a/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/TychoConstants.java
+++ b/tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/TychoConstants.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2008, 2022 Sonatype Inc. and others.
+ * Copyright (c) 2008, 2023 Sonatype Inc. and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -59,6 +59,14 @@ public interface TychoConstants {
     String PROP_CLASSIFIER = "maven-classifier";
 
     String PROP_REPOSITORY = "maven-repository";
+
+    String PROP_WRAPPED_GROUP_ID = "maven-wrapped-groupId";
+
+    String PROP_WRAPPED_ARTIFACT_ID = "maven-wrapped-artifactId";
+
+    String PROP_WRAPPED_VERSION = "maven-wrapped-version";
+
+    String PROP_WRAPPED_CLASSIFIER = "maven-wrapped-classifier";
 
     String PROP_PGP_KEYS = "pgp.publicKeys";
 
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/resolver/ResolverTests.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/resolver/ResolverTests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2021 Christoph Läubrich and others.
+ * Copyright (c) 2023 Christoph Läubrich and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -16,6 +16,7 @@ import java.util.List;
 
 import org.apache.maven.it.Verifier;
 import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
+import org.junit.Ignore;
 import org.junit.Test;
 
 public class ResolverTests extends AbstractTychoIntegrationTest {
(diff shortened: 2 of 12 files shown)


ptziegler:
Unrelated to this issue: The dependencies are marked as optional in the Manifest, but the `<optional>true</optional>` entry is missing in the consumer pom. I assume that's a bug?

laeubi:
"Optional" has a strange meaning in maven and currently Tycho has no way to carry that information for p2 items, I'm alos not sure if it would be useful, but if you have a strong use-case for that we should use a seperate issue for that.
Just keep in mind that if an item is optional can not always be determined, e.g. if you import different packages from an artifact some can be marked optional while others are mandatory.

tycho-its/src/test/java/org/eclipse/tycho/test/resolver/ResolverTests.java
```diff
@@ -104,4 +104,12 @@ public void testConsiderResolutionWithUsesDirectiveIfVanillaResolutionFails() th
 		verifier.executeGoal("compile");
 		verifier.verifyErrorFreeLog();
 	}
+
+	@Test
+	public void testResolveRebundledMavenDependencies() throws Exception {
+		Verifier verifier = getVerifier("target.maven.wrapAsBundle", true);
+		verifier.executeGoal("dependency:tree");
+		verifier.verifyErrorFreeLog();
+		verifier.verifyTextInLog("com.squareup.okhttp3:okhttp:jar:4.10.0:compile");
```
laeubi, line 113: As explained, `dependency:tree` will most likely never show wrapped bundles as their original ones. but we can have this test with `@Ignore` explaining that this is a missing feature, actually it would be great if Tycho could delay the wrapping part further down in the build and e.g. only perform it when the item is packed into a product, updatesite or used in a test.

ptziegler:
>  I'm alos not sure if it would be useful, but if you have a strong use-case for that we should use a seperate issue for that.

I tend to agree on that... I can't imagine many situations where this would be a problem. And even then, the user can always explicitly ban a dependency via `<exclusion>...</exclusion>`.

As a fix for the consumer-pom, the original GAV is added as a property to the IU. In a new field, in order to preserve the distinction between rebundled and original artifact.

tycho-its/src/test/java/org/eclipse/tycho/test/resolver/ResolverTests.java
ptziegler: That sounds like a good solution. The CycloneDX plugin, for example, requires a valid list of dependencies to generate a bill-of-materials. Currently, those only work in a limited capacity.

tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/UpdateConsumerPomMojo.java
```diff
@@ -299,6 +303,22 @@ private String mapTychoPackagingTypeToMaven(String packaging) {
 
 	private boolean handleSystemScopeDependency(Dependency dep) {
 		if (dep.getGroupId().startsWith(TychoConstants.P2_GROUPID_PREFIX) || isEmbeddedJar(dep)) {
+			Dependency originalDependency = null;
+			if (dep instanceof ArtifactDescriptor) {
```
laeubi, line 307: This should be implemented as a `ArtifactCoordinateResolver`.

tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/TychoConstants.java
```diff
@@ -64,6 +64,8 @@ public interface TychoConstants {
 
     String PROP_PGP_SIGNATURES = "pgp.signatures";
 
+    String PROP_WRAPPED_ORIGINAL_GAV = "wrapped.originalGav";
```
laeubi, line 67: we already have some "maven-groupId etc. properties (see above), so I think we should use the maven prefix here as well, maybe `maven-gav-hint`, or even just duplicate all of them using

- `maven-wrapped-groupId`
- `maven-wrapped-artifactId`
- ...

That way it is much more flexible /extensible and probably one can even reuse the existing Properties advice and just pass the prefix (`maven-` or `maven-wrapped-`)

laeubi:
I think for resolving, you can simply enhance [P2ArtifactCoordinateResolver](https://github.com/eclipse-tycho/tycho/blob/master/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/reverseresolve/P2ArtifactCoordinateResolver.java).

laeubi:
ECA validation is currently broken: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2636

ptziegler:
> I think for resolving, you can simply enhance [P2ArtifactCoordinateResolver](https://github.com/eclipse-tycho/tycho/blob/master/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/reverseresolve/P2ArtifactCoordinateResolver.java).

Good point. Though I just want to mention that this then means the original GAV is only restored when `mapP2Dependencies` is set. Which I suppose is perfectly valid...

tycho-bundles/org.eclipse.tycho.embedder.shared/src/main/java/org/eclipse/tycho/TychoConstants.java
ptziegler: Splitting the constants into the prefix (`maven-` and `maven-wrapped`) and property (`groupId`, ...) makes them look very messy.
I think it's cleaner to simply pass a map of properties. This way I can still reuse the existing class.

laeubi:
> > I think for resolving, you can simply enhance [P2ArtifactCoordinateResolver](https://github.com/eclipse-tycho/tycho/blob/master/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/reverseresolve/P2ArtifactCoordinateResolver.java).
> 
> Good point. Though I just want to mention that this then means the original GAV is only restored when `mapP2Dependencies` is set. Which I suppose is perfectly valid...

That's how it is supposed to work at least :-)

laeubi (approved):
Thanks for the test-case and fix!

PR #2055, opened 2023-02-06
Improve FTP reconnection logic

### LINKED ISSUE #2054: FTP connection should be kept alive/re-initialized during build process
If a lot of time might pass by between FTP calls, the FTP connection might be dropped abruptly.  
A good enhancement might be using a background thread sending `NOOP` commands to keep the connection alive, or re-connect automatically when a `FTPConnectionClosedException` is thrown.

lppedd:
Very small change that makes reconnection actually work when needed.  
The client, besides verifying if the socket connection is open, tries to send a NOOP command.
Both checks must succeed, otherwise a new connection is created.

Closes #2054

Note: @laeubi in debug I see two instances of `FtpTransportProtocolHandler` gets created, any idea how's that possible?

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
@@ -28,7 +28,7 @@ public class FtpTransportProtocolHandler implements TransportProtocolHandler, Di
      * FTP clients with an open connection.
      * Every client should be disposed at the end of the build.
      */
-    private final Map<String, FTPClient> clients = new ConcurrentHashMap<>(8);
+    private static final Map<String, FTPClient> CLIENTS = new ConcurrentHashMap<>(8);
 
     @Requirement
     private Logger logger;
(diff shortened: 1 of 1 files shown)


laeubi:
> Note: @laeubi in debug I see two instances of FtpTransportProtocolHandler gets created, any idea how's that possible?

Maven uses different classlaoders (realms) so components are maybe created multiple times depending on the realm they live in ...

lppedd:
@laeubi oh, that sounds ok. Tho, should the `clients` map be static then? 
I don't like static stuff but if it avoids created unnecessary connections...

```java
/**
* FTP clients with an open connection.
* Every client should be disposed at the end of the build.
*/
private final Map<String, FTPClient> clients = new ConcurrentHashMap<>(8);
```

laeubi:
> @laeubi oh, that sounds ok. Tho, should the `clients` map be static then? I don't like static stuff but if it avoids creating unnecessary connections...
> 
> ```java
> /**
> * FTP clients with an open connection.
> * Every client should be disposed at the end of the build.
> */
> private final Map<String, FTPClient> clients = new ConcurrentHashMap<>(8);
> ```

That won't work, as different classlaoders will result in different classes that have different statics, I think we have to live with that for now.

laeubi:
Well if it helps anything, a static might be fine here.

laeubi:
ECA checks are currently broken ... see https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2636

PR #2057, opened 2023-02-06
Allow configuring ClassLoader order in OsgiSurefireBooter

joeshannon:
The composite loader searches the three loaders in order. If another "copy" of the JUnit Platform classes is available (and visible to the test class loader) it can take priority over the classes in the Surefire booter plugin causing a conflict.

### DIFF
--- a/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractEclipseTestMojo.java
+++ b/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractEclipseTestMojo.java
@@ -528,6 +528,27 @@ public abstract class AbstractEclipseTestMojo extends AbstractTestMojo {
     @Parameter
     private int reactorConcurrencyLevel;
 
+    public enum ClassLoaderOrder {
+        booterFirst, testProbeFirst
+    }
+
+    /**
+     * The test runtime is configured with a composite class loader. This defines the order in which
+     * the loaders are searched, and it may need to be configured depending on the resolved
+     * classpath of the project.
+     * <p>
+     * Available values are:
+     * <ul>
+     * <li><code>booterFirst</code> - the loader of the surefire booter (including bundled test
+     * platform) is searched first</li>
+     * <li><code>testProbeFirst</code> - the loader of the test class's plugin is searched
+     * first</li>
+     * </ul>
+     * Defaults to <code>booterFirst</code>.
+     */
+    @Parameter(defaultValue = "booterFirst")
+    private ClassLoaderOrder classLoaderOrder;
+
     @Component
     private ProviderHelper providerHelper;
 
@@ -838,6 +859,7 @@ public abstract class AbstractEclipseTestMojo extends AbstractTestMojo {
         wrapper.setProperty("rerunFailingTestsCount", String.valueOf(rerunFailingTestsCount));
         wrapper.setProperty("printBundles", String.valueOf(printBundles));
         wrapper.setProperty("printWires", String.valueOf(printWires));
+        wrapper.setProperty("classLoaderOrder", classLoaderOrder.toString());
         Properties mergedProviderProperties = getMergedProviderProperties();
         mergedProviderProperties.putAll(provider.getProviderSpecificProperties());
         Map<String, String> providerPropertiesAsMap = propertiesAsMap(mergedProviderProperties);
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,12 @@ This page describes the noteworthy improvements provided by each release of Ecli
 
 ## 4.0.0 (under development)
 
+### Class loading changes for Eclipse based tests
+
+Due to reported class loading clashes, the ordering of class loading has been modified in Eclipse based tests.
+The previous loading can be restored by a new `classLoaderOrder` parameter.
+This applies to `tycho-surefire-plugin:test` and `tycho-surefire-plugin:plugin-test`.
+
 ### new bnd-test mojo
 
 Tycho now has a new mojo `tycho-surefire-plugin:bnd-test` to easily execute tests using the [bnd-testing](https://bnd.bndtools.org/chapters/310-testing.html) framework.
--- a/tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/java/org/eclipse/tycho/surefire/osgibooter/OsgiSurefireBooter.java
+++ b/tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/java/org/eclipse/tycho/surefire/osgibooter/OsgiSurefireBooter.java
@@ -138,8 +138,9 @@ public class OsgiSurefireBooter {
                 new PrintStreamLogger(System.out));
         // API indicates we should use testClassLoader below but surefire also tries
         // to load surefire classes using this classloader
-        RunResult result = ProviderFactory.invokeProvider(null, createCombinedClassLoader(testPlugin), reporterFactory,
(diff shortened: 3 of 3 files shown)


joeshannon:
This relates to #1773

laeubi (changes requested):
Please use a new property (`classloaderOrder`) with the values `testProbeFirst` (old behavior) and `booterFirst` (changed behavior)), and then have a configuration option in the mojo for this.

lppedd:
@laeubi a clarification: should the new parameter have as default the new or old behavior?

laeubi:
I would go for the new behaviour as default, if it causes too much issues switch back to the old

joeshannon:
Sorry was a bit busy yesterday to look at this.

The good news is that I think I have found the cause of the issue (one of the causes, there may be others). As mentioned in https://github.com/eclipse-tycho/tycho/pull/1773#issuecomment-1409027030 we are now seeing multiple versions of the jupiter classes in the test runtime.

The switch to use the Maven central JUnit jars looks to be the cause, if we compare the two Manifests: `junit-jupiter-api_5.9.1.jar` (2022-09 tp) and `org.junit.jupiter.api_5.8.1.v20211018-1956.jar` (2022-03 tp).

The Maven central junit-jupiter-api contains the following entry:
```
Require-Capability: org.junit.platform.engine;filter:="(&(org.junit.platform.engine=junit-jupiter)(version>=5.9.1)(!(version>=6)))";effective:=active,osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
````
which is not present in the `org.junit.jupiter.api` bundle.

So it looks like this is causing all the additional jupiter jars to be added as dependences to the Maven project when Tycho resolves projects at the start of the build.

Since these packages are all already bundled into the Tycho surefire booter/fragments, this causes the conflict.

The classloader ordering switch works around this but it feels like a more robust fix would be to prevent this situation by excluding these plugins from the test runtime to prevent including these libraries twice.

I haven't looked into the history but presumably there is some OSGi/class loading reason why the booter requires JUnit attached as custom Tycho fragments.

**Comment by laeubi** [comment]
the "classic" plugin-test is an attempt to simulate the behavior of PDE Plugin Tests and "merge" the world of maven+osgi+surefire what works up to a certain degree.

If I understand correctly you are using it with `eclipse-plugin` packaging in which case I can recommend to try out the new [bnd-test-mojo](https://tycho.eclipseprojects.io/doc/master/TestingBundles.html#bnd-testing) that uses a different approach reusing the junit-launcher framework *inside* an OSGi framework, what should generally work better regarding isolation of the classloaders, but currently has no special support for "UI Harness" Tests (this does not mean it can't work with the UIHarness, it is just that no one has ever tried that), but if you do not depend on this it should be safer to use.

**Comment by lppedd** [comment]
Maybe it's worth going for the "easy" fix now and continue the investigation later on if necessary?

**Comment by joeshannon** [comment]
Yes it would probably make sense to do that. Just looking at it now.
Is it possible to add the parameter as an enum to prevent custom validation/checking of a string? It looks like that has been done for some other parameters in the project.

If so, it would also be good to place in a location both `AbstractEclipseTestMojo` and `OsgiSurefireBooter` can access to prevent duplication.

**Comment by lppedd** [comment]
> Is it possible to add the parameter as an enum to prevent custom validation/checking of a string

If you mean as a Mojo parameter, then yes, Maven has an [EnumConverter](https://github.com/eclipse/sisu.plexus/blob/master/org.eclipse.sisu.plexus/src/org/codehaus/plexus/component/configurator/converters/basic/EnumConverter.java) for this use case.

> OsgiSurefireBooter

What's the call stack to here? I was looking at how it's used and wondering how we get there from `AbstractEclipseTestMojo`.

**Comment by laeubi** [comment]
Values are passed by the `surefire.properties` to the external process.

File: tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractEclipseTestMojo.java [hunk 1]
```diff
@@ -528,6 +528,18 @@ public abstract class AbstractEclipseTestMojo extends AbstractTestMojo {
     @Parameter
     private int reactorConcurrencyLevel;
 
+    public enum ClassLoaderOrder {
```
joeshannon, line 531: It would be nice if this enum didn't have to be duplicated.

lppedd: @joeshannon I would just use a string in `OsgiSurefireBooter`. Imo, we can assume the value is valid once it gets there.

laeubi: Yes no need for an enum on the booter level, as this is quite low-level.

lppedd: Also, you already throw an exception if it's not in the valid range of values.

tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractEclipseTestMojo.java
```diff
@@ -528,6 +528,18 @@ public abstract class AbstractEclipseTestMojo extends AbstractTestMojo {
...
+        booterFirst, testProbeFirst
+    }
+
+    /**
+     * The test runtime is configured with a composite class loader. This defines the order in which
+     * the loaders are searched. This may need to be configured depending on the classpath of the
+     * project. Available values "booterFirst" and "testProbeFirst"
+     */
+    @Parameter(defaultValue = "booterFirst")
```
lppedd, line 540: Always good to be tell what's the default value in the docs.

```suggestion
    /**
     * The test runtime is configured with a composite class loader.
     * This defines the order in which the loaders are searched, and it may need to be configured depending on the classpath of the project.
     * <p>
     * Available values are <code>booterFirst</code>, <code>testProbeFirst</code>.
     * Defaults to <code>booterFirst</code>.
     */
    @Parameter(defaultValue = "booterFirst")
```

tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractEclipseTestMojo.java
joeshannon: Yes that makes sense, have updated.

lppedd:
That's perfect @joeshannon, thanks!  
Imo this can become a mergeable PR now.

tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractEclipseTestMojo.java
```diff
@@ -528,6 +528,27 @@ public abstract class AbstractEclipseTestMojo extends AbstractTestMojo {
...
+     * <ul>
+     * <li><code>booterFirst</code> - the loader of the surefire booter (including bundled test
+     * platform) is searched first</li>
+     * <li><code>testProbeFirst</code> - the loader of the test class's plugin is searched
+     * first</li>
+     * </ul>
+     * Defaults to <code>booterFirst</code>.
+     */
+    @Parameter(defaultValue = "booterFirst")
+    private ClassLoaderOrder classLoaderOrder;
```
laeubi, line 550: Can you please ad a note to the release_notes.md about the new option?

lppedd:
Should I bring in the latest changes in the reproduced PR before merging?

PR #2061, opened 2023-02-07
Add a check that builds (including tests) the 'build-individual-bundles'

laeubi:


laeubi:
Check is failing by intention so merge this now...

PR #2084, opened 2023-02-10
Investigate Maven 3.9.0

laeubi:


akurtakov:
What's the issue with null annotations?

laeubi:
It seems maven no longer use them, and I can't see any wider use in Tycho nor we loosing anything so it seems better to just remove them see also https://stackoverflow.com/questions/4963300/which-notnull-java-annotation-should-i-use it seems there is no common standard to use anyways. So I think `Objects.requireNonNull()` is more appropriate (and if I understand @stephan-herrmann JDT understand this if we ever want to enable null analysis).

akurtakov:
I can only hope that jakarta.annotation is gonna absorb all these.

laeubi:
> I can only hope that jakarta.annotation is gonna absorb all these.

Probably but still then theses four places won't change the world, so if we decide it should be used we really need to use it everywhere...

akurtakov:
Yeah, my question originated by the fact that maven 3.9.0 ships javax.annotation-api-1.3.2.jar in libs thus I wondered why this is needed.

laeubi:
I have not checked in details why this is no longer a (transitive) dependency, so maybe it would equally work to add an explicit one (they maybe changed it to provided or optional or ...), but unless we have any benefit it might be better to just remove that at all as it seems not used elsewhere.

lppedd:
Talking about nullability annotations, I use https://github.com/osundblad/intellij-annotations-instrumenter-maven-plugin to avoid manually checking for preconditions.

I'm an avid user of `@NotNull/@Nullable` and other annotations, especially because the IDE (currently only IntelliJ) can use them to infer inspections data.

laeubi:
> especially because the IDE (**currently only IntelliJ**) can use them to infer inspections data.

And the highlighted part makes then essentially useless.. I really don't get why everyone needs to invent its own "nullable" annotations here, that's a complete mess. Anyways this PR is not really about the annotations, the benefit of those for Tycho are currently just non existent.

tycho-core/src/main/java/org/eclipse/tycho/core/ee/EEVersion.java
```diff
@@ -15,8 +15,6 @@
 
 import java.util.Objects;
 
-import javax.annotation.Nonnull;
```
stephan-herrmann, line 18: No longer using `javax.annotation.Nonnull` is a good move, but removal looks like a step backwards, no?

stephan-herrmann:
> I can only hope that jakarta.annotation is gonna absorb all these.

They will still have a very long road ahead, before they can "absorb" existing annotations:
* [specification](https://github.com/jakartaee/common-annotations-api/blob/master/spec/src/main/asciidoc/spec.adoc#jakartaannotationnullable) is very far from sufficient as a standard
* apparently not intended as TYPE_USE annotations, which would be the minimum requirement these days
* to the best of my knowledge not coordinated with prior art in checkers-framework, JDT, nor jspecify.

I'm open to discuss with them. If that would *finally* become the standard (which long ago JSR-305 had promised without delivering) - I'm all for it, but looking from today I don't see that.

tycho-core/src/main/java/org/eclipse/tycho/core/ee/EEVersion.java
laeubi: Given that (I haven counted them) there are many hundreds of file not suing it compared to only these few its jsut a baby-step :-)

stephan-herrmann:
> I'm an avid user of `@NotNull/@Nullable` and other annotations, especially because the IDE (currently only IntelliJ) can use them to infer inspections data.

If there is anything in this regard which IntelliJ supports, but not JDT, please file an issue against JDT/Core. Last time I looked they were still stuck on the inferior _declaration annotations_ (as opposed to TYPE_USE annotations).

laeubi:
I think most of the failures at the moment are cause because https://github.com/takari/polyglot-maven does not support Maven 3.9.0, so if anyone likes to step up and propose a fix?
I'll also try to convince the maven-devs to probably "adopt" the code so it became part of maven-core, but this can be a longer process...

laeubi:
Will require:
- https://github.com/takari/polyglot-maven/pull/256

laeubi:
Alright, Tycho can now build with Maven 3.9.0, all integration tests run and platform checks are green.

laeubi:
Adjustments have men merged to master now, and we can use the dependabot branch again as-is

PR #2099, opened 2023-02-15
Make maven-compat test dependency in tycho-core

akurtakov:
It's used only in tests so this is first step for removing it from dependency tree.


laeubi:
@akurtakov "bundleLocation not found" is a long outstanding issue with some tests (probably because of sync issues in the maven3.8. resolver) so you can ignore this.

akurtakov:
OK, I've just made naive effort to reduce things that could potentially go wrong aka less surprises . Will wait for the build and push after that.

PR #2119, opened 2023-02-16
Switch maven-compat to maven-core in tycho-source-plugin

akurtakov:


akurtakov:
This was the last compat runtime dependency I have found.

laeubi:
Thanks! Now I think we need only get rid of the old MavenResolver and use Aether instead directly I'll try to take a look the next days...

akurtakov:
What about jsr330 migration?

laeubi:
> What about jsr330 migration?

This would strictly require Maven 3.9.x (because 3.8.x uses old injector that do not support Java 17) and beside that it is "more modern" I think we don't have any benefit (yet) from using it.

PR #2133, opened 2023-02-17
Add support for (semi-) automated version bumps

laeubi:
Currently work in progress but to see if restructuring breaks anything...


lppedd:
The only thing I don't really like is the failure treated as success.

laeubi:
> The only thing I don't really like is the failure treated as success.

What do you mean? Due to technical reasons the build still fails, it just automates what users previously would apply manually after such failure then one has to run the build again.

PR #2179, opened 2023-03-01
[backport] Add support for pomless bnd-workspace build

laeubi:
BND Workspace is similar to the eclipse-structured layout as actually all configuration files contain enough information to perform the build without any pom configuration.

This changes adds support for the BND Workspace layout, a demo project and a help-page to sues this, also mixed layout builds are possible but not part of this change and will be subject to future improvements.


laeubi:
@akurtakov I reverted the mojo version change as well here and now the check seems fine again.

PR #2197, opened 2023-03-04
Add option to configure eager resolve in the TargetPlatformConfiguration

laeubi:
Once there was the idea to switch between a "new" and "old" resolver style but this did not worked well as one only can switch this globally with a system property and it is actually confusing as not the resolver itself is different by the time when resolving happens.

This is now an attempt to unify this under a new (configurable) way that works by specify in the target platform what mode is actually desired for a given configuration.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -226,15 +226,27 @@ This can be disabled with the following configuration in the pom:
 
 ### Migration guide 3.x > 4.x
 
-### New Tycho Resolver
+### New delayed target platform resolving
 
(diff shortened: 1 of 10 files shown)


laeubi:
FYI @akurtakov @mickaelistria I now reworked that feature and it builds for both the full as partial platform build.

beside that, it is now configurable through the target platform, so can be switched on/off in a profile, or even for single projects!

So I think this is ready to be merged and we can give it a try for 2023-09, I'll just wait for the final GA announcement as I assume then no builds for the old stream would happen anymore.

PR #2218, opened 2023-03-10
[Setup] Use Java-17 JRE and remove unnecessary auto-build workaround

HannesWell:


HannesWell:
Just submit this since the verification build does not check the Oomph setup any way.

Issue #2214, opened 2023-03-09
Cannot delete corrupted file


lppedd:
The following line doesn't seem to do anything because the file is allocated by the process somewhere.

https://github.com/eclipse-tycho/tycho/blob/343d48d36d783b86eda0233389dda7b805398a0c/tycho-core/src/main/java/org/eclipse/tycho/core/maven/MavenDependencyInjector.java#L264

So the file is not actually downloaded again, as in a deeper call `location.exists()` will return true.

laeubi:
@lppedd is this still an issue?

lppedd:
@laeubi yes, it's still an issue unfortunately.

laeubi:
we plan a 3.0.5 bufix release so if you want to provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue and probably a fix it might can be included.

PR #2220, opened 2023-03-11
Synchronize access to FTP clients

### LINKED ISSUE #2213: Make FtpTransportProtocolHandler thread safe
`FTPClient` is not thread safe per se, so when trying to run a Maven build with `-T *`, FTP commands will not work properly.

@laeubi I'll take care of this soon.

lppedd:
I've used locking primitives only here, but I'm open to rewriting in other ways if it's required.

Rationale: Apache's `FTPClient` is not thread-safe, and that means multiple threads could access an instance and issue multiple FTP commands concurrently. This won't work as one `FTPClient` is associated to one connection, and commands must be issued in sequence. This change synchronizes on the cached `FTPClient` instances, allowing different open connections to retrieve data concurrently.

Closes #2213

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FtpTransportProtocolHandler.java
@@ -16,6 +16,8 @@ import java.time.Duration;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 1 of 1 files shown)


laeubi:
@lppedd I have merged this now, probabbly you can think about having a thread local that opens one connection per Thread (maybe limiting to some kind of maximum per server)or some kind of pool of connections, just an idea if you like to play around with that.

lppedd:
@laeubi I'll consider this too, thanks! As of now I'm trying to understand #2214, not with good results unfortunately.

PR #2260, opened 2023-03-19
Propagation of suggested version and add more alias for baseline-compare

laeubi:


HannesWell:
I think maintaining those extensions in the sources does not scale Well. There should also be a configuration Option so a User can adjust this.

Additionaly we could also ask git If a file is a Text -file. I gits knowledge in this regard is quite advanced and one can customize it in the .gitattribute file

laeubi:
Users are not mandated to use git ... also the problem occurs at the first place with these "advanced" git feature that magically modifies "text" files ... so in general it would be best to avoid that all together and one won't get any problems.

Tycho should just support a set of usual items we have checked that they are sane, at best emf would offer some kind of comparator that can compare the files even semantically, not sure if such thing exits.

Having configuration option is possible but would require a lot of effort to make this really generic as we are not only doing text-compares, so if you can configure one, you most probably want to configure others as well, then this is extensible so configuration might also needs to be extensible (what maven currently do not offer).

PR #2271, opened 2023-03-21
Enhance comparators and tests

HannesWell:
The tests, at least locally succeeded and I hope they do now as well in the CI.

Furthermore they now require much less resources (which avoids copying) and makes the changes more obvious.
This also simplifies adding more test cases.


laeubi:
This will mean that you can't execute the test anymore standalone what is crucial for debugging / investigation I think. Also setups that *delete* a file seem no longer possible.

HannesWell:
> This will mean that you can't execute the test anymore standalone what is crucial for debugging / investigation I think

OK, that's a point. But making it work in the CI is as crucial as running it standalone. So I'll change it that both works.

> Also setups that _delete_ a file seem no longer possible.

How do you mean that? In the CI it isn't possible anymore to delete a file?

laeubi:
> But making it work in the CI is as crucial as running it standalone. So I'll change it that both works.

As this changes quite a lot things I'm not sure I can spot what makes it run on the CI? So probably it would be good to make a first minimal change to enable it on CI and make enhancements a separate PR.

> How do you mean that? In the CI it isn't possible anymore to delete a file?

You currently copy the "base" bundle for each test, but the intend was to support a wide range of baseline problems. One such problem might be that a file is deleted that previously was in the baseline, if you copy all content from the base this won't work, so even though its a bit more work to maintain things I think it is much more flexible like its currently implemented as one can control in details what is included and what not.

HannesWell:
> > But making it work in the CI is as crucial as running it standalone. So I'll change it that both works.
> 
> As this changes quite a lot things I'm not sure I can spot what makes it run on the CI? So probably it would be good to make a first minimal change to enable it on CI and make enhancements a separate PR.

I'm pretty sure it is the before-test method that now builds the baseline in advance and moves it to a location that is not deleted during the copy-operation (which prunes its target).




> > How do you mean that? In the CI it isn't possible anymore to delete a file?
> 
> You currently copy the "base" bundle for each test, but the intend was to support a wide range of baseline problems. One such problem might be that a file is deleted that previously was in the baseline, if you copy all content from the base this won't work, so even though its a bit more work to maintain things I think it is much more flexible like its currently implemented as one can control in details what is included and what not.

That would still have been possible. In the consumer passed to the buildBaselineProject method one could modify the copied 'reactor' project in any way intended, of course one could also delete a file.

Any way. I now reverted that part so that the test-cases can be executed as the are standalone and in the CI. :)

tycho-baseline-plugin/src/main/java/org/eclipse/tycho/baseline/FeatureBaselineComparator.java
```diff
@@ -388,37 +384,28 @@ private static boolean propertyEquals(String baselineValue, String projectValue)
...
 			char c1 = s1.charAt(i);
 			char c2 = s2.charAt(i);
 			if (c1 != c2) {
-				break;
+				return i;
 			}
 		}
-		return i;
-
+		return -1;
```
laeubi, line 399: I don't think this is correct, if a > b but there is no difference in the prefix it must then return the length.

tycho-its/src/test/java/org/eclipse/tycho/test/apitools/ApiToolsTest.java
```diff
@@ -47,16 +47,15 @@ public void testVerify() throws Exception {
...
-		} catch (VerificationException e) {
-			verifier.verifyTextInLog("1 API ERRORS");
-			verifier.verifyTextInLog("0 API warnings");
-			verifier.verifyTextInLog("The type bundle.InterfaceA has been removed from api-bundle");
-			verifier.verifyTextInLog("The type bundle.ClassA has been removed from api-bundle");
-			// TODO: check with api-filter
-			// TODO: check with second plugin with BREE?
-		}
+
+		assertThrows(VerificationException.class, () -> verifier.executeGoals(List.of("clean", "verify")));
```
laeubi, line 51: The assert should contain a message why this is expected.

laeubi (commented):
Just two minor comments.

laeubi:
> That would still have been possible. In the consumer passed to the buildBaselineProject method one could modify the copied 'reactor' project in any way intended, of course one could also delete a file.

Okay then lets do this in a second PR after this is merged!

tycho-baseline-plugin/src/main/java/org/eclipse/tycho/baseline/FeatureBaselineComparator.java
HannesWell: That's right. That was also the behavior before.
Changed it, thanks for pointing out.

tycho-its/src/test/java/org/eclipse/tycho/test/apitools/ApiToolsTest.java
HannesWell: Done.

HannesWell:
> > That would still have been possible. In the consumer passed to the buildBaselineProject method one could modify the copied 'reactor' project in any way intended, of course one could also delete a file.
> 
> Okay then lets do this in a second PR after this is merged!

OK, great. Since the test-projects then will be deleted I also reverted the formatting unifications for this PR to keep it smaller.

PR #2281, opened 2023-03-24
Add integration test for tycho-apitools-plugin:verify mojo

HannesWell:
This adds an integration test for the `tycho-apitools-plugin:verify` mojo.

On my Windows computer this currently fails saying that the baseline must not be null.
Furthermore I currently have no good idea how encode the repository URL as a relative path. At the moment I use `file:///${maven.multiModuleProjectDirectory}/repositories/api-tools`, but that fails on Windows since it mixes forward and backward slashes. But it should succeed on Unix.
Besides that the test-case needs some assertions.

FYI @laeubi.


laeubi:
> Furthermore I currently have no good idea how encode the repository URL as a relative path

The usual approach in the Itest is to have a variable (e.g. `${baselineRepo}`) in the pom, then set this in the verifier via 

```
File repo = new File(verifier.getBaseDir(), "..some path...");
verifier.setCliArgument("-DbaselineRepo"+=repo.getAbsoloutePath())
```

laeubi:
@HannesWell this seem to break the generate test-case with:

> .api_description not found

is a change in this area expected / intended? I currently can't see why this would interact here....

HannesWell:
> @HannesWell this seem to break the generate test-case with:
> 
> > .api_description not found
> 
> is a change in this area expected / intended? I currently can't see why this would interact here....

No that was accidental. Thank you for fixing this in https://github.com/eclipse-tycho/tycho/pull/2282.

Btw. if you want you can also push extra commits directly to my branch do avoid the need for an extra PR. :)

PR #2282, opened 2023-03-24
Pr 2281 api tools

laeubi:
@HannesWell I have adjusted your test so it actually runs on linux (and should as well on windows), you might want to integrate this in your pr, the api check still fails, I'll investigate a bit more on this and probably push it to this branch.


laeubi:
I think I sorted this out now and the test are passing now!
Let me know if I should squash merge this, merge as two commits or you like to just integrate this into your pr.

HannesWell:
Thank you! I just integrated your commit into my PR https://github.com/eclipse-tycho/tycho/pull/2281.

PR #2300, opened 2023-03-25
Generate modified reactor projects on the fly in BaselineMojoTest

HannesWell:
As discussed in https://github.com/eclipse-tycho/tycho/pull/2271, this changes the BaselineMojoTest to generate the 'reactor'-projects for each case on the fly.


tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
```diff
@@ -139,4 +196,24 @@ private Verifier getBaselineProject(String project) throws Exception {
...
+		replaceStringInFile(projectRoot.resolve("META-INF/MANIFEST.MF"), //
+				"Bundle-Version: " + from, "Bundle-Version: " + to);
+		UnaryOperator<String> mavenVersion = v -> """
+				\t<artifactId>my.api.bundle</artifactId>
+				\t<version>""" + v + "</version>";
+		replaceStringInFile(projectRoot.resolve("pom.xml"), mavenVersion.apply(from), mavenVersion.apply(to));
+	}
+
+	private static void replaceStringInFile(Path file, String target, String replacement) throws IOException {
+		if (!"\n".equals(System.lineSeparator())) { // Text-Blocks only use \n
```
laeubi, line 210: I don't think this is a good idea, the method should just do the replacement and the caller should take care of do the replacement.

laeubi:
@HannesWell can you probably finish this as I like to add some tests here.

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
```diff
@@ -139,4 +196,24 @@ private Verifier getBaselineProject(String project) throws Exception {
 		verifier.addCliOption(project + "/pom.xml");
 		return verifier;
 	}
+
+	private static void bumpPluginProjectVersion(Path projectRoot, String from, String to) throws IOException {
```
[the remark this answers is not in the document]
HannesWell: Applied your suggestion, thanks for that. The ModelWriter and the Manifest class format the resulting files differently than they are before.
But I think that is also a good opportunity to test the specialized comparators.

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
HannesWell: The problem is that TextBlocks always use `\n` to separate newlines. Therefore on Windows they cannot match the content of the checkout files, which are usually checkedout using `\r\n` and at the same time, the replacement should also use \r\n on Windows. This ensures that the content of the generated files is the same as if they were all committed to the git repo.

I split this method into two so that we can have a 'pure' version and one that does this nasty replacement, but avoids code repetition in the callers.

laeubi: Then textblocks are probably not the right choice here ;-)
It defeats a bit their purpose if we need to handle them special afterwards and one should then avoid them in this context.

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
```diff
@@ -139,4 +206,42 @@ private Verifier getBaselineProject(String project) throws Exception {
...
+	private static void setPomArtifactVersion(Path projectRoot, String version) throws IOException {
+		File pom = projectRoot.resolve("pom.xml").toFile();
+		Model model = new DefaultModelReader().read(pom, null);
+		model.setVersion(version);
+		new DefaultModelWriter().write(pom, null, model);
+	}
+
+	private static final Attributes.Name BUNDLE_VERSION = new Attributes.Name(Constants.BUNDLE_VERSION);
+
+	private static void setManifestBundleVersion(Path projectRoot, String version) throws IOException {
```
laeubi, line 219: Same here, enhancing this to other use case wil most probably require setting any manifest property.

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
```diff
@@ -74,23 +120,36 @@ public void testAddMethod() throws Exception {
...
-	 * @throws Exception
 	 */
 	@Test
-	public void testAddHeader() throws Exception {
+	public void testAddHeader() throws Throwable {
+
 		// if version is not bumped this should fail
-		Verifier verifier = buildBaselineProject("add-header", true);
+		Verifier verifier = buildBaselineProject(true, r -> {
+			replaceStringInFileWithPlatformsLineSeparator(r.resolve("META-INF/MANIFEST.MF"), "Bundle-Version: 1.0.0",
```
laeubi, line 148: This should not use string replace, see comment about generic Manifest modification method.

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
```diff
@@ -55,17 +75,43 @@ public void testUnchangedApi() throws Exception {
...
+	public void testAddMethod() throws Throwable {
+
 		// test adding a default method to the "public" interface
-		Verifier verifier;
-		verifier = buildBaselineProject("add-method", true);
-		verifyBaselineProblem(verifier, "ADDED", "METHOD", "concat(java.lang.String,java.lang.String)", "1.0.0",
-				"1.1.0");
+		Verifier verifier1 = buildBaselineProject(true, r -> {
+			Path aClass = r.resolve(Path.of("src", "my", "api", "bundle", "MyApiInterface.java"));
+			replaceStringInFileWithPlatformsLineSeparator(aClass, """
```
laeubi, line 83: This here indicates one can use the java AST outside of eclipse:
https://stackoverflow.com/a/20918431

this would avoid using string replace operations at all.

laeubi (changes requested):

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
```diff
@@ -139,4 +206,42 @@ private Verifier getBaselineProject(String project) throws Exception {
 		verifier.addCliOption(project + "/pom.xml");
 		return verifier;
 	}
+
+	private static void setPomArtifactVersion(Path projectRoot, String version) throws IOException {
```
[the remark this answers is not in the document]
HannesWell: Sounds good. Done.

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
HannesWell: Done as well.

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
HannesWell: That's right. 👍🏽 Done.

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
HannesWell: Looks promising, I started to do that, but running out of time. Will complete it later today.
Then the string replaces should not be necessary at all. 👍🏽

laeubi: I think we are not in hurry but this can greatly simplify test creation and make it more "semantic", so I think its worth to be patient :-)

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
```diff
@@ -55,17 +91,39 @@ public void testUnchangedApi() throws Exception {
...
+			modifyJavaSourceFile(aClass, cu -> {
+				TypeDeclaration apiInterface = (TypeDeclaration) cu.types().get(0);
+				addMethod(apiInterface, "concat", m -> {
+					setSimpleReturnType(m, "String");
+					addSimpleTypeParameter(m, "String", "a");
+					addSimpleTypeParameter(m, "String", "b");
+				});
+			});
+		});
+		verifyBaselineProblem(verifier1, "ADDED", "METHOD", "concat(java.lang.String,java.lang.String)", "1.0.0",
```
HannesWell, line 108: I'm unsure why in the meantime (in a previous state of this PR) the baseline plugin suggested 2.0.0 instead of 1.1.0. 
But since this used to be 1.1.0 I think it is fine to leave that unchanged.

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
HannesWell: Done.

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
HannesWell: Is completed now. ;)

HannesWell:
For me this is ready. @laeubi any more remarks?

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
```diff
@@ -55,17 +91,48 @@ public void testUnchangedApi() throws Exception {
...
+				TypeDeclaration internalInterface = (TypeDeclaration) cu.types().get(0);
+				addNoArgsVoidMethod(internalInterface, "newMethodButItIsInternal");
+			});
+
+			modifyProjectManifest(r, Map.of(Constants.BUNDLE_VERSION, "1.0.1"));
+			modifyProjectPom(r, p -> p.setVersion("1.0.1"));
+		};
+
+		Verifier verifier2 = buildBaselineProject(true, addMethodInternal);
+		verifyBaselineProblem(verifier2, "ADDED", "METHOD", "newMethodButItIsInternal()", "1.0.1", "2.0.0");
```
laeubi, line 133: this should be then as above example, not sure why this was split into two separate methods, the intend is that we test the extension mode, so it should run once without extension (then an baseline problem must be reported) and then with extension (then no baseline problem should be reported because it is an internal method)

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
```diff
@@ -139,4 +224,78 @@ private Verifier getBaselineProject(String project) throws Exception {
...
 	}
+
+	private static void modifyProjectPom(Path projectRoot, Consumer<Model> pomModifier) throws IOException {
+		File pomFile = projectRoot.resolve("pom.xml").toFile();
+		Model model = new DefaultModelReader().read(pomFile, null);
+		pomModifier.accept(model);
+		new DefaultModelWriter().write(pomFile, null, model);
+	}
+
+	private static void modifyProjectManifest(Path projectRoot, Map<String, String> newEntries) throws IOException {
```
laeubi, line 235: Why not just a `Consumer<Manifest>`? It seems much more flexible to do anything one wants with the manifest. Also the name "newEntries" is confusing when it actually can replace / delete items.

laeubi (commented):
Looks good but one of the test assertions seems a bit strange now.

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
HannesWell: Changed it to use a `Consumer<Manifest>`.

tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java
HannesWell: I split the previous single case into two test-cases since two different scenarios are tested. AFAICT it is usually recommended to test one scenario in one test case. This also makes test failures simpler to understand and in this case simpler to debug the test cases in standalone mode since the resources for that case are not overwritten for the second scenario respectively they cannot interfere (although they should not).
I left build-with-extension scenario in that case for convenience to not have to replicate all of the setup.

Or what are you suggesting exactly?

laeubi: Compared to the "add method testcase" it now requires a major version bump:

> verifyBaselineProblem(verifier2, "ADDED", "METHOD", "newMethodButItIsInternal()", "1.0.1", "2.0.0");

but the other test only requires a minor version bump

> verifyBaselineProblem(verifier1, "ADDED", "METHOD", "concat(java.lang.String,java.lang.String)", "1.0.0",	"1.1.0");

HannesWell: Ah got it.
That a major version bump is expected was already the case before, but I'm not sure why. Maybe the bnd-analyser considers the interface to be a provider-type?
That's what the log says:
```
[DEBUG] visitClassFile(): path=my/api/bundle/MyApiInterface.class resource=:C:\dev\git\eclipse.tycho\tycho-its\target\projects\BaselineMojoTest\testAddInternalMethod\baseline\api-bundle\target\my.api.bundle-1.0.1.jar(my/api/bundle/MyApiInterface.class):
[DEBUG] visitClassFile(): path=my/api/bundle/internal/AnInternalInterface.class resource=:C:\dev\git\eclipse.tycho\tycho-its\target\projects\BaselineMojoTest\testAddInternalMethod\baseline\api-bundle\target\my.api.bundle-1.0.1.jar(my/api/bundle/internal/AnInternalInterface.class):
[DEBUG] visitClassFile(): path=my/api/bundle/pkg1/AnotherApiInterface.class resource=:C:\dev\git\eclipse.tycho\tycho-its\target\projects\BaselineMojoTest\testAddInternalMethod\baseline\api-bundle\target\my.api.bundle-1.0.1.jar(my/api/bundle/pkg1/AnotherApiInterface.class):
[DEBUG] visitClassFile(): path=my/api/bundle/MyApiInterface.class resource=:C:\dev\git\eclipse.tycho\tycho-its\target\projects\BaselineMojoTest\testAddInternalMethod\baseline\api-bundle\target\my.api.bundle-1.0.1.jar(my/api/bundle/MyApiInterface.class):
[DEBUG] visitClassFile(): path=my/api/bundle/internal/AnInternalInterface.class resource=:C:\dev\git\eclipse.tycho\tycho-its\target\projects\BaselineMojoTest\testAddInternalMethod\baseline\api-bundle\target\my.api.bundle-1.0.1.jar(my/api/bundle/internal/AnInternalInterface.class):
[DEBUG] visitClassFile(): path=my/api/bundle/pkg1/AnotherApiInterface.class resource=:C:\dev\git\eclipse.tycho\tycho-its\target\projects\BaselineMojoTest\testAddInternalMethod\baseline\api-bundle\target\my.api.bundle-1.0.1.jar(my/api/bundle/pkg1/AnotherApiInterface.class):
[DEBUG] parseClassFile(): path=my/api/bundle/MyApiInterface.class resource=:340:
[DEBUG] parseClassFile(): path=my/api/bundle/internal/AnInternalInterface.class resource=:142:
[DEBUG] parseClassFile(): path=my/api/bundle/pkg1/AnotherApiInterface.class resource=:138:
[DEBUG] collecting Contracts  from -define-contract
[DEBUG] begin aQute.bnd.classindex.ClassIndexerAnalyzer@70df4f7a
[DEBUG] end
[DEBUG] visitClassFile(): path=my/api/bundle/MyApiInterface.class resource=:340:
[DEBUG] visitClassFile(): path=my/api/bundle/internal/AnInternalInterface.class resource=:142:
[DEBUG] visitClassFile(): path=my/api/bundle/pkg1/AnotherApiInterface.class resource=:138:
[DEBUG] visitClassFile(): path=my/api/bundle/MyApiInterface.class resource=:340:
[DEBUG] visitClassFile(): path=my/api/bundle/internal/AnInternalInterface.class resource=:142:
[DEBUG] visitClassFile(): path=my/api/bundle/pkg1/AnotherApiInterface.class resource=:138:
[DEBUG] attrs for my.api.bundle version="1.0.0"
[DEBUG] attrs for my.api.bundle.internal version="1.0.0";x-internal:=true
[DEBUG] attrs for my.api.bundle.pkg1 version="1.0.0"
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] ?Delta                     ?Type                      ?Name                      ?Project Version          ?Baseline Version         ?Suggested Version        ?
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] ?MAJOR                     ?PACKAGE                   ?my.api.bundle.internal    ?1.0.0                    ?1.0.0                    ?2.0.0                    ?
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] ?..MAJOR                   ?INTERFACE                 ?my.api.bundle.internal.AnInternalInterface                                                              ?
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] ?....ADDED                 ?METHOD                    ?newMethodButItIsInternal()                                                                              ?
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] ?MICRO                     ?MANIFEST                  ?my.api.bundle-1.0.1.jar   ?1.0.1                    ?1.0.0                    ?1.0.1                    ?
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] ?..REMOVED                 ?HEADER                    ?Bundle-Version:1.0.0                                                                                    ?
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] ?..ADDED                   ?HEADER                    ?Bundle-Version:1.0.1                                                                                    ?
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] ?MICRO                     ?RESOURCES                 ?my.api.bundle-1.0.1.jar   ?1.0.1                    ?1.0.0                    ?1.0.1                    ?
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] ?..CHANGED                 ?RESOURCE                  ?META-INF/maven/org.eclipse.tychoits/my.api.bundle/pom.properties                                        ?
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] ?properties files differ                                                                                                                                       ?
[ERROR] ?version: baseline='1.0.0' != reactor='1.0.1'                                                                                                                  ?
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] ?..CHANGED                 ?RESOURCE                  ?META-INF/maven/org.eclipse.tychoits/my.api.bundle/pom.xml                                               ?
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] ?..CHANGED                 ?CLASS                     ?my/api/bundle/internal/AnInternalInterface.class                                                        ?
[ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
[ERROR] C:\dev\git\eclipse.tycho\tycho-its\target\projects\BaselineMojoTest\testAddInternalMethod\baseline\api-bundle [0:0]: Baseline problems found! Project version: 1.0.1, baseline version: 1.0.0, suggested version: 2.0.0, + /my.api.bundle.internal/my.api.bundle.internal.AnInternalInterface/newMethodButItIsInternal()
```

PR #2337, opened 2023-04-08
Fix product type test on mac by accept finding things at any deep

laeubi:
I think this will not hard and should finally fix this test to run on mac


laeubi:
Only two failures on mac left!

HannesWell (approved):
Considering the different folder Layout on Mac this looks reasonable.
Great!

PR #2343, opened 2023-04-10
Allow resolving mixed builds in eager resolve mode

lppedd:
This reuse the same concept used in `TychoReactorReader`, more or less.

`InstallableUnitGenerator` takes for granted a JAR file has been built already for a project that is inside the same reactor.  
This is not always the case.

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/InstallableUnitGenerator.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/InstallableUnitGenerator.java
@@ -319,21 +319,22 @@ public class InstallableUnitGenerator {
 				// TODO in case of "java-source" type, we might want to generate the source IU
 				// based on the parent artifact!
 				File file = artifact.getFile();
-				if (isValidFile(file)) {
-					lastModified = file.lastModified();
-					String type = artifact.getType();
-					if (PackagingType.TYPE_ECLIPSE_PLUGIN.equals(type)
-							|| PackagingType.TYPE_ECLIPSE_TEST_PLUGIN.equals(type) || "bundle".equals(type)) {
-						List<IPublisherAction> actions = getPublisherActions(PackagingType.TYPE_ECLIPSE_PLUGIN, file,
-								artifact.getVersion(), artifact.getArtifactId());
-						return units = publisher.publishMetadata(actions);
-					} else if (PackagingType.TYPE_ECLIPSE_FEATURE.equals(type)) {
-						List<IPublisherAction> actions = getPublisherActions(PackagingType.TYPE_ECLIPSE_FEATURE, file,
-								artifact.getVersion(), artifact.getArtifactId());
-						return units = publisher.publishMetadata(actions);
-					} else {
-						boolean isBundle = false;
-						boolean isFeature = false;
+				lastModified = file.lastModified();
+				String type = artifact.getType();
+				if (PackagingType.TYPE_ECLIPSE_PLUGIN.equals(type)
+					|| PackagingType.TYPE_ECLIPSE_TEST_PLUGIN.equals(type) || "bundle".equals(type)) {
+					List<IPublisherAction> actions = getPublisherActions(PackagingType.TYPE_ECLIPSE_PLUGIN, file,
+							artifact.getVersion(), artifact.getArtifactId());
+					return units = publisher.publishMetadata(actions);
+				} else if (PackagingType.TYPE_ECLIPSE_FEATURE.equals(type)) {
+					List<IPublisherAction> actions = getPublisherActions(PackagingType.TYPE_ECLIPSE_FEATURE, file,
+							artifact.getVersion(), artifact.getArtifactId());
+					return units = publisher.publishMetadata(actions);
+				} else {
+					boolean isBundle = false;
+					boolean isFeature = false;
+
+					if (file.getName().toLowerCase().endsWith(".jar") && file.isFile()) {
 						try (JarFile jarFile = new JarFile(file)) {
 							Manifest manifest = jarFile.getManifest();
 							isBundle = manifest != null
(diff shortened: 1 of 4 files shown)


lppedd:
This is related to #2341.

laeubi (changes requested):
As already said, this is nothing Tycho should/would support.

laeubi:
> @laeubi there is nothing related to IDEs here. This simply allows resolving non-Tycho packagings without requiring them to have been built already.

If they are not build, Tycho can't use them because there is no generic way to know what packages they export, what bundle or version they will provide and so on.

lppedd:
@laeubi obviously that is not the typical use-case.

IntelliJ IDEA injects a custom `WorkspaceReader` that returns a project's base directory in case that project doesn't have an artifact associated already. This demo PR shows that by letting Tycho work with a directory instead of a JAR file only, the entire flow works smoothly.

In any other context where there are no custom `WorkspaceReader`s injected, Tycho will work as before.

laeubi:
A base directory is not enough. The general contract for a mixed build is that the project produces a jarfile and this jar file is a bundle, everything else is invalid and not supported.

So please provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue with a plain `mvn clean verify`, and then we probably can change things showing that this use-case now works.

lppedd:
@laeubi the problem here is that you see Tycho only from the build perspective. Not all tools that use Maven will use it to build.

Some tools only need to configure themselves based on how Maven generate its model.
In this specific case what happens is the Maven model will register a "mixed" jar-packaged project as a dependency of another tycho-packaged project, that's it. No build will ever occur.

laeubi:
As said Tycho can't take the burden of other tools, if they need special treatment it must be handled there.

If no build ever occurs, the default Tycho mode will be enough and will never give any errors. Eager resolve is only for special cases where the Targetplatform is known in advance and one needs references to reactor projects via transitive dependencies and any other use-case it not supported, especially if it is not possible to write an integration test for this as this is completely internal behavior of Tycho and will change anytime in any way.

> This change allows the Mavel model configuration to finish without generating errors

This seems very unlikely by the provided error you see, that says you depend on another module providing a bundle but Tycho can't know this before the bundle is build and therefore would be never satisfiable.

If you want something similar to what was done before (resolve as much dependencies as currently possible), this can be done by an ow extension (possibly reusing Tycho internal API), but this is not what Tycho resolving process does.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/InstallableUnitGenerator.java
```diff
@@ -343,16 +344,37 @@ public synchronized Collection<IInstallableUnit> getUnits(Artifact artifact) {
...
-							List<IPublisherAction> actions = getPublisherActions(PackagingType.TYPE_ECLIPSE_FEATURE,
-									file, artifact.getVersion(), artifact.getArtifactId());
-							return units = publisher.publishMetadata(actions);
+					}
+
+					// TychoReactorReader or another custom WorkspaceReader might have returned
+					// a "dummy" file for the project's artifact if the project is in the reactor.
+					// Usually that file is the project's base directory
+					if (file.isDirectory()) {
+						final File manifestFile = new File(file, "META-INF/MANIFEST.MF");
```
laeubi, line 353: If you want such behavior you should simply use the `eclipse-plugin` / `eclipse-feature` packaging for that project.

tycho-core/src/main/java/org/eclipse/tycho/core/TargetPlatformConfiguration.java
```diff
@@ -156,11 +156,7 @@ public void setPomDependencies(PomDependencies pomDependencies) {
 
     public PomDependencies getPomDependencies() {
         if (pomDependencies == null) {
-            if (isRequireEagerResolve()) {
-                return PomDependencies.ignore;
-            } else {
-                return PomDependencies.consider;
-            }
+            return PomDependencies.consider;
```
laeubi, line 159: This is an incompatible change, please the the documentation, consider is only the default for eager resolve is not configured.

tycho-core/src/main/java/org/eclipse/tycho/p2resolver/P2ResolverImpl.java
```diff
@@ -227,9 +227,7 @@ protected P2ResolutionResult resolveDependencies(Collection<IInstallableUnit> ro
             if (project != null && p2ResolverFactoryImpl != null && pomDependencies != PomDependencies.ignore) {
                 TargetPlatformConfiguration configuration = p2ResolverFactoryImpl.getProjectManager()
                         .getTargetPlatformConfiguration(project);
-                if (!configuration.isRequireEagerResolve()) {
-                    data.setAdditionalUnitStore(p2ResolverFactoryImpl.getPomUnits().createPomQueryable(project));
-                }
+                data.setAdditionalUnitStore(p2ResolverFactoryImpl.getPomUnits().createPomQueryable(project));
```
laeubi, line 230: This is only valid in lazy mode, because in eager mode all dependencies are already resolved beforehand and can't be read from the pom (again).

lppedd:
I'll address points here.

> if they need special treatment it must be handled there

There is no way to handle mostly anything outside of Tycho as it is so "self contained" that everybody would have to replicate what Tycho does, which makes no sense.

> this can be done by an ow extension (possibly reusing Tycho internal API)

Cannot be done because of the very changes I had to make here. See `PomDependencies.ignore` > `PomDependencies.consider`.
This is an implementation detail that would be impossible to change externally, and would conflict in the use of `tycho-spi` definitions such as `TychoResolver` in a custom extension.

> If you want such behavior you should simply use the eclipse-plugin / eclipse-feature packaging for that project

Might be true, but a POM-specified dependency should be registered in the model without throwing errors if no build is occurring.

See https://github.com/openjdk/jmc/blob/91329f1b8d9896d5160e22262f06385bf43e960b/application/pom.xml#L151-L155 for the dependency declaration related to the error shown above.

> This seems very unlikely by the provided error you see

I've tested the changes and everything is imported correctly. Obviously tested on JMC and other internal projects.

> consider is only the default for eager resolve is not configured

That change is needed otherwise `PomInstallableUnitStore` will never return a useful query result.
```
if (considerPomDependencies == PomDependencies.ignore) {
    return EMPTY_RESULT;
}
```

lppedd:
> Also as said, you don't want a "Resolver"

I'm not entirely sure about this. The matter is I need the resolution process against p2 repositories too, I need all the stack. I don't need only inter-workspace resolution. A `GraphBuilder` only handles connections between reactor projects.

> if you want eager resolve of pom dependencies use PomDependencies = consider or PomDependencies = wrap in your target platform

This actually helped, thanks. As this is done through IDE code, I had to add a configurable system property in this style
```
String value = getStringValue(configuration.getChild(POM_DEPENDENCIES), session, "tycho.target.pomDependencies", null);
```
Which is then set before execution
```
@Override
public void customizeUserProperties(
  final Project project,
  final MavenProject mavenProject,
  final Properties properties
) {
  properties.setProperty("tycho.target.eager", "true");
  properties.setProperty("tycho.target.pomDependencies", "consider");
}
```

The good side is I have reverted what you mentioned as a breaking change in `TargetPlatformConfiguration#getPomDependencies`.
Keeping the removal of `if (!configuration.isRequireEagerResolve()) { ... }` in `P2ResolverImpl` seems to be required still tho. The import fails if I reinstate the condition.

> because it won't work if no Manifest is there

Well yes, there is a trade-off to be made, since it is impossible to handle all the cases.

> As said you should provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test)

It's impossible to provide an integration test as the import process is handled through Maven Embedder in the IDE context.

laeubi:
@lppedd with this PR https://github.com/eclipse-tycho/tycho/pull/2418 Tycho now can detect `bnd-process` and generates a preliminary manifest if eager resolution is requested (you still need to enable to consider pom dependencies just in case), probably you want to try that out if it helps with your issue, but depending on your exact configuration it might need some enhancements (e.g. if you use bnd:jar or similar).

PR #2345, opened 2023-04-11
Allow setting target's pomDependencies via user property

lppedd:
As per title, this allow setting the `pomDependencies` value via property, e.g., `-Dtycho.target.pomDependencies=consider`.


laeubi:
Sorry for they delay, this looks good, thanks for adding property support here.

PR #2379, opened 2023-04-29
Update documentation of Tycho-compiler/sufire-plugin's 'useJDK' property and minor improvments

HannesWell:
Update documentation of the tycho-compiler/sufire-plugin's 'useJDK' property to fix https://github.com/eclipse-tycho/tycho/issues/2377.

Additionally apply some minor code improvements in the Toolchains processing, for example
- Replace deprecated DefaultJavaToolChain by JavaToolchainImpl
- Don't walk the entire directory when searching `JDK_HOME/bin` for a toolname case insensitive.

PR #2380, opened 2023-05-02
Minor simplifications and clean-ups in Toolchain processing

HannesWell:
Replace deprecated DefaultJavaToolChain by JavaToolchainImpl.

Initially part of https://github.com/eclipse-tycho/tycho/pull/2379, but there seem to be some bugs in this change.
@laeubi do you directly have any hot hint?

### DIFF
--- a/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
+++ b/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
@@ -882,12 +882,12 @@ public abstract class AbstractOsgiCompilerMojo extends AbstractCompilerMojo impl
 
     @Override
     public List<ClasspathEntry> getClasspath() throws MojoExecutionException {
-        ReactorProject reactorProject = DefaultReactorProject.adapt(project);
         List<ClasspathEntry> classpath;
         String dependencyScope = getDependencyScope();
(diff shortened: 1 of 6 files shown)


laeubi:
@HannesWell I have no immediate idea, especially as these are maven classes, you cna try to even split this more into some simplifications that are maybe less critical, beside that, in the past it has worked well to abstract the maven parts (e.g by doing as much as possible in tycho-core) instead of any of the mojos,, I even using reflection on some places :-\

laeubi:
@HannesWell do you plan to work on this or should we postpone this until Maven 4?

HannesWell:
I temporarily reverted the replacement of the deprecated DefaultJavaToolChain. Lets see if this makes a difference. I also checked again the other changes and actually they look save, but I might have overseen something.

HannesWell:
> I temporarily reverted the replacement of the deprecated DefaultJavaToolChain. Lets see if this makes a difference. I also checked again the other changes and actually they look save, but I might have overseen something.

OK, that worked. But I don't understand why it breaks something if we just use the super-class.
Lets try it again and the previous failure was just a temporary issue.
If it still fails, lets submit this with the deprecated `DefaultJavaToolChain` in place.

HannesWell:
Looks like this worked now.
Could it be that the test-runners where using a too old Maven version?
Any way, @laeubi any object to merge this now?

laeubi:
> Could it be that the test-runners where using a too old Maven version?

I have no clue, as it works now I don't see a reason to further wait with this.

PR #2418, opened 2023-05-17
Detect metadata of bnd-process

laeubi:


HannesWell:
Could it be that this caused a regression in the M2E build?
https://github.com/eclipse-m2e/m2e-core/pull/1376
https://ci.eclipse.org/m2e/job/m2e/job/master/583/

The day before the build succeeded without issues.

laeubi:
The build complains about a baselineproblem so I'm not sure how this could be related?

HannesWell:
The new Bundle-ClassPath header contains more entries than before and than expected and the change in the mentioned PR should not have a influence on this.

laeubi:
Thats really strange, as everything is supposed to happen in memory and Tycho only using the P2 metadata I'm curious how this can influence how BND process the jar afterwards ..

laeubi:
I tracked this down to to the fact that returning the additional units seem to modify the model in a way so that the dependency-copy goal see more dependencies, what results in more jars downloaded as desired...

laeubi:
Alright the "magic" is in `org.eclipse.tycho.core.maven.TychoMavenLifecycleParticipant.afterProjectsRead(MavenSession)` where we inject reactor project dependencies, due to the metadata acquired we now know that (from OSGi POV) m2e-archetypes requires m2e-runtime, so a new compile dependency is injected, what pull in all dependencies of m2e-runtime... the fix seems quite straight forward to just skip adding additional dependencies for non-tycho project:
- https://github.com/eclipse-tycho/tycho/pull/2423

HannesWell:
> the fix seems quite straight forward to just skip adding additional dependencies for non-tycho project:
> Skip the injection of additional projects if not managed by Tycho #2423

Sounds reasonable, thanks!
I assume a similar problem cannot happen for a Tycho Project, since there it is desired to have the OSGi dependencies.

PR #2421, opened 2023-05-18
PR for issue 1711

vaclavHala:
I had to extract the part which does variable resolution from TargetDefinitionResolver into the .shared bundle
so it is available both in TargetDefinitionResolver but also in Mojos which deal with repository location.

Also had to make some public access to TargetDefinition.Repository implementation in TargetDefinitionFile.
This is very ugly, I feel the problem here is the Repository in TargetDefinition and TargetDefinitionFile
should in fact not be the same class as one represents the repository element as present in .target file
(potentially with unresolved variables) and the other is resolved repository with valid URI.
Changing this would be a bigger refactoring I felt would be too invasive to do here.

I did not find any test for update-target and am not sure how to write one (since it updates files in place)
but at least tested that one manually and seems to work fine.
The other affected plugin, target validation Mojo, is tested in IT along with testing the locations resolve correctly.


vaclavHala:
not sure what is up with the ECA, I already have that signed and contributed PR to this repo before https://github.com/eclipse-tycho/tycho/pull/622

laeubi:
@vaclavHala is this still important for you? 

If yes I would suggest to first create a PR to make the 2.7. branch producing snapshots again and the backport:

- https://github.com/eclipse-tycho/tycho/pull/2425

PR #2425, opened 2023-05-22
PR for issue 1711 - master

vaclavHala:


tycho-core/src/test/java/org/eclipse/tycho/p2resolver/TargetDefinitionFileTest.java
```diff
@@ -44,7 +43,7 @@ public void testTarget() throws Exception {
 
         InstallableUnitLocation location = (InstallableUnitLocation) locations.get(0);
         assertEquals(1, location.getRepositories().size());
-        assertEquals(URI.create("https://download.eclipse.org/eclipse/updates/3.5/"),
+        assertEquals("https://download.eclipse.org/eclipse/updates/3.5/",
```
laeubi, line 46: Are the changes in this file necessary? Why using sting instead of URI for equals?

tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetDefinitionResolverService.java
```diff
@@ -117,6 +118,11 @@ private void debugCacheMiss(ResolutionArguments arguments) {
         }
     }
 
+    public URI resolveRepositoryLocation(String location) {
```
laeubi, line 121: Can we probably move the code for resolving variables from `TargetDefinitionResolver` into `TargetDefinitionResolverService` then? Also I think there should be exactly one method to replace variables, the parsing into an URL should happen at the relevant place, or we can have helper methods that call the generic one like resolveAsURI, resolveAsFile resolveAs...

tycho-extras/target-platform-validation-plugin/src/main/java/org/eclipse/tycho/extras/tpvalidator/TPValidationMojo.java
```diff
@@ -194,8 +198,9 @@ private void validateTarget(File targetFile) throws TPError {
             for (Location location : targetDefinition.getLocations()) {
                 if (location instanceof InstallableUnitLocation p2Loc) {
                     for (Repository repo : p2Loc.getRepositories()) {
-                        ref.addArtifactRepository(repo.getLocation());
-                        ref.addMetadataRepository(repo.getLocation());
+                        var repoUri = definitionResolver.resolveRepositoryLocation(repo.getLocation());
```
laeubi, line 201: please don't use `var` for trivial types.

tycho-its/src/test/java/org/eclipse/tycho/test/target/TargetVariableResolutionTest.java
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	@After
+	public void stopServer() throws Exception {
+		server.stop();
+	}
+
+	@Test
+	public void repositoryUrlCanContainEnvVarVariable() throws Exception {
+		Verifier verifier = getVerifier("target.variables-env", false);
```
laeubi, line 34: can we use subfolders instead of dash to separate different use cases e.g.
```suggestion
		Verifier verifier = getVerifier("target.variables/env", false);
```
just in case more tests are added this is easier to manage.

tycho-its/src/test/java/org/eclipse/tycho/test/target/TargetVariableResolutionTest.java
```diff
@@ -0,0 +1,49 @@
...
+		verifier.setEnvironmentVariable("MY_MIRROR", baseurl);
+		verifier.executeGoals(Arrays.asList("package"));
+		verifier.verifyErrorFreeLog();
+		verifier.verifyTextInLog("validate-target-platform");
+	}
+
+	@Test
+	public void repositoryUrlCanContainSystemPropertyVariable() throws Exception {
+		Verifier verifier = getVerifier("target.variables-sysprop", false);
```
laeubi, line 43:
```suggestion
		Verifier verifier = getVerifier("target.variables/sysprop", false);
```

tycho-targetplatform/src/main/java/org/eclipse/tycho/targetplatform/TargetDefinitionFile.java
```diff
@@ -563,6 +563,10 @@ public static boolean isTargetFile(File file) {
 	            && !file.getName().startsWith(".polyglot.");
 	}
 
+	public static TargetDefinition.Repository repository(String id, String uri) {
```
laeubi, line 566: Why is this required and one can not just call the constructor directly?

laeubi (commented):
This looks fine in general but some adjustments should be performed!

laeubi:
@vaclavHala also please add an entry to the release notes so people are aware of that feature and link the Issue in your commit message.

tycho-core/src/test/java/org/eclipse/tycho/p2resolver/TargetDefinitionFileTest.java
vaclavHala: Yes it is necessary because I had to change `Repository.getLocation`  return type from `URI` to `String`.
Since URI will never be .equal() to any object of type String I had to change the expected value to also be String

tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetDefinitionResolverService.java
vaclavHala: > Can we probably move the code for resolving variables from TargetDefinitionResolver into TargetDefinitionResolverService then?

Well yes, but then we have to pass `TargetDefinitionResolverService` into `TargetDefinitionResolver` as some sort of callback for resolving variables because the `resolve*` methods are needed deep in the code of `TargetDefinitionResolver`.
Maybe you had something like this in mind?
``` java
interface TargetDefinitionVariableResolver {
  String resolveVariable(String raw);
}

class TargetDefinitionResolverService implements TargetDefinitionVariableResolver{
  String resolveVariable(String raw){ 
    // move code for env_var, systep_prop and project_loc here
  }
  
  CompletableFuture<TargetDefinitionContent> resolveFromArguments(ResolutionArguments arguments) {
    ...
    var resolver = new TargetDefinitionResolver(this, arguments.environments, ...);
    resolver.resolveContent(...)
}

class TargetDefinitionResolver {
  final TargetDefinitionVariableResolver varResolver;
  TargetDefinitionResolver(TargetDefinitionVariableResolver varResolver, ...){}

  TargetDefinitionContent resolveContentWithExceptions(TargetDefinition definition, ...){
    ...
    } else if (locationDefinition instanceof PathLocation pathLocation) {
        String resolvePath = varResolver.resolvePath(pathLocation.getPath(), definition);
    ...
  }
}
```

personally I prefer having the resolution in `TargetDefinitionResolver` and delegating to it from `TargetDefinitionResolverService` where necessary but it is your call

> Also I think there should be exactly one method to replace variables,

ok I can do that, I did it as a separate method because I didn't want repository location to support `project_loc` as that makes no sense to me but I guess there is no harm in having that option available and just never user rather than creating special value for it

tycho-targetplatform/src/main/java/org/eclipse/tycho/targetplatform/TargetDefinitionFile.java
vaclavHala: I found this less intrusive than making the private type public, but if you have no problem with that I'll just increase visibility on the Repository class and remove this

tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetDefinitionResolverService.java
laeubi: Having a `TargetDefinitionVariableResolver` makes sense, maybe even as a separate component (it seems the mojos only reference `TargetDefinitionResolverService` because the actually want to resolve variables).

>  I did it as a separate method because I didn't want repository location to support project_loc as that makes no sense to me but I guess there is no harm in having that option available and just never user rather than creating special value for it

Yes we should just support "variables" the user has to decide what is the best one to use...

tycho-targetplatform/src/main/java/org/eclipse/tycho/targetplatform/TargetDefinitionFile.java
laeubi: But where is this method actually used?

vaclavHala: in `UpdateTargetMojo.LatestVersionLocation#getRepositories` when creating copy of repository with its location resolved

laeubi: In this case it would be better to have a private implementation in `UpdateTargetMojo` that can delegate everything to original item but return a different URL ...

tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetDefinitionResolverService.java
vaclavHala: What is the plexus way of doing `@PostConstruct`?

I created the `TargetDefinitionVariableResolver` but now I need to get the logger from `mavenContext` after injecting it
``` java
@Component(role = TargetDefinitionVariableResolver.class)
public class TargetDefinitionVariableResolver {

    @Requirement
    private MavenContext mavenContext;
    @Requirement // this does not work
    private MavenLogger logger;
    
    @PostConstruct // this does not work either
    private void postConstruct(){
      // I expect the DI framework to call this method after it injected all members
      this.logger = mavenContext.getLogger();
    }
```
of course I could get the logger at the start of the `resolve` method (so it sets the logger the first time it is called and does effectively noop on each subsequent call) but that seems ugly

vaclavHala: well both `org.eclipse.tycho.core.shared.MavenContext` and `org.eclipse.tycho.core.shared.MavenLogger` were used in the original `TargetDefinitionResolver`, I just moved the code from it into `TargetDefinitionVariableResolver` without doing any further changes.

If you wish I can replace the `logger` there by injected `org.codehaus.plexus.logging.Logger`.
As for the `mavenContext` it was used for
`mavenContext.getSessionProperties()` and `mavenContext.getProjects()` - what should I replace these by to avoid the `MavenContext`?

thanks for the post construct hint

tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetDefinitionVariableResolver.java
```diff
@@ -0,0 +1,99 @@
+package org.eclipse.tycho.p2resolver;
```
laeubi, line 1: Please add a header from the file you have extracted/copied this

tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetDefinitionVariableResolver.java
```diff
@@ -0,0 +1,99 @@
...
+@Component(role = TargetDefinitionVariableResolver.class)
+public class TargetDefinitionVariableResolver implements Initializable {
+
+    private static final Pattern SYSTEM_PROPERTY_PATTERN = createVariablePatternArgument("system_property");
+    private static final Pattern PROJECT_LOC_PATTERN = createVariablePatternArgument("project_loc");
+    private static final Pattern ENV_VAR_PATTERN = createVariablePatternArgument("env_var");
+
+    @Requirement
+    private MavenContext mavenContext;
+    private MavenLogger logger;
```
laeubi, line 24: Please use `org.codehaus.plexus.logging.Logger` like this
```suggestion
    @Requirement
    private Logger logger;
```

that way we don't need the `Initializable` nor the `MavenContext`

tycho-extras/tycho-version-bump-plugin/src/main/java/org/eclipse/tycho/versionbump/UpdateTargetMojo.java
```diff
@@ -178,4 +192,26 @@ public String getVersion() {
 
     }
 
+    private static final class Repository implements TargetDefinition.Repository {
```
laeubi, line 195: You might want to choose a more descriptive name e.g.
```suggestion
    private static final class ResolvedRepository implements TargetDefinition.Repository {
```

laeubi (commented):
Thanks for the adjustments I think it looks quite good now just some smaller adjustments/hints here.

tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetDefinitionVariableResolver.java
vaclavHala: I replaced the logger as you suggest, but we still do need the MavenContext. It was not used just to get to the logger, I also need it to `mavenContext.getProjects()` and `mavenContext.getSessionProperties()`. What are suitable replacements for these?

laeubi: You can take a look at `DefaultMavenContext` class,

mavenContext.getSessionProperties() --> `Session.getProperties()`
mavenContext.getProjects() --> `Session.getProjects()`

but if you like you can just retain this for now and only replace the logger, its just good to remove references to the MavenContext where it is easily possible (like loggers).

vaclavHala: I looked there but if I understand you correctly that would mean instead of `MavenContext` I'd have to inject `LegacySupport` and copy/paste the `get(Session|Propeties|Projects)` methods from `DefaultMavenContext` into the `TargetDefinitionVariableResolver` which hardly seems as an improvement

laeubi: If you want access to the session, you can add `@SessionScoped` to the component and add a constructor like this:

```
@Inject
public MySessionScopedComponent(MavenSession mavenSession) {
    this.mavenSession = mavenSession;
}
```

vaclavHala: Ok I did that but now there is a new problem... 
In unit tests I need to provide some `TargetDefinitionVariableResolver`. 
Before I just passed in real impl with mocked `MavenContext` I already had available.
Now I'd need to provide mocked `MavenSession` which seems like a lot of trouble, so I opted to rather turn `TargetDefinitionVariableResolver` into an interface and have its implementation
``` java
@SessionScoped
@Component(role = TargetDefinitionVariableResolver.class)
public class DefaultTargetDefinitionVariableResolver implements TargetDefinitionVariableResolver {
    @Requirement
    private Logger logger;

    @Inject
    public DefaultTargetDefinitionVariableResolver(MavenSession session) {
        this.mavenSession = session;
    }
```
Now this works great in the unit tests, but it started causing trouble in tests inherited from `AbstractTychoMojoTestCase` as the real impl with injeted session is used here (any way to place mocks into the plexus container in these tests?) and apparently the `MavenSession` that gets injected into it is different than the session prepared inside `AbstractTychoMojoTestCase#setUp`. In particular I tracked it down to this discrepancy - when the failure of the test occurs there is stack trace like this
```
DefaultTargetDefinitionVariableResolver.projects() line: 102    
DefaultTargetDefinitionVariableResolver.findProjectLocation(String) line: 70    
0x0000000801817800.apply(Object) line: not available    
DefaultTargetDefinitionVariableResolver.resolvePattern(String, Pattern, Function<String,String>) line: 110  
DefaultTargetDefinitionVariableResolver.resolve(String) line: 61    
TargetDefinitionResolver.resolvePath(String, TargetDefinition) line: 294    
TargetDefinitionResolver.resolveContentWithExceptions(TargetDefinition, IProvisioningAgent) line: 146   
TargetDefinitionResolver.resolveContent(TargetDefinition, IProvisioningAgent) line: 102 
TargetDefinitionResolverService.resolveFromArguments(ResolutionArguments) line: 93  
0x00000008018131d0.apply(Object) line: not available    
ConcurrentHashMap<K,V>.computeIfAbsent(K, Function<? super K,? extends V>) line: 1708   
TargetDefinitionResolverService.getTargetDefinitionContent(TargetDefinition, List<TargetEnvironment>, ExecutionEnvironmentResolutionHints, IncludeSourceMode, IProvisioningAgent) line: 68  
TargetPlatformFactoryImpl.resolveTargetDefinitions(TargetPlatformConfigurationStub, ExecutionEnvironmentResolutionHints) line: 214  
TargetPlatformFactoryImpl.createTargetPlatform(TargetPlatformConfigurationStub, ExecutionEnvironmentResolutionHandler, List<ReactorProject>) line: 164  
TargetPlatformFactoryImpl.createTargetPlatform(TargetPlatformConfigurationStub, ExecutionEnvironmentConfiguration, List<ReactorProject>) line: 137  
TargetPlatformFactoryImpl.createTargetPlatform(TargetPlatformConfigurationStub, ExecutionEnvironmentConfiguration, List) line: 88   
ReactorRepositoryManagerImpl.computePreliminaryTargetPlatform(ReactorProject, TargetPlatformConfigurationStub, ExecutionEnvironmentConfiguration, List<ReactorProject>) line: 71    
P2DependencyResolver.computePreliminaryTargetPlatform(MavenSession, MavenProject, List<ReactorProject>) line: 205   
...
```
and in `P2DependencyResolver` its `MavenSession#getProjects` returns
```
MavenProject: bootclasspath:bootclasspath:0.0.1 @ /data/codasip/git/vaclavHala_tycho/tycho-core/target/projects/bootclasspath/pom.xml,
MavenProject: bootclasspath:bundle1:0.0.1 @ /data/codasip/git/vaclavHala_tycho/tycho-core/target/projects/bootclasspath/bundle1/pom.xml,
MavenProject: bootclasspath:bundle2:0.0.1 @ /data/codasip/git/vaclavHala_tycho/tycho-core/target/projects/bootclasspath/bundle2/pom.xml
```
whereas in `DefaultTargetDefinitionVariableResolver` its `MavenSession#getProjects` returns
```
MavenProject: unknown:empty-project:0 @ 
```

**TL;DR**
I'd prefer to stick with `MavenContext` for now as replacing it by `MavenSession` seems to have far reaching consequences I don't fully understand and don't feel able to properly handle

laeubi: Just make sure the project is a `TychoPlexusTestCase` then you can simply `lookup(TargetDefinitionVariableResolver.class)`

vaclavHala: Sorry I can't see how that helps me, all the failing tests already implement `PlexusTestCase` which probides the `lookup` method. Problem is not how to get to `TargetDefinitionVariableResolver`, but how to setup the test such that anyone who looks up the resolver using the method you show will get some mock implementation instead of the real one. I have pushed commit showing the problem

laeubi: You should never need to use any mocks to test your changes, only test the user visible API and/or using real integration test projects.

laeubi: Beside that, if you want to change the session values in a test you can override `org.eclipse.tycho.testing.TychoPlexusTestCase.modifySession(MavenSession)`

laeubi: Regarding failures like:

`Caused by: java.lang.NullPointerException: Cannot invoke "java.io.File.getName()" because the return value of "org.eclipse.tycho.ReactorProject.getBasedir()" is null`

one should simply add a null-check at that location.

vaclavHala: > one should simply add a null-check at that location.

I don't think I agree here, these tests worked fine before but started breaking when I changed `MavenContext` in the variable resolver for `MavenSession` to look up current projects. The `MavenSession` that gets injected provides different list of projects than `MavenContext`. Projects provided by the `MavenSession` (as shown in my comment above) do not have the `basedir` set which causes these NPE failures.

I agree mocking is probably not the way to do here. However simply injecting the `MavenSession` leads to the test failures for reasons I do not currently understand and don't think I could address in reasonable time. So since using `MavenSession` instead of `MavenContext` is not a matter of simple 1:1 replacement I suggest we do not go through with it in this PR and keep the `MavenContext` for now

laeubi: > So since using MavenSession instead of MavenContext is not a matter of simple 1:1 replacement I suggest we do not go through with it in this PR and keep the MavenContext for now

That would be okay.

vaclavHala:
@laeubi I pushed some new commits, I believe all your comments are now addressed. Please let me know if you need me to do anything more before this can be merged.

Also, what needs to be done for this new feature to be backported into 2.7.x so it is included in 2.7.5 release? I can update https://github.com/eclipse-tycho/tycho/pull/2421 to make it match changes in this PR as closely as possible

tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetDefinitionVariableResolver.java
```diff
@@ -0,0 +1,23 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2022 SAP SE and others.
```
laeubi, line 2:
```suggestion
 * Copyright (c) 2023 Vaclav Hala and others.
```

tycho-core/src/main/java/org/eclipse/tycho/p2resolver/DefaultTargetDefinitionVariableResolver.java
```diff
@@ -0,0 +1,112 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2022 SAP SE and others.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ * Contributors:
+ *    Vaclav Hala - initial API and implementation
```
laeubi, line 10:
```suggestion
 *    Vaclav Hala - extraction code from TargetDefinitionResolver into dedicated component
```

tycho-its/src/test/java/org/eclipse/tycho/test/target/TargetVariableResolutionTest.java
```diff
@@ -0,0 +1,61 @@
+/*******************************************************************************
+ * Copyright (c) 2011, 2022 SAP SE and others.
```
laeubi, line 2:
```suggestion
 * Copyright (c) 2023 Vaclav Hala and others.
```

laeubi (approved):
Looks good, some header adjustments for completeness required then its fine to merge.

laeubi:
@vaclavHala also please rebase/squash all changes into one commit if possible.

> Also, what needs to be done for this new feature to be backported into 2.7.x so it is included in 2.7.5 release? I can update https://github.com/eclipse-tycho/tycho/pull/2421 to make it match changes in this PR as closely as possible

If code has diverged to much its most often easier to "start from scratch" and just try to insert the interesting parts accordingly also its most often good to not perform refactoring in backports, even though this might lead to duplication or not so clean code, but for a backport smaller local changes are usually better.

>  included in 2.7.5 release

I fear a new release of the 2.7.5 is quite unlikely if we can get the CI working (again) there should be a SNAPSHOT to test at laest, but you might want to backport it to 3.x first there will be a release after the 2023-06 eclipse release (about two week from now on).

vaclavHala:
I pushed new commit with the header changes you wanted and everything squashed.

> included in 2.7.5 release

I meant next release in the 2.7.x line, I see I mistyped the exact version as 2.7.5 already exists, so 2.7.6.
We'd like to see this in the 2.7.x as we are not yet on java 17 and I'm not sure how feasible upgrading to that would be at the moment for us. So that said 3.x does not help us as that also requires java 17.

For now please let's just merge this into master and we will discuss internally how we want to proceed with the backporting/upgrading first

laeubi:
This looks good now and is ready to be merged.

> We'd like to see this in the 2.7.x as we are not yet on java 17 and I'm not sure how feasible upgrading to that would be at the moment for us. 

Please note that you only need java 17 to run the maven build, the code can still be all java 11 (or 1.8 or ..) so if you see any problems regarding this usecase let us know!

PR #2434, opened 2023-05-23
Synchronize Maven Target support shared with M2E

HannesWell:
Copy changes from https://github.com/eclipse-m2e/m2e-core/pull/1391

### DIFF
--- a/.github/workflows/verify-platform.yml
+++ b/.github/workflows/verify-platform.yml
@@ -45,7 +45,7 @@ jobs:
     - name: Set up Maven
       uses: stCarolas/setup-maven@v4.5
       with:
-        maven-version: 3.8.6
+        maven-version: 3.9.1
     - name: Build Tycho
       env:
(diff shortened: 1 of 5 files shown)


HannesWell:
Looks like `org.eclipse.aether.spi.synccontext.SyncContextFactory` is only available since maven-resolver-spi 1.7, but Maven 3.8.7 only contains maven-resolver-spi 1.6.3.
In order to not require Maven 3.9 for Tycho is suggest to simply stay at the old class, that is deprecated with Maven 3.9.x.
The discrepancy to m2e is IMHO ok.

laeubi:
Tycho is supposed to use Maven 3.9 but it is not yet enforced.

HannesWell:
> Tycho is supposed to use Maven 3.9 but it is not yet enforced.

Understand, updated the verify-platform workflows to use Maven 3.9.1.
It looks like until now it also worked with 3.8.6

PR #2442, opened 2023-05-26
Update to ECJ 3.34

### LINKED ISSUE #2441: Fix for broken JDT null-analysis in tycho 3.x
Hi,

it would be nice to have a Tycho 3.x version with the final JDT 2023-06 artifacts with the fix for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1007 as the problem is also visible in tycho as https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1093 demonstrates.

If a new 3.x version is not possible: is it possible to overide the jdt/ecj version the tycho compiler plugin uses?

cdietrich:
Update to ECJ 3.34
Fixes #2441

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,7 +86,7 @@
 		<surefire-version>3.0.0-M5</surefire-version>
 
 		<equinoxVersion>3.18.300</equinoxVersion>
-		<ecjVersion>3.33.0</ecjVersion>
+		<ecjVersion>3.34.0</ecjVersion>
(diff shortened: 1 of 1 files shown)


cdietrich:
The ECJ 3.34 release is not available yet

jdt is also used in tycho-apitools-plugin/pom.xml
so these need to be bumped too?

laeubi:
We usually try to update to the latest versions if possible so it should be changed there too.

PR #2480, opened 2023-06-08
Add support for automatic adding repository references to p2 sites

laeubi:
This adds two new options to automatically include repository references from the target / pom repositories.

FYI @akurtakov @merks @vrubezhny  @mickaelistria 

This should ease the usage of `<filterProvided>` as it automatically derive the referenced repositories from the target file or pom configuration, so one usually want to use:

```
<plugin>
	<configuration>
		<filterProvided>true</filterProvided>
		<addPomRepositoryReferences>true</addPomRepositoryReferences>
		<addIUTargetRepositoryReferences>true</addIUTargetRepositoryReferences>
	</configuration>
</plugin>
```

that is both self-contained but minimal in the sense of that it excludes everything already available by one of the sites used in the projects target definition (what actually is the initial intent of a "targetplatform" to specify what is already available in the runtime).

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -226,6 +226,29 @@ The parameters of the `tycho-apitools-plugin:generate` goal have been completed 
 The `tycho-p2-repository-plugin:assemble-repository` mojo has now a new configuration parameter `filterProvided` that (if enabled) filter units and artifacts that are already present in one of the referenced repositories.
 That way one can prevent including items that are already present in the same form in another repository.
 
+If you want to include repository references automatically, there are two other new options:
+
+- `addPomRepositoryReferences` - all P2 repositories from the pom are added as a reference
+- `addIUTargetRepositoryReferences` - all P2 repositories defined in target files IU-location types are added as a reference
+
+so now one can produce a self-contained update-site that only includes what is not already available from the target content used by specify:
+
+
+```
+<plugin>
+	<groupId>org.eclipse.tycho</groupId>
+	<artifactId>tycho-p2-repository-plugin</artifactId>
+	<version>${tycho-version}</version>
+	<configuration>
+		<includeAllDependencies>true</includeAllDependencies>
+		<filterProvided>true</filterProvided>
+		<addPomRepositoryReferences>true</addPomRepositoryReferences>
+		<addIUTargetRepositoryReferences>true</addIUTargetRepositoryReferences>
+	</configuration>
+</plugin>
+
+```
+
 ### New parameter in tycho-packaging-plugin:package-plugin
 
 The `tycho-packaging-plugin:package-plugin` mojo has now a new configuration parameter `deriveHeaderFromSource` (default true), that allows Tycho to discover additional headers declared in the source (e.g. from annotations).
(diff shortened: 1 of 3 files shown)


akurtakov:
Looks quite promising.

merks:
It seems quite useful.  

That being said, "careless" use of this can lead to problems.  (You know how often folks just blindly do things without really understanding the detailed implications.)  E.g., often Eclipse projects refer to locations that are transient, e.g.,  Orbit stable/milestone drops and Platform I-Builds.  When such projects produce release build repositories, they're going to have to be careful to refer to locations that are **permanently** meaningful.  I.e., they'll want to refer to release locations whereas often the builds should also be "testing" against "latest" locations to be sure that the future release of that location will work.

We saw this in the WWD discussion where this approach would work well except perhaps for this location:

https://download.eclipse.org/lsp4e/snapshots

Also the issue of using moving-target locations yet requiring very specific versions in the IUs is problematic...

laeubi:
I think we are already moving to a "new world" where one effectively always using the latest version (java, maven, any 3rd party libs) will become mandatory so having "historic" sites that work forever is a promise that might not be possible to hold anymore (especially fro smaller projects with less resources).

Therefore I'm less concerned in general, it has to "work" until the next release in most cases, everything else might require additional efforts that not everyone seem to want/can handle. The first thing after a bug-report is often to ask the user "are you using the latest version/ibuild/...".

So my primary goal with all those features (m2e-targets, automatic derived updatesites, more flexible build structures and integration with other tool-suites) is more to allow people to have faster and easier releases to deliver the next valuable increment than to stay-in-stone for ever.

merks:
The point is that if one produces a release repository and that repository stops working right after June 14th (when the Platform deletes 4.28-I-Builds, or when the exact required version of some lsp4e IU is no longer in their latest repo) then the users wanting the latest, installing the latest, and/or updating to the latest but failing to do so will fail to achieve the goal of always using the latest. 

But we digress from the specifics of this enhancement itself which is absolutely is cool and useful.  I hope it will be used with an understanding of the implications of using it, which is best achieved by documenting those implications.

laeubi:
AN Update-site is just a collection of things, so referencing the snapshot will not be a problem in general I think, for example I use the snapshot updatesite of m2e, the only "drawback" is that I more often get update-notifications.

regarding I-Builds it is of course bad if they vanish, I think I already suggested somewhere to have an "latest-ibuild" URL that just point to a composite maybe, nerveless I would expect projects are always using the latest released (like m2e) and not the I-Build repo or maybe a (stable url) snapshot repo.

akurtakov:
We already have https://download.eclipse.org/eclipse/updates/I-builds/ which is supposed to serve that need.

laeubi:
> We already have https://download.eclipse.org/eclipse/updates/I-builds/ which is supposed to serve that need.

Any reason we do not use this in the aggregator parent? Currently we always update the URL name there all the time...

akurtakov:
I think because for very short period (between RC2 and release) there is a chance to build with 4.28-I-builds for RC2a/b release while master is being open and using 4.29-I-builds.

Issue #34, opened 2021-04-09
Bug 569937 - Support obr packaging


laeubi:
Christoph Laeubrich CLA Friend 2020-12-28 01:44:33 EST

OSGI has a native repository format [1], we should support building this with tycho as an alternative to the eclipse-repository format.

[1] https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.repository.html
[2] https://github.com/osgi/bindex/tree/885fa197903fe981c10038f69c0dc9a4260ebac6
[3] https://bnd.bndtools.org/chapters/250-resolving.html

[tag] [reply] [−] Comment 1 Christoph Laeubrich CLA Friend 2020-12-28 01:48:16 EST

[4] https://felix.apache.org/documentation/subprojects/apache-felix-osgi-bundle-repository.html

[tag] [reply] [−] Comment 2 Mickael Istria CLA Friend 2021-01-04 08:42:12 EST

Tycho is for building Eclipse artifacts. Before working on publishing to this format, we should work on making Eclipse IDE able to install content from such OBR repositories, and that is more something for p2. Please consider reopening bug 393648 and adding support for OBR in p2 first.

[tag] [reply] [−] Comment 3 Christoph Laeubrich CLA Friend 2021-01-04 08:52:01 EST

From your mentioned bug-report it seems it is not directly related to P2 but to PDE to support this in a target file right?
Supporting it in P2 would be just to resolve dependencies from other P2 Units and even though that's clearly possible it sounds a bit uncommon to me?

BTW Tycho supports some features that PDE/Eclipse won't (multiple target files, special entries in category.xml, ...) so it seams that at least in the past it was okay to have tycho support other things first :-)

Its also a bit of a chicken-egg problem, if we can't produce OBR how would someone want to use it? But of course I think for a real solution there will be needs to support it in tycho and in the IDE!

[tag] [reply] [−] Comment 4 Mickael Istria CLA Friend 2021-01-04 09:09:08 EST

(In reply to Christoph Laeubrich from comment #3)
> From your mentioned bug-report it seems it is not directly related to P2 but
> to PDE to support this in a target file right?
> Supporting it in P2 would be just to resolve dependencies from other P2
> Units and even though that's clearly possible it sounds a bit uncommon to me?

It was initially about P2 being able to consume OBR bundle, so it would automatically cascade to Tycho and PDE and others having this feature.
[...]

PR #2485, opened 2023-06-13
Add support for generation of OSGi repository format

laeubi:
OSGi specifies a standardized repository format [1] to provide bundles and other resources. This adds support for generating such standardized format out of the content of the P2 repository alongside with the p2 format to improve interoperability of tools.

[1] https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.repository.html

Fix https://github.com/eclipse-tycho/tycho/issues/34


mickaelistria:
Cool! So this feature is enabled by default for eclipse-repository packaging type?

laeubi:
It is disabled by default, you can see here how to enable it:

https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#building-osgi-repositories-with-tycho-p2-repository-pluginassemble-repository

My plan would be to enable it for the i-builds (if possible) so we have a larger repository for investigation/testing, ideally it would also be enabled for the eclispe releases but I have no idea if the release repos are actually build by Tycho...

mickaelistria:
> It is disabled by default

Why not turning it on by default? Does it has a major impact on performance?

> My plan would be to enable it for the i-builds (if possible) so we have a larger repository for investigation/testing, ideally it would also be enabled for the eclispe releases but I have no idea if the release repos are actually build by Tycho...

You can try twaeking https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.releng.tychoeclipsebuilder/eclipse.platform.repository/pom.xml

laeubi:
> Why not turning it on by default? Does it has a major impact on performance?

I found that there is always anyone unhappy with that and i haven't measured the performance yet but for lager sides it might add a few seconds as this currently needs to read all bundle/feature jars and analyze them then serialize back to XML ...

mickaelistria:
> I found that there is always anyone unhappy

Keeping everyone happy is an impossible requirement; sometimes people are unhappy for no good reason.

laeubi:
Since its a major release it would be fine to make it the default, we then should just add a note to the migration guide how to disable the feature, so if you like just open a PR that makes it enabled by default :-)

PR #2509, opened 2023-06-14
Versionbumps for 2023-06

laeubi:


p2-maven-plugin/pom.xml
```diff
@@ -28,7 +28,7 @@
 		<dependency>
 			<groupId>org.eclipse.platform</groupId>
 			<artifactId>org.eclipse.equinox.p2.core</artifactId>
-			<version>2.9.200</version>
+			<version>2.10.0</version>
```
laeubi, line 31: As a reminder, probably keep p2 core at the previous version....

Issue #2331, opened 2023-04-06
Tycho 3.x.x doesn't resolve Composite P2 repository in the same way as Tycho 2.5.x


BenShelbourne:
We are in the process of migrating from Tycho 2.5.0 -> 3.0.4 (latest). The problem we are seeing is that we use an internal composite p2 repository which contains other internal p2 repositories. By adding this as a `<repository>` we were able to access everything in the composite repository and all sub-repositories but now Tycho 3 requires authentication for each sub-repository inside the composite repository. We don't want to have to repeat the same authentication details with different ID's for each sub-repository so how can we get the same behaviour seen in 2.5.0, in 3.0.4 ?

laeubi:
Can you explain the structure of your repo/subrepository?

Usually it should work to only use the "shortest path" as an ID to reuse the authentication for other repositories.

laeubi:
So are they all form the same base url and is this base url configured with the credentials?

We have https://github.com/eclipse-tycho/tycho/blob/master/tycho-its/src/test/java/org/eclipse/tycho/test/target/PasswordProtectedM2RepositoryTest.java testcase, maybe you can enhance it to reproduce the issue you see?

BenShelbourne:
So the url for each sub-repository is `<internal-composite-repository-url>/packages/package-A`. The `compositeArtifacts.xml` and `compositeContent.xml` in the bundle points to them via a relative path like so:
```
<?xml version='1.0' encoding='UTF-8'?>
<?compositeArtifactRepository version='1.0.0'?>
<repository name='Composite Repository'
    type='org.eclipse.equinox.internal.p2.artifact.repository.CompositeArtifactRepository' version='1.0.0'>
  <properties size='2'>
    <property name='p2.timestamp' value='1668781002'/>
    <property name='p2.atomic.composite.loading' value='false'/>
  </properties>
  <children size='3'>
    <child location='../../packages/package-A'/>
    <child location='../../packages/package-B'/>
    <child location='../../packages/package-Z'/>
  </children>
</repository>
```
We have the credentials for the bundles in our `settings.xml` but the credentials are the same for all repositories. I can take a look at the tescases. Do you know if anything might have changed from Tycho 2.5.0 until now regarding how you resolve composite repositories? I appreciate 2.5.0 is an *old* version!

laeubi:
You should check if you have configured it for the host not for the site, e.g. assume your server is:

`http://internal-composite-repository-url/packages/package-A`

[...]

PR #2552, opened 2023-06-19
Issue #2331 - Add integration test for password composite p2 repo

BenShelbourne:
I've used 2 of the repositories already checked in to create a composite repository matching the structure we use and then copied the existing `PasswordProtectedP2RepositoryTest.java` to create something for the composite and it fails with a `ProvisionException` which is what we see.


tycho-its/src/test/java/org/eclipse/tycho/test/target/PasswordProtectedCompositeP2RepositoryTest.java
```diff
@@ -0,0 +1,57 @@
+package org.eclipse.tycho.test.target;
```
laeubi, line 1: please add a copyright header to new files

laeubi:
@BenShelbourne I have been working on this a bit but it seems there is an error in the test setup, after fixing the authentication I now get

`No repository found at http://localhost:34287/packages/e352`

so it seems the reference is not correctly mapped, you can reproduce this locally with 
`server = HttpServer.startServer(null, null);`
where it fails the same way!

BenShelbourne:
> @BenShelbourne I have been working on this a bit but it seems there is an error in the test setup, after fixing the authentication I now get
> 
> `No repository found at http://localhost:34287/packages/e352`
> 
> so it seems the reference is not correctly mapped, you can reproduce this locally with `server = HttpServer.startServer(null, null);` where it fails the same way!

I've updated the relative paths in the xml but I'm still seeing `ProvisionException: Unable to read repository at...`

tycho-its/src/test/java/org/eclipse/tycho/test/target/PasswordProtectedCompositeP2RepositoryTest.java
```diff
@@ -1,3 +1,4 @@
+// Copyright (C) 2023 Arm Limited (or its affiliates).
```
laeubi, line 1: Please use

```
/*******************************************************************************
 * Copyright (c) 23 Arm Limited and others.
 * This program and the accompanying materials
 * are made available under the terms of the Eclipse Public License 2.0
 * which accompanies this distribution, and is available at
 * https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *
 * Contributors:
 *    Arm Limited (or your name here) - initial API and implementation
 *******************************************************************************/
```

laeubi:
@BenShelbourne can you squash all into a single commit and make sure you have signed the ECA?

> I've updated the relative paths in the xml but I'm still seeing ProvisionException: Unable to read repository at...

Also when you set username/password to `null`?

BenShelbourne:
> @BenShelbourne can you squash all into a single commit and make sure you have signed the ECA?
> 
> > I've updated the relative paths in the xml but I'm still seeing ProvisionException: Unable to read repository at...
> 
> Also when you set username/password to `null`?

I can do the squash yes, what's an ECA? 😅 

When I set username/password to null the test passes but the server I'm working with requires username/password so we can't not have it. Smells like something to do with how the credentials are used?

laeubi:
> I can do the squash yes, what's an ECA? sweat_smile

You should see it here: https://accounts.eclipse.org/legal/eca/validation/160062

if you don't have an eclipse account just create one and make sure you have enterer your Github username in the account details, then you should see ECA > Sign.

> When I set username/password to null the test passes but the server 

Thats good, and just to verify the site itself is working regardless the credentials issue.

> but the server I'm working with requires username/password so we can't not have it. Smells like something to do with how the credentials are used?

Yes I found out two issues I will address independently:

1. For such a setup, we need to record that a subsite of the original is accessed and use credentials from that "parent" site
2. The preemptive auth currently does not work because Java HttpClient ignores the header if an authenticator is set, that makes that we authenticate two times, that not the issue here but should be fixed.

BenShelbourne:
> You should see it here: https://accounts.eclipse.org/legal/eca/validation/160062
> 
> if you don't have an eclipse account just create one and make sure you have enterer your Github username in the account details, then you should see ECA > Sign.

Thank you!

> Yes I found out two issues I will address independently:
> 
>     1. For such a setup, we need to record that a subsite of the original is accessed and use credentials from that "parent" site

This was the behaviour in Tycho 2.x and I tracked the change to this commit https://github.com/eclipse-tycho/tycho/commit/b1b992c87 and specifically these lines

```
            // cache indices of p2 repositories in the local Maven repository
            RemoteRepositoryCacheManager cacheMgr = new RemoteRepositoryCacheManager(transport, mavenContext);
            agent.registerService(CacheManager.class, cacheMgr);
```
but I don't really know the depths of this so might be a complete red herring!

laeubi:
@BenShelbourne thanks for the test-case it is now passing successful on master.

Please also try out the current [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) build to check with your real builds.

BenShelbourne:
Thank you for fixing this. I can confirm that we can now resolve dependencies from our composite p2 repository in the latest snapshot build.

BenShelbourne:
Although we are still seeing a failure in the `assemble-repository` with resolving credentials
```
[ERROR] An error occurred while transferring artifact canonical: osgi.bundle,x.y.z from repository <internal-server-url>/packages/x.y.z:
[ERROR]    authentication failed for <internal-server-url>/packages/x.y.z.jar
```
But the authentication works for compiling the projects that are being assembled?

laeubi:
Can you maybe enhance the testcase to show that error?

BenShelbourne:
Enahnced the test and it shows the above error but in the `publish-repository` not `assemble-repository`. The error is consistent however https://github.com/eclipse-tycho/tycho/pull/2582

PR #2558, opened 2023-06-20
Propagate authentication data from parent to child

laeubi:
Fix https://github.com/eclipse-tycho/tycho/issues/2331

@BenShelbourne with this patch I can successfully execute your testcase from:

- https://github.com/eclipse-tycho/tycho/pull/2552

**BUT** I think the test do not really performs a download at all, only metadata is resolved!

So we should finish this PR and your testcase and then add another one (or enhance the existing) to actually download a file I think.

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/MavenAuthenticator.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/MavenAuthenticator.java
@@ -18,11 +18,13 @@ import java.net.PasswordAuthentication;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.Stack;
 import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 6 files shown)


BenShelbourne:
> **BUT** I think the test do not really performs a download at all, only metadata is resolved!
> 
> So we should finish this PR and your testcase and then add another one (or enhance the existing) to actually download a file I think.

How can I test actually downloading a file? Is there any examples in the current tests where we do this?

BenShelbourne:
Does it have to reference it in source code or will a reference in the MANIFEST suffice? The project it loads `target.httpAuthentication` has a MANIFEST that requires org.eclipse.osgi. Is that not enough?

laeubi:
The manifest should be enough we can also assume that it works and you try when the new snapshot is available with your real build?

BenShelbourne:
I have just built the latest locally to do this test and it fails. Although appears to be in a different way and along the lines of what you were suggesting. 

```
[ERROR] An error occurred while transferring artifact canonical: osgi.bundle,<x.y.z> from repository <internal-server-url>/packages/1.0.0:
[ERROR]    download from <internal-server-url>/packages/1.0.0/plugins/<x.y.z>.jar failed
```
So it appears that we can now authenticate in to the server but fail to download anything

laeubi:
@BenShelbourne it seems I don't see this locally because of cached items, so no download is attempted ... at least we can see that a download is performed and still fails, so I'll prepare another fix for that.

PR #2600, opened 2023-06-29
Make comparison of newlines in text files more precise

HannesWell:
At the moment the `TextComparator` ignores simplify all kind of newline characters (`\r` and `\n`), which has the effect that, although a new line to a text file was added or removed, the comparator considers that file as unchanged.

This PR enhances the `TextComparator` to only ignore a difference in newline characters if there is is an exactly matching pair of `\r` or `\r\n` in baseline respectively reactor.

This PR also adds tests for the `TextComparator` and moves the `showDiffDetails` property to `ComparisonData` (in the first commit) to simplify testing.

### DIFF
--- a/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/TextComparator.java
+++ b/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/TextComparator.java
@@ -12,9 +12,9 @@
  *******************************************************************************/
 package org.eclipse.tycho.zipcomparator.internal;
 
-import java.io.EOFException;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -36,6 +36,14 @@ public class TextComparator implements ContentsComparator {
 
     static final String HINT = "txt";
 
+    private static final char CR = '\r';
+    private static final char LF = '\n';
+
+    // Possible new lines:
+    // \n -- unix style
+    // \r\n -- windows style
+    // \r -- old Mac OS 9 style, recent Mac OS X/macOS use \n
+
     @Override
     public ArtifactDelta getDelta(ComparatorInputStream baseline, ComparatorInputStream reactor, ComparisonData data)
             throws IOException {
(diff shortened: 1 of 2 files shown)


laeubi:
> At the moment the TextComparator ignores simplify all kind of newline characters (\r and \n), which has the effect that, although a new line to a text file was added or removed, the comparator considers that file as unchanged.

If newlines do not matter for the delta should then additional ones also don't matter? Can you explain in what case an additional newline is relevant? In the end the user can always bump the version anyways if it is considered a relevant change.

tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/TextComparator.java
```diff
@@ -36,62 +36,59 @@ public class TextComparator implements ContentsComparator {
...
+            if (baselineNewLine < 0 || reactorNewLine < 0) {
+                return false;
             }
-            throw new EOFException();
-        }
-
-        boolean hasNext() {
-            skipNewLines();
-            return index < bytes.length;
+            // Both sliders are at either "\n" or "\r\n"
```
laeubi, line 73: Please note that a single `\r` is a valid newline as well!

HannesWell:
> > At the moment the TextComparator ignores simplify all kind of newline characters (\r and \n), which has the effect that, although a new line to a text file was added or removed, the comparator considers that file as unchanged.
> 
> If newlines do not matter for the delta should then additional ones also don't matter? Can you explain in what case an additional newline is relevant? In the end the user can always bump the version anyways if it is considered a relevant change.

They do matter. At least from my point of view, the new line skip was introduced for text files to deal with the different default line-ending on Windows and Linux and because it is very common to use git's auto-crlf feature where on Windows on checks in Linux-style and checks-out windows style line-endings.
So I think the goal should be to check if there is a content difference or if only the line-endings style changed.
And yes I would consider just a blank new line a content change, even if one for example just appends a new-line at the end of the file.

And considering added/removed newlines is for example relevant if you adjust a doc or license file. Of course one can bump the version, but if that is not necessary I believe nobody thinks about that in advance and will just wonder why the changed doc/license file is not in the build jars, while they are present in the IDE.

laeubi:
>  I believe nobody thinks about that in advance and will just wonder why the changed doc/license file is not in the build jars, while they are present in the IDE.

Maybe because its a strange license that is considered different because of more line endings ;-)

Anyways it seems the testcases do not pass (anymore) with that change on some platforms.

tycho-artifactcomparator/src/main/java/org/eclipse/tycho/artifactcomparator/ArtifactComparator.java
```diff
@@ -19,22 +19,22 @@
 
 public interface ArtifactComparator {
 
-    public static class ComparisonData {
+    public static record ComparisonData(List<String> ignoredPattern, boolean writeDelta, boolean showDiffDetails) {
```
laeubi, line 22: Maybe this change can be a separate PR.

tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/TextComparator.java
```diff
@@ -36,62 +36,59 @@ public class TextComparator implements ContentsComparator {
 
     static final String HINT = "txt";
 
+    private static final char CR = '\r';
+    private static final char LF = '\n';
+
+    // Possible new lines:
+    // \n -- unix style
+    // \r\n -- windows style
+    // \r -- whatever style (ignored?!)
```
laeubi, line 45: this is macos https://stackoverflow.com/questions/38688516/line-break-represented-as-r-on-mac

tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/TextComparator.java
```diff
@@ -36,62 +36,59 @@ public class TextComparator implements ContentsComparator {
...
-            this.bytes = bytes;
-        }
-
-        byte next() throws EOFException {
-            if (hasNext()) {
-                byte b = bytes[index];
-                index++;
-                return b;
+    public static boolean isEqualTextIngoreNewLine(byte[] baseline, byte[] reactor) {
+        int iB = 0;
```
laeubi, line 62: can we use more descriptive names here?

laeubi:
@HannesWell do you plan to work on this and possibly backport to Tycho 4? I currently plan to prepare a bugfix release so so we might want to include it there.

tycho-artifactcomparator/src/main/java/org/eclipse/tycho/artifactcomparator/ArtifactComparator.java
HannesWell: Sure. Here we go: https://github.com/eclipse-tycho/tycho/pull/2610

tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/TextComparator.java
HannesWell: According to [Wikipedia's Mac OS 9 article](https://en.wikipedia.org/wiki/Mac_OS_9), the latest release of Mac OS 9 is 9.2.2 / December 5, 2001, which was 21 years ago. Therefore I think this is not really relevant anymore.

Nevertheless my intention was that this works as if one would use a `BufferedReader` to read the lines one by one using BufferedReader.readLine() and compare the line content, just more efficient (i.e. without creating Strings all the time).
And the javadoc of BufferedReader.readLine() says:
```
Reads a line of text. A line is considered to be terminated by any oneof a line feed ('\n'), a carriage return ('\r'),
a carriage return followed immediately by a line feed, or by reaching the end-of-file(EOF).
```
So if equivalence is the goal, `\r` needs to be considered.

tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/TextComparator.java
HannesWell: Sure.

HannesWell:
> @HannesWell do you plan to work on this and possibly backport to Tycho 4? I currently plan to prepare a bugfix release so so we might want to include it there.

I adjusted the logic so that \r now should be treated as regular line-ending and in the latest run all tests passed unchanged.

https://github.com/eclipse-tycho/tycho/pull/2610 should be merged before this

PR #2622, opened 2023-07-14
Make the produces repository format configurable

laeubi:
This now introduces a new SPI RepositoryGenerator that allows to generate a certain repository format with a generic mojo.


laeubi:
Replaced with https://github.com/eclipse-tycho/tycho/pull/2642

PR #2624, opened 2023-07-17
Generate a manifest for all tycho jars

laeubi:
Currently we sometimes generate a manifest and sometimes not using different settings, this should be unified to simply generate a manifest for any jar we produce in Tycho.


laeubi:
Closing this now in favor of activate this individually...

PR #2626, opened 2023-07-17
Fix resolving of project if target do not contains JUnit

laeubi:
Currently Tycho fails when the JUnit-Container is used but no JUnit is in the target platform. Actually the JUnit Container itself implies already some set of "target content" that should be supplied by Tycho.


laeubi:
Current error shown by the unit test is:

```
[ERROR]   You requested to install 'org.eclipse.equinox.p2.iu; junit-jupiter-api [5.0.0,6.0.0)' but it could not be found
[ERROR] 
[ERROR] See https://wiki.eclipse.org/Tycho/Dependency_Resolution_Troubleshooting for help.

```

laeubi:
Now the test at least passes the resolving phase ... but compile fails as it only loads items from the target but do not resolve the JUnitBundles alternatively as maven artifacts.

Issue #2625, opened 2023-07-17
Could not resolve dependency with tycho > 2.7.5


tivervac:
I'm trying to build a target platform with Tycho 3.0.5, but it fails. Without any config changes, this TP builds fine for using Tycho 2.7.5.

The command I'm using is `./mvnw --update-snapshots -Dtycho.disableP2Mirrors=true clean install -f pom-target-2021-03.xml -Dmaven.repo.local=$(pwd)/.repository -X`

The error I'm getting is (full stacktrace at the bottom)

```
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project com.sigasi.hdt.dependencies.feature: Could not resolve dependencies for project com.sigasi.hdt.dependencies:com.sigasi.hdt.dependencies.feature:eclipse-feature:5.3.0-SNAPSHOT: The following artifacts could not be resolved: org.eclipse.xtext:org.eclipse.xtext.purexbase:jar:2.29.0, org.eclipse.xtext:org.eclipse.xtext.purexbase.ide:jar:2.29.0: Could not find artifact org.eclipse.xtext:org.eclipse.xtext.purexbase:jar:2.29.0 in central (https://repo.maven.apache.org/maven2)
```

True, due to a likely build mistake, `org.eclipse.xtext.purexbase:jar:2.29.0` is [not available](https://repo.maven.apache.org/maven2/org/eclipse/xtext/org.eclipse.xtext.purexbase/) on maven central. 2.28.0 is available and building with that version does work so it seems like Tycho really wants to get the dependency from maven central. However, I don't understand why it wants to do so, my target platform is setup with the following repository

```xml
<repository>
	<id>tmf</id>
	<url>https://download.eclipse.org/modeling/tmf/xtext/updates/releases/2.29.0</url>
	<layout>p2</layout>
</repository>
```

And that repository does contain purexbase as expected

![image](https://github.com/eclipse-tycho/tycho/assets/3389524/e178a88e-7830-40b5-958b-e47ddb8df935)

The reason it's trying to download purexbase (which is never mentioned in our code base) is likely because we include this line in our feature.xml

```
<includes id="org.eclipse.xtext.sdk" version="2.29.0.qualifier" />
```

That transitively includes purexbase: https://github.com/eclipse/xtext/blob/main/org.eclipse.xtext.sdk.feature/feature.xml

Running with `-X` doesn't add much more, except for this odd snippet

```
[...]

PR #2634, opened 2023-07-18
Intregration test for 2625

tivervac:
Tried to boil it down as much as possible. On my machine running `mvn clean verify -Dtycho-version=3.0.5` in `projects-its/issue2625/com.sigasi.hdt.test`

See #2625

### DIFF
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/brokenp2data/BrokenP2DataTest.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/brokenp2data/BrokenP2DataTest.java
@@ -13,7 +13,7 @@ public class BrokenP2DataTest extends AbstractTychoIntegrationTest {
 	public void test() throws Exception {
 		Verifier verifier = getVerifier("brokenp2data");
 
-		verifier.executeGoals(List.of("clean", "verify", "-Dtycho-version=3.0.5"));
+		verifier.executeGoals(List.of("clean", "verify"));
 		verifier.verifyErrorFreeLog();
 	}
 }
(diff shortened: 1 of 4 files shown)


laeubi:
Can you please reference this in a TestClass so the test is actually executed?

tivervac:
@laeubi Done! I've also proofread the Contributing document, mostly because I didn't know where to look for the Java test class

laeubi:
> I've also proofread the Contributing document, mostly because I didn't know where to look for the Java test class

Many thanks, would you mind to open a separate PR for the contribution document, I think this should be merged right away and is independent from the rest.

tivervac:
Sure, see https://github.com/eclipse-tycho/tycho/pull/2639

tivervac:
It was available, not sure why it was failing. I've moved things around, let's see if that works better

laeubi:
@tivervac I'm working on a fix for the problem, but the the build fails:

> The Maven artifactId (currently: "feature") must be the same as the feature ID (currently: "tycho-its-project.maven.brokenp2data.feature")

so I think a little adjustment is required here!

tivervac:
How about this?

tycho-its/src/test/java/org/eclipse/tycho/test/brokenp2data/BrokenP2DataTest.java
```diff
@@ -0,0 +1,19 @@
...
+import org.junit.Test;
+
+// See #2625
+public class BrokenP2DataTest extends AbstractTychoIntegrationTest {
+
+	@Test
+	public void test() throws Exception {
+		Verifier verifier = getVerifier("brokenp2data");
+
+		verifier.executeGoals(List.of("clean", "verify", "-Dtycho-version=3.0.5"));
```
laeubi, line 16:
```suggestion
		verifier.executeGoals(List.of("clean", "verify"));
```

Issue #2666, opened 2023-07-27
SharedHttpCacheStorage doesnt resolve redirect correctly if the uri that is given isn't normalized


jcompagner:
The getCacheEntry function:
https://github.com/eclipse-tycho/tycho/blob/master/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/SharedHttpCacheStorage.java#L92

should normalize the incoming url, so that will not become an extra redirect that the HttpClient of java does (and then 2 redirects must be followed what the code of Tycho doesn't do and then you get a stack overflow because getCachedEntry is constantly called by itself through the redirect code.

```
	public CacheEntry getCacheEntry(URI uriParam, Logger logger) throws FileNotFoundException {
		URI uri = uriParam.normalize();
```

that is the code that fixes it, the param is renamed and the an variable uri is made that is the normalized version of it.
Then the Java HttpClient doesn't treat this as a extra redirect.

laeubi:
@jcompagner can this be closed now or do you still see issues?

jcompagner:
yes this one is fixed 4.0.1 works fine now for our product build

PR #2669, opened 2023-07-28
SharedHttpCacheStorage doesnt resolve redirect correctly if the uri that is given isn't normalized #2666

jcompagner:
Fix for SharedHttpCacheStorage doesnt resolve redirect correctly if the uri that is given isn't normalized #2666 

make sure taht the incoming uri param is normalized when used further
up. This way it is not an extra redirect (so there are 2 redirects to
the final source)

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/SharedHttpCacheStorage.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/SharedHttpCacheStorage.java
@@ -90,14 +90,15 @@ public class SharedHttpCacheStorage implements HttpCache {
      */
 	@Override
 	public CacheEntry getCacheEntry(URI uri, Logger logger) throws FileNotFoundException {
-        CacheLine cacheLine = getCacheLine(uri);
+		URI normalized = uri.normalize();
+		CacheLine cacheLine = getCacheLine(normalized);
 		if (!cacheConfig.isUpdate()) { // if not updates are forced ...
             int code = cacheLine.getResponseCode();
             if (code == HttpURLConnection.HTTP_NOT_FOUND) {
-                throw new FileNotFoundException(uri.toASCIIString());
+				throw new FileNotFoundException(normalized.toASCIIString());
             }
             if (code == HttpURLConnection.HTTP_MOVED_PERM) {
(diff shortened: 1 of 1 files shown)


p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/SharedHttpCacheStorage.java
```diff
@@ -89,7 +89,8 @@ protected boolean removeEldestEntry(final Map.Entry<File, CacheLine> eldest) {
      *             if the URI is know to be not found
      */
 	@Override
-	public CacheEntry getCacheEntry(URI uri, Logger logger) throws FileNotFoundException {
+	public CacheEntry getCacheEntry(URI uriParam, Logger logger) throws FileNotFoundException {
+		URI uri = uriParam.normalize();
```
laeubi, line 93:
```suggestion
		URI normalized = uriParam.normalize();
```

laeubi (changes requested):
Can you please rebase your change on master?

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/SharedHttpCacheStorage.java
jcompagner: Do you want that rename? because then we need a few more changes (everywhere that "uri" is used)

jcompagner: i pushed a rename

laeubi: Yes even if that requires more changes it should make the code more clear what URI is actually used here!

laeubi (approved):
Looks good

laeubi:
@jcompagner sadly Github seem to be confused now:

"This branch cannot be rebased due to too many changes", do you think you can create a new branch from master (maybe 

so we need to wait until github issue is resolved see
- https://github.com/orgs/community/discussions/51612#discussioncomment-6554628

jcompagner:
https://github.com/eclipse-tycho/tycho/pull/2672

Issue #2710, opened 2023-08-13
Still "No digest algorithm is available to verify download of " during assemble-repository


dhendriks:
Follow-up of #1482.

The issue was reduced, but not completely solved. We still get several such warnings in Tycho 4.0.1:

```
[2023-08-11T08:49:37.917Z] [INFO] --- tycho-p2-repository:4.0.1:assemble-repository (default-assemble-repository) @ org.eclipse.escet.product ---
[2023-08-11T08:49:37.917Z] [WARNING] Parameter 'qualifier' (user property 'buildQualifier') is read-only, must not be used in configuration
[2023-08-11T08:50:10.010Z] [WARNING] [349e7efd-6b6e-442a-abc7-000cf4144b2c][extension>org.eclipse.tycho:tycho-maven-plugin:4.0.1] No digest algorithm is available to verify download of osgi.bundle,org.apache.commons.commons-text.source,1.10.0.
[2023-08-11T08:50:10.010Z] [WARNING] [349e7efd-6b6e-442a-abc7-000cf4144b2c][extension>org.eclipse.tycho:tycho-maven-plugin:4.0.1] No digest algorithm is available to verify download of binary,org.eclipse.escet.product.headless.feature_root.gtk.linux.x86_64,1.0.0.v20230811-083501-dev.
[2023-08-11T08:50:10.010Z] [WARNING] [349e7efd-6b6e-442a-abc7-000cf4144b2c][extension>org.eclipse.tycho:tycho-maven-plugin:4.0.1] No digest algorithm is available to verify download of osgi.bundle,com.github.com-github-javabdd.com.github.javabdd.source,6.0.0.
[2023-08-11T08:50:10.010Z] [WARNING] [349e7efd-6b6e-442a-abc7-000cf4144b2c][extension>org.eclipse.tycho:tycho-maven-plugin:4.0.1] No digest algorithm is available to verify download of binary,org.eclipse.escet.product.feature_root,1.0.0.v20230811-083501-dev.
[2023-08-11T08:50:10.010Z] [WARNING] [349e7efd-6b6e-442a-abc7-000cf4144b2c][extension>org.eclipse.tycho:tycho-maven-plugin:4.0.1] No digest algorithm is available to verify download of binary,org.eclipse.escet.product.executable.gtk.linux.x86_64,1.0.0.v20230811-083501-dev.
[...]

PR #2711, opened 2023-08-13
tycho-its reproducer p2Repository.downloadVerifyNoDigestAlgo (#2710)

dhendriks:
My first merge request to contribute a tycho-its. Hope it works.

Addresses #2710

### DIFF
--- a/tycho-api/src/main/java/org/eclipse/tycho/TychoConstants.java
+++ b/tycho-api/src/main/java/org/eclipse/tycho/TychoConstants.java
@@ -18,7 +18,7 @@ import java.util.regex.Pattern;
 
 public interface TychoConstants {
 
-    public static final String ECLIPSE_LATEST = "https://download.eclipse.org/releases/2023-09/";
+    public static final String ECLIPSE_LATEST = "https://download.eclipse.org/releases/2023-12/";
 
     static final String ANY_QUALIFIER = "qualifier";
 
--- a/tycho-extras/tycho-document-bundle-plugin/src/main/java/org/eclipse/tycho/extras/docbundle/runner/ConvertSchemaToHtmlRunner.java
+++ b/tycho-extras/tycho-document-bundle-plugin/src/main/java/org/eclipse/tycho/extras/docbundle/runner/ConvertSchemaToHtmlRunner.java
@@ -41,12 +41,15 @@ import org.eclipse.pde.internal.core.PDECoreMessages;
 import org.eclipse.pde.internal.core.XMLDefaultHandler;
 import org.eclipse.pde.internal.core.builders.SchemaTransformer;
 import org.eclipse.pde.internal.core.ischema.ISchema;
+import org.eclipse.pde.internal.core.ischema.ISchemaDescriptor;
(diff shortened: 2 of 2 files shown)


b/tycho-its/src/test/java/org/eclipse/tycho/test/p2Repository/DownloadVerifyNoDigestAlgo.java
```diff
@@ -0,0 +1,32 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+import org.apache.maven.it.Verifier;
+import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
+import org.junit.Assert;
+
+public class DownloadVerifyNoDigestAlgo extends AbstractTychoIntegrationTest throws IOException {
```
laeubi, line 22:
```suggestion
public class DownloadVerifyNoDigestAlgoTest extends AbstractTychoIntegrationTest throws IOException {

```

dhendriks:
For some reason my new test is not picked up. Any idea why?

dhendriks:
I think I found the problem. Let's see whether the next build works.

tycho-its/src/test/java/org/eclipse/tycho/test/p2Repository/DownloadVerifyNoDigestAlgoTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class DownloadVerifyNoDigestAlgoTest extends AbstractTychoIntegrationTest {
+    @Test
+    public void test() throws Exception {
+        Verifier verifier = getVerifier("p2Repository.downloadVerifyNoDigestAlgo", false);
+        verifier.setLogFileName("p2Repository.downloadVerifyNoDigestAlgo.log");
+        verifier.executeGoal("verify");
+        verifier.verifyErrorFreeLog();
+        String logContents = Files.readString(Paths.get("p2Repository.downloadVerifyNoDigestAlgo.log"));
+        assertFalse(logContents.contains("No digest algorithm is available to verify download"));
```
laeubi, line 31: `AbstractTychoIntegrationTest` already contains some helper methods to check if log messages are (not) there.

laeubi:
> The new integration test thus shows that Tycho currently gives the warnings that we don't want. What is the next step?

Please squash/rebase your change so it could be uses as a basis for a possible fix. Do you maybe want to give it a try reducing the message for some of the items? At least for feature/products I think it could be a similar fix as applied for plugins already. For m2e target items it should be a bit different and reusing the maven checksums.

dhendriks:
> Please squash/rebase your change so it could be uses as a basis for a possible fix.

Done.

> Do you maybe want to give it a try reducing the message for some of the items?

It already took me way more time than expected to report the issue, make a minimal reproducer, and get that in the right form as a tycho-its (4+ hours). At the moment I don't have the time to actually go fixing this. Especially since I know absolutely nothing about the internals of Tycho, so it would be a huge effort to figure out where it is, what is going on, and how to address that. I was kind of hoping some Tycho developer could look at it?

PR #2714, opened 2023-08-19
Make sure the last modified timestamp is set correctly

laeubi:
Linux stores modification time with nano-seconds precision, but we currently set the timestamp of the cached file only in milliseconds. This can lead to the situation that the source is considered a few nanoseconds older than the cache file and therefore the cache is regenerated every time.


HannesWell (approved):
LGTM

Issue #2716, opened 2023-08-20
"'build.plugins.plugin.version' for org.codehaus.mojo:build-helper-maven-plugin is missing" in case of multiple source directories


dhendriks:
If I have multiple source directories in my plugin, I get the following warning:

```
$ mvn clean verify -Dtycho-version=4.0.1
Picked up _JAVA_OPTIONS: -Djava.net.preferIPv4Stack=true
[INFO] Scanning for projects...
[WARNING]
[WARNING] Some problems were encountered while building the effective model for tycho-its-project.compliler:tycho-its-project.compliler.multipleSourceDirs.bundle:eclipse-plugin:1.0.0-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for org.codehaus.mojo:build-helper-maven-plugin is missing.
[WARNING]
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING]
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING]
[INFO] Tycho Version:  4.0.1 (26e55402ca6295ab784674c5053c364ae7b428e7)
[INFO] Tycho Mode:     extension
[INFO] Tycho Builder:  maven
...
```

That is, I have a `src` and `src-gen` folder in my bundle. These folders are configured in `.classpath` and `build.properties`. The bundle has no `pom.xml`, as I use pomless build.

PR #2718, opened 2023-08-20
Tycho-its reproducer for issue 2716.

dhendriks:
See
- https://github.com/eclipse-tycho/tycho/issues/2716
- https://github.com/eclipse-tycho/tycho/discussions/2717


dhendriks:
The build fails, as expected, since the reproducer is meant to show the issue.

laeubi:
@dhendriks as mentioned in the discussion this can be fixed by adding a managed plugin the the main pom.

PR #2741, opened 2023-09-01
Consider Source-Unit version when adding sources to a P2-repo

HannesWell:


laeubi (approved):
Looks good any chance we can have an itest to verify this works (and will work in the future)?

HannesWell:
> Looks good any chance we can have an itest to verify this works (and will work in the future)?

Done. Extended the existing `IncludeAllSourcesTest.testSourceInclusion()` test case to also cover sources of artifacts with multiple versions.

PR #2742, opened 2023-09-01
Add an option to prevent repo-references from being added to a p2-repo

HannesWell:
Add a new parameter named 'excludedRepositoryReferences' to the AssembleRepositoryMojo to prevent (mainly the automatic) addition of repository references with matching locations.
This can be used with ANT-style or Java RegEx patterns as follows:
```
<excludedRepositoryReferences>
	<excluded>https://foo.bar.org/component1/**</excluded>
	<excluded>%regex[http(s)?:\/\/bar\.org\/hidden.*]</excluded>
</excludedRepositoryReferences>
```

The is especially convenient in combination with the automatic addition of IU Target-Repository or POM-Repository references, if some but not all repos should be added.

One use-case of this is for example https://github.com/eclipse-m2e/m2e-core/pull/1222.

### DIFF
--- a/tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
+++ b/tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
@@ -19,6 +19,8 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import org.apache.maven.model.Repository;
@@ -29,6 +31,7 @@ import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.codehaus.plexus.util.FileUtils;
+import org.codehaus.plexus.util.MatchPatterns;
 import org.eclipse.tycho.PackagingType;
 import org.eclipse.tycho.ReactorProject;
 import org.eclipse.tycho.TychoConstants;
@@ -73,6 +76,12 @@ import aQute.bnd.repository.fileset.FileSetRepository;
  */
 @Mojo(name = "assemble-repository", defaultPhase = LifecyclePhase.PACKAGE, threadSafe = true)
 public class AssembleRepositoryMojo extends AbstractRepositoryMojo {
+
+    public static class RepositoryReferenceFilter {
+        List<String> exclude;
+        List<String> include;
+    }
+
     private static final Object LOCK = new Object();
     /**
      * <p>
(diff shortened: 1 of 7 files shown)


tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
```diff
@@ -217,6 +219,26 @@ public class AssembleRepositoryMojo extends AbstractRepositoryMojo {
...
+     * </pre>
+     * 
+     * The first pattern uses an {@code ANT} pattern and the second pattern uses a
+     * {@code Java RegEx} {@link Pattern} (enclosed in {@code %regex[<the-regex-pattern>]} to
+     * prevent matching locations references from being added as references to the assembled
+     * repository.
+     * </p>
+     */
+    @Parameter
+    private List<String> excludedRepositoryReferences = List.of();
```
laeubi, line 240: we should name the parameter something like `<filterReferences>` (just in case someone later wants to add support for `<includes>` or other fancy stuff ...)

laeubi (commented):
Looks good any chance we can have an itest to verify this works (and will work in the future)?

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
HannesWell: > we should name the parameter something like `<filterReferences>` (just in case someone later wants to add support for `<includes>` or other fancy stuff ...)

Can you please explain in more detail about what potential methods you are thinking?
Do you mean that one can add positive and negative filters? 

For example maven-surefire also has [excludes](https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#excludes) and [includes](https://maven.apache.org/surefire/maven-surefire-plugin/test-mojo.html#includes) as dedicated parameters.
If we later support inclusion filters adding a `includedRepositoryReferences` paramter is probably simpler to use than having a `filterReferences` parameter with complex objects that have a extra inclusion/exclusion sub-parameter.

HannesWell:
> Looks good any chance we can have an itest to verify this works (and will work in the future)?

Will try that.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
laeubi: yes we should avoid adding more and more dedicated parameters if the belong together.

HannesWell: OK. Did that now. And since you mentioned it I also added support for inclusions.
It is now a bit more verbose, but I don't know if/how it could be done more compact:
```
<repositoryReferenceFilters>
	<filter>
		<type>exclusion</type>
		<locationPattern>https://foo.bar.org/hidden/**</locationPattern>
		<patternType>ant</patternType><!-- If not specified, defaults to 'ant'-->
	</filter>
	<filter>
		<type>inclusion</type>
		<locationPattern>http(s)?:\/\/bar\.org\/.*</locationPattern>
		<patternType>regex</patternType>
	</filter>
</repositoryReferenceFilters>
```

laeubi: I don't think it has to be that complicated:

1. we don't need different pattern types, (if at all) a simple prefix matching should be enough
2. `repositoryReferenceFilters` could simply has one list of excludes and one of includes like this:

```
<repositoryReferenceFilters>
	<exclude>https://foo.bar.org/hidden/**</exclude>
	<include>https://foo.bar.org/hidden/butthis</include>
</repositoryReferenceFilters>
```

If you think its to complicated to define as a POJO you can use `PlexusConfiguration` as the parameter type and then get simply the user specified XML there.

HannesWell: > I don't think it has to be that complicated:
> 
> 1. we don't need different pattern types, (if at all) a simple prefix matching should be enough

I suggest we simply use the distinction by prefix build into plexus.util.MatchPatterns.
Actually it be more work to forbid it.

> 2. `repositoryReferenceFilters` could simply has one list of excludes and one of includes like this:
> 
> ```
> <repositoryReferenceFilters>
> 	<exclude>https://foo.bar.org/hidden/**</exclude>
> 	<include>https://foo.bar.org/hidden/butthis</include>
> </repositoryReferenceFilters>
> ```

Actually I wanted it that simple but didn't knew/expect that this would work.
I changed it now accordingly and also adapted the example to show the usage for a single and multi element list.

HannesWell:
> Looks good any chance we can have an itest to verify this works (and will work in the future)?

Added a test-case and simplified the existing reference test to use the P2Repo Xpath Tools.

HannesWell:
We could consider to just filter those repositories that are added automatically. If one adds a repo-reference in the `category.xml` explicitly, I find it hard to think of a use-case where one should later want to have it filtered out, since they have no other purpose than being added to the repo.

laeubi:
> We could consider to just filter those repositories that are added automatically. 

Sounds reasonable.

laeubi (approved):
Looks good

HannesWell:
> > We could consider to just filter those repositories that are added automatically.
> 
> Sounds reasonable.

Changed it that way now and also removed the plural s from filters since I think it is one filter that has different parameters/components.

HannesWell:
Platform Aggregator build failure is likely due to the incomplete transistion to Eclipse 4.30, thus merging.

PR #2744, opened 2023-09-03
Suppress addition of unnecessary repo-refereces when assembling p2-repos

HannesWell:
The main goal of this PR is to prevent the addition of repository references that don't provide any relevant artifact or are full duplicates (relating the relevant artifacts) to other repositories and thus reducing the number of references which speeds up querying the repo since less other repositories have to be queried.

I think this becomes more relevant since one can now add repository references automatically and for example not all repos referenced in the TP are actually relevant for a p2-repository build for a project.
While with one can explicitly filter references with https://github.com/eclipse-tycho/tycho/pull/2742, the user should not need to carry the burden to check manually which repos are relevant for the assembled repository. Instead Tycho should be smart enough to check that (and with this PR it can).

In detail  his PR consists of three commits
1. Minor code enhancements in p2-repository assembling code (will move that to a separate PR once settled)
2. Suppress addition of unnecessary repo-refereces when assembling p2-repos
3. Print added repository-references as info, when assembly a p2-repository

Adding some information about added repo-references in the log seems to be reasonable since it is now not obvious anymore with auto-addition, explicit and implicit filtering which references actually end up in the repo.

Similar to https://github.com/eclipse-tycho/tycho/pull/2742#issuecomment-1704421421 we could consider to activate this auto-filtering only for those references that are added automatically, because one might have added an effectively empty repo on purpose.
The logging could also only activated if automatic addition is enabled.

Btw. for M2E this reduces the list of added references from 9 to 6:
```
https://download.eclipse.org/eclipse/updates/4.28/
https://download.eclipse.org/egit/updates-6.6/
https://download.eclipse.org/modeling/emf/emf/builds/release/2.34.0
https://download.eclipse.org/webtools/downloads/drops/R3.28.0/R-3.28.0-20221120050827/repository/
https://download.eclipse.org/wildwebdeveloper/releases/1.3.0/
https://download.eclipse.org/tm4e/releases/0.8.1/
https://download.eclipse.org/lsp4e/releases/0.23.0/
https://download.eclipse.org/lsp4j/updates/releases/0.21.0/
https://download.eclipse.org/cbi/updates/license/
```
to
```
https://download.eclipse.org/lsp4j/updates/releases/0.21.0/
https://download.eclipse.org/eclipse/updates/4.28/
https://download.eclipse.org/webtools/downloads/drops/R3.28.0/R-3.28.0-20221120050827/repository/
https://download.eclipse.org/wildwebdeveloper/releases/1.3.0/
https://download.eclipse.org/modeling/emf/emf/builds/release/2.34.0
https://download.eclipse.org/tm4e/releases/0.8.1/
```

Currently this includes work-arounds for the following pending P2 PRs that are not available before December.
- https://github.com/eclipse-equinox/p2/pull/311
- https://github.com/eclipse-equinox/p2/pull/312
- https://github.com/eclipse-equinox/p2/pull/314

### DIFF
--- a/tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
+++ b/tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
@@ -21,7 +21,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
@@ -75,8 +74,17 @@ import aQute.bnd.repository.fileset.FileSetRepository;
 public class AssembleRepositoryMojo extends AbstractRepositoryMojo {
 
     public static class RepositoryReferenceFilter {
+        /**
+         * If {@link #filterProvided} is {@code true} and repository references are added
+         * automatically via {@link #addIUTargetRepositoryReferences} or
+         * {@link #addPomRepositoryReferences}, then this property controls if from the
+         * automatically added ones only references to those repositories are added, that provide
+         * relevant content, which is not provided by any other referenced repositories. If this is
+         * set to {@code false} all automatically added references are added as they are available.
+         */
+        public boolean addOnlyProviding = true;
         /** The list of location patterns that exclude matching repository references. */
-        List<String> exclude = List.of();
+        public List<String> exclude = List.of();
     }
 
     private static final Object LOCK = new Object();
@@ -224,20 +232,32 @@ public class AssembleRepositoryMojo extends AbstractRepositoryMojo {
     private boolean addIUTargetRepositoryReferences;
 
     /**
-     * A list of patterns to exclude automatically derived repository references from being added to
-     * the assembled repository.
+     * Filters to exclude automatically derived repository references from being added to the
+     * assembled repository.
+     * 
      * <p>
+     * Repository references can be filtered based on their location URI using a list of exclusion
+     * pattern:<br>
      * The location of a reference must not be matched by any pattern, in order to be eventually
-     * added to the assembled repository. An arbitrary number of patterns can be specified.<br>
-     * The specified filters are only applied to those repository references derived from the
+     * added to the assembled repository. An arbitrary number of patterns can be specified.
+     * </p>
+     * <p>
+     * If the sub-property {@code addOnlyProviding} is set to {@code true} (the default), references
+     * to repositories that don't provide any relevant unit are excluded from being added to the
+     * assembled repository.
+     * </p>
+     * <p>
+     * All those filters are only applied to those repository references derived from the
      * target-definition or pom file, when {@link #addIUTargetRepositoryReferences} respectively
-     * {@link #addPomRepositoryReferences} is set to {@code true}.
+     * {@link #addPomRepositoryReferences} is set {@code true}. References explicitly listed in the
+     * repository file ({@code category.xml}) are always added.
      * </p>
      * <p>
      * Configuration example 1
      * 
      * <pre>
      * &lt;repositoryReferenceFilter&gt;
+     *   &lt;addOnlyProviding&gt;true&lt;/addOnlyProviding&gt;
      *   &lt;exclude&gt;https://foo.bar.org/hidden/**&lt;/exclude&gt;
      * &lt;/repositoryReferenceFilter&gt;
      * </pre>
@@ -246,6 +266,7 @@ public class AssembleRepositoryMojo extends AbstractRepositoryMojo {
      * 
      * <pre>
      * &lt;repositoryReferenceFilter&gt;
+     *   &lt;addOnlyProviding&gt;false&lt;/addOnlyProviding&gt;
      *   &lt;exclude&gt;
      *     &lt;location&gt;https://foo.bar.org/hidden/**&lt;/location&gt;
      *     &lt;location&gt;%regex[http(s)?:\/\/foo\.bar\.org\/secret\/.*]&lt;/location&gt;
@@ -259,7 +280,8 @@ public class AssembleRepositoryMojo extends AbstractRepositoryMojo {
      * {@code %regex[<the-regex-pattern>]}). <br>
      * The third pattern is a negated (enclosed in {@code ![<the-negated-pattern>]}), which
      * effectively makes it an <em>inclusion</em> pattern that all references must match in order to
-     * be added.
+     * be added. Unlike in the first example, in the second example all references that pass the
+     * location filter are added, regardless of if the provide any unit or not.
      * </p>
      */
     @Parameter
@@ -312,14 +334,15 @@ public class AssembleRepositoryMojo extends AbstractRepositoryMojo {
                         .map(Category::getRepositoryReferences)//
                         .flatMap(List::stream)//
                         .map(ref -> new RepositoryReference(ref.getName(), ref.getLocation(), ref.isEnabled()))//
-                        .collect(Collectors.toCollection(ArrayList::new));
+                        .toList();
                 Predicate<String> autoReferencesFilter = buildRepositoryReferenceLocationFilter();
+                List<RepositoryReference> autoRepositoryRefeferences = new ArrayList<>();
                 if (addPomRepositoryReferences) {
                     getProject().getRepositories().stream() //
                             .filter(pomRepo -> "p2".equals(pomRepo.getLayout()))
                             .filter(pomRepo -> autoReferencesFilter.test(pomRepo.getUrl()))
                             .map(pomRepo -> new RepositoryReference(pomRepo.getName(), pomRepo.getUrl(), true))
-                            .forEach(repositoryReferences::add);
+                            .forEach(autoRepositoryRefeferences::add);
                 }
                 if (addIUTargetRepositoryReferences) {
                     projectManager.getTargetPlatformConfiguration(getProject()).getTargets().stream()
@@ -328,14 +351,15 @@ public class AssembleRepositoryMojo extends AbstractRepositoryMojo {
                             .flatMap(iu -> iu.getRepositories().stream())
                             .filter(iuRepo -> autoReferencesFilter.test(iuRepo.getLocation()))
                             .map(iuRepo -> new RepositoryReference(null, iuRepo.getLocation(), true))
-                            .forEach(repositoryReferences::add);
+                            .forEach(autoRepositoryRefeferences::add);
                 }
                 DestinationRepositoryDescriptor destinationRepoDescriptor = new DestinationRepositoryDescriptor(
                         destination, repositoryName, compress, xzCompress, keepNonXzIndexFiles,
-                        !createArtifactRepository, true, extraArtifactRepositoryProperties, repositoryReferences);
+                        !createArtifactRepository, true, extraArtifactRepositoryProperties, repositoryReferences,
+                        autoRepositoryRefeferences);
                 mirrorApp.mirrorReactor(sources, destinationRepoDescriptor, projectSeeds, getBuildContext(),
                         includeAllDependencies, includeAllSources, includeRequiredPlugins, includeRequiredFeatures,
-                        filterProvided, profileProperties);
+                        filterProvided, repositoryReferenceFilter.addOnlyProviding, profileProperties);
                 if (generateOSGiRepository) {
                     XMLResourceGenerator resourceGenerator = new XMLResourceGenerator();
                     resourceGenerator.name(repositoryName);
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -28,8 +28,11 @@ Repositories can contain references to other repositories (e.g. to find addition
 ```
 ### new option to filter added repository-references when assembling a p2-repository
 
-The repository references automatically added to a assembled p2-repository (via `tycho-p2-repository-plugin`'s `addIUTargetRepositoryReferences` or `addPomRepositoryReferences`) 
-can now be filtered by their location using exclusion and inclusion patterns and therefore allows more fine-grained control which references are added.
+If filtering provided artifacts is enabled, the repository references automatically added to a assembled p2-repository
+(via `tycho-p2-repository-plugin`'s `addIUTargetRepositoryReferences` or `addPomRepositoryReferences`) can now be filtered by their location
+using exclusion and inclusion patterns and therefore allows more fine-grained control which references are added.
+Additionally the automatically added references can be filter based on if they provide any of the filtered units or not.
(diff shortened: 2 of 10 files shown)


tycho-core/src/main/java/org/eclipse/tycho/core/TargetPlatformConfiguration.java
```diff
@@ -136,11 +135,8 @@ public String getTargetPlatformResolver() {
     }
 
     public synchronized List<TargetDefinitionFile> getTargets() {
-        for (Iterator<Supplier<File>> iterator = lazyTargetFiles.iterator(); iterator.hasNext();) {
-            Supplier<File> supplier = iterator.next();
-            targets.add(supplier.get().toURI());
-            iterator.remove();
-        }
+        lazyTargetFiles.forEach(supplier -> targets.add(supplier.get().toURI()));
```
laeubi, line 138: This is not equivalent, assume loading one lazy target fails (and the method is called multiple times) this will add target definitions multiple times, while before the lazy one is removed as soon as it was successfully queried.

laeubi (commented):
I think this must then be configurable (at best with the ongoing `<filter>` configuration) because we can't decide on the rationale if/how a user wants to add additional repositories, even though PDE currently not support this fully (see https://github.com/eclipse-pde/eclipse.pde/issues/720 ) Tycho does so "unnecessary" depends on the usecase, and if the updatesites are really not required one probably should never add them to the target platform anyways ...

laeubi:
What do you think about having

```
<repositoryReferenceFilters>
    ...
   <strictlyRequired>true/false (default)</strictlyRequired>
</repositoryReferenceFilters>
```

that do this filtering with `strictlyRequired=true`?

tycho-core/src/main/java/org/eclipse/tycho/core/TargetPlatformConfiguration.java
HannesWell: Indeed, makes sense.
I thought about using the more modern `removeIf()`, but unfortunately the `removeIf()` method in the default implementation in Collection and in `ArrayList` behave differently in case an exception is thrown (i.e. the former has removed all matches before and the latter just does nothing effectively since the actual removal is postponed to the end).
Thus I'll just keep it as it is.

HannesWell:
> What do you think about having
> 
> ```
> <repositoryReferenceFilters>
>     ...
>    <strictlyRequired>true/false (default)</strictlyRequired>
> </repositoryReferenceFilters>
> ```
> 
> that do this filtering with `strictlyRequired=true`?

The schema sounds good and I think this fits well to https://github.com/eclipse-tycho/tycho/pull/2742.
But I think the name should be more expressive, e.g. `excludeNotStrictlyRequired`?
I assume that in most cases `excludeNotStrictlyRequired=true` is what users want and therefore would suggest to make that the default.

> and if the updatesites are really not required one probably should never add them to the target platform anyways ...

Sometimes yes, but I think there are scenarios where it is not so obvious. For example in the case of m2e (and probably many other eclipse projects) the license repo is not required since the license is inlined during the build. Or one could have a p2-repo from which test dependencies are fetched. That should probably then also not be added to a 'product' code repo.

HannesWell:
Furthermore I think that instead of pruning the references based on the content of the artifact repository, it actually needs to be done based on the metadata repository. Otherwise the content provided by transitive references is not considered.

Which in general brings me to another idea I had:
Wouldn't it also be reasonable variant to keep the metadata (potentially with the full closure depending on the plugin-config) as it is and only remove the artifacts and reference the artifact repositories (and not the corresponding metadata repo).
I would assume that this would speed up the resolution process since then probably less IUs are available,
the closure of the assembled repo is a sub-set of the closure of all references and probably in most cases even a real sub-set, and less metadata repos would have to be queried.

laeubi:
> But I think the name should be more expressive, e.g. excludeNotStrictlyRequired?

better add an exhaustive documentation to the parameter that trying to express to much with the name

> I assume that in most cases excludeNotStrictlyRequired=true is what users want and therefore would suggest to make that the default.

Its hard to guess what "most users" want, but this can't be the default for Tycho 4 at least and for 5 then needs a documentation in the migration guide.

> For example in the case of m2e (and probably many other eclipse projects) the license repo is not required since the license is inlined during the build.

This is a very special case and one can either filter that, or simply put the license repo in the pom and exclude pom repositories in the first place.

> Furthermore I think that instead of pruning the references based on the content of the artifact repository, it actually needs to be done based on the metadata repository. 

Yes the metadata is the "key" in fact not every requirement results in an artifact.

> Wouldn't it also be reasonable variant to keep the metadata (potentially with the full closure depending on the plugin-config) as it is and only remove the artifacts and reference the artifact repositories (and not the corresponding metadata repo).

You can add a new option `<includeMetadata>` but be aware that if one references a snapshot/latest repository the metadata might change or is probably not valid anymore then.

HannesWell:
> > But I think the name should be more expressive, e.g. excludeNotStrictlyRequired?
> 
> better add an exhaustive documentation to the parameter that trying to express to much with the name

I think it is best to have both. An expressive name and a good description.

And since we have the `exclude` filter for locations `excludeNotStrictlyRequired` fits IMHO quite well.

> 
> > I assume that in most cases excludeNotStrictlyRequired=true is what users want and therefore would suggest to make that the default.
> 
> Its hard to guess what "most users" want, but this can't be the default for Tycho 4 at least and for 5 then needs a documentation in the migration guide.

Yes that's right. Sounds fair. Lets do it that way.

> > Wouldn't it also be reasonable variant to keep the metadata (potentially with the full closure depending on the plugin-config) as it is and only remove the artifacts and reference the artifact repositories (and not the corresponding metadata repo).
> 
> You can add a new option `<includeMetadata>` but be aware that if one references a snapshot/latest repository the metadata might change or is probably not valid anymore then.

Yes that's right. But I think this is a general 'issue' of `filterProvided` respectively relying on references. You should only do that with stable repositories.

laeubi:
> An expressive name and a good description.

Longer names are not necessarily more expressive (but harder to remember / type), e.g. many mojos have a `skip` flag even though in many cases it could have been `skipTest`, `skipCompile`, `skipExecutionOfMyMojo`, ...

Especially for boolean "flags" using the negation is often confusing, e.g. `excludeNotStrictlyRequired` is unclear to me what it means (without reading the docs) if I set it to `true` because it contains two "negations", `exclude` and `not` so does this means a double negation so I better should choose `false` here?!? Also for me filtering always implies exclusion already (we are inside a `<repositoryReferenceFilters>` tag) , so I would assume an exclusion for any option here by default (the opposite of filtering would be selection as in SQL where you first `SELECT` some set and later on filter it with `WHERE`)

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
```diff
@@ -78,8 +74,19 @@
...
+         * the assembled repository.
+         */
+        boolean excludeNotStrictlyRequired = true;
+        /** The list of location patterns that exclude matching repository references. */
+        List<String> exclude = List.of();
+        /**
+         * The list of location patterns that includes only matching repository references (if this
+         * list is not empty).
+         */
+        List<String> include = List.of();
```
laeubi, line 89: I think this must be made more clear, what happens if I exclude `A*` but include `A/B/*` will `A/B/C` contained in the repository? Or can/should I always only choose either `exclude` or `include`?

laeubi: Maybe `<matches>` instead of exclude would be better (the user already knows we are in a filtering expression so exclusion is implicit) would be better

HannesWell: > I think this must be made more clear, what happens if I exclude `A*` but include `A/B/*` will `A/B/C` contained in the repository? Or can/should I always only choose either `exclude` or `include`?

You can choose both.
I have described that in the doc of the `repositoryReferenceFilter` parameter.
The inclusions are applied first, if any are specified, which means among all available references it is first checked which match the inclusion filter and all others are discarded (if no inclusion is specified just all available refs are considered for the next step). On the remaining references in step two all reference that match an exclusion are discarded too.
So basically the inclusion is just an negated exclusion.

Therefore in the case you described A/B/C will not be contained in the assembled repo.

HannesWell:
> > An expressive name and a good description.
> 
> Longer names are not necessarily more expressive (but harder to remember / type), e.g. many mojos have a `skip` flag even though in many cases it could have been `skipTest`, `skipCompile`, `skipExecutionOfMyMojo`, ...

It is of course reasonable to strip of the context, but I didn't suggest `excludeNotStrictlyRequiredRepositoryReferencesWhenAssemblingARepository` :P

> 
> Especially for boolean "flags" using the negation is often confusing, e.g. `excludeNotStrictlyRequired` is unclear to me what it means (without reading the docs) if I set it to `true` because it contains two "negations", `exclude` and `not` so does this means a double negation so I better should choose `false` here?!? Also for me filtering always implies exclusion already (we are inside a `<repositoryReferenceFilters>` tag) , so I would assume an exclusion for any option here by default (the opposite of filtering would be selection as in SQL where you first `SELECT` some set and later on filter it with `WHERE`)

That's a points. I choose exclude as it fits the underlying process since the references are first added and then removed. But of course this does not have to be reflected here.
So maybe the opposite `includeOnlyStrictlyRequired`?

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
laeubi: > The inclusions are applied first, if any are specified, which means among all available references it is first checked which match the inclusion filter and all others are discarded (if no inclusion is specified just all available refs are considered for the next step). On the remaining references in step two all reference that match an exclusion are discarded too.

Hm... I would expect it the other way round as by default everything is included already, so maybe `matches` is a better wording here?

HannesWell: > Hm... I would expect it the other way round as by default everything is included already, so maybe `matches` is a better wording here?

It probably quite personal what one expects. On the other hand it would be somehow strange as well to eventually include things that matched a exclusion?

But since inclusion and exclusion basically just have an inverted sign, what about adding a syntax to the pattern to invert it and just have exclusions/excluding filter? E.g. similar to the style `%regex[<pattern>]` we could for example use `![<pattern>]` to negate a pattern.
And instead of 
```
<repositoryReferenceFilter>
    <exclude>https://foo.bar/hidden</exclude>
    <include>https://foo.bar</include>
<repositoryReferenceFilter>
```
we would only have something like
```
<repositoryReferenceFilter>
    <exclude>https://foo.bar/hidden</exclude>
    <exclude>![https://foo.bar]</exclude>
<repositoryReferenceFilter>
```
Plus the parameter for the new automated removal of unused references.

Btw. regarding expecting exclusion during filtering: E.g. in the Java-Stream API the `filter()` method expects a Predicate which all elements must match in order to pass the filter.
So I think it is not so clear that filter means always exclusion (although I would also tend more towards exclusion).

laeubi: > It probably quite personal what one expects. On the other hand it would be somehow strange as well to eventually include things that matched a exclusion?

The problem often is I want to block everything except one, for example in `.gitignore` it could also be quite challenging:
https://stackoverflow.com/questions/5533050/gitignore-exclude-folder-but-include-specific-subfolder

> Btw. regarding expecting exclusion during filtering: E.g. in the Java-Stream API the filter() method expects a Predicate which all elements must match in order to pass the filter.

Yes that's very confusing and hit me several times already :-\

laeubi:
@HannesWell would be good if we can finish this I want to perform a new release end of month so this is included then.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
HannesWell: I think this is now solved with https://github.com/eclipse-tycho/tycho/pull/2849.

HannesWell:
This should now be almost ready. The only missing thing are the adjustment of the release notes.

I now change the name of the property to `addOnlyProviding`, a full configuration looks like:
```
<plugin>
	<configuration>
		<repositoryReferenceFilters>
			<addOnlyProviding>true</addOnlyProviding>
		</repositoryReferenceFilters>
	</configuration>
</plugin>
```

tycho-core/pom.xml
```diff
@@ -258,13 +258,16 @@
...
 		<dependency>
 			<groupId>commons-codec</groupId>
 			<artifactId>commons-codec</artifactId>
 			<version>1.16.0</version>
 		</dependency>
-
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-lang3</artifactId>
+			<version>3.12.0</version>
```
HannesWell, line 269: For some reason this was nessary to avoid classloading Problems at runtime. But actually this is provided by Maven itself (and marked as provided Dependency of maven-core).

HannesWell: Replaced the usage of StringUtils with standard java.

HannesWell:
This is now actually complete.
But at my final local testing with https://github.com/eclipse-m2e/m2e-core/pull/1222, I noticed that the two fields of RepositoryReferenceFilter where not properly populated for the mojo if both are used together.
So although I have used the following configuration the exclude list was empty when the mojo was executed.
```
<plugin>
	<configuration>
		<includeAllSources>true</includeAllSources>
		<filterProvided>true</filterProvided>
		<addIUTargetRepositoryReferences>true</addIUTargetRepositoryReferences>
		<repositoryReferenceFilters>
			<addOnlyProviding>true</addOnlyProviding>
			<exclude>
				<location>https://download.eclipse.org/tools/orbit/**</location>
				<location>![%regex[http(s)?:\/\/download\.eclipse\.org\/.*]]</location>
			</exclude>
		</repositoryReferenceFilters>
	</configuration>
```
So either I'm doing something wrong or `addOnlyProviding` needs to be a separate attribute. Then the list of exclusions could also be a simple List<String>.

HannesWell:
> So either I'm doing something wrong or `addOnlyProviding` needs to be a separate attribute. Then the list of exclusions could also be a simple List.

I was doing something wrong. I used the old configuration where filters was used as plural.
The parameter is named `repositoryReferenceFilter` now without trailing s.

laeubi:
Good you find that out, so is this ready to be merged?

HannesWell:
> Good you find that out, so is this ready to be merged?

Yes it finally is.

PR #2753, opened 2023-09-05
Add an option to filter repo-references before being added to a p2-repo

HannesWell:
Back-port of https://github.com/eclipse-tycho/tycho/pull/2742 for Tycho 4.0.x

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,29 @@
 
 This page describes the noteworthy improvements provided by each release of Eclipse Tycho.
 
+## 4.0.3
+
+### new option to filter added repository-references when assembling a p2-repository
+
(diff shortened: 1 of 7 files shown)


RELEASE_NOTES.md
```diff
@@ -2,6 +2,29 @@
 
 This page describes the noteworthy improvements provided by each release of Eclipse Tycho.
 
+## 4.0.3
```
laeubi, line 5: just in case it is sufficient for a backport to only have a short summary of what was done  but the "long form" is fine as well)

HannesWell: I was wondering about that point because actually this feature is then not new in 5 because it will be first introduced in 4.0.3.
But since they are on different branches it will not appear on master with 4.0.3.
Nevertheless it could be confusing if one has used a feature and now reads that it is new in a later version.
Do you plan to 'forward' port the Release notes on a 4.0.3 or at least 5 release to the master?

laeubi: Usually I copy them after the realese from the release branch to the master.

PR #2841, opened 2023-09-19
Fix DependencyComputer missing bundles when they have different versions

laeubi:
Currently the DependencyComputer sorts only by bundle name (even though the comment for the code mentions the version as well) this can lead to the situation that if a bundle requires the same bundle in different versions (e.g. because it has actually different package names like javax -> jakarta) it only gets one of both on the classpath even though the P2 input suggests both.

This simply enhances the used key in the mak with the version as already suggested by the java comment.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/DependencyComputer.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/DependencyComputer.java
@@ -119,6 +119,11 @@ public class DependencyComputer {
         public Version getVersion() {
             return getRevision().getVersion();
         }
+
+        @Override
+        public String toString() {
+            return "DependencyEntry [module=" + module + ", rules=" + rules + "]";
+        }
     }
 
     private final class VisiblePackages {
@@ -182,7 +187,7 @@ public class DependencyComputer {
         // sort by symbolicName_version to get a consistent order
         Map<String, BundleRevision> resolvedImportPackages = new TreeMap<>();
         for (BundleRevision bundle : visiblePackages.getParticipatingModules()) {
-            resolvedImportPackages.put(bundle.getSymbolicName(), bundle);
+            resolvedImportPackages.put(bundle.getSymbolicName() + "_" + bundle.getVersion(), bundle);
         }
         for (BundleRevision bundle : resolvedImportPackages.values()) {
             addDependencyViaImportPackage(bundle, added, visiblePackages, entries);


tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/DependencyComputer.java
```diff
@@ -182,7 +187,7 @@ public List<DependencyEntry> computeDependencies(ModuleRevision module) {
         // sort by symbolicName_version to get a consistent order
         Map<String, BundleRevision> resolvedImportPackages = new TreeMap<>();
         for (BundleRevision bundle : visiblePackages.getParticipatingModules()) {
-            resolvedImportPackages.put(bundle.getSymbolicName(), bundle);
+            resolvedImportPackages.put(bundle.getSymbolicName() + "_" + bundle.getVersion(), bundle);
```
HannesWell, line 190: Wouldn't

HannesWell, line 190: Wouldn't it be simpler (and maybe faster) to do something like instead?
```
List<BundleRevision> bundles = new ArrayList<>(visiblePackages.getParticipatingModules());
bundles.sort(Comparator.comparing(Bundle::getSymbolicName).thenComparing(Bundle::getVersion));
```

HannesWell: Or use a `TreeSet` if duplicates should be filtered.

laeubi: One don't even need to create a map / list at all I think and could merge it with the loop below, but to be honest I first want this ugly bug fixed before restructure the code to have a minimal change here.

HannesWell: Understand. But just appending the version is not absolutly save.
Imaginge somebody has a Bundle `foo` with version `1` and another Bundle with name `foo_1` and version 2. Then this does not work.

laeubi: > Imagine somebody has a Bundle foo with version 1 and another Bundle with name foo_1 and version 2. Then this does not work.

Why? this will result in `foo_1`, and `foo_1_2` keys what are distinct strings :-)

HannesWell: Hm. OK. It will probably work.

PR #2848, opened 2023-09-23
Use 2023-09 simrel

akurtakov:
As default repo in plugins that have such and in tests.

### DIFF
--- a/tycho-metadata-model/src/main/java/org/eclipse/tycho/model/classpath/JUnitClasspathContainerEntry.java
+++ b/tycho-metadata-model/src/main/java/org/eclipse/tycho/model/classpath/JUnitClasspathContainerEntry.java
@@ -29,8 +29,8 @@ public interface JUnitClasspathContainerEntry extends ClasspathContainerEntry {
     static final JUnitBundle JUNIT4_PLUGIN = JUnitBundle.of("org.junit", "[4.13.0,5.0.0)",
             "org.apache.servicemix.bundles", "org.apache.servicemix.bundles.junit");
 
-    static final JUnitBundle HAMCREST_CORE_PLUGIN = JUnitBundle.of("org.hamcrest.core", "[1.1.0,2.0.0)",
-            "org.apache.servicemix.bundles", "org.apache.servicemix.bundles.hamcrest");
+    static final JUnitBundle HAMCREST_CORE_PLUGIN = JUnitBundle.of("org.hamcrest", "[2.0.0,3.0.0)", "org.hamcrest",
+            "hamcrest");
 
     static final JUnitBundle JUNIT_JUPITER_API_PLUGIN = JUnitBundle.of("junit-jupiter-api", "[5.0.0,6.0.0)",
             "org.junit.jupiter", "junit-jupiter-api");
--- a/tycho-testing-harness/src/main/java/org/eclipse/tycho/test/util/EnvironmentUtil.java
+++ b/tycho-testing-harness/src/main/java/org/eclipse/tycho/test/util/EnvironmentUtil.java
@@ -36,7 +36,7 @@ public class EnvironmentUtil {
 
     private static final String MAVEN_HOME_INFO = "Maven home:";
 
-    public static final String ECLIPSE_LATEST = "https:////download.eclipse.org/releases/2022-12/";
+    public static final String ECLIPSE_LATEST = "https:////download.eclipse.org/releases/2023-09/";
 
     private static final Properties props;
 
--- a/tycho-its/projects/api-tools/pom.xml
+++ b/tycho-its/projects/api-tools/pom.xml
@@ -12,7 +12,7 @@
 		<module>bundle2</module>
 	</modules>
 	<properties>
-		<target-platform>https://download.eclipse.org/releases/2023-06/</target-platform>
+		<target-platform>https://download.eclipse.org/releases/2023-09/</target-platform>
 	</properties>
 	<repositories>
 		<repository>
--- a/tycho-its/projects/document-bundle-plugin/build-help-index/pom.xml
+++ b/tycho-its/projects/document-bundle-plugin/build-help-index/pom.xml
@@ -6,7 +6,7 @@
 	<version>1.0.0-SNAPSHOT</version>
 	<packaging>eclipse-plugin</packaging>
 	<properties>
-		<target-platform>https://download.eclipse.org/releases/2023-03/</target-platform>
+		<target-platform>https://download.eclipse.org/releases/2023-09/</target-platform>
 		<tycho-version>4.0.0-SNAPSHOT</tycho-version>
 	</properties>
 	<repositories>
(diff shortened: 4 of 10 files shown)


laeubi:
@akurtakov what do you think about moving this to TychoConstants as it is also used in several places in Tycho (no test) code?

tycho-testing-harness/src/main/java/org/eclipse/tycho/test/util/EnvironmentUtil.java
```diff
@@ -36,7 +36,7 @@ public class EnvironmentUtil {
 
     private static final String MAVEN_HOME_INFO = "Maven home:";
 
-    public static final String ECLIPSE_LATEST = "https:////download.eclipse.org/releases/2022-12/";
+    public static final String ECLIPSE_LATEST = "https:////download.eclipse.org/releases/2023-09/";
```
laeubi, line 39: do you know hy we have the four times `/` here? it seems to not harm but looks strange...

akurtakov: I would guess 2 // for the protocol and 2 more to escape them somewhere in the code.

laeubi: This looks suspicious... if we need escaping it should not be done in the constant itself ...

tycho-metadata-model/src/main/java/org/eclipse/tycho/model/classpath/JUnitClasspathContainerEntry.java
```diff
@@ -29,8 +29,8 @@ public interface JUnitClasspathContainerEntry extends ClasspathContainerEntry {
     static final JUnitBundle JUNIT4_PLUGIN = JUnitBundle.of("org.junit", "[4.13.0,5.0.0)",
             "org.apache.servicemix.bundles", "org.apache.servicemix.bundles.junit");
 
-    static final JUnitBundle HAMCREST_CORE_PLUGIN = JUnitBundle.of("org.hamcrest.core", "[1.1.0,2.0.0)",
-            "org.apache.servicemix.bundles", "org.apache.servicemix.bundles.hamcrest");
+    static final JUnitBundle HAMCREST_CORE_PLUGIN = JUnitBundle.of("org.hamcrest", "[2.0.0,3.0.0)", "org.hamcrest",
```
laeubi, line 32: the specified maven artifact is not a bundle!

akurtakov: Hmm, this looks like a bundle to me: 
```
Manifest-Version: 1.0
Implementation-Title: hamcrest
Automatic-Module-Name: org.hamcrest
Bundle-SymbolicName: org.hamcrest
Implementation-Version: 2.2
Bundle-ManifestVersion: 2
Bnd-LastModified: 1571259132000
Import-Package: javax.xml.namespace;resolution:=optional,javax.xml.xpa
 th;resolution:=optional,org.w3c.dom;resolution:=optional,org.hamcrest
 ;version="[2.2,3)",org.hamcrest.beans;version="[2.2,3)",org.hamcrest.
 collection;version="[2.2,3)",org.hamcrest.comparator;version="[2.2,3)
 ",org.hamcrest.core;version="[2.2,3)",org.hamcrest.internal;version="
 [2.2,3)",org.hamcrest.number;version="[2.2,3)",org.hamcrest.object;ve
 rsion="[2.2,3)",org.hamcrest.text;version="[2.2,3)",org.hamcrest.xml;
 version="[2.2,3)"
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.7))"
Tool: Bnd-3.4.0.201707252008
Implementation-Vendor: hamcrest.org
Export-Package: org.hamcrest;version="2.2";uses:="javax.xml.namespace,
 org.hamcrest.collection,org.hamcrest.core,org.hamcrest.internal,org.w
 3c.dom",org.hamcrest.beans;version="2.2";uses:="org.hamcrest",org.ham
 crest.collection;version="2.2";uses:="org.hamcrest",org.hamcrest.comp
 arator;version="2.2";uses:="org.hamcrest",org.hamcrest.core;version="
 2.2";uses:="org.hamcrest",org.hamcrest.internal;version="2.2";uses:="
 org.hamcrest",org.hamcrest.io;version="2.2";uses:="org.hamcrest",org.
 hamcrest.number;version="2.2";uses:="org.hamcrest",org.hamcrest.objec
 t;version="2.2";uses:="org.hamcrest",org.hamcrest.text;version="2.2";
 uses:="org.hamcrest",org.hamcrest.xml;version="2.2";uses:="javax.xml.
 namespace,org.hamcrest,org.w3c.dom"
Bundle-Name: hamcrest
Bundle-Version: 2.2
Created-By: 1.8.0_232-ea (Private Build)
```
What issue do you have with it ?

laeubi: hm... it seems the `hamcrest` vs `hamcrest-core` has tricked me here :-\

So everything looks fine here :+1:

akurtakov:
@laeubi Is there a chance you can help with the last failing test? I'm kind of lost on this one.

akurtakov:
I've tried by removing vintage engine from the junit5 bundles and the test works fine so what happens is that vintage engine fails and that way it even blocks other engines from looking up for tests.

laeubi:
I have now created
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/837

So we need to somehow convince Tycho to include junit4 when vintage engine is included, I'm just a bit confused why it is not select anymore automatically...

laeubi:
I now extracted the easy part here:

- https://github.com/eclipse-tycho/tycho/pull/2862

akurtakov:
@laeubi Do you agree with using 
```
<configuration>
					<excludeJUnit5Engines>
						<engine>junit-vintage</engine>
					</excludeJUnit5Engines>
				</configuration>
```
to workaround the issue for now?

laeubi:
I have now tested it and at first everything seems okay but then surefire fails, I can fix the test by adding

```
<dependency>
		  <groupId>org.apache.servicemix.bundles</groupId>
		  <artifactId>org.apache.servicemix.bundles.junit</artifactId>
		  <version>4.13.2_1</version>
		  <scope>test</scope>
</dependency>
```

as an explicit dependency ... so it seems that something goes wrong maybe when injecting the dependencies even though I can't see how/why this happens need to debug a bit further.

laeubi:
The junit dependency is currently injected as

```
 <dependency>
      <artifactId>org.apache.servicemix.bundles.junit</artifactId>
      <version>4.13.2.1</version>
      <type>eclipse-plugin</type>
      <scope>system</scope>
      <systemPath>/tmp/file not yet available5847474459691235710.tmp</systemPath>
    </dependency>
```

so for some reason p2 things that the file is not (yet) there and then surefire can not use it of course ... will try to provide a fix for that...

laeubi:
This should fix the issue we see here:
- https://github.com/eclipse-tycho/tycho/pull/2867

akurtakov:
Jenkins succeeded. Merging.

PR #2850, opened 2023-09-23
Support assembling and archiving repositories in parallel

HannesWell:
Assembling and arching repositories is a relatively long running task and repository-projects are usually build as one of the last projects in the reactor. So the chances are relatively high, that (if multiple projects are present) multiple repositories are assembled at the same. Therefore it should be possible to build multiple repositories in parallel and there should not be one global lock for each repository-related mojo.

### DIFF
--- a/tycho-api/src/main/java/org/eclipse/tycho/FileLockService.java
+++ b/tycho-api/src/main/java/org/eclipse/tycho/FileLockService.java
@@ -37,4 +37,10 @@ public interface FileLockService {
      * advisory only, i.e. all processes must use the same locking mechanism.
      */
     Closeable lock(File file, long timeout);
+
+    /**
+     * Locks the given file for this JVM to protect read/write access from multiple threads in this
+     * JVM on it.
+     */
+    Closeable lockVirtually(File file);
 }
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/locking/FileLockServiceImpl.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/locking/FileLockServiceImpl.java
@@ -19,30 +19,56 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.codehaus.plexus.component.annotations.Component;
 import org.eclipse.tycho.FileLockService;
+import org.eclipse.tycho.LockTimeoutException;
 
 @Component(role = FileLockService.class)
 public class FileLockServiceImpl implements FileLockService {
+    record FileLocks(FileLockerImpl fileLocker, Lock vmLock) {
+    }
 
-    private final Map<Path, FileLockerImpl> lockers = new ConcurrentHashMap<>();
+    private final Map<Path, FileLocks> lockers = new ConcurrentHashMap<>();
 
     @Override
     public Closeable lock(File file, long timeout) {
-        FileLockerImpl locker = getFileLocker(file.toPath());
+        FileLocks locks = getFileLocker(file.toPath());
+        FileLockerImpl locker = locks.fileLocker();
+        try {
+            if (!locks.vmLock().tryLock(timeout, TimeUnit.MILLISECONDS)) {
+                throw new LockTimeoutException("lock timeout: Could not acquire lock on file " + locker.lockMarkerFile
+                        + " for " + timeout + " msec");
+            }
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new LockTimeoutException("Interrupted", e);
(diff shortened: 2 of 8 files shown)


tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AbstractRepositoryMojo.java
```diff
@@ -13,11 +13,33 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
+import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.eclipse.tycho.core.maven.AbstractP2Mojo;
 
 public abstract class AbstractRepositoryMojo extends AbstractP2Mojo {
+    private static final Map<File, Lock> REPOSITORY_LOCK = new ConcurrentHashMap<>();
```
laeubi, line 27: Tycho has a `FileLockService` can we probably reuse that here?

HannesWell: I looked into that, but since the FileLockService creates a lock-file in the directory to lock this is not usable in cases where for example a repository is zipped, because then the lock file ends up in the zipped repository.
Furthermore a inter-process file locking does not seem to be necessary because the resources are within the projects being build and the goal is only to prevent illegal concurrent access in the same build-process. One should not run multiple build processes at the same time for the same or even sub-directories any way. So having a file-locking seems to be too much for me and an in-memory locking is sufficient.

HannesWell: But with https://github.com/eclipse-tycho/tycho/pull/2853 we could simply add another method like `FileLockService.lockForThisProcess` or similar that only does an ordinary in-process/memory locking without creating any files.

laeubi: > I looked into that, but since the FileLockService creates a lock-file in the directory to lock this is not usable in cases where for example a repository is zipped, because then the lock file ends up in the zipped repository.

Have you tested this? The zipping of repository is taking place in a separate step where the lock file should already be deleted (or can be deleted).

> One should not run multiple build processes at the same time for the same or even sub-directories any way

I think you are aware of [murphy's law](https://de.wikipedia.org/wiki/Murphy%E2%80%99s_Law) ;-)

> So having a file-locking seems to be too much for me and an in-memory locking is sufficient.

If we can reuse something that offers even more we should use that instead of inventing another approach that offers weaker guarantees.

laeubi: Now you have refactored the lock do you think it could be used here?

HannesWell: > > I looked into that, but since the FileLockService creates a lock-file in the directory to lock this is not usable in cases where for example a repository is zipped, because then the lock file ends up in the zipped repository.
> 
> Have you tested this? The zipping of repository is taking place in a separate step where the lock file should already be deleted (or can be deleted).

I haven't tried it yet. But with this zipping the repo would also be guarded with a lock-file, which would create the file before the zipping starts and deletes it only afterwards. And in between the directory content is zipped. Therefore expect that the lock-file will be included. But I'll try it.



> > One should not run multiple build processes at the same time for the same or even sub-directories any way
> 
> I think you are aware of [murphy's law](https://en.wikipedia.org/wiki/Murphy%27s_law) ;-)

It could definitively happen, but one should not do it. So its then the users fault. :)



> > So having a file-locking seems to be too much for me and an in-memory locking is sufficient.
> 
> If we can reuse something that offers even more we should use that instead of inventing another approach that offers weaker guarantees.

In generally that's right. But as written above, I'm not yet sure if this is really suitable. If it is not suitable, I would just add the method as something like `lockVirtually()` to the `FileLockerService`.

laeubi: I think we do not loose anything using the function as-is ... any complication like "virtual" make it just confusing so if we zip the content it should be easy to exclude the lockfile anyways.

HannesWell: I would see the opposite as more complicated since the exclusion then would need to be kept in sync if the lock-file changes one day.
And which file is locked eventually is an implementation detail of the locker and consumer not interested in inter-process locking should not have to care about that.

PR #2852, opened 2023-09-24
Assemble repository for all environments at once

HannesWell:
This is helpful for https://github.com/eclipse-tycho/tycho/pull/2744 since the full resulting repository is available on the first pass.

Additionally it can improve the performance since all IUs are only processed once instead of once per environment (which applies for all IUs without environment filter) and it avoids multiplication of print-outs for each environment.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
@@ -9,6 +9,7 @@
  *
  * Contributors:
  *     SAP SE - initial API and implementation
+ *     Hannes Wellmann - Assemble repository for all environments in one pass
  *******************************************************************************/
 package org.eclipse.tycho.p2tools;
 
@@ -29,6 +30,8 @@ import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.equinox.internal.p2.director.PermissiveSlicer;
 import org.eclipse.equinox.internal.p2.director.Slicer;
 import org.eclipse.equinox.internal.p2.metadata.IRequiredCapability;
+import org.eclipse.equinox.internal.p2.metadata.InstallableUnit;
+import org.eclipse.equinox.internal.p2.metadata.RequiredCapability;
 import org.eclipse.equinox.p2.core.IProvisioningAgent;
 import org.eclipse.equinox.p2.core.ProvisionException;
 import org.eclipse.equinox.p2.internal.repository.tools.RepositoryDescriptor;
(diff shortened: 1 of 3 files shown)


tycho-core/src/main/java/org/eclipse/tycho/p2tools/copiedfromp2/MirrorApplication.java
```diff
@@ -396,23 +399,32 @@ private IArtifactMirrorLog getLog(File location, String root) {
...
+            pcr.addAll(sourceIUs);
+            IProvisioningPlan plan = planner.getProvisioningPlan(pcr, null, monitor);
+            registry.removeProfile(profileId);
+            queryables.add(plan.getAdditions());
+            IProvisioningPlan installerPlan = plan.getInstallerPlan();
+            if (installerPlan != null) {
+                queryables.add(installerPlan.getAdditions());
+            }
+        }
+        return new CompoundQueryable<>(queryables.toArray(EMPTY));
```
HannesWell, line 427: As far as I can tell, this code path is actually never called by Tycho, but I added it for completes.

tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
```diff
@@ -82,13 +84,19 @@ protected IArtifactRepository initializeDestination(RepositoryDescriptor toInit,
 
     @Override
     protected Slicer createSlicer(SlicingOptions options) {
-        Map<String, String> context = options.getFilter();
         boolean includeOptionalDependencies = options.includeOptionalDependencies();
         boolean onlyFilteredRequirements = options.followOnlyFilteredRequirements();
-        boolean considerFilter = context != null && context.size() > 1;
+        List<IInstallableUnit> selectionContexts = environments.stream().map(env -> {
```
HannesWell, line 89: The case of no environments has to be considered here and in the super-class.

tycho-core/src/main/java/org/eclipse/tycho/p2tools/copiedfromp2/MirrorApplication.java
laeubi: @HannesWell the `MirrorApplication` is a copy from P2 (with some adjustments from Tycho) so if you think there is something in general to improve it might be worth to contribute it to P2 as well.

But if the method is not used and serves no purpose we can also remove it.

laeubi:
@HannesWell some of the test now (again?) fail with

> Caused by: java.lang.NoSuchMethodError: 'org.apache.maven.toolchain.java.DefaultJavaToolChain org.eclipse.tycho.core.maven.ToolchainProvider.findMatchingJavaToolChain(org.apache.maven.execution.MavenSession, java.lang.String)'

tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
HannesWell: Done.

HannesWell:
> @HannesWell some of the test now (again?) fail with
> 
> > Caused by: java.lang.NoSuchMethodError: 'org.apache.maven.toolchain.java.DefaultJavaToolChain org.eclipse.tycho.core.maven.ToolchainProvider.findMatchingJavaToolChain(org.apache.maven.execution.MavenSession, java.lang.String)'

Still no clue why this happens. At least the workflows run with maven 3.9.2...

tycho-core/src/main/java/org/eclipse/tycho/p2tools/copiedfromp2/MirrorApplication.java
HannesWell: Yes I noticed that already and I have the intention to contribute this change to P2 as well. I could then try to incooporate the other changes as well or at least provide suitable possibilities for extensions so that Tycho does not need a copy anymore.
But for now I would only do it here to have it quicker.

laeubi: > Tycho does not need a copy anymore.
> But for now I would only do it here to have it quicker.

I think it is fine to have a copy here, maintaining things at the other projects has not worked out very good (same for ApiAnalysisApplication ...) as it often required considerable amount of changes/workarounds (and discussions) for little gain.

HannesWell: I suggest we finish this first and if it works as desired we'll see if I manage to contribute it back to P2.
In case that works we can then use the enhanced classes and if even later some other changes are necessary again, nothing prevents us from adding the copy again. :)

HannesWell:
So hopefully it is now working. At least the tests that failed in the last build now succeeded locally.
The missing pies was to copy `PermissiveSlicer.isApplicable(IRequirement)` so that `Slicer.isApplicable(IRequirement)` could be replaced too to query all environment contexts.

tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
```diff
@@ -103,17 +108,49 @@ protected boolean isApplicable(IInstallableUnit iu, IRequirement req) {
...
                         }
-                        return !considerFilter || filter == null || filter.isMatch(selectionContext);
+                        return !considerFilter || filter == null || matchesSelectionContext(filter);
                     }
                 }
-                return super.isApplicable(req);
+                return permissiveSlicer_isApplicable(req);
+            }
+
+            private boolean permissiveSlicer_isApplicable(IRequirement req) {
```
laeubi, line 117: why not use the existing `isApplicable(IRequirement req)` and let it not throw USOE?

HannesWell: Done.

laeubi (approved):

PR #2853, opened 2023-09-24
Introduce FileLockService.Locking to leverage try-with-resource-blocks

HannesWell:
This allows to simplify the usage of the FileLocker/FileLockingService to:

try (var locking = fileLockService.lock(fileToProtect)) {
  // do something
}

With this changes the FileLocker interface is only used in tests. Therefore my suggestion is to remove it completely and just use the `FileLockService.Locking` instead.


tycho-api/src/main/java/org/eclipse/tycho/FileLockService.java
```diff
@@ -20,11 +20,34 @@
  */
 public interface FileLockService {
 
+    interface Locking extends AutoCloseable {
```
laeubi, line 23: Why do we need a new interface and can not simply use the existing `FileLocker` one?

HannesWell: As said above, actually the `FileLocker` interface is not necessary anymore with this auto-closable style.
And from the name `FileLocker` describes something that can lock something (in this case a file), the implementation of Locking represent a locked state (it could also be named `LockedFile` or similar). Therefore I would just remove FileLocker and add something that reflects the state representation better and just extends `AutoClosable`.
That being said, I mainly introduced `Locking` just because `AutoClosable` throws an Exception. But using Closable instead only throws an `IOException` which is already handled in almost all callers anyway.

For now I removed the `FileLocker` interface entirly and only kept `FileLockerImpl` as an implementation detail of `FileLockServiceImpl` that is queried in the `FileLockServiceTest`. But if there would be a real isLocked checked in the test, even `FileLockServiceImpl` could become a hidden internal class of `FileLockServiceImpl`, probably stripped down to the minimum.

laeubi: This interface originally used the Equinox resource API and was somehow driven by what this API offers, so from my side if Tycho does not use this functionality anymore it could be removed.

Issue #2858, opened 2023-09-27
Default lifecycle mapping for tycho packaging produces exception for bundles with nested jars


jstroleny:
Use following its project [sourceBundle.nestedJars](https://github.com/eclipse-tycho/tycho/tree/master/tycho-its/projects/sourceBundle.nestedJars) and import it into Eclipse and enable Maven nature to it.

This bundle uses multiple nested jars as output, each have its own `nameOfJar-classes` directory in `target` folder.
When imported into Eclipse with Maven nature enabled, it causes errors (maven build problem, see below).

When bundle with single jar is imported with same steps (e.g. [packaging.consumer.pom](https://github.com/eclipse-tycho/tycho/tree/master/tycho-its/projects/packaging.consumer.pom/bundle)) there is no error.

```
Failed to execute mojo org.eclipse.tycho:tycho-packaging-plugin:4.0.2:package-plugin {execution: default-package-plugin} (org.eclipse.tycho:tycho-packaging-plugin:4.0.2:package-plugin:default-package-plugin:package)

org.eclipse.core.runtime.CoreException: Failed to execute mojo org.eclipse.tycho:tycho-packaging-plugin:4.0.2:package-plugin {execution: default-package-plugin}
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.executeMojo(MavenExecutionContext.java:340)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.lambda$0(MavenExecutionContext.java:291)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.executeBare(MavenExecutionContext.java:394)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.execute(MavenExecutionContext.java:275)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.execute(MavenExecutionContext.java:290)
	at org.eclipse.m2e.core.project.configurator.MojoExecutionBuildParticipant.build(MojoExecutionBuildParticipant.java:57)
	at org.eclipse.m2e.core.internal.builder.MavenBuilderImpl.lambda$2(MavenBuilderImpl.java:153)
	at java.base/java.util.LinkedHashMap.forEach(LinkedHashMap.java:721)
	at org.eclipse.m2e.core.internal.builder.MavenBuilderImpl.build(MavenBuilderImpl.java:133)
[...]

PR #2859, opened 2023-09-27
Fix creation of output directory in packaging mojo for bundles with nested jars

jstroleny:
Fixes https://github.com/eclipse-tycho/tycho/issues/2858


laeubi (approved):
Looks good.

laeubi:
Thnaks for providing this fix!

PR #2862, opened 2023-09-28
Introduce a public static constant for the current eclipse release

laeubi:
This should make it less cumbersome to update things in the future.


akurtakov:
Jenkins succeeded

PR #2878, opened 2023-10-02
fix #2877

### LINKED ISSUE #2877: org.eclipse.tycho:tycho-versions-plugin:bump-versions does not honor SNAPSHOT suffix
When I use the bump-versions mojo as described [here](https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#new-tycho-versions-plugin-mojos), then the qualifier in the manifest and/or the snapshot suffix in the pom.xml of the bundle that get its version bumped is wiped out.

mlehmannm:


tycho-versions-plugin/src/main/java/org/eclipse/tycho/versions/engine/Versions.java
```diff
@@ -17,9 +17,9 @@
 import org.osgi.framework.Version;
 
 public class Versions {
-    private static final String SUFFIX_QUALIFIER = ".qualifier";
+    public static final String SUFFIX_QUALIFIER = ".qualifier";
```
laeubi, line 20: We probably should move those constants to the [TychoConstants](https://github.com/eclipse-tycho/tycho/blob/master/tycho-api/src/main/java/org/eclipse/tycho/TychoConstants.java) class.

mlehmannm: I can do this in another PR. Would you backport this PR to 4.0.x first?

laeubi: Just add it to this pr no need for several ones. It would also be good to improve the commit message a bit (e.g. describe what is the problem and what was changed).

mlehmannm: Will do.

laeubi (approved):
Looks good, a testcase would be good to ensure this will not breakin the future.

mlehmannm:
> Looks good, a testcase would be good to ensure this will not breakin the future.

Where is a good place to add a testcase?

laeubi:
@mlehmannm we have some integration test for the version plugin here:

https://github.com/eclipse-tycho/tycho/tree/master/tycho-its/src/test/java/org/eclipse/tycho/test/versionsplugin

mlehmannm:
Providing a testcase is really hard, because one would need a repository with at least an eclipse-plugin and the same eclipse-plugin as source with changes so that the baseline-mojo needs to increment the version of that eclipse-plugin.

laeubi:
You might want to take a look at the [BaselineMojoTest](https://github.com/eclipse-tycho/tycho/blob/master/tycho-its/src/test/java/org/eclipse/tycho/test/baseline/BaselineMojoTest.java) that creates a baseline repo first (you might even want to add your testcase there) and then modify the source sightly to generate a baseline error.

laeubi:
@mlehmannm whats the status here? I plan to publish a bugfix of Tycho4 soon and it seems good to have your fix included.

mlehmannm:
Sorry for the delay. I was on vacation and had a busy week afterwards. I plan to provide a test for the fix, but it might take some time, but I think you can include that fix as it is mostly trivial.

Issue #2709, opened 2023-08-11
Tycho 4.0.1 Unable to read repository at https://download.eclipse.org/releases/latest behind proxy


mneffgen:
I am behind a corporate proxy and tried to build the itp01 demo with Tycho 4.
While the build succeeded with Tycho 3.0.5 and 2.7.5 with my proxy configuration in the maven settings, it seems that the proxy is ignored by tcho 4.0.1.

Kummallinen:
I'm having the same issue, however it does work some of the time

laeubi:
There is a (passing!) test for proxy support:

https://github.com/eclipse-tycho/tycho/blob/master/tycho-its/src/test/java/org/eclipse/tycho/test/TYCHO279HttpProxy/ProxySupportTest.java

so if anyone is able to reproduce the problem with that test it would help alot.

Torbjorn-Svensson:
> There is a (passing!) test for proxy support:
> 
> https://github.com/eclipse-tycho/tycho/blob/master/tycho-its/src/test/java/org/eclipse/tycho/test/TYCHO279HttpProxy/ProxySupportTest.java
> 
> so if anyone is able to reproduce the problem with that test it would help alot.

PR created that updates the testcase to emulate a corporate network with a proxy.
Hopefully, you can use this to identify why tycho fails to use the configured proxy.

mneffgen:
It turnedd out, that my proxy configuration is the problem.
It does not work with protocol specific configurations.

Working:
<proxies>
	<proxy>
		<id>Proxy</id>
		<active>true</active>
		<host>192.168.8.50</host>
		<port>3128</port>
	</proxy>
</proxies>

Not working:
<proxies>
	<proxy>
		<id>Proxy HTTP</id>
		<active>true</active>
		<protocol>http</protocol>
		<host>192.168.8.50</host>
		<port>3128</port>
	</proxy>
	<proxy>
		<id>Proxy HTTPS</id>
		<active>true</active>
		<protocol>https</protocol>
		<host>192.168.8.50</host>
		<port>3128</port>
	</proxy>
</proxies>

akurtakov:
Is this still a problem with latest Tycho version?

PR #2895, opened 2023-10-06
Update testcase to emulate corporate proxy

Torbjorn-Svensson:
The testcase wrongly assumed that the client can resolve the domain name.
Some corporate networks with proxy have internal name servers that does not include public records and this, resolving any non-internal domain names will fail. The proxy servers are having one leg in the real world and can thus resolve public records.

I've only updated the test case as I'm unable to build of tycho for some unknown reason.

To validate that the testcase did the intended, I had to:
1. Remove every module in `/pom.xml` except `tycho-its` and `tycho-testing-harness`.
2. Remove every java file in `/tycho-its/src/test/java/` except `/tycho-its/src/test/java/org/eclipse/tycho/test/TYCHO279HttpProxy/ProxySupportTest.java`.
3. Build using `mvn -Dtycho-build-version=4.0.3  clean package`.

Hope this fulfills the request for help in https://github.com/eclipse-tycho/tycho/issues/2709#issuecomment-1722854049.

### DIFF
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/TYCHO279HttpProxy/ProxySupportTest.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/TYCHO279HttpProxy/ProxySupportTest.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
+import javax.servlet.http.HttpServletRequest;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerFactory;
@@ -39,6 +40,10 @@ import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.handler.ContextHandler;
 import org.eclipse.jetty.server.handler.ResourceHandler;
+import org.eclipse.jetty.servlet.ServletContextHandler;
+import org.eclipse.jetty.servlet.ServletHandler;
+import org.eclipse.jetty.servlet.ServletHolder;
+import org.eclipse.jetty.servlet.ServletMapping;
 import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
 import org.junit.After;
 import org.junit.Before;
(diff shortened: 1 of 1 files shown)


laeubi:
To build/test you usually do:

`mvn clean install -DskipTests -T1C`

then cd into tycho-its then call

`mvn clean install -Pits -Dtest=<classname of your test to run>`

laeubi:
> The testcase wrongly assumed that the client can resolve the domain name.

Why is this wrong? I'm not sure if such thing can ever work in a unit test reliable as name service is different from proxy service.

So if you like to have that we either also need to mock a DNS server or one needs to configure a mirror or similar.

Torbjorn-Svensson:
> > The testcase wrongly assumed that the client can resolve the domain name.
> 
> Why is this wrong? I'm not sure if such thing can ever work in a unit test reliable as name service is different from proxy service.
> 
> So if you like to have that we either also need to mock a DNS server or one needs to configure a mirror or similar.

Well, if you want the real deal, then no, it would not work, but I'm using a different approach.
In the test, I instead of stating the web-server that has the content and let the proxy just tag along, I define a different URL that the proxy knows that it should replace the the real URL. This way, I'm forcing the request to go though the proxy in order to actually resolve to anything, meaning that I emulate that the client cannot do the resolving while the proxy server "can" (obviously it can't but it instead replaces the request).
Does this make sense?

```
client
  -> request http://fake-non-existing-domain/random/path/ from proxy

proxy 
  -> get the request
  -> rewrite URL to http://localhost:12345/test/
  -> send the request of http://localhost:12345/test/
  -> forward the reply back to "client" as data from http://fake-non-existing-domain/random/path/
```

Obviously, `fake-non-existing-domain` can be anything that the client cannot access without the proxy. In my case, I assume that there is nothing that matches this string.

Do you think this way to emulate the scenario with a corporate network that requires to go thought a proxy to get out is valid?

Now that I'm thinking about this a 2nd time, I might have missed to replace all links to `http://localhost:...` with the `http://fake-non...` part in the reply.

laeubi:
Alright, but now the test fails with a message that proxy was **not** used while I would expect that it should fail with `http://fake-non-existing-domain/random/path/` not found instead ...

laeubi:
@Torbjorn-Svensson there was a fix in the proxy support recently do you want to rebase your change and see if it fixes your issue?

laeubi:
@Torbjorn-Svensson can you rebase and resolve conflicts?

laeubi:
I'm closing this as no feedback from the original author, if this is still interesting just open a PR with an up-to-date version.

PR #2899, opened 2023-10-07
Call the API tools directly without using ApiAnalysisApplication

laeubi:
This replaces ApiAnalysisApplication usage by a more direct approach that do not require us to pass file lists around and directly class the ApiTools.
This is also much more flexible in regard to debugging and additional printouts but still retains that we can consume dynamic implementation from P2 repositories.

Currently this is work-in-progress:

1. For full features we need a solution for the problem described here: https://github.com/eclipse-pde/eclipse.pde/issues/785
2. Currently API problems are just printed for demo purpose and feature validation
3. The ApiTools currently determine the profile internally, we should supply the profile generated by Tycho

### DIFF
--- a/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysis.java
+++ b/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysis.java
@@ -1,0 +1,274 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Christoph Läubrich and others.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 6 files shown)


laeubi:
This is ready now and gives much better user feedback. Instead of "ApiApplication failed" and require the user to scan the logs one now will be presented in the error message with the details about the failure:
grafik

beside that the API analysis is now faster because we don't need to build / compile the project to perform the API analysis (in this case before 25 seconds now 15 seconds).

This also adds a new parameter "debug" that prints additional information about the used API baseline:

```
<plugin>
	<artifactId>tycho-apitools-plugin</artifactId>
	<configuration>
		<debug>true</debug>
	</configuration>
</plugin>
```

PR #2919, opened 2023-10-13
disable vm install job in jdt

laeubi:
See https://github.com/eclipse-pde/eclipse.pde/issues/782#issuecomment-1761335390

### DIFF
--- a/tycho-apitools-plugin/pom.xml
+++ b/tycho-apitools-plugin/pom.xml
@@ -69,6 +69,11 @@
 		    <artifactId>org.eclipse.jdt.core</artifactId>
 		    <version>3.35.0</version>
 		</dependency>
+		<dependency>
+		    <groupId>org.eclipse.jdt</groupId>
+		    <artifactId>org.eclipse.jdt.launching</artifactId>
+		    <version>3.20.100</version>
+		</dependency>
 		<!-- libs -->
 		<dependency>
 		  <groupId>org.ow2.asm</groupId>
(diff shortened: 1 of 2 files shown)


tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisMojo.java
```diff
@@ -162,6 +162,7 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 			} catch (Exception e) {
 				throw new MojoFailureException("Can't fetch dependencies!", e);
 			}
+			System.setProperty("DetectVMInstallationsJob.disabled", "true");
```
iloveeclipse, line 165: This can be too late, not sure at which point in time this code is triggered...

mickaelistria: This code is triggered (job is scheduled) when the jdt.launching bundle starts, which can be pretty early as it's required by jdt.core to resolve the classpath IIRC.
Wouldn't it be possible to set the flag before starting the VM (or at least the Equinox container) ?

laeubi: 1. This *is* before Equinox Container starts
2. I now using the preference instead (so no unforeseen side effects)
3. I have debugged that this prevents the job from being started.

So if setting it on the Jenkins Job directly works as system-property, this should then work here as well.

PR #2927, opened 2023-10-15
Remove duplicated ProductFile2

HannesWell:
@laeubi is there a reason to keep `ProductFile2` duplicated or is this unintended?


laeubi:
It could be that previously where we have some bundle world and some maven world it was easier to copy than share this. If its really a duplicate it should be possible to only use one variant now.

HannesWell:
Understand.
Then lets see what the build says and if it suceeds I assume it is safe to submit it.

HannesWell:
All builds are green.
Then lets go with this one.

PR #2929, opened 2023-10-15
Add support for Product Update-site names

HannesWell:
This a work-around to have https://github.com/eclipse-equinox/p2/pull/353 available now and can be reverted once the mentioned change in P2 is available in Tycho.
The reason for this is that I would like to use name-support in the product-definition already for the 2023-12 SimRel.
I'll take care about reverting this once the P2 updates for 2023-12 are available in Tycho.


HannesWell:
I just added a test-case and with that this is ready for me.

The test-case also needs to be added to the master branch, but I suggest to do that as part of the update to the next p2 release, which also allows to revert all implementation changes done in Tycho in this PR.

PR #2969, opened 2023-10-30
Add schema-to-html mojo as a replacement for ant ConvertSchemaToHTML

laeubi:
Currently platform build contains ant targets to run ConvertSchemaToHTML for documentation purpose.

This adds as a first step a new mojo
tycho-document-bundle-plugin:schema-to-html that offers the same functionality and should act as a base to further improve the brittle manually maintained references the current process requires and already allows to move things from the ant build to the maven build files.

### DIFF
--- a/tycho-extras/tycho-document-bundle-plugin/pom.xml
+++ b/tycho-extras/tycho-document-bundle-plugin/pom.xml
@@ -9,7 +9,8 @@
 - Contributors:
 -    IBH SYSTEMS GmbH - initial API and implementation
 -->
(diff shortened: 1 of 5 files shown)


laeubi:
For further improvements we would need for example:
- https://github.com/eclipse-pde/eclipse.pde/pull/845

to also search schemas in dependencies.

laeubi:
With
- https://github.com/eclipse-pde/eclipse.pde/pull/846

we will be able to get rid of most of the copied code and can further develop this in PDE where the actual function belongs.

Issue #2977, opened 2023-10-31
Virtual configuration IU in p2.inf does not work in 4.0.3 (works in 3.0.5)


mdaloia:
### Discussed in https://github.com/eclipse-tycho/tycho/discussions/2972

<div type='discussions-op-text'>

<sup>Originally posted by **mdaloia** October 30, 2023</sup>
Hi,

We have a "virtual" IU provided and required in a p2.inf file of a bundle to add a touchpoint configuration (it really doesn't matter what it does) that used to work in 3.0.5. We are migrating to 4.0.3 and found that it fails with:
```
[ERROR] Cannot resolve project dependencies:
[ERROR]   Software being installed: p2.inf 1.0.0.qualifier
[ERROR]   Missing requirement: p2.inf 1.0.0.qualifier requires 'org.eclipse.equinox.p2.iu; configure.p2.inf 0.0.0' but it could not be found
```

I have reduced it to an example project: https://github.com/mdaloia/tycho-p2-inf

(**Environment used**: Maven 3.9.4 / Java 17)

If I build it with
- `mvn clean package -Dtycho.localArtifacts=ignore -Dtycho-version=4.0.3`
it fails with the mentioned error and no `target/p2content.xml` is created.

I tried adding `-Dtycho.resolver.classic=false` as explained in "[mixed reactor setups require the new resolver now](https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#mixed-reactor-setups-require-the-new-resolver-now)"  but I got the same error.

But if I only change the version:
- `mvn clean package -Dtycho.localArtifacts=ignore -Dtycho-version=3.0.5`
it works and in the `target/p2content.xml` file I see the `<unit id='configure.p2.inf'>` and the `<required namespace='org.eclipse.equinox.p2.iu' name='configure.p2.inf' range='0.0.0'/>`

This is because I have `<pomDependencies>consider</pomDependencies>` and it shows this message:
```
[INFO] The following requirements are not satisfied yet and must be provided through pom dependencies:
[INFO]    - org.eclipse.equinox.p2.iu; configure.p2.inf 0.0.0
```
If I remove this configuration it fails with the same error as using 4.0.3.
</div>

mdaloia:
@laeubi Unfortunately the changes made in PR #2989 partially fixes it.

If you have another bundle in which the `Require-Bundle` references the bundle that has the `p2.inf` with the `provides` & `requires` the same error appears:

```
[...]

PR #2978, opened 2023-10-31
Add a reproducer IT for issue #2977

mdaloia:
Virtual configuration IU in p2.inf does not work in 4.0.3 (works in 3.0.5)

If we add this statement in the `VirtualUnitTest`:
```
verifier.addCliOption("-Dtycho-version=3.0.5");
```
the test works ok and does not fail.

### DIFF
--- a/tycho-its/projects/p2Inf.virtualUnit/bundle/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/p2Inf.virtualUnit/bundle/META-INF/MANIFEST.MF
@@ -1,0 +1,7 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: pvu.bundle
+Bundle-SymbolicName: pvu.bundle;singleton:=true
+Bundle-Version: 1.0.0.qualifier
+Bundle-Vendor: TEST
+Bundle-RequiredExecutionEnvironment: JavaSE-17
--- a/tycho-its/projects/p2Inf.virtualUnit/bundle/META-INF/p2.inf
+++ b/tycho-its/projects/p2Inf.virtualUnit/bundle/META-INF/p2.inf
@@ -1,0 +1,11 @@
+# Create the virtual IU
+units.0.id=configure.pvu.bundle
+units.0.version=1.0.0
+units.0.provides.1.namespace=org.eclipse.equinox.p2.iu
+units.0.provides.1.name=configure.pvu.bundle
+units.0.provides.1.version=1.0.0
+
+# Require in this bundle the created virtual IU
+requires.0.namespace=org.eclipse.equinox.p2.iu
+requires.0.name=configure.pvu.bundle
+requires.0.range=0.0.0
--- a/tycho-its/projects/p2Inf.virtualUnit/bundle/build.properties
+++ b/tycho-its/projects/p2Inf.virtualUnit/bundle/build.properties
@@ -1,0 +1,2 @@
+bin.includes = META-INF/,\
+               .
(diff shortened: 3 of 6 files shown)


laeubi:
@mdaloia thanks for your testcase, I assume these is the expected failure:
https://ci.eclipse.org/tycho/job/tycho-github/job/PR-2978/1/testReport/junit/org.eclipse.tycho.test.p2Inf/VirtualUnitTest/testVirtualUnitRequirementDoesNotFailBuild/

mdaloia:
@laeubi yes, the error, as mentioned in the issue #2977 is:

```
[ERROR]   Software being installed: pvu.bundle 1.0.0.qualifier
[ERROR]   Missing requirement: pvu.bundle 1.0.0.qualifier requires 'org.eclipse.equinox.p2.iu; configure.pvu.bundle 0.0.0' but it could not be found
```

I added it as a regression test, I expected it to fail. Let me know if that it is ok or should I modify it somehow.

Thanks!

laeubi:
I'll try to investigate why this fails.

laeubi:
I have now checked this and it seems the `p2content.xml` in both cases is the same (except the build qualifier) for Tycho 4 and Tycho 3 so it seems not be a problem of generated metadata, but probably somewhere the provided capability is lost.

laeubi:
I now have created
- https://github.com/eclipse-tycho/tycho/pull/2988

for further debugging, but this also seems sane, need to investigate further...

laeubi:
The preliminary units look like this what seems to be sane as well:

```
<?xml version='1.0' encoding='UTF-8'?>
<units size='2'>
  <unit id='configure.pvu.bundle' version='1.0.0' singleton='false'>
    <provides size='1'>
      <provided namespace='org.eclipse.equinox.p2.iu' name='configure.pvu.bundle' version='1.0.0'/>
    </provides>
    <touchpoint id='null' version='0.0.0'/>
  </unit>
  <unit id='pvu.bundle' version='1.0.0.qualifier' generation='2'>
    <update id='pvu.bundle' range='[0.0.0,1.0.0.qualifier)' severity='0'/>
    <properties size='2'>
      <property name='org.eclipse.equinox.p2.name' value='pvu.bundle'/>
      <property name='org.eclipse.equinox.p2.provider' value='TEST'/>
    </properties>
    <provides size='4'>
      <provided namespace='org.eclipse.equinox.p2.iu' name='pvu.bundle' version='1.0.0.qualifier'/>
      <provided namespace='osgi.bundle' name='pvu.bundle' version='1.0.0.qualifier'/>
      <provided namespace='osgi.identity' name='pvu.bundle' version='1.0.0.qualifier'>
        <properties size='1'>
          <property name='type' value='osgi.bundle'/>
        </properties>
      </provided>
      <provided namespace='org.eclipse.equinox.p2.eclipse.type' name='bundle' version='1.0.0'/>
    </provides>
    <requires size='2'>
      <requiredProperties namespace='osgi.ee' match='(&amp;(osgi.ee=JavaSE)(version=17))'>
        <description>
          pvu.bundle
        </description>
      </requiredProperties>
      <required namespace='org.eclipse.equinox.p2.iu' name='configure.pvu.bundle' range='0.0.0'/>
    </requires>
    <artifacts size='1'>
      <artifact classifier='osgi.bundle' id='pvu.bundle' version='1.0.0.qualifier'/>
    </artifacts>
    <touchpoint id='org.eclipse.equinox.p2.osgi' version='1.0.0'/>
    <touchpointData size='1'>
      <instructions size='2'>
        <instruction key='zipped'>
          true
        </instruction>
        <instruction key='manifest'>
          Bundle-SymbolicName: pvu.bundle;singleton:=true&#xA;Bundle-Version: 1.0.0.qualifier&#xA;
        </instruction>
      </instructions>
    </touchpointData>
  </unit>
</units>
```

laeubi:
Looking at the project resolved target platform it contains the bundle itself (with the additional requirement) but not the units from `p2.inf` so the advice get applied, but the additional units seem to be lost.

laeubi:
@mdaloia thanks for the testcase I have now added this and a fix to this PR:
- https://github.com/eclipse-tycho/tycho/pull/2989

laeubi:
Backport is now here:
- https://github.com/eclipse-tycho/tycho/pull/2991

so we can close this now!

PR #2989, opened 2023-11-03
Pr 2978

laeubi:
This includes the test and a fix for
- https://github.com/eclipse-tycho/tycho/pull/2978

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetPlatformFactoryImpl.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetPlatformFactoryImpl.java
@@ -41,8 +41,8 @@ import org.apache.maven.project.MavenProject;
 import org.codehaus.plexus.logging.Logger;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.URIUtil;
-import org.eclipse.equinox.internal.p2.artifact.repository.simple.SimpleArtifactDescriptor;
 import org.eclipse.equinox.p2.core.IProvisioningAgent;
 import org.eclipse.equinox.p2.core.ProvisionException;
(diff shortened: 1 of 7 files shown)


laeubi:
Jenkins testfailure is an instability from
- https://github.com/eclipse-equinox/p2/pull/364

mdaloia:
@laeubi that was a really fast resolution!!! Thank you very much.

Is there any plan to release a 4.0.4 soon? Just asking to plan ahead on my side.

tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetPlatformFactoryImpl.java
```diff
@@ -291,6 +296,31 @@ public P2TargetPlatform createTargetPlatform(TargetPlatformConfigurationStub tpC
         return targetPlatform;
     }
 
+    private void gatherP2InfUnits(ReactorProject reactorProject, Set<IInstallableUnit> externalUIs) {
+        if (reactorProject == null) {
+            return;
+        }
+        AdviceFileAdvice advice;
+        if (PackagingType.TYPE_ECLIPSE_PLUGIN.equals(reactorProject.getPackaging())) {
+            advice = new AdviceFileAdvice(reactorProject.getArtifactId(), Version.parseVersion("1.0.0"),
```
mdaloia, line 305: Is this hardcoded version `1.0.0` irrelevant? what if the declared unit sets its version to anything else?
(same below on the `else` branch of the `if`)

laeubi:
The plan was to release it last week :-D

But I considered this important enough to delay it by some days, if nothing crucial came up I'll probably make a release at the weekend or on Monday.

tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetPlatformFactoryImpl.java
laeubi: The version is only used to check if the advice should be applied to a unit, as we are not interested in the advice itself but only the additional units it defines the version has actually no influence here.

Issue #3019, opened 2023-11-08
ApiAnalysisMojo fails with API error if build.properties has missing 'output.. = bin/' entry


jukzi:
Without "output.. = bin/" all builds fail

for example
https://github.com/eclipse-platform/eclipse.platform/pull/841
https://github.com/eclipse-platform/eclipse.platform.ui/pull/1286
https://github.com/eclipse-equinox/p2/pull/382

jukzi:
probably caused by
https://github.com/eclipse-tycho/tycho/pull/2996

jukzi:
@akurtakov can you please help?

jukzi:
pde only creates the file and uses JDT to interpret the javaproject output. JDT then adds the default value.

laeubi:
Correct so there is no default, you could be happy that Tycho is not that strict here and always includes the default folder, because according to the spec a missing `output.. = bin/` would mean the bundle should be empty == contain no classes at all.

jukzi:
In eclipse.jdt.debug\org.eclipse.jdt.debug\build.properties the workaround does not help 
There is already a output.jdimodel.jar
Adding a `output.. = bin/` would introduce a warning `Redundant entry for output folder 'bin/'. It does not correspond to any source folder in 'source..' entry.`

laeubi:
> Adding a output.. = bin/ would introduce a warning Redundant entry for output folder 'bin/'. It does not correspond to any source folder in 'source..' entry.

this build configuration is just odd... what it does is putting everything from `bin/`  into a jdimodel.jar

`output.jdimodel.jar = bin/ `

this jar then is used as the bundle content, if you look at the jar it looks like this:

![grafik](https://github.com/eclipse-tycho/tycho/assets/1331477/d45682cb-4048-4f2b-bc1c-391efa810d1a)


No idea why one needs `jdimodel.jar` and can't put it directly in the bundle, searching for `jdimodel.jar` in java files did not reveal any usage where it might be used (and even if in such case the bundle itself could be used), but as it uses `Eclipse-BundleShape: dir` it seem to want extract it somehow on the file system (again, no clue why the bundle itself is not suitable here).

But in general it should work, so is there a minimal PR where one can see the problem?

laeubi:
The problem might be here that multiple source(jar) folders map to the same output folder, will investigate this.

laeubi:
[...]

PR #3020, opened 2023-11-08
Add testcase to reproduce failing behavior

laeubi:
See 
- https://github.com/eclipse-tycho/tycho/issues/3019


laeubi:
Test now reproduces the behavior, will try to reduce the test case and propose a fix now.

PR #3028, opened 2023-11-10
Fix warnings and enable save actions to fix them

akurtakov:


akurtakov:
This seems to be unstable test.

PR #3039, opened 2023-11-14
Add a reproducer to show that PR #2978 was a partial fix

mdaloia:
The fix of PR #2978 was a partial fix for issue #2977. It fixed the case of a single bundled but when the bundle providing/requiring the virtual IU is required by another bundle it fails with the same error as before even with Tycho 4.0.4:

```
[ERROR]   Software being installed: pvumb.bundle2 1.0.0.qualifier
[ERROR]   Missing requirement: pvumb.bundle1 1.0.0.qualifier requires 'org.eclipse.equinox.p2.iu; configure.pvumb.bundle1 0.0.0' but it could not be found
[ERROR]   Cannot satisfy dependency: pvumb.bundle2 1.0.0.qualifier depends on: osgi.bundle; pvumb.bundle1 0.0.0
```

A workaround is to use:

```
requires.0.optional=true
requires.0.greedy=true
```

This PR provides a new Integration Test case that shows this scenario. Previous one was refactored a little bit to avoid duplications.

### DIFF
--- a/tycho-its/projects/p2Inf.virtualUnit.multiBundle/bundle1/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/p2Inf.virtualUnit.multiBundle/bundle1/META-INF/MANIFEST.MF
@@ -1,0 +1,7 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: pvumb.bundle1
+Bundle-SymbolicName: pvumb.bundle1;singleton:=true
+Bundle-Version: 1.0.0.qualifier
+Bundle-Vendor: TEST
+Bundle-RequiredExecutionEnvironment: JavaSE-17
--- a/tycho-its/projects/p2Inf.virtualUnit.multiBundle/bundle1/META-INF/p2.inf
+++ b/tycho-its/projects/p2Inf.virtualUnit.multiBundle/bundle1/META-INF/p2.inf
@@ -1,0 +1,11 @@
+# Create the virtual IU
+units.0.id=configure.pvumb.bundle1
+units.0.version=1.0.0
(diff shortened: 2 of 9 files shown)


laeubi:
@mdaloia thanks for the test-case, in general I wonder why 

> A workaround is to use:

you think this is a workaround, can you maybe explain why a hard requirement to artificial IUs is used in your setup? This seems more that only one bundle should have it and then add that to a feature (or even let the feature host the `p2.inf`) instead of several bundles requiring this IU?

mdaloia:
@laeubi our case to use this artificial IU is to add some p2 instructions that changes the `--launcher.library` only for Linux product. The p2.inf with these instructions are in a plugin (not a feature). I don't know why other plug-ins are requiring it (I guess that it must be because the host/root plugin is required by others plug-ins so transitively it is required).
I thought on creating a fragment also as another way (I haven't tested if it would have worked or not) but it seemed too cumbersome having another bundle just for a p2.inf file.

laeubi:
> I don't know why other plug-ins are requiring it (I guess that it must be because the host/root plugin is required by others plug-ins so transitively it is required)

Sounds like a possible option ...

> The p2.inf with these instructions are in a plugin (not a feature)

If your product is based on features can you probably try to move the p2.inf there (just put it to the root + bin includes) to see how that behaves?

laeubi:
I now created
- https://github.com/eclipse-tycho/tycho/pull/3121

to discover extra IUs from other bundles as well, lets see if that fixes the issue.

laeubi:
This PR is included here:
- https://github.com/eclipse-tycho/tycho/pull/3121

And backported here:
- https://github.com/eclipse-tycho/tycho/pull/3144

Issue #3040, opened 2023-11-14
Target platform "system_property" feature doesn't work for update sites with addIUTargetRepositoryReferences


kysmith-csg:
I have in my target definition file an update site defined as

```xml
    <location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
        <repository location="https://download.eclipse.org/releases/${system_property:eclipse.version}"/
        <unit id="org.eclipse.sdk.feature.group" version="0.0.0"/>
    </location>
```

where `eclipse.version` is defined in `.mvn/maven.config` as `-Declipse.version=2023-09`. This works fine during target resolution and the build will succeed.

In the same reactor build, I have an update site project that has the following property set:
```xml
            <plugin>
                <groupId>org.eclipse.tycho</groupId>
                <artifactId>tycho-p2-repository-plugin</artifactId>
                <version>${tycho.version}</version>
                <configuration>
                    <addIUTargetRepositoryReferences>true</addIUTargetRepositoryReferences>
                </configuration>
            </plugin>
```

If I have the location site defined with the `${system_property:eclipse.version}`, the build will now fail because I guess this mojo doesn't resolve the properties and just uses the url literally.

> [ERROR] Failed to execute goal org.eclipse.tycho:tycho-p2-repository-plugin:4.0.4:assemble-repository (default-assemble-repository) on project com.example.my-project: Execution default-assemble-repository of goal org.eclipse.tycho:tycho-p2-repository-plugin:4.0.4:assemble-repository failed: Illegal character in path at index 39: https://download.eclipse.org/releases/${system_property:eclipse.version}

laeubi:
@kysmith-csg thats an interesting case indeed, do you like to try providing a PR for that (at best 
with an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue)? 

We should already have an abstraction about the variable resolving itself but if you have any problems let me know.

kysmith-csg:
#3042

laeubi:
I assume this is fixed now.

@kysmith-csg Can youtry the current [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) build if you still see any issues?

kysmith-csg:
It does work now with the latest 4.0.5-SNAPSHOT now, thanks.

content.xml in my eclipse-repository project
```xml
  <references size='4'>
[...]

PR #3042, opened 2023-11-15
Resolve target platform variables during p2 assembly (#3040)

kysmith-csg:
This updates the p2-repository-plugin:assemble-repository mojo to resolve any variables in the target definition locations just as the target-platform-configuration is able to do.

I modified the existing tests for target variables to include this because they are related and should both work together.


tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
```diff
@@ -349,8 +353,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
                             .flatMap(tpFile -> tpFile.getLocations().stream())
                             .filter(InstallableUnitLocation.class::isInstance).map(InstallableUnitLocation.class::cast)
                             .flatMap(iu -> iu.getRepositories().stream())
-                            .filter(iuRepo -> autoReferencesFilter.test(iuRepo.getLocation()))
-                            .map(iuRepo -> new RepositoryReference(null, iuRepo.getLocation(), true))
+                            .filter(iuRepo -> autoReferencesFilter.test(varResolver.resolve(iuRepo.getLocation())))
+                            .map(iuRepo -> new RepositoryReference(null, varResolver.resolve(iuRepo.getLocation()),
+                                    true))
```
HannesWell, line 358: Lets save the second variable substitution.
```suggestion
                            .map(iuRepo -> varResolver.resolve(iuRepo.getLocation())).filter(autoReferencesFilter)
                            .map(location -> new RepositoryReference(null, location, true))
```

HannesWell (commented):
In general this looks good, just one suggestion below.

Furthermore it would be good to verify in the Test that the URL is properly resolved.
You could do it like in `RepoRefLocationP2RepositoryIntegrationTest`.

Is the plan to also apply this to Tycho 5? Because if you had opened the PR against the master it could have probably be back-ported automatically.

kysmith-csg:
> Is the plan to also apply this to Tycho 5?

I guess so yeah. I'm just more used to fixing things the opposite way (lower branches first then merge-forward).

HannesWell:
> > Is the plan to also apply this to Tycho 5?
> 
> I guess so yeah. I'm just more used to fixing things the opposite way (lower branches first then merge-forward).

Understandable. But for back-ports from Tycho 5 to 4 an automated process exists where you just have to apply a lable and some GH actions do the rest.
AFAIK for the opposite way there is only the manual process.

Can you please also update the test case that runs the modified test-build to check for the resolved reference as indicated in my previous comment?
Then this change would be perfectly fine for me.

HannesWell (approved):
Pefect. Thanks.
Can you please squash all the there commits into one with a suitable Message/description for the overall change?

laeubi:
@kysmith-csg also please rebase this on top of master, Tycho 4 is more ore less in maintenance mode so we usually do not implement new features to Tycho 4 but rather implement in Tycho 5 branch (where we can change everything) and then decide if it is save to be backported or requires some adjustments (e.g. keeping a previous default), that way it can't happen that something is lost for Tycho 5.

kysmith-csg:
> we usually do not implement new features to Tycho 4 but rather implement in Tycho 5 branch

I disagree that this is a new feature - it is most definitely a bug. But I agree that it should be on latest first then backported. I'm not super familiar with the system you guys have and didn't know they can be backported automatically. Maybe helpful to add to the contributing guide?

I'm not familiar at all with rebasing branches. I'll try to use the UI because seemingly I can change the branch. If it gets too messed up I'll just open a new PR.

kysmith-csg:
Yeah. This is why we can't trust UIs :) remaking the PR...

laeubi:
> I disagree that this is a new feature - it is most definitely a bug.

Whatever we classify it, we need it in both branches as the bug is present in both of them, we have an automated way to backport (compatible) thing from master > Tycho 4 branch so usually it is easier to show that the bug still exits on master and then let it automatically backport :-)

HannesWell:
> Maybe helpful to add to the contributing guide?

I think so. Do you want to provide a PR?
There is already a [backporting-section](https://github.com/eclipse-tycho/tycho/blob/master/CONTRIBUTING.md#backporting) but that is more general



> Yeah. This is why we can't trust UIs :) remaking the PR...

You probably would have to 'move' your commit to the master branch locally and force-push it. Moving can be done by switching to the master branch and cherry-picking the commit.
Any way, your approach worked as well 👍🏽

PR #3043, opened 2023-11-15
Adapt to P2 enhancements in Eclipse 2023-12

HannesWell:
And add tests for support of Product Update-site names from the tycho-4.0.x branch.

This can be submitted once the Eclipse 2023-12 SimRel, which includes the new P2 is released.

The back-port to Tycho 4 is done explicitly in https://github.com/eclipse-tycho/tycho/pull/3125, since slightly different things have to be done.


pom.xml
```diff
@@ -78,6 +78,12 @@
 		<clean-plugin.version>2.4.1</clean-plugin.version>
 		<surefire-plugin.version>${surefire-version}</surefire-plugin.version>
 	</properties>
+	<repositories>
+		<repository>
+			<id>Eclipse-SNAPSHOTS</id>
+			<url>https://repo.eclipse.org/content/repositories/eclipse-snapshots</url>
+		</repository>
+	</repositories>
```
HannesWell, line 86: This must be removed before submission.

HannesWell: Done.

HannesWell:
The test was failing because through some dependency paths (I didn't investigated which one exactly) old and new versions of the `LocalMetadataRepository` class ended up in tycho.core and the old version of the class of course didn't implement `removeReferences()`, which is why the test failed (so they are working).
In order to fix that I just bumped all p2 dependencies to their latest version with this PR.

PR #3049, opened 2023-11-17
Resolve target definition variables during p2 repository assembly

kysmith-csg:
Recreation of #3042 to account for new base branch (master vs v4.0.x).

PR #3051, opened 2023-11-17
Backport #3049 to 4.0.x

kysmith-csg:
This backports #3049 from master to 4.0.x due to merge conflicts.
Original issue #3040.


laeubi:
@kysmith-csg thanks for this fix, the new snapshot should be available in a few hours so you can test if this.

PR #3055, opened 2023-11-18
Remove the sisu-osgi/sisu-equinox-embedder

laeubi:
Tycho has no longer any use of it as we have migrated completely to the usage of the osgi-connect technique that is much more flexible.


laeubi:
@akurtakov any objections as PL?

akurtakov:
+1
Probably not for 4.x though

laeubi:
> Probably not for 4.x though

Yes only for 5.x

PR #3085, opened 2023-11-23
Fixed non-quotes element in TargetPlatform.md

vogella:


vogella:
Looks like the check does not find xml tags in regular text. Sorry for the noise @laeubi

laeubi:
No problem, the issue is more that we do not have a check for building the docs.

PR #3087, opened 2023-11-23
Fixed non-quoted element in TargetPlatform.md

vogella:


vogella:
Now running `mvn -U -V -e -B -ntp clean install site site:stage --file tycho/pom.xml -T1C -DskipTest` to see if that is the lsat change.

vogella:
Local build with the command was successful. @laeubi please merge

Issue #2797, opened 2023-09-14
src/main/resources/META-INF/plexus/components.xml default lifecycle bindings use very old plugin versions


sratz:
The default lifecycle bindings
https://github.com/eclipse-tycho/tycho/blob/7f55e0b96415a33482bd14341fc192f0a21e26e1/tycho-maven-plugin/src/main/resources/META-INF/plexus/components.xml
use quite old versions of maven plugins, defined in the super pom:
https://github.com/eclipse-tycho/tycho/blob/7f55e0b96415a33482bd14341fc192f0a21e26e1/pom.xml#L75-L78

This, for example causes warnings such as
```
[INFO] ---------------------------[ eclipse-plugin ]---------------------------
[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but this         *
[WARNING] * project contains the following plugin(s) that have goals not  *
[WARNING] * marked as thread-safe to support parallel execution.          *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against Apache Maven.                           *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked as thread-safe in com.foo.bar:
[WARNING]   org.apache.maven.plugins:maven-deploy-plugin:2.5
[WARNING]
[WARNING] Enable debug to see precisely which goals are not marked as thread-safe.
[WARNING] *****************************************************************
```

Should these be updated to the latest versions?

Is there a way to have this done automatically (or via dependabot)?

laeubi:
In general it is good practice to not rely on the default lifecycle bindings versions but define the one you want in your main pom file.

Nerveless I don't see a reason to not upgrade to the latest version.

sratz:
Typically, I don't want to influence the bindings that Tycho choses, if I don't want to change the configuration of those plugins.

However
```
     <pluginManagement>
       <plugins>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-deploy-plugin</artifactId>
           <version>3.1.1</version>
         </plugin>
       </plugins>
     </pluginManagement>
```
in the super pom appears to be enough to force a newer version (without influencing the bindings).

laeubi:
> @laeubi These should upgraded to their latest versions , no need to carry old/ancient versions of maven plugins

[...]

PR #3093, opened 2023-11-27
Update the default plugin versions used in tycho lifecycle

laeubi:
Fix https://github.com/eclipse-tycho/tycho/issues/2797


laeubi:
@akurtakov what do you think should we backport this to tycho 4.0.4 ?

akurtakov:
> @akurtakov what do you think should we backport this to tycho 4.x ?

Absolutely!

laeubi:
Applied a possible fix for
- https://issues.apache.org/jira/browse/MINSTALL-169

PR #3125, opened 2023-12-05
[Tycho-4] Adapt to P2 enhancements in Eclipse 2023-12

HannesWell:
Back-port of https://github.com/eclipse-tycho/tycho/pull/3043, that also reverts the temporary changes of `PublishProductToolImpl` made in https://github.com/eclipse-tycho/tycho/pull/2929, which are obsolete now that the new P2 version has support for names of update-sites built in.


HannesWell:
> @HannesWell some test fail now :-\

Looks like this is a consequence of https://github.com/eclipse-equinox/p2/pull/310, but tycho still uses `pack.gz` repositories:
https://github.com/eclipse-tycho/tycho/tree/master/tycho-core/src/test/resources/repositories/packgz

So good news is, it is not a bug, the bad news the tests have to be rewritten...

Issue #3197, opened 2023-12-09
All statisfied requirements map to dependencies in InstallableUnitSlicer.computeDirectDependencies


merks:
There are two cases where a requirement should not map to a dependency.

1. When `Requirement.getMax() == 0`, i.e., for so-called negative requirements.
2. When `IRequirement.getFilter() != null` and the filter is no applicable.

Mapping those two cases inappropriately can and does result in dependency cycles that break a build (for builds that worked for Tycho 2.x and 3.x).

Currently `org.eclipse.tycho.p2maven.InstallableUnitSlicer.computeDirectDependencies(Collection<IInstallableUnit>, IQueryable<IInstallableUnit>)` does not have enough context to determine if a filter is applicable or not, so the best we can do in the absence of such information is to assume the filter isn't applicable.

laeubi:
@merks can you explain what context are needed? ws/os and maybe system properties?

laeubi:
@merks I now have analyzed the problem and it is actually a bit more complex as initially thought:

1. first you are right that negative requirements are handled as regular ones, but the more general problem is that Tycho does not consider the `req.getMax()` at all, this is quite easy to fix and I'll provide a PR for that.
2. You are also right that filters are not considered, I have a solution to evaluate them but this can't be applied at the moment because it revealed another problem here see below.
3. The root cause of all evil (beside the rather trivial problem 1.)  is that currently we compute the project dependencies for each project and then hold them in a map. Based on that we now compute the inter-dependencies and projects to consider as an additional requirement. Now the following can (and does happen, e.g. for SWT build):

[...]

PR #3198, opened 2023-12-09
Consider a requirement's filter and max when computing dependencies

merks:
https://github.com/eclipse-tycho/tycho/issues/3197


laeubi:
@merks thanks for looking into this, sadly this now fails to discover dependencies from SWT>Fragement (at least I suspect that's what the error in "Individual Bundles Build" seem to cause). so I think one can not completely ignore any requirement that has a filter.

laeubi:
@merks thanks for the testcase this was included here:
- https://github.com/eclipse-tycho/tycho/pull/3204

and should support full profile properties filtering.

merks:
Changes this this have superseded this attempt, and the following test case is much simpler:

https://github.com/eclipse-tycho/tycho/pull/3227

PR #3204, opened 2023-12-10
Pass the profile properties so they can be considered in the resolve

laeubi:
This includes the testcase and preliminary work from
- https://github.com/eclipse-tycho/tycho/pull/3198

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/DirectDependenciesResult.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/DirectDependenciesResult.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Christoph Läubrich and others.
+ * This program and the accompanying materials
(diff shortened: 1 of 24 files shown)


laeubi:
No the individual bundles build fails :-\
Lets see if the integration test reveal a failure to give some more indication what could be the cause...

laeubi:
Thinking more about the failure I think it is caused by the fact that one needs to check if any of the configured target environments can satisfy the filter...

merks:
Yes, if you look at test cases like this one you can see the environments are in the context IU:

image

So probably we should make a context IU for each environment combination specified along with the profile properties:

image

and then check that the filter matches one of such context IUs.

This PR appears to have lost the getMax filtering.   It's of course highly relevant to not create a dependency for a negative requirement.  The getMax() == 0 requirements could perhaps be filtered out in the first loop where the self requirements are being removed.

laeubi:
Work continues here:
- https://github.com/eclipse-tycho/tycho/pull/3239

Issue #2930, opened 2023-10-15
When trying to resolve a target platform from a p2 repo to be accessed over an authenticated mirror, Tycho fails to authenticate against the mirror.


kevloral:
I use Maven 3.9.1. Authentication worked fine when using Tycho 2.7.5 but began to fail with Tycho 3.0.0 and keeps failing using the latest release (4.0.3)

I have a Sonatype Nexus server that acts as a mirror for some p2 repositories. The mirror is declared like this in my Maven _settings.xml_ file:

```
<mirror>
	<id>nexus-p2</id>			
	<name>Nexus P2</name>
	<url>https://servidor.jma:8443/nexus/content/groups/global-p2/</url>
	<mirrorOf>eclipse_2022-06,eclipse_2023-06,mylyn_3.25,swtbot_snapshots,tpd_nightly</mirrorOf>
	<layout>p2</layout>
    <mirrorOfLayouts>p2</mirrorOfLayouts>
</mirror>
```

Credentials are also declared for that mirror in the same _settings.xml_ file (password hidden here for obvious reasons):

```
<server>
	<id>nexus-p2</id>
	<username>jm.arnesto</username>
	<password>********</password>
</server>
```

Now, if I try to build using Tycho 2.7.5, it works fine. However, as soon as I change the Tycho version (and only the version), I get the following error when trying to access the _eclipse_2023-06_ repository over the mirror:

```
Caused by: org.apache.maven.plugin.PluginExecutionException: Execution default of goal org.eclipse.tycho.extras:tycho-eclipserun-plugin:4.0.3:eclipse-run failed: Failed to load p2 repository with ID 'eclipse_2023-06' from location https://download.eclipse.org/releases/2023-06
[..]
Caused by: java.lang.RuntimeException: Failed to load p2 repository with ID 'eclipse_2023-06' from location https://download.eclipse.org/releases/2023-06
[..]
Caused by: org.eclipse.equinox.p2.core.ProvisionException: Unable to read repository at https://servidor.jma:8443/nexus/content/groups/global-p2/.
[..]
Caused by: org.eclipse.equinox.internal.p2.repository.AuthenticationFailedException
    at org.eclipse.tycho.p2maven.transport.SharedHttpCacheStorage$CacheLine.fetchFile (SharedHttpCacheStorage.java:259)
    at org.eclipse.tycho.p2maven.transport.SharedHttpCacheStorage$2.getCacheFile (SharedHttpCacheStorage.java:136)
[...]

PR #3221, opened 2023-12-10
Fix and test cases for #2930

kevloral:
Two test cases added to class _PasswordProtectedP2RepositoryTest_ for trying to reproduce the bug:

- _testMirror_: the test server is accessed over a mirror without authentication.
- _testAuthMirror_: the test server is accessed over a mirror with authentication. This the one that should fail.

 As things stand, they both pass, so they need improving.

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/DefaultMavenRepositorySettings.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/DefaultMavenRepositorySettings.java
@@ -18,6 +18,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 
 import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.artifact.repository.ArtifactRepositoryPolicy;
@@ -157,5 +158,10 @@ public class DefaultMavenRepositorySettings implements MavenRepositorySettings, 
 			settings = new Settings();
 			mirrors = Collections.emptyList();
 		}
+	}
+
+	@Override
+	public Stream<MavenRepositoryLocation> getMirrors() {
+		return mirrors.stream().map(m -> new MavenRepositoryLocation(m.getId(), URI.create(m.getUrl())));
 	}
 }
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/DefaultRepositoryIdManager.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/DefaultRepositoryIdManager.java
@@ -118,9 +118,10 @@ public class DefaultRepositoryIdManager implements IRepositoryIdManager {
 
     @Override
 	public Stream<MavenRepositoryLocation> getKnownMavenRepositoryLocations() {
-        return knownMavenRepositoryIds.entrySet().stream()
-                .map(e -> new MavenRepositoryLocation(e.getValue(), e.getKey()));
-    }
+		// Returns both repository and mirror locations
+		return Stream.concat(knownMavenRepositoryIds.entrySet().stream()
+				.map(e -> new MavenRepositoryLocation(e.getValue(), e.getKey())), settings.getMirrors());
+	}
 
 	@Override
 	public MavenRepositorySettings getSettings() {
--- a/tycho-api/src/main/java/org/eclipse/tycho/MavenRepositorySettings.java
+++ b/tycho-api/src/main/java/org/eclipse/tycho/MavenRepositorySettings.java
@@ -14,6 +14,7 @@
 package org.eclipse.tycho;
 
 import java.net.URI;
+import java.util.stream.Stream;
 
 /**
  * Provides the mirror configuration and credentials from the Maven settings for loading remote p2
@@ -59,5 +60,10 @@ public interface MavenRepositorySettings {
      * Returns the configured credentials for the given repository, or <code>null</code>.
      */
     Credentials getCredentials(MavenRepositoryLocation location);
+
+    /**
+     * Returns all configured mirror locations.
+     */
+    Stream<MavenRepositoryLocation> getMirrors();
 
 }
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/target/PasswordProtectedP2RepositoryTest.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/target/PasswordProtectedP2RepositoryTest.java
@@ -13,7 +13,6 @@
 package org.eclipse.tycho.test.target;
 
 import java.io.File;
-import java.util.Properties;
 
 import org.apache.maven.it.Verifier;
 import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
@@ -29,15 +28,57 @@ public class PasswordProtectedP2RepositoryTest extends AbstractTychoIntegrationT
 	private HttpServer server;
 	private String p2RepoUrl;
 
+	private HttpServer mirror;
+	private String p2MirrorUrl;
+
+	private HttpServer authMirror;
+	private String p2AuthMirrorUrl;
+
 	@Before
 	public void startServer() throws Exception {
 		server = HttpServer.startServer("test-user", "test-password");
 		p2RepoUrl = server.addServer("foo", ResourceUtil.resolveTestResource("repositories/e342"));
+
+		mirror = HttpServer.startServer();
+		p2MirrorUrl = mirror.addServer("bar", ResourceUtil.resolveTestResource("repositories/e342"));
+
+		authMirror = HttpServer.startServer("mirror-user", "mirror-password");
+		p2AuthMirrorUrl = authMirror.addServer("bar", ResourceUtil.resolveTestResource("repositories/e342"));
 	}
 
 	@After
 	public void stopServer() throws Exception {
+		authMirror.stop();
+		mirror.stop();
 		server.stop();
+	}
+
+	/**
+	 * Tries to access a p2 repository over an authenticated mirror.
+	 * 
+	 * @throws Exception
+	 */
+	@Test
+	public void testAuthMirror() throws Exception {
+		Verifier verifier = createVerifier("settings-auth-mirror.xml");
+		verifier.setSystemProperty("p2.authMirror", p2AuthMirrorUrl);
+		verifier.addCliOption("-P=repository");
+		verifier.executeGoal("package");
+		verifier.verifyErrorFreeLog();
+	}
+
+	/**
+	 * Tries to access a p2 repository over a mirror with no authentication.
+	 * 
+	 * @throws Exception
+	 */
+	@Test
+	public void testMirror() throws Exception {
+		Verifier verifier = createVerifier("settings-mirror.xml");
+		verifier.setSystemProperty("p2.mirror", p2MirrorUrl);
+		verifier.addCliOption("-P=repository");
+		verifier.executeGoal("package");
+		verifier.verifyErrorFreeLog();
 	}
 
 	@Test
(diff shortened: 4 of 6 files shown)


laeubi:
@kevloral thnaks for working on this, if I understand correctly from your example you have the case that there is one (not authenticated) p2 site that is mirrored but the mirror is password protected.

I would suggest that your _copy_ the `PasswordProtectedP2RepositoryTest` and maybe resources (for example as `PasswordProtectedMirrorP2RepositoryTest`) and replicate it as much as possible, e.g. you can use the latest eclipse as the not password protected site as in your real example, then add a setting to mirror that to the local test server and then assert that the test server was accessed.

kevloral:
Not exactly: the p2-repo that is accessed through the mirror is always password protected (_test-user/test-password_). It is the mirror that may be protected or not:

- _testMirror_: it is trying to access a password protected p2-repo (_test-user/test-password_) through a mirror. The mirror is not password protected.
- testAuthMirror: it is trying to access a password protected p2-repo (test-user/test-password) through a mirror. The mirror is password protected (_mirror-user/mirror-password_). Note that, even though both are password protected, the username and passwords are different (just in case).

Anyway, these test cases pass even though I expected them not to. Therefore, I am going to add some logging to the _MavenAuthenticator_ class to see if the problem is actually in there (and, also, to try to find out why those tests pass). Once I know more, I will change those tests if needed or add others that manage to reproduce the bug.

kevloral:
Two more test cases added. The second one (_testTargetDefinitionAuthMirror_) successfully reproduces the bug:

- _testTargetDefinitionMirror_: this one tries to resolve a target definition from a p2 repo accessed over a mirror with no authentication. It passes.

- _testTargetDefinitionAuthMirror_: this one tries to resolve a target definition from a p2 repo accessed over an authenticated mirror. It fails, as this is exactly the case that is being reported on the associated bug.

kevloral:
Fixed. 

When _MavenAuthenticator_ tried to get the credentials for a given URI (method _getServerCredentials_), it asked _DefaultRepositoryIdManager_ for the list of known maven locations (method _getKnownMavenRepositoryLocations_) so it could compare the URIs of those locations with the one credentials are being requested for (method _uriPrefixMatch_). 

That list of known locations did not include mirrors, even though methods _getEffectiveLocation_ and _getEffectiveLocationAndPrepareLoad_ do return mirrors when needed. Therefore, credentials were never returned by method method _getServerCredentials_ when the URI belonged to a mirror because the URIs never matched.

Now, whenever a new mapping is added to _DefaultRepositoryIdManager_ (method _addMapping_), the corresponding mirror, if any, is stored. When the list of known locations is requested, method _getKnownMavenRepositoryLocations_ returns both repos and mirrors and, therefore, method _getServerCredentials_ now return the credentials because the URIs do match.

tycho-its/src/test/java/org/eclipse/tycho/test/target/PasswordProtectedP2RepositoryTest.java
```diff
@@ -85,6 +135,8 @@ private Verifier createVerifier(String settingsFile, String settingsSecurityFile
 				new File("projects/target.httpAuthentication/" + settingsFile));
 		Properties systemProperties = verifier.getSystemProperties();
 		systemProperties.setProperty("p2.repo", p2RepoUrl);
+		systemProperties.setProperty("p2.mirror", p2MirrorUrl);
```
laeubi, line 138: it is required to modify the existing test-case here?
Can it maybe duplicated into one that uses a mirror and one that do not use a mirror?

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/DefaultRepositoryIdManager.java
```diff
@@ -52,6 +57,15 @@ public void addMapping(String mavenRepositoryId, URI location) {
...
                     + "', now is '" + mavenRepositoryId + "'");
         }
+
+		// Checks if a mirror exists for the given location
+		MavenRepositoryLocation repositoryLocation = new MavenRepositoryLocation(mavenRepositoryId, key);
+		MavenRepositoryLocation mirrorLocation = settings.getMirror(repositoryLocation);
+		if (mirrorLocation != null) {
+			// A mirror exists. Adds its id to the map of known mirror ids.
+			URI mirrorKey = normalize(mirrorLocation.getURL());
+			knownMavenMirrorIds.put(mirrorKey, mirrorLocation.getId());
```
laeubi, line 67: instead of having an extra map, can we probably just add it to the existing `knownMavenRepositoryIds` map?
Also this should include the above check about duplicate mapped ids.

laeubi (changes requested):
Many thanks for the test-case and working on a fix, it looks good in general but could maybe be improved a bit.

Beside that it would be good if you

1. target the main branch, the change can then be backported automatically to tycho-4 branch
2. squash your commits into one with a combined commit message that includes what was changed.

tycho-its/src/test/java/org/eclipse/tycho/test/target/PasswordProtectedP2RepositoryTest.java
kevloral: Do you mean implementing these new tests into a whole new class? I mean, the mirrors are initialized and stopped in common methods (_startServer_ and _StopServer_), so I didn't think adding these two new properties to _createVerifier_ would be a problem. After all, these new properties are only used to resolve their value in the settings file used by the specific test. If they are not present, they do nothing.

Now, could we generalize the _createVerifier_ method so the properties it needs to set are passed as arguments from the individual tests? That might be interesting, even though I am not sure the added complexity would be worth it.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/DefaultRepositoryIdManager.java
kevloral: Sure, I can use just one map. I implemented it this way because, initially, I was tempted to propose modifying the _IRepositoryIdManager_ interface so that repositories were not mixed with mirrors in the _getKnownMavenRepositoryLocations_ method.

Regarding duplicate ids when dealing with mirrors, the same mirror may be used for several repository locations and that is something I don't think there is any need to warn the user about.

tycho-its/src/test/java/org/eclipse/tycho/test/target/PasswordProtectedP2RepositoryTest.java
kevloral: By the way, I don't know if you find tests _testMirror_, _testAuthMirror_ and _testTargetDefinitionMirror_ worth merging or if you think I should remove them altogether. After all, the only one that actually reproduces the bug is _testTargetDefinitionAuthMirror_.

kevloral:
I will follow your suggestions and prepare a new PR targeting main instead. Thanks a lot for your comments.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/DefaultRepositoryIdManager.java
laeubi: Thinking a bit more about that I wonder if we need to store the mirror IDs at all... the id map is actually there because in some cases we don't know the id of the server, but for a _mirror_ we can simply fetch all mirrors from the current maven session like we do in `DefaultMavenRepositorySettings.initialize()`.

So I think `MavenRepositorySettings` should get a new method `MavenRepositorySettings#getMirrors()` and this can then be used in `getKnownMavenRepositoryLocations()` to simply return all mirror `id` -> `url` mappings.

laeubi:
> I will follow your suggestions and prepare a new PR targeting main instead. Thanks a lot for your comments.

If you squash all your commits into one and then rebase on `main` you can also just edit this PR I think.

laeubi:
> By the way, I don't know if you find tests testMirror, testAuthMirror and testTargetDefinitionMirror worth merging or if you think I should remove them altogether. 

We should just keep them, even though they do not show the bug behavior they prove these cases already work.

laeubi:
@kevloral can you rebase and adjust it to use the `MavenRepositorySettings#getMirrors()` aproach? I want to make a new release this month and it would be good to include your changes there!

kevloral:
> @kevloral can you rebase and adjust it to use the `MavenRepositorySettings#getMirrors()` aproach? I want to make a new release this month and it would be good to include your changes there!

Of course. I will make that change as well as the others you suggested in the next few days.

tycho-its/src/test/java/org/eclipse/tycho/test/target/PasswordProtectedP2RepositoryTest.java
kevloral: The system properties containing the mirror URLs (_p2.mirror_ and _p2.authMirror_) are no longer set in the _createVerifier_ method but in the tests themselves. In addition, the _createVerifier_ method has been slightly simplified by using the _setSystemProperty_ method belonging to the _Verifier_ class.

tycho-api/src/main/java/org/eclipse/tycho/MavenRepositorySettings.java
```diff
@@ -60,4 +60,9 @@ public String toString() {
      */
     Credentials getCredentials(MavenRepositoryLocation location);
 
+    /**
+     * Returns all configured mirror locations.
+     */
+    Iterable<MavenRepositoryLocation> getMirrors();
```
laeubi, line 66:
```suggestion
    Stream<MavenRepositoryLocation> getMirrors();
```

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/DefaultMavenRepositorySettings.java
```diff
@@ -158,4 +159,14 @@ public void initialize() throws InitializationException {
 			mirrors = Collections.emptyList();
 		}
 	}
+
+	@Override
+	public Iterable<MavenRepositoryLocation> getMirrors() {
```
laeubi, line 164:
```suggestion
	public Stream<MavenRepositoryLocation> getMirrors() {
```

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/DefaultMavenRepositorySettings.java
```diff
@@ -158,4 +159,14 @@ public void initialize() throws InitializationException {
...
+		List<MavenRepositoryLocation> result = new ArrayList<>();
+		for (Mirror mirror : mirrors) {
```
laeubi, line 166:
```suggestion
		return mirrors.stream().map(mirror->new MavenRepositoryLocation(mirror.getId(), URI.create(mirror.getUrl()));
```

laeubi:
@kevloral thanks for the changes, just one more adjustment and I think its fine, beside that, can you please squash + rebase the changes into one commit so we can easier backport the changes?

kevloral:
Sure. I will replace that Iterable with a _Stream_ and squash the commits into one. By the way, it is already rebased on _main_ even though my branch is still called [kevloral:tycho-4.0.x](https://github.com/kevloral/tycho/tree/tycho-4.0.x).

laeubi:
@kevloral I plan to do a bugfix release soon and your change seems important enough to include to you think you can finish it?

kevloral:
> @kevloral I plan to do a bugfix release soon and your change seems important enough to include to you think you can finish it?

Sorry for the delay. Other commitments got in the way. As you requested, I have replaced the _Iterable_ with a _Stream_ and squashed the commits into one. I guess it is ready to be merged now. 

Thanks a lot for your help.

laeubi:
@kevloral thanks for the testcase and bugfix!

PR #3229, opened 2023-12-11
Document (better) referencedRepositoryMode in TargetPlatformConfigura…

LorenzoBettini:
…tionMojo.java

This should hopefully improve the documentation of referencedRepositoryMode (it was not clear to me which values to specify)


laeubi:
@LorenzoBettini as you seem to using that option already it is currently off by default, but maybe you thing we should make it on by default for Tycho 5?

LorenzoBettini:
@laeubi OK. So, if I understand correctly in that PR I should make `include` the default (I seem to understand it is enough to make that element the first one in the enum), change the Javadoc accordingly and update the release notes for 5 accordingly (of course, I guess also a few IT will have to be updated). Is that correct?

PR #3246, opened 2023-12-12
make "include" the default enum value

LorenzoBettini:


LorenzoBettini:
@laeubi on my fork, GitHub Actions shows several IT failures, which I don't understand https://github.com/LorenzoBettini/tycho/actions/runs/7180800199/job/19556904813

The failure seems to be due to

```
[ERROR] unsupported URI https:/download.eclipse.org/releases/2023-12/202312061001/plugins/org.eclipse.emf.ecore_2.35.0.v20230829-0934.jar
java.lang.IllegalArgumentException: unsupported URI https:/download.eclipse.org/releases/2023-12/202312061001/plugins/org.eclipse.emf.ecore_2.35.0.v20230829-0934.jar
    at jdk.internal.net.http.common.Utils.newIAE (Utils.java:286)
    at jdk.internal.net.http.HttpRequestBuilderImpl.checkURI (HttpRequestBuilderImpl.java:85)
    at jdk.internal.net.http.HttpRequestBuilderImpl.uri (HttpRequestBuilderImpl.java:71)
    at jdk.internal.net.http.HttpRequestBuilderImpl.uri (HttpRequestBuilderImpl.java:43)
    at org.eclipse.tycho.p2maven.transport.Java11HttpTransportFactory.createTransport (Java11HttpTransportFactory.java:78)
    at org.eclipse.tycho.p2maven.transport.SharedHttpCacheStorage$CacheLine.fetchFile (SharedHttpCacheStorage.java:242)
    at org.eclipse.tycho.p2maven.transport.SharedHttpCacheStorage$2.getCacheFile (SharedHttpCacheStorage.java:133)
    at org.eclipse.tycho.p2maven.transport.HttpTransportProtocolHandler.getFile (HttpTransportProtocolHandler.java:50)
    at org.eclipse.tycho.p2maven.transport.TychoRepositoryTransport.stream (TychoRepositoryTransport.java:141)
    at org.eclipse.tycho.p2maven.transport.TychoRepositoryTransport.download (TychoRepositoryTransport.java:104)
...
```

I don't understand where that invalid URI comes from; I guess that with the default use of referenced repositories the problem shows up.

akurtakov:
@LorenzoBettini It's most likely my fault caused by https://github.com/eclipse-tycho/tycho/pull/3224 which https://github.com/eclipse-tycho/tycho/pull/3244 should fix or at least reduce the issue significantly. 
Sorry about that, I'm trying to resolve it ASAP!

akurtakov:
Java enums don't have default values so I'm closing this PR as it is not changing anything.

LorenzoBettini:
@akurtakov isn't the default value for enum 0, which corresponds to the first element of an enum (unless a value is specified, and then automatically incremented)? If there's no default value then why, without my PR, the default is `ignore`?

akurtakov:
It is explicitly set to ignore at https://github.com/eclipse-tycho/tycho/blob/cf8cffa66da35bf674fc61e62af3668df5540a35/tycho-core/src/main/java/org/eclipse/tycho/core/TargetPlatformConfiguration.java#L122 .
Enums has ordinal value as you think but the one with ordinal 0 is not the default as there is no way to express that in Java:
* EnumType enumVar; - construct will give you "error: variable enumVar might not have been initialized" compile time error by default
* EnumType enumVar = null; - will throw NPE at execution time
* EnumType enumVar = EnumType.NAME_1; - will explicitly select one of the enum values and is the only usable construct.

LorenzoBettini:
@akurtakov I had missed that file: I was looking at the target platform configuration mojo where no default was specified and I thought maven would use the first value of the enum. I forgot that mojo is only for documentation purpose.. I'll work on another PR.

laeubi:
>  I was looking at the target platform configuration mojo where no default was specified and I thought maven would use the first value of the enum.

The default for maven is `null` in such case.

> I forgot that mojo is only for documentation purpose.. 

Yes it is a bit unfortunate at the moment  but I haven't found a clean maven solution yet, maybe I can convince the maven devs for Maven4 to offer something better :-)

> I'll work on another PR.

Thanks, please also include an release note changes in such PR!

PR #3295, opened 2023-12-20
Restoring the option to ignore p2 mirrors via the Maven settings

vogella:
Commit 7e6d5953f34af4e2f6dc3ad45aab7c995a1fbc96 has removed the option to use the settings.xml to disable the usage of p2 mirrors.

See https://github.com/eclipse-tycho/tycho/wiki/Frequently-asked-questions#how-do-i-disable-p2-mirrors

Restores this option and give the command line option priority.

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteArtifactRepositoryManagerAgentFactory.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteArtifactRepositoryManagerAgentFactory.java
@@ -12,6 +12,7 @@
  *******************************************************************************/
 package org.eclipse.tycho.p2maven.transport;
 
+import org.apache.maven.plugin.LegacySupport;
 import org.codehaus.plexus.component.annotations.Component;
 import org.codehaus.plexus.component.annotations.Requirement;
 import org.codehaus.plexus.logging.Logger;
@@ -34,6 +35,9 @@ public class RemoteArtifactRepositoryManagerAgentFactory implements IAgentServic
 	@Requirement
 	MavenAuthenticator authenticator;
 
+	@Requirement
+	protected LegacySupport mavenContext;
+
 	@Override
 	public Object createService(IProvisioningAgent agent) {
 		IArtifactRepositoryManager plainRepoManager = (IArtifactRepositoryManager) new ArtifactRepositoryComponent()
@@ -45,14 +49,36 @@ public class RemoteArtifactRepositoryManagerAgentFactory implements IAgentServic
 	}
 
 	private boolean getDisableP2MirrorsConfiguration() {
-		String key = "tycho.disableP2Mirrors";
-		String value = System.getProperty(key);
-		if (value != null) {
-			logger.info("Using " + key
+		String deprecatedKey = "tycho.disableP2Mirrors";
+		String deprecatedValue = getMirrorProperty(deprecatedKey);
+
+		if (deprecatedValue != null) {
+			logger.info("Using " + deprecatedKey
 					+ " to disable P2 mirrors is deprecated, use the property eclipse.p2.mirrors instead, see https://tycho.eclipseprojects.io/doc/"
 					+ TychoVersion.getTychoVersion() + "/SystemProperties.html for details.");
-			return Boolean.parseBoolean(value);
+			return Boolean.parseBoolean(deprecatedValue);
+		}
+
+		String key = "eclipse.p2.mirrors ";
+		String value = getMirrorProperty(key);
+
+		if (value != null) {
+			// eclipse.p2.mirrors false -> disable mirrors
+			return !Boolean.parseBoolean(value);
 		}
 		return false;
+
+	}
+
+	private String getMirrorProperty(String key) {
+		String value = System.getProperty(key);
+		if (key == null && mavenContext.getSession() != null) {
+			key = mavenContext.getSession().getSystemProperties().getProperty(key);
+
+			if (key == null) {
+				key = mavenContext.getSession().getUserProperties().getProperty(key);
+			}
+		}
+		return value;
 	}
 }


laeubi:
Please use [eclipse.p2.mirrors](https://tycho.eclipseprojects.io/doc/main/SystemProperties.html#p2) property instead here.

Also it seems your are working on an outdated master branch.

vogella:
> Please use [eclipse.p2.mirrors](https://tycho.eclipseprojects.io/doc/main/SystemProperties.html#p2) property instead here.

eclipse.p2.mirrors is the new tycho.disableP2Mirrors? Do we want to support both or only the latest? Shall I also update the FAQ? https://github.com/eclipse-tycho/tycho/wiki/Frequently-asked-questions#how-do-i-disable-p2-mirrors	

> Also it seems your are working on an outdated master branch.
Ah, you moved to main, will update

p2-maven-plugin/pom.xml
```diff
@@ -207,6 +207,11 @@
 			<groupId>commons-net</groupId>
 			<artifactId>commons-net</artifactId>
 		</dependency>
+		<dependency>
```
laeubi, line 210: p2 plugin should not depend on tycho-core

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteArtifactRepositoryManagerAgentFactory.java
```diff
@@ -34,6 +35,9 @@ public class RemoteArtifactRepositoryManagerAgentFactory implements IAgentServic
 	@Requirement
 	MavenAuthenticator authenticator;
 
+	@Requirement
+	MavenContext mavenContext;
```
laeubi, line 39: please use `LegacySupport` to access the session directly to get its properties

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteArtifactRepositoryManagerAgentFactory.java
```diff
@@ -45,14 +49,33 @@ public Object createService(IProvisioningAgent agent) {
 	}
 
 	private boolean getDisableP2MirrorsConfiguration() {
-		String key = "tycho.disableP2Mirrors";
-		String value = System.getProperty(key);
-		if (value != null) {
-			logger.info("Using " + key
+		String deprecatedKey = "tycho.disableP2Mirrors";
+		String deprecatedValue = System.getProperty(deprecatedKey);
```
laeubi, line 53: its seems a good idea to refactor this and the follogiwn check into a method that:

1. first check for the property in the System.getProperty
2. then if we have a session
3. if session the use system/user properties from there

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteArtifactRepositoryManagerAgentFactory.java
```diff
@@ -45,14 +49,33 @@ public Object createService(IProvisioningAgent agent) {
...
+		String key = "eclipse.p2.mirrors ";
+		String value = System.getProperty(key);
+
+		if (value == null) {
+			value = mavenContext.getSessionProperties().getProperty(key);
+		}
+
+
+		if (value != null) {
+			// eclipse.p2.mirrors false -> disable mirrors
```
laeubi, line 75: One problem is that this will not really fully disable P2 mirrors so we probably need a P2 change here as well to make it really configurable.

laeubi (changes requested):

laeubi:
@vogella the next bugfix release will happen at end of this month, do you plan to finish your work here so it can be included?

vogella:
Looking into it now

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteArtifactRepositoryManagerAgentFactory.java
vogella: Is this the correct order? Should user property not have priority over system session property and system property?

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteArtifactRepositoryManagerAgentFactory.java
vogella: Not sure what to do here.

p2-maven-plugin/pom.xml
vogella: Done

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteArtifactRepositoryManagerAgentFactory.java
vogella: Done

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteArtifactRepositoryManagerAgentFactory.java
vogella: I updated with your suggestion, let me know if I should change the order or not

vogella:
> @vogella the next bugfix release will happen at end of this month, do you plan to finish your work here so it can be included?

Thanks for the ping and happy new year to you.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteArtifactRepositoryManagerAgentFactory.java
laeubi: if you search p2 code for `eclipse.p2.mirrors` you will see that this is only evaluated once as a system property, so even if you specify it Tycho will ignore mirrors, but P2 will still try to fetch the mirror list.

p2-maven-plugin/pom.xml
laeubi: Github still show a change here...

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteArtifactRepositoryManagerAgentFactory.java
```diff
@@ -45,14 +49,41 @@ public Object createService(IProvisioningAgent agent) {
...
 	private boolean getDisableP2MirrorsConfiguration() {
-		String key = "tycho.disableP2Mirrors";
-		String value = System.getProperty(key);
-		if (value != null) {
-			logger.info("Using " + key
+		String deprecatedKey = "tycho.disableP2Mirrors";
+		String deprecatedValue = System.getProperty(deprecatedKey);
+
+		if (deprecatedValue == null) {
+			deprecatedValue = mavenContext.getSession().getSystemProperties().getProperty(deprecatedKey);
```
laeubi, line 56: The session might be null so you should extract it and make sure it is only used when not null.

vogella:
Is tomorrow still ok? Today I'm working outside Eclipse world

laeubi:
Yes we still have some days until release, I just don'T want to delay it forever as theres always something new :-D

vogella:
I updated the change

laeubi (approved):
Looks good now and build is green.

PR #3299, opened 2023-12-22
Improve artifact comparison delta results

laeubi:
Currently only the text is written as detailed information into the artifactcomparision directory.

This is often not enough information to find the actual difference, to mitigate this the handling of the writing is now delegated to the delta and the following improvements are made:

- class file comparison writes the class file as well as the disassemble
- debug info is included in the disassemble
- zip comparator writes the baseline and build jars

Fix https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1688

### DIFF
--- a/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/artifactcomparator/ArtifactDelta.java
+++ b/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/artifactcomparator/ArtifactDelta.java
@@ -12,6 +12,9 @@
  *******************************************************************************/
 package org.eclipse.tycho.artifactcomparator;
 
+import java.io.File;
+import java.io.IOException;
+
 import org.eclipse.tycho.zipcomparator.internal.SimpleArtifactDelta;
 
 /**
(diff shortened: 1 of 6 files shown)


laeubi:
@iloveeclipse @srikanth-sankaran I'm not sure the current approach is 100% desired, while debug symbols might be useful for analysis, this will also lead to comparator errors if debug symbols change what seems undesirable.

I can change this to use a different settings for compare and one for the debug output. In general we can even use anything else from JDT that seems better suited to compare classfiles+generate diffs just let me know if there is something available, the comments in the gives some hints but they are likely outdated:

https://github.com/eclipse-tycho/tycho/blob/3cfab28b0c382d0b7f24bc208bc82787d91f7030/tycho-artifactcomparator/src/main/java/org/eclipse/tycho/zipcomparator/internal/ClassfileComparator.java#L36-L39

laeubi:
Closing in favor of:
- https://github.com/eclipse-tycho/tycho/pull/3300

after that is merged we might (or not) start a new attempt to include debugging infos, maybe for the cases where one really wants to find out the difference using the raw class files and a dedicated tool works better anyways.

PR #3320, opened 2024-01-05
Only use ci-friendly version if there is such property defined

laeubi:
Currently it can happen that one uses one of the "trigger properties" that enable Tycho pomless feature to assume that a cifriendly version is to be used even though ci-friendly versions are actually not used at all.

This tries to differentiate the two cases by checking if there is actually a property defined that could be used in a ci-friendly way.

### DIFF
--- a/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
+++ b/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
@@ -67,7 +67,8 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
 
     private static final String PARENT_POM_DEFAULT_VALUE = System.getProperty(TYCHO_POMLESS_PARENT_PROPERTY, "..");
     private static final String QUALIFIER_SUFFIX = ".qualifier";
-    private static final String MODEL_PARENT = "TychoMapping.model.parent";
+
+    private Map<Path, ParentModel> parentModelCache = new HashMap<Path, ParentModel>();
 
     @Requirement
     protected PlexusContainer container;
@@ -79,7 +80,7 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
     private boolean extensionMode;
     @SuppressWarnings("unused")
     private File multiModuleProjectDirectory;
-    private String snapshotFormat;
+    private String snapshotProperty;
 
     @Override
     public File locatePom(File dir) {
@@ -155,7 +156,7 @@ public abstract class AbstractTychoMapping implements Mapping, ModelReader {
         model.setPackaging(getPackaging());
         initModel(model, artifactReader, artifactFile);
         if (model.getParent() == null) {
-            model.setParent(findParent(artifactFile.getParent(), options));
+            model.setParent(findParent(artifactFile.getParent(), options).parentReference());
         }
         if (model.getVersion() == null && model.getParent() != null) {
             //inherit version from parent if not given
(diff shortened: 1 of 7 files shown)


tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
```diff
@@ -198,24 +198,27 @@ protected Parent findParent(Path projectRoot, Map<String, ?> projectOptions) thr
         Model parentModel = parentPom.getReader().read(parentPom.getPomFile(), options);
         Parent parentReference = new Parent();
         String groupId = parentModel.getGroupId();
-        if (groupId == null) {
+        Parent grandParent = parentModel.getParent();
```
tivervac, line 201: Should we just get the grandparent, or keep going up until we find a parent that has a groupId?

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
```diff
@@ -198,24 +198,27 @@ protected Parent findParent(Path projectRoot, Map<String, ?> projectOptions) thr
...
+            groupId = grandParent.getGroupId();
         }
         parentReference.setGroupId(groupId);
         parentReference.setArtifactId(parentModel.getArtifactId());
         String version = parentModel.getVersion();
-        if (version == null) {
+        if (version == null && grandParent != null) {
             // must be inherited from grandparent
-            version = parentModel.getParent().getVersion();
+            version = grandParent.getVersion();
```
tivervac, line 211: Same question

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
```diff
@@ -342,18 +345,41 @@ private static void setLocation(Model model, Path modelSource) {
...
+                return unqualifiedVersion + "${" + snapshotProperty + "}";
             }
             return unqualifiedVersion + "-SNAPSHOT";
         }
         return pomVersion;
     }
 
+    private boolean modelHasProperty(String property, Model model, Path projectRoot) {
+        if (property == null) {
+            //nothing we can check asume it is NOT present...
```
tivervac, line 363:
```suggestion
            //nothing we can check, assume it is NOT present...
```

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
```diff
@@ -342,18 +345,41 @@ private static void setLocation(Model model, Path modelSource) {
...
+        Properties properties = model.getProperties();
+        String string = properties.getProperty(property);
+        if (string != null) {
+            return true;
+        }
+        if (model.getParent() == null) {
+            try {
+                ParentModel findParent = findParent(projectRoot.getParent(), Map.of());
+                return modelHasProperty(property, findParent.parentModel(),
+                        projectRoot.resolve(findParent.parentReference().getRelativePath()));
```
tivervac, line 375:
```suggestion
                ParentModel parent = findParent(projectRoot.getParent(), Map.of());
                return modelHasProperty(property, parent.parentModel(),
                        projectRoot.resolve(parent.parentReference().getRelativePath()));
```

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
laeubi: I think a parent reference must always contain a GAV you can only leave it out for the item itself.

tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/AbstractTychoMapping.java
laeubi: Done

tivervac (approved):

laeubi:
This bugfix was gently sponsored by  [Sigasi](https://www.sigasi.com/).

PR #3325, opened 2024-01-06
Move components to JSR330

laeubi:
See https://github.com/eclipse/sisu.plexus/wiki/Plexus-to-JSR330

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -506,6 +506,20 @@
 					<artifactId>maven-invoker-plugin</artifactId>
 					<version>3.6.1</version>
 				</plugin>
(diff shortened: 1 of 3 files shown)


akurtakov:
After giving a thought to this one, if we start with core component we will have to preserve both plexus and jsr330 annotations so plexus components in the rest of codebase can find it.
Alternatively, conversion can start from "leaf" components and should work.
Yet another alternative is to have a "big bang" conversion.

laeubi:
It should be possible to mix things, so we can just start with one component at a time (leaving mojos as is for now).

Issue #2462, opened 2023-06-01
JUnit 5 Test Suite not working - "No tests found"


Phillipus:
I'm trying to execute a JUnit 5 Test Suite using Tycho Surefire. The suite is this:

```java
@Suite
@SelectClasses({
    FooTest1.class,
    FooTest2.class
})

@SuiteDisplayName("All Tests")
public class AllTests {
}
```

This runs in Eclipse itself, but not using Maven/Tycho. Error is "Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:3.0.4:test (default-test) on project test.package.tests: No tests found."

Attached is a simple test project. Run with `mvn clean verify`

Tested on Tycho 3.0.4 and 4.0.0-SNAPSHOT

[test.zip](https://github.com/eclipse-tycho/tycho/files/11628117/test.zip)

ptziegler:
I had a brief look at it the other day, because we ran into a similar situation and I think there are two reasons why the JUnit 5 test suites don't work:

1) Test suites are executed via the SuiteTestEngine, which is provided by `platform-suite-engine`. This bundle is not included in the embedded JUnit5 runner (see org.eclipse.tycho.surefire.junit59).

2) Even if this bundle is added, the services are not merged correctly. Thus, the test engine is not discovered via SPI.

laeubi:
As always it would be great to provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue.

Nerveless one main problem *might* be that Tycho has a pre-check that, if triggered, skips the execution all together, so the first step should be to check if the suite is detected at that stage already.

Phillipus:
> As always it would be great to provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue.

I tried, I really did, but gave up in sheer frustration in getting the workflow to run.

https://github.com/eclipse-tycho/tycho/pull/2469

ptziegler:
I can have a look at it this afternoon as well. Given that I played around with it for a little bit, there should be a small test project I can add to the test suite.

> Nerveless one main problem might be that Tycho has a pre-check that, if triggered, skips the execution all together, so the first step should be to check if the suite is detected at that stage already.

[...]

PR #3342, opened 2024-01-10
Provide tycho-surefire integration test for JUnit 5 suites

HeikoKlare:
This is an initial draft for an integration test for using JUnit 5 suites in tycho-surefire.
It is supposed to demonstrate the behavior discussed in 
- #2462

### DIFF
--- a/tycho-its/projects/surefire.junit5suite/bundle.test/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/surefire.junit5suite/bundle.test/META-INF/MANIFEST.MF
@@ -1,0 +1,8 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: JUnit5 Suite Test Plug-in
+Bundle-SymbolicName: bundle.test.junit5suite
+Bundle-Version: 1.0.0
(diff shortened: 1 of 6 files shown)


laeubi:
@HeikoKlare the referenced issue seems odd, do you maybe want to reference another issue or an issue in a different repository?

HeikoKlare:
> the referenced issue seems odd, do you maybe want to reference another issue or an issue in a different repository?

Thanks! Yes, sorry for that. I corrected the referenced issue.

HeikoKlare:
To ease investigation of the issue, here is the same test setup with pure Maven surefire: [bundle.test.zip](https://github.com/eclipse-tycho/tycho/files/13888591/bundle.test.zip)
That configuration works fine with test suite as long `junit-platform-suite-engine` is on the classpath.

laeubi:
@HeikoKlare 

is this supposed to fix the mentioned issue:
- https://github.com/eclipse-tycho/tycho/pull/3358

HeikoKlare:
The test case provided here was only supposed to demonstrate the issue documented in:
- #2462

The test case is also integrated into and thus subsumed by the PR supposed to fix the issue:
- #3358

So in case #3358 gets merged, this becomes obsolete.

PR #3350, opened 2024-01-11
Support javac as a compiler for Tycho

laeubi:
Currently Tycho is strictly using ecj as a compiler but in general it might be there are situations one wants to use javac for compilation even though it might not has full features supported (e.g. package restrictions).

This is an attempt to make it possible to use javac at a very basic level.

### DIFF
--- a/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
+++ b/tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
@@ -573,6 +573,7 @@ public abstract class AbstractOsgiCompilerMojo extends AbstractCompilerMojo impl
         final List<String> classpath = new ArrayList<>();
         Set<String> seen = new HashSet<>();
         Set<String> includedPathes = new HashSet<>();
+        boolean useAccessRules = JDT_COMPILER_ID.equals(compilerId);
         for (ClasspathEntry cpe : getClasspath()) {
             Stream<File> classpathLocations = Stream
                     .concat(cpe.getLocations().stream(),
@@ -581,7 +582,7 @@ public abstract class AbstractOsgiCompilerMojo extends AbstractCompilerMojo impl
                     .filter(AbstractOsgiCompilerMojo::isValidLocation).distinct();
             classpathLocations.forEach(location -> {
                 String path = location.getAbsolutePath();
-                String entry = path + toString(cpe.getAccessRules());
+                String entry = path + toString(cpe.getAccessRules(), useAccessRules);
                 if (seen.add(entry)) {
                     includedPathes.add(path);
                     classpath.add(entry);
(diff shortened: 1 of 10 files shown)


mickaelistria:
This is an interesting feature also for people who want to build/use plugins leveraging very new Java features which aren't yet supported by ECJ.

laeubi:
I now made the (very simple) testcase pass by just omit any custom options Tycho adds if not ECJ is selected as compiler.

tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
```diff
@@ -851,14 +855,22 @@ private void configureBootClassPath(CompilerConfiguration compilerConfiguration,
                 if (includeParent != null) {
                     Xpp3Dom[] includes = includeParent.getChildren("include");
                     if (includes.length > 0) {
-                        compilerConfiguration.addCompilerCustomArgument("-bootclasspath", scanBootclasspath(
+                        addCompilerCustomArgument(compilerConfiguration, "-bootclasspath", scanBootclasspath(
```
szarnekow, line 858: Bootclasspath is a standard javac option. Skipping this might cause surprises.

tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
```diff
@@ -738,7 +742,7 @@ protected CompilerConfiguration getCompilerConfiguration(List<String> compileSou
                 if (jreClasspathEntry.isModule()) {
                     Collection<String> modules = jreClasspathEntry.getLimitModules();
                     if (!modules.isEmpty()) {
-                        compilerConfiguration.addCompilerCustomArgument("--limit-modules", String.join(",", modules));
+                        addCompilerCustomArgument(compilerConfiguration, "--limit-modules", String.join(",", modules));
```
szarnekow, line 745: This is a standard javac option, too

tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
```diff
@@ -581,7 +582,7 @@ public List<String> getClasspathElements() throws MojoExecutionException {
                     .filter(AbstractOsgiCompilerMojo::isValidLocation).distinct();
             classpathLocations.forEach(location -> {
                 String path = location.getAbsolutePath();
-                String entry = path + toString(cpe.getAccessRules());
+                String entry = path + toString(cpe.getAccessRules(), useAccessRules);
                 if (seen.add(entry)) {
```
szarnekow, line 586: For all but the first entry `seen.add(entry)` will be false. The included paths and the classpath will miss entries.

tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
```diff
@@ -573,6 +573,7 @@ public List<String> getClasspathElements() throws MojoExecutionException {
         final List<String> classpath = new ArrayList<>();
```
szarnekow, line 573: @laeubi This class `AbstractOsgiCompilerMojo` does not contain any JavaDoc on the class declaration itself. It makes me wonder if it is even supposed to work with javac, given that the plain Java compiler does not have any understanding of OSGI semantics.

tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
laeubi: It seems not to be a standard plexus compiler option so I'm not sure how to handle this gracefully...

tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
laeubi: It seems not to be a standard plexus compiler option so I'm not sure how to handle this gracefully...

tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
laeubi: > For all but the first entry `seen.add(entry)` will be false. The included paths and the classpath will miss entries. (with javac that is)

I'm not sure I understand what you want to suggest here, the whole "see" part is that a path could apear multiple times (e.g. with different environments) so it is to make the list not contain it more than once ...

tycho-compiler-plugin/src/main/java/org/eclipse/tycho/compiler/AbstractOsgiCompilerMojo.java
laeubi: Compilation actually do not require much of OSGi, ECJ supports "restrictions" on classpath entries but `javac` not as far as I know so thats the main difference here, also the target and source levels are derived from the manifest if possible.

laeubi:
@szarnekow the `javac` support is really basic, I think the best would be to merge this first (I already tested it with a smaller Tycho project) and enhance it afterwards by e.g. adding another testcase that makes use of some of the yet not supported option.

Another open issue is that we should upgrade to catch up with the latest changes from the maven-compiler-plugin here:
- https://github.com/eclipse-tycho/tycho/issues/3351

PR #3358, opened 2024-01-12
Enable JUnit 5 test suite engine in tycho-surefire #2462

HeikoKlare:
Currently, Tycho is not capable of running JUnit 5 test suites, since the required `SuiteTestEngine` is not provided while running tests.

This change adds the `SuiteTestEngine` to tycho-surefire executions using JUnit Platform 1.8 (the first with suite support) or newer. To this end, the `SuiteTestEngine` is always enabled when running JUnit 5 tests. It also adds a regression test for executing a test suite based on JUnit Jupiter 5.9.

I do not see any problems of simply making the `SuiteTestEngine` _always_ available (other than the `VintageTestEngine`), but let me know if you think different. The problem when adding further fragments like `junit59withvintage` for when the suite API is on the classpath is the combinatorial complexity, as you would need `junitX`, `junitXwithvintage`, `junitXwithsuite`, `junitXwithsuiteandvintage`.

Fixes https://github.com/eclipse-tycho/tycho/issues/2462
Closes #3342

### DIFF
--- a/tycho-its/projects/surefire.junit59suite/bundle.test/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/surefire.junit59suite/bundle.test/META-INF/MANIFEST.MF
@@ -1,0 +1,8 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: JUnit5 Suite Test Plug-in
+Bundle-SymbolicName: bundle.test.junit59suite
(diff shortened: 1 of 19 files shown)


laeubi (approved):
@akurtakov It looks al sane to me, but as you have done the surefire parts in the past you probably want to take a second look.

akurtakov (approved):
Change looks good to me.

PR #3359, opened 2024-01-12
Adjust Oomph setup for master->main branch change

HeikoKlare:


HeikoKlare:
I've successfully tested the updated setup locally.

laeubi:
Maybe @merks can review as well.

merks:
That looks correct if you are renaming the master branch to main.

akurtakov:
It already happened thus merging.

HannesWell:
Looks good.
I just wanted to mention that due to the change in the stream name one has to fix existing workspace setups, i.e. the previously referenced stream does not exist anymore and has to be changed to the main stream.
Not ideal, but better do it now than to keep it inconsistent.

For the record to fix
grafik
one has to delete the broken stream reference and just drag-and-drop the new one from the setup in the workspace.
In the end the first line should be replaced by the second line.
```
<stream href="index:/org.eclipse.setup#//@projectCatalogs[name='org.eclipse']/@projects[name='tycho']/@streams[name='master']"/>
  <stream href="index:/org.eclipse.setup#//@projectCatalogs[name='org.eclipse']/@projects[name='tycho']/@streams[name='main']"/>
```

HeikoKlare:
Thanks for the hint, Hannes! I did not think about that and did not face the problem, as I had not preexisting setup. Sorry for this inconvience. But as you mentioned, it is probably best to have this rename done consistently.

PR #3360, opened 2024-01-12
Log skipped TargetEnvironments only in debug

akurtakov:
This info is not helpful in majority of cases as setting a platform filter in the manifest is an explicit action which the developer does so there is no need to be reminded of that at the very beginning of the build where Maven usually gives warnings like missing plugin versions and etc. that makes it quite often considered as potential problem. Log in debug mode instead.


akurtakov:
@laeubi wdyt?

laeubi (approved):

PR #3378, opened 2024-01-17
Checking for value instead of key in

vogella:
RemoteArtifactRepositoryManagerAgentFactory

Fixes #3375

PR #3384, opened 2024-01-19
Also consider profile properties when looking for disable mirror keys

laeubi:
I also created
- https://github.com/eclipse-equinox/p2/pull/431

as currently Tycho does ignore any mirrors (by modify the repository under the hood), but the mirror list is still loaded in such a case!

Replaces:
- https://github.com/eclipse-tycho/tycho/pull/3378


laeubi:
@vogella @Phillipus please validate

Phillipus:
@laeubi I've tested this with both options `tycho.disableP2Mirrors` and `eclipse.p2.mirrors` in my `settings.xml` file and profile and it is working as expected.

Thanks for this, sorry I didn't take it further in the original PR, your implementation is much better than what I would have done!

laeubi:
Backported here:
- https://github.com/eclipse-tycho/tycho/pull/3387

vogella:
Thanks @laeubi for fixing it and @Phillipus for identifying the issue and helping with testing it.

You guys are cool.

PR #3390, opened 2024-01-19
Remove warning about unspecified environment

laeubi:
Currently if not target platform environment is specified Tycho emits a warning "No explicit target runtime environment configuration. Build is platform dependent." beside from that nothing ever seems go wrong and actually Tycho builds *can* be environment agnostic. Also there is not really a way to tell Tycho that it is okay, so either the user has to use a specific one or configure many ones that make resolving slower.

This now removes this warning, if with modern Tycho there are any problems that arise from a missing environment configuration one should better give a specific error/hint at this place instead of warn "just in case".


akurtakov (approved):
Looks good to me.

PR #3397, opened 2024-01-22
Speedup repository assembly

laeubi:
Main improvement is that `FinalTargetPlatformImpl#getMetadataRepository()` previously copied on each call the whole target and also the index for it got rebuild on each call, what can have a impact if called very often.

In the bad case it can take > 1 minute to build a repository where now it is down to just 10 seconds.


laeubi:
With the batching for the RecreateRepositoryApplication now one can even cut the time by half, my testcase goes down from `1:38` > `0:05` now.

PR #3417, opened 2024-01-26
Move p2 dependency tree logic from tycho-p2-plugin to tycho-core

ptziegler:
This moves the algorithm which is responsible for calculating the IU dependency tree used by the dependency-tree Mojo to tycho-core, so that it can be used by other projects.

For a given project, this tree can be calculated using the P2DependencyTreeGenerator, which can be injected into a Mojo as a simple Plexus component.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/p2/tools/P2DependencyTreeGenerator.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/p2/tools/P2DependencyTreeGenerator.java
@@ -1,0 +1,209 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Patrick Ziegler and others.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


ptziegler:
Prerequisite to #3258

laeubi:
@ptziegler can this maybe even be moved to `tycho-p2-maven-plugin`? or does it require any class from `tycho-core` that is not in `tycho-api`?

ptziegler:
The "tycho-p2-maven-plugin"?

laeubi:
> The "tycho-p2-maven-plugin"?

https://github.com/eclipse-tycho/tycho/tree/main/p2-maven-plugin

ptziegler:
Thanks. I was only looking at the projects starting with "tycho-" and completely missed that one 😅 

But no, I need e.g. the TychoProjectManager to calculate the initial set of IUs that are required by a given Maven project.
While it's not strictly necessary (I could just precompute the list and then pass it as an additional argument), every method who wants to use this component would then have to effectively copy-paste the same code snippet.

image

laeubi:
Right I though I have moved tychoproject to api already but it seems not the case...

PR #3442, opened 2024-01-31
Add IT to reproduce missing launcher

henryju:
Follow up of https://github.com/eclipse-tycho/tycho/discussions/3440


laeubi:
@henryju thanks for the example I have looked into it but the problem lies a bit deeper, I could manage it to have the launcher included but then the runtime can't startup because some bundles need to be started... without P2 we emulate this and setup everything but with a P2 install this is not done and usually a bit hard to guess because *if* there is a product used one for sure want to have its settings used.

One option would be to allow specify a product, another that we have some flag that tells if the p2 provisioned should contain some usual defaults or we require the user to setup things manually (and just enhance the documentation)

As you have brought up this usecase I wonder if you have any preferences here?

henryju:
IMO having a way to specify a product would be nice, and kind of consistent what we can do in Eclipse UI when configuring a run config:
image

laeubi:
Here is an example how to publish a product (what can even be almost empty):

https://github.com/eclipse-pde/eclipse.pde/blob/c07f514aead5cad864d02d4540254c4cff78cb1d/build/org.eclipse.pde.build.tests/pom.xml#L115-L128

that the can be used later on here:

https://github.com/eclipse-pde/eclipse.pde/blob/c07f514aead5cad864d02d4540254c4cff78cb1d/build/org.eclipse.pde.build.tests/pom.xml#L144

would that be suitable for you?

Do you want to enhance your testcase and the documentation here:

https://github.com/eclipse-tycho/tycho/blob/main/src/site/markdown/TestingBundles.md#tycho-surefire-plugin

to describe the way to use p2installed runtime?

henryju:
Is what you are suggesting already doable with Tycho 4.0.5? I tried to define a dummy product in my test project:

`test.product`
```xml
<?pde version="3.5"?>

<product name="A product to define the runtime of the tests" uid="org.sonarlint.eclipse.tests.product" id="org.eclipse.sdk.ide" application="org.eclipse.ui.ide.workbench" version="1" type="features" includeLaunchers="true" autoIncludeRequirements="true">

   <configIni use="">
   </configIni>

   <plugins>
   </plugins>

   <configurations>
      <plugin id="org.eclipse.core.runtime" autoStart="true" startLevel="0" />
      <plugin id="org.eclipse.equinox.common" autoStart="true" startLevel="2" />
      <plugin id="org.eclipse.equinox.event" autoStart="true" startLevel="2" />
      <plugin id="org.eclipse.equinox.simpleconfigurator" autoStart="true" startLevel="1" />
   </configurations>

</product>
```

Then use this product for tests:

`pom.xml`
```xml
  <build>
    <plugins>
      <plugin>
        <artifactId>tycho-surefire-plugin</artifactId>
        <configuration>
          <product>org.sonarlint.eclipse.tests.product</product>
          <testRuntime>p2Installed</testRuntime>
        </configuration>
      </plugin>
      <plugin>
        <artifactId>tycho-p2-publisher-plugin</artifactId>
        <executions>
          <execution>
            <id>publish-products-for-tests</id>
            <goals>
              <goal>publish-products</goal>
            </goals>
            <phase>pre-integration-test</phase>
          </execution>
        </executions>
      </plugin>
    </plugins>
  </build>
```

But this is failing with error:
> [INFO] --- tycho-p2-publisher:4.0.5:publish-products (publish-products-for-tests) @ org.sonarlint.eclipse.core.tests ---
> [ERROR] Failed to execute goal org.eclipse.tycho:tycho-p2-publisher-plugin:4.0.5:publish-products (publish-products-for-tests) on project org.sonarlint.eclipse.core.tests: Publisher failed. Verify your target-platform-configuration and executionEnvironment are suitable for proper resolution: Unable to locate feature 'org.eclipse.equinox.executable'. This feature is required for native product launchers.

I tried to follow the hint and added `<unit id="org.eclipse.equinox.executable.feature.group" version="0.0.0"/>` to my target platform, but this doesn't seem to make any difference.

For the record, my target platform file now looks like:

```xml
<target name="sonarlint-eclipse-build" sequenceNumber="6">
  <locations>
    <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
      <unit id="org.eclipse.cdt.feature.group" version="0.0.0"/>
      <unit id="org.eclipse.jdt.feature.group" version="0.0.0"/>
      <unit id="org.eclipse.m2e.feature.feature.group" version="0.0.0"/>
      <unit id="org.eclipse.platform.ide" version="0.0.0"/>
      <unit id="org.eclipse.egit.feature.group" version="0.0.0"/>
      <unit id="org.eclipse.equinox.core.sdk.feature.group" version="0.0.0"/>
      <!-- Needed for test environment -->
      <unit id="org.eclipse.equinox.executable.feature.group" version="0.0.0"/>
      <repository location="https://download.eclipse.org/releases/2022-06/"/>
    </location>
  </locations>
  <targetJRE path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
</target>
```

laeubi:
You need to use the current [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) build (but it will be released soon).

henryju:
I was able to make things [work](https://github.com/SonarSource/sonarlint-eclipse/pull/614/commits/78b7b214a6c42f3370d99b18b4d6beab8d8bbee9) using the latest Tycho release. Thanks. I will try to contribute IT and documentation as you suggested, but I am off next week so this might only come later.

PR #3484, opened 2024-02-07
Use the running environment if possible in API analysis

laeubi:
Fix https://github.com/eclipse-platform/eclipse.platform.ui/issues/1654


tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisMojo.java
```diff
@@ -324,6 +324,26 @@ private Collection<Path> getBaselineBundles() throws MojoFailureException {
 		return baselineBundles;
 	}
 
+	/**
+	 * THis method selected the a Target environment bes suited for the current
```
BeckerWdf, line 328:
```suggestion
	 * This method selected the a target environment best suited for the current
```

laeubi: Done

PR #3499, opened 2024-02-10
Automatically discover dependencies from a published product

laeubi:
Currently the publis-products mojo is only fully usable in the eclipse-repository packaging type but there are other use-cases where it becomes interesting to publish the product metadata and further use it, for example in an p2 installed runtime. Even though it works to use the mojos it is quite inconvenient as one has to specify all its requirements manually as extra dependencies.

This adds a new PublishProduct P2 unit providers that collect the dependencies from a product and supply them automatically as requirements to the project so the mojo can be used without having missed dependencies from the target.


akurtakov:
Do you have some exampleto show what becomes possible/easier with this change that wasn't before?

laeubi:
This is used here: 
- https://github.com/eclipse-equinox/p2/pull/458

laeubi:
Also used here now:
- https://github.com/eclipse-pde/eclipse.pde/pull/1149

PR #3515, opened 2024-02-16
Make the slicing options of the mirror-target-platform mojo configurable

laeubi:
Currently the mirror-target-platform mirrors everything it could gather,
that could probably be to much, one maybe even want to configure even
more things like optional dependencies.

This now adds a new `<options>` parameter that allows to fine tune the
slicing options used for mirror the target.


p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/SlicingOptions.java
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Holding the slicing option for a permissive slicer, the defaults are:
+ * <ul>
+ * <li>includeOptionalDependencies = true</li>
+ * <li>everythingGreedy = true</li>
+ * <li>forceFilterTo = true</li>
+ * <li>considerStrictDependencyOnly = false</li>
+ * <li>followOnlyFilteredRequirements = false</li>
+ * <li>filter = null</li>
```
tivervac, line 27: These options are not very clear to me. Could you add some description of what they're supposed to do? I also don't know how I could configure a HashMap (`filter`) in a `pom.xml`

laeubi: Yes these are rather special options I think if the mojo is really used more one would want a dedicated description on it.

target-platform-configuration/src/main/java/org/eclipse/tycho/target/MirrorTargetPlatformMojo.java
```diff
@@ -74,6 +75,35 @@ public class MirrorTargetPlatformMojo extends AbstractMojo {
...
+    /**
+     * Allows configuration of additional slicing options for example like this:
+     * 
+     * <pre>
+     * &lt;options&gt;
+    *   &lt;!-- should optional dependencies be included (true) or ignored (false), defaults to true --&gt;
+    *   &lt;includeOptionalDependencies&gt;true/false&lt;/includeOptionalDependencies&gt;
+    *   &lt;!-- should requirements be considered always greedy (true) or only if specified (false), defaults to true --&gt;
+    *   &lt;everythingGreedy&gt;true/false&lt;/everythingGreedy&gt;
+    *   &lt;!-- should only strict dependencies be consodered (true) or all dependencies (false), defaults to false --&gt;
```
tivervac, line 87:
```suggestion
    *   &lt;!-- should only strict dependencies be considered (true) or all dependencies (false), defaults to false --&gt;
```

target-platform-configuration/src/main/java/org/eclipse/tycho/target/MirrorTargetPlatformMojo.java
```diff
@@ -74,6 +75,35 @@ public class MirrorTargetPlatformMojo extends AbstractMojo {
...
+    *   &lt;!-- a strict dependency is onw with a strict version range (e.g. [1.2,1.2]) and usually maps to items included in a feature --&gt;
```
tivervac, line 88:
```suggestion
    *   &lt;!-- a strict dependency is one with a strict version range (e.g. [1.2,1.2]) and usually maps to items included in a feature, default is false --&gt;
```

target-platform-configuration/src/main/java/org/eclipse/tycho/target/MirrorTargetPlatformMojo.java
```diff
@@ -74,6 +75,35 @@ public class MirrorTargetPlatformMojo extends AbstractMojo {
...
+    *   &lt;considerStrictDependencyOnly&gt;true/false&lt;/considerStrictDependencyOnly&gt;
+    *   &lt;!-- should only items that have a filter be considerd (true) or all of them (false), default is false --&gt;
```
tivervac, line 90:
```suggestion
    *   &lt;!-- should only items that have a filter be considered (true) or all of them (false), default is false --&gt;
```

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/SlicingOptions.java
laeubi: I enhanced the description now!

target-platform-configuration/src/main/java/org/eclipse/tycho/target/MirrorTargetPlatformMojo.java
```diff
@@ -74,6 +75,35 @@ public class MirrorTargetPlatformMojo extends AbstractMojo {
...
+    *   &lt;followOnlyFilteredRequirements&gt;true/false&lt;/followOnlyFilteredRequirements&gt;
+    *   &lt;!-- if no filter context is defined and there is a filter, it is forced to be asume to be matched (true) or non matching (false), default is true --&gt;
```
tivervac, line 92:
```suggestion
    *   &lt;!-- if no filter context is defined and a filter is encountered (e.g., os = win32 on a requirement), the filter will always match (true) or fail (false), default is true --&gt;
```

target-platform-configuration/src/main/java/org/eclipse/tycho/target/MirrorTargetPlatformMojo.java
```diff
@@ -74,6 +75,35 @@ public class MirrorTargetPlatformMojo extends AbstractMojo {
...
     private boolean includeCategories = true;
 
+    /**
+     * Allows configuration of additional slicing options for example like this:
+     * 
+     * <pre>
+     * &lt;options&gt;
+    *   &lt;!-- should optional dependencies be included (true) or ignored (false), defaults to true --&gt;
+    *   &lt;includeOptionalDependencies&gt;true/false&lt;/includeOptionalDependencies&gt;
+    *   &lt;!-- should requirements be considered always greedy (true) or only if specified (false), defaults to true --&gt;
```
tivervac, line 85:
```suggestion
    *   &lt;!-- should requirements be considered always greedy (true) - i.e., install even if there are no usages - or only if specified (false), defaults to true --&gt;
```

tivervac (approved):

PR #3516, opened 2024-02-16
Add p2-aware model converter for CycloneDX SBOM generation

ptziegler:
This change consists of two parts. The first part consists of generating a valid package URL for a given p2 artifact. The PURL contains the symbolic name, version and classifier of the artifact key, as well as the repository it is located in.
The second part handles the dependencies between two PURLs. For the sake of consistency, this dependency tree matches the tree calculated by the dependency-tree mojo.


ptziegler:
Luckily, RELEASE_NOTES was the only conflicting file.

laeubi:
> Luckily, RELEASE_NOTES was the only conflicting file.

Yes I havent yet found a good way to let it merge Release notes sucessfully as a backport :-(

Issue #3501, opened 2024-02-11
Authentication fails when trying to resolve a target platform over an authenticated mirror that is a composite repository


kevloral:
Related to #2930, authentication still fails when the mirror that requires authentication is a composite repository . As a reminder, the mirror is declared like this in my Maven _settings.xml_ file:

```
<mirror>
	<id>nexus-p2</id>			
	<name>Nexus P2</name>
	<url>https://servidor.jma:8443/nexus/content/groups/global-p2/</url>
	<mirrorOf>eclipse_2022-06,eclipse_2023-06,mylyn_3.25,swtbot_snapshots,tpd_nightly</mirrorOf>
	<layout>p2</layout>
    <mirrorOfLayouts>p2</mirrorOfLayouts>
</mirror>
```

And its credentials are declared like this :
```

<server>
	<id>nexus-p2</id>
	<username>jm.arnesto</username>
	<password>********</password>
</server>
```

After the fixes we implemented in #2930, authentication works for the mirror itself, but still fails for the child locations it declares:

```
<?xml version='1.0' encoding='UTF-8'?>
<?compositeArtifactRepository version='1.0.0'?>
<repository name='generated-composite-metadata-repository' type='org.eclipse.equinox.internal.p2.artifact.repository.CompositeArtifactRepository' version='1'>
  <properties size='1'>
    <property name='p2.timestamp' value='1693774779425'/>
  </properties>
  <children size='7'>
    <child location='../../repositories/eclipse_2023-06'/>
    <child location='../../repositories/basis_23'/>
    <child location='../../repositories/eclipse_2022-06'/>
    <child location='../../repositories/basis_22'/>
    <child location='../../repositories/mylyn_3.25'/>
    <child location='../../repositories/swtbot_snapshots'/>
    <child location='../../repositories/tpd_nightly'/>
  </children>
</repository>
```

I will implement a test that reproduces the bug and propose a fix for it in a PR.

laeubi:
@kevloral yes composited that points to childs below the base dir are a bit tricky.

> I will implement a test that reproduces the bug and propose a fix for it in a PR.

Thanks, I plan to make a new bugfix release soon, do you have an estimate so we can probably wait for your fix to be included?

akurtakov:
As thsi bugfix release is aimed for 4.31 Platform build to not use snapshot Tycho 4.x release would have to be out next Monday (Feb 19th) at the very latest in order to have couple of days to propagate things to the RC1 build.

[...]

PR #3521, opened 2024-02-20
Authentication fails when trying to resolve a target platform over an authenticated mirror that is a composite repository

kevloral:
Test created for reproducing the bug as described in #3501:

- _testTargetDefinitionAuthMirror_: tries to resolve a target definition from a composite p2 repository accessed over an authenticated composite mirror.

On the other hand, the following tests pass and have been created just for completion:

- _testAuthMirror_: tries to access a composite p2 repository over an authenticated composite mirror.
- _testMirror_: tries to access a composite p2 repository over a composite mirror with no authentication.
- _testRepositoryEncrypted_: tries to access an authenticated composite p2 repository whose password is encrypted.
- _testTargetDefinition_: tries to resolve a target platform from a composite p2 repository.
- _testTargetDefinitionEncrypted_: tries to resolve a target definition from an authenticated composite p2 repository whose password is encrypted.
- _testTargetDefinitionMirror_: tries to resolve a target definition from a composite p2 repository accessed over a composite mirror with no authentication.

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteMetadataRepositoryManager.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteMetadataRepositoryManager.java
@@ -63,7 +63,7 @@ class RemoteMetadataRepositoryManager implements IMetadataRepositoryManager {
 	public IMetadataRepository loadRepository(URI location, int flags, IProgressMonitor monitor)
 			throws ProvisionException, OperationCanceledException {
 		URI effectiveLocation = translateAndPrepareLoad(location);
-		authenticator.enterLoad(location);
+		authenticator.enterLoad(effectiveLocation);
 		try {
 			IMetadataRepository loadedRepository = delegate.loadRepository(effectiveLocation, flags, monitor);
 			failIfRepositoryContainsPartialIUs(loadedRepository, effectiveLocation);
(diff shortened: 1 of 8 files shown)


kevloral:
The problem lied in the fact that  _RemoteMetadataRepositoryManager_ was passing the original location to the _MavenAuthenticator_, instead of the effective one:
https://github.com/eclipse-tycho/tycho/blob/9920425716a2caae97d0a6af31431512a75ae584/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteMetadataRepositoryManager.java#L66

On the other hand, _RemoteArtifactRepositoryManager_ was already passing the effective URI instead of the original one, so no change was needed there:
https://github.com/eclipse-tycho/tycho/blob/9920425716a2caae97d0a6af31431512a75ae584/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteArtifactRepositoryManager.java#L139

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RemoteMetadataRepositoryManager.java
```diff
@@ -63,7 +63,7 @@ public IMetadataRepository loadRepository(URI location, IProgressMonitor monitor
 	public IMetadataRepository loadRepository(URI location, int flags, IProgressMonitor monitor)
 			throws ProvisionException, OperationCanceledException {
 		URI effectiveLocation = translateAndPrepareLoad(location);
-		authenticator.enterLoad(location);
+		authenticator.enterLoad(effectiveLocation);
```
laeubi, line 66: Good catch!

laeubi (approved):

PR #3538, opened 2024-02-29
Add test case for PublishFeaturesAndBundlesMojo marking bundle unpacked

HeikoKlare:
The `PublishFeaturesAndBundlesMojo` produces a p2 repository in which plugins are marked as "zipped" and will thus be unpacked upon installation when a feature including them does not include the plugin with "unpack=false". The added test demonstrates this behavior and ensures that only the feature will properly be marked as "zipped" but not the included plugin.

Since the mojo uses p2's `FeaturesAndBundlesPublisherApplication`, this test will be influcend by the p2 fix: https://github.com/eclipse-equinox/p2/pull/477

Note that this is only for demonstration purposes and _not_ a proper test case yet. In particular, I was not yet able to properly check the resulting p2 repository, which is why the test simply scans the generated `content.xml` for "zipped" markers.


laeubi:
@HeikoKlare mayn thanks for the test-case, just in case creating a plain maven project in the `itest` module is most often easier than a "real" unit test.

Also as the eclipse release is done and it is fixed there, I think we can wait for the release, otherwise we need to fork many of the feature classes from P2 what I like to avoid (for now).

HeikoKlare:
Thanks for the hint, @laeubi!

I hope I understood your proposal correctly: I have rewritten the test case in the `tycho-its` / integration tests, which made the test case itself much simpler.

I have also improved the assertion logic by replacing the rundimentary scanning of `content.xml` file contents with checking the XML model for the (non-)existence of appropriate entries. The test also succeeds now due to being built against current p2.

laeubi:
Perfect, thanks!

PR #3549, opened 2024-03-06
tycho-p2-director:director: Fix handling of destination on macOS

sratz:


laeubi (commented):
Can we probably have a new testcase instead of adjusting the existing one? Both "variants" should work  in my opinion.

tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/p2/director/DirectorMojo.java
```diff
@@ -378,9 +392,15 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         args.addNonNull("-iuProfileproperties", iuProfileproperties);
         args.addNonNull("-flavor", flavor);
         args.addNonNull("-bundlepool", bundlepool);
-        args.addNonNull("-p2.os", p2os);
-        args.addNonNull("-p2.ws", p2ws);
-        args.addNonNull("-p2.arch", p2arch);
+        if (p2os != null || p2ws != null || p2arch != null) {
+            if (p2os == null || p2ws == null || p2arch == null) {
+                throw new MojoFailureException(
+                        "If any of p2os / p2ws / p2arch is specified, all of them must be specified.");
```
laeubi, line 398: can we mention what parameter is missing, e.g. 
```suggestion
         String msg =  "All of p2os / p2ws / p2arch must be specified, ";
         if (p2os!=null) {
            msg+= "p2os="+p2os+ "given";
         } else {
            msg+= "p2os is missing";
         }
         ...
```
That is a bit cumbersome to wirte but users will like good and clear error messages without doubts.

tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/provisioning/ProvisionedInstallationBuilder.java
```diff
@@ -126,7 +126,7 @@ private void executeDirector(TargetEnvironment env) throws MojoFailureException
         for (String iu : ius) {
             command.addUnitToInstall(iu);
         }
-        command.setDestination(DirectorRuntime.getDestination(effectiveDestination, env));
+        command.setDestination(effectiveDestination);
```
laeubi, line 129: This now no longer takes the environment into account.

laeubi (commented):
Thanks for working on this, this looks good in general just some small remarks.

tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/provisioning/ProvisionedInstallationBuilder.java
sratz: This is considered, (the effective destination already contains the Contents/Eclipse stuff.

The actual problem is here now:
`work` in `tycho-surefire-plugin`
https://github.com/sratz/tycho/blob/361fd9881e7bbc58046ea3c791fb5679cb1e6f2e/tycho-its/projects/surefire.p2InstalledRuntime/extProductTestDirector/pom.xml#L52
behav
is now interpreted differently than
`destination` in `tycho-p2-director-plugin`
https://github.com/sratz/tycho/blob/361fd9881e7bbc58046ea3c791fb5679cb1e6f2e/tycho-its/projects/surefire.p2InstalledRuntime/extProductTestDirector/pom.xml#L62

and the 'magic' is now not consistent anymore :(

I now get your pint in the issue's discussion better now.

This `Contents/Eclipse` layout on mac is real mess and makes things really complicated.

I'll have to think about this some more.

laeubi: @sratz maybe Tycho can reuse the "logic" from the Director somehow so they behave equally?

laeubi:
@sratz I plan to do a bugfix release maybe end of month, do you think we can include this one here?

sratz:
> @sratz I plan to do a bugfix release maybe end of month, do you think we can include this one here?

Yes, I think that should be possible, I am currently looking for a solution to this issue again.

sratz:
I thought about this some more and changed the pull request again to what you initially proposed in the issue.

The processing of `work` in `tycho-surefire-plugin`  and `destination` in `tycho-p2-director-plugin` is now consistent and the same rules apply.

I think the benefits of applying this in some standard scenario (e.g. https://github.com/eclipse-tycho/tycho/blob/d23230d9ea1c03e64184f3c69bfe69499e815ef8/tycho-its/projects/surefire.p2InstalledRuntime/extProductTestDirector/pom.xml) outweigh the fact that `tycho-p2-director-plugin:director` is not a drop-in-replacement for the legacy `eclipse -application org.eclipse.equinox.p2.director -destination <destination> ...`.

I tried to document these peculiarities of path name handling in the mojo parameter descriptions.

laeubi (approved):
Looks fine and tests are green, @sratz if you are happy with that let me know and I'll put this on the merge train!

sratz:
> Looks fine and tests are green, @sratz if you are happy with that let me know and I'll put this on the merge train!

I have one more question:

I am trying to understand the intention behind the `installAllEnvironments` option:
https://github.com/eclipse-tycho/tycho/blob/95b912e2b1a13f798ad1bf8554d4c0702593a120/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractEclipseTestMojo.java#L700

Was the idea to run the director N times and install IUs for all platforms into the same product/installation?

Has this ever worked?

The way I see it, the environment options (os/ws/arch) end up in the director call inside the `IProfile`, but only if that profile does not exist, yet:
https://github.com/eclipse-tycho/tycho/blob/c66c62895c9298468184f6e17ab82b2de0904de0/tycho-core/src/main/java/org/eclipse/tycho/p2tools/copiedfromp2/DirectorApplication.java#L824-L844

So AFAIK `installAllEnvironments` is a no-op, and just runs the (same) director multiple times.

I am asking here, because if we could remove that option, it would simplify the handling of the target env even more inside `ProvisionedInstallationBuilder`

From what I can tell  `installAllEnvironments` is not used in production anywhere.

sratz:
> the difference is if you have native fragments, you can install all of them while this si not possible with one classic call.

When I follow the information flow of os/ws/arch, I don't see how that would've ever worked in practice:

- `org.eclipse.tycho.surefire.provisioning.ProvisionedInstallationBuilder.executeDirector(TargetEnvironment)`
- `org.eclipse.tycho.p2.tools.director.shared.DirectorRuntime.Command.setEnvironment(TargetEnvironment)`
- Director arguments `-p2.os / -p2.ws / -p2.arch`

And then inside https://github.com/eclipse-tycho/tycho/blob/c66c62895c9298468184f6e17ab82b2de0904de0/tycho-core/src/main/java/org/eclipse/tycho/p2tools/copiedfromp2/DirectorApplication.java#L824-L844 the only place where this information is used again is when creating a new profile, when it does not does not yet exist.

So, the first `org.eclipse.tycho.surefire.provisioning.ProvisionedInstallationBuilder.install(TargetEnvironment)` makes use of the environment, when creating the IProfile,  but all the subsequent calls don't and they essentially just repeat the first invocation.

> Yes I think we can remove that option.

Ok, then I'll do that in a separate pull-request to clean things up.

I think we can then merge this one.

laeubi:
> So, the first org.eclipse.tycho.surefire.provisioning.ProvisionedInstallationBuilder.install(TargetEnvironment) makes use of the environment, when creating the IProfile, but all the subsequent calls don't and they essentially just repeat the first invocation.

Yes the "key" was a new profile each time so they end up installed in the folder but as said in the end this was not really something useful at all...

> Ok, then I'll do that in a separate pull-request to clean things up.

Whatever you think fits best if it make things easier you can also do it in this PR.

sratz:
> Whatever you think fits best if it make things easier you can also do it in this PR.

Then I'll do it here. Also keeps the discussions together in one issue/PR.

laeubi (approved):
Many thanks!

PR #3568, opened 2024-03-11
Remove publish-osgi-ee from the default life-cycle

laeubi:
This serves both as a reminder as well as to track progress with the fact that the osgi-ee in P2 repositories is actually not useful.

Required tasks:

- [ ] Check that Tycho can work without published EE
- [ ] Check that such updatesites can be consumed by PDE / P2
- [ ] Write release notes with migration for Tycho 4 > 5 that explains how to restore old behaviour


sratz:
Would this make workarounds such as

https://github.com/eclipse-tycho/tycho/blob/fd3e5f87eeb430750bb33f3877818a698661bbfc/demo/justj/product/pom.xml#L40

https://github.com/eclipse-embed-cdt/org.eclipse.epp.packages/blob/60afb0e0dd98aba881f2fe6d656c67e448c492ef/packages/org.eclipse.epp.package.cpp.product/pom.xml#L38

obsolete?

laeubi:
@sratz I now have created:
- https://github.com/eclipse-tycho/tycho/pull/3816

this will allow to include JustJ in a product without any special / additional configuration or workarounds.

PR #3635, opened 2024-03-12
assemble-repository: Prevent sources from being included inadvertently

### LINKED ISSUE #3522: product / repository contains source jars by default
After migration from Tycho 4.0.5 to 4.0.6 I noticed that our product update site suddenly contains the source jar files even though none of the config was changed.

I bisected that to #3456, specifically https://github.com/eclipse-tycho/tycho/pull/3456/files#diff-9a9726aa0be8197a6082b700e8219a18531b5f9ba103b2f9fc5f162b154794eaR74

The related P2 change https://github.com/eclipse-equinox/p2/pull/446 explicitly mentions that this new feature should be disabled by deafult.

Shouldn't

https://github.com/eclipse-tycho/tycho/pull/3456/commits/2053d3761e621a8c40deb9062dbe2b0a404497fa#diff-9a9726aa0be8197a6082b700e8219a18531b5f9ba103b2f9fc5f162b154794eaR74

set this to `false`?

sratz:
In the presence of products, P2 adds a virtual `tooling.source.default`
IU which optionally depends on all sources.

If these sources are available in the target platform, this would cause
them to be picked up even if assemble-repository was configured with

```
  <includeAllDependencies>true<includeAllDependencies>
```

but with

```
  <includeAllSources>false</includeAllSources>.
```

This scenario has become much more likely with recent changes to P2 [1].

We do not want includeAllDependencies to imply includeAllSources, so
we actively prevent this from happening by ignoring the

```
  <required namespace='org.eclipse.equinox.p2.eclipse.type'
            name='source'
            range='0.0.0'
            optional='true'
            multiple='true'
            greedy='false'/>
```

requirements if `includeAllSources` is not specified.

Fixes #3522.

[1] https://github.com/eclipse-equinox/p2/pull/446

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
@@ -54,6 +54,7 @@ import org.eclipse.equinox.p2.repository.artifact.IArtifactRepository;
 import org.eclipse.equinox.p2.repository.artifact.IArtifactRepositoryManager;
 import org.eclipse.equinox.p2.repository.metadata.IMetadataRepository;
 import org.eclipse.equinox.p2.repository.metadata.IMetadataRepositoryManager;
+import org.eclipse.equinox.spi.p2.publisher.PublisherHelper;
 import org.eclipse.tycho.TargetPlatform;
 import org.eclipse.tycho.p2.tools.DestinationRepositoryDescriptor;
 import org.eclipse.tycho.p2.tools.RepositoryReference;
@@ -118,6 +119,7 @@ public class TychoMirrorApplication extends org.eclipse.tycho.p2tools.copiedfrom
         boolean considerOnlyStrictDependency = options.considerStrictDependencyOnly();
         return new PermissiveSlicer(repository, filters.get(0), includeOptionalDependencies,
                 options.isEverythingGreedy(), evalFilterTo, considerOnlyStrictDependency, onlyFilteredRequirements) {
+
             @Override
             protected boolean isApplicable(IInstallableUnit iu, IRequirement req) {
                 if ((includeRequiredBundles || includeRequiredFeatures) && QueryUtil.isGroup(iu)
@@ -147,6 +149,20 @@ public class TychoMirrorApplication extends org.eclipse.tycho.p2tools.copiedfrom
                 if (considerOnlyStrictDependency && !RequiredCapability.isStrictVersionRequirement(req.getMatches())) {
                     return false;
                 }
+
+                if (!includeAllSource && req.getMin() == 0 && !req.isGreedy()
+                        && req instanceof IRequiredCapability capability
+                        && PublisherHelper.NAMESPACE_ECLIPSE_TYPE.equals(capability.getNamespace())
+                        && PublisherHelper.TYPE_ECLIPSE_SOURCE.equals(capability.getName())) {
+                    // When dealing with published products, these products always include a dependency to
+                    // 'tooling.source.default', which in turn optionally + non-greedily depends on up all sources.
+                    // Since https://github.com/eclipse-equinox/p2/pull/446 the target platform contains
+                    // all the sources. When 'includeAllDependencies' is true, they would be picked up unless
+                    // we prevent this explicitly by also explicitly looking at 'includeAllSources'.
+                    // See https://github.com/eclipse-tycho/tycho/issues/3522
+                    return false;
+                }
+
                 //deal with filters
                 IMatchExpression<IInstallableUnit> filter = req.getFilter();
                 if (considerFilter) {
--- a/tycho-its/projects/product.productRepository/pom.xml
+++ b/tycho-its/projects/product.productRepository/pom.xml
@@ -1,0 +1,102 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>tycho-its-project.product.productRepository</groupId>
+  <artifactId>product.product</artifactId>
+  <version>1.0.0-SNAPSHOT</version>
+  <packaging>eclipse-repository</packaging>
+
+  <repositories>
+    <repository>
+      <id>e431</id>
+      <layout>p2</layout>
+      <!-- Important: This update site must have already been generated with the new P2 that adds the source as optional dependencyies with <filter>(org.eclipse.update.install.sources=true)</filter> metadata for the problem to show -->
+      <url>https://download.eclipse.org/eclipse/updates/4.31/R-4.31-202402290520/</url>
+    </repository>
+  </repositories>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.eclipse.tycho</groupId>
+        <artifactId>tycho-maven-plugin</artifactId>
+        <version>${tycho-version}</version>
+        <extensions>true</extensions>
+      </plugin>
+
+      <plugin>
+        <groupId>org.eclipse.tycho</groupId>
+        <artifactId>target-platform-configuration</artifactId>
+        <version>${tycho-version}</version>
+        <configuration>
+          <environments>
+            <environment>
+              <os>linux</os>
+              <ws>gtk</ws>
+              <arch>x86_64</arch>
+            </environment>
+          </environments>
+        </configuration>
+      </plugin>
+
+      <plugin>
+        <groupId>org.eclipse.tycho</groupId>
+        <artifactId>tycho-p2-repository-plugin</artifactId>
+        <version>${tycho-version}</version>
+        <configuration>
+          <includeAllDependencies>true</includeAllDependencies>
+        </configuration>
+      </plugin>
+
+      <plugin>
+        <groupId>org.eclipse.tycho</groupId>
+        <artifactId>tycho-p2-director-plugin</artifactId>
+        <version>${tycho-version}</version>
+        <executions>
+          <execution>
+            <id>materialize-products</id>
+            <goals>
+              <goal>materialize-products</goal>
+            </goals>
+          </execution>
+        </executions>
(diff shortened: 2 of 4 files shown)


tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
```diff
@@ -118,6 +119,10 @@ protected Slicer createSlicer(SlicingOptions options) throws ProvisionException
         boolean considerOnlyStrictDependency = options.considerStrictDependencyOnly();
         return new PermissiveSlicer(repository, filters.get(0), includeOptionalDependencies,
                 options.isEverythingGreedy(), evalFilterTo, considerOnlyStrictDependency, onlyFilteredRequirements) {
+
+            private static final String TOOLING_SOURCE_IU_ID = PublisherHelper.createDefaultConfigUnitId("source",
```
laeubi, line 123: I'm not sure this is the correct approach also the test failuresindicate that.

tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
```diff
@@ -118,6 +119,10 @@ protected Slicer createSlicer(SlicingOptions options) throws ProvisionException
...
+                    "tooling");
+
             @Override
             protected boolean isApplicable(IInstallableUnit iu, IRequirement req) {
                 if ((includeRequiredBundles || includeRequiredFeatures) && QueryUtil.isGroup(iu)
```
laeubi, line 128: I think this is actually the place where you want to filter things, this test if a requirement for a IU is applicable and you don't want the new requirement to be considered.

laeubi (changes requested):
It goes in the right direction but basically I would 

1. check here if this is a requirement for a source bundle
2. Construct a context IU with `org.eclipse.update.install.sources=true/false`
3. then check if they match in `isApplicable(IInstallableUnit iu, IRequirement req)`

that way the requirements will just be filtered out.

sratz:
> 1\. check here if this is a requirement for a source bundle
> 2\. Construct a context IU with `org.eclipse.update.install.sources=true/false`

I'm not sure I understand this part.
Where is 'here'?
What's a 'context IU'?
Also, as the problem is not the `org.eclipse.update.install.sources`, but the fact that a published product contains a `tooling.source.default` IU which optionally depends on all sources.

> 3\. then check if they match in `isApplicable(IInstallableUnit iu, IRequirement req)`

I think this is the important aspect. We don't want to filter
`tooling.source.default` itself in `isApplicable(IInstallableUnit iuIRequirement req)`
but only filter
`tooling.source.default`'s dependencies in `isApplicable(IInstallableUnit iuIRequirement req)`.

I have moved the condition to `isApplicable(IInstallableUnit iuIRequirement req)`, let's see what the test suite says.

laeubi:
I must confess I'm not sure `tooling.source.default` is really the problem. This has not changed recently but what has changed is that each bundle now has the `org.eclipse.update.install.sources` requirement.

sratz:
> I must confess I'm not sure `tooling.source.default` is really the problem. This has not changed recently but what has changed is that each bundle now has the `org.eclipse.update.install.sources` requirement.

It has not changed, but before `org.eclipse.update.install.sources` was introduced those sources weren't part of the target platform.

Now they are, and the effect of `tooling.source.default` now pulls them in.

sratz:
An alternative to the current proposal of looking at `tooling.source.default` could be to provide a filtered view on the project's target platform with `org.eclipse.update.install.source=true` IUs excluded. Then, using that filtered target platform to initiate the mirroring. But I am not sure I know how to do that.

laeubi:
As said this are not _IUs_ but _Requirements_ ... You can create a so called "context IU" from a map of properties with `InstallableUnit.contextIU(context)` in your case maybe `InstallableUnit.contextIU(Map.of(org.eclipse.update.install.source, includeAllSources))`, if you have that you can use 

```
IMatchExpression<IInstallableUnit> filter = req.getFilter();
return filter == null || filter.isMatch(contextIU);
```

for example to check if a requirement is matching that context IU.

sratz:
> As said this are not _IUs_ but _Requirements_ ... You can create a so called "context IU" from a map of properties with `InstallableUnit.contextIU(context)` in your case maybe `InstallableUnit.contextIU(Map.of(org.eclipse.update.install.source, true))`, if you have that you can use
> 
> ```
> IMatchExpression<IInstallableUnit> filter = req.getFilter();
> booelan matches =  filter != null && filter.isMatch(contextIU);
> ```
> 
> for example to check if a requirement is matching that context IU and if yes return false from is applicable if includeAllSources is false.

I tried this:

```java
            private static final IInstallableUnit SOURCEC_CONTEXT = InstallableUnit
                    .contextIU(Map.of("org.eclipse.update.install.source", "true"));

            @Override
            protected boolean isApplicable(IInstallableUnit iu, IRequirement req) {
                if (!includeAllSource) {
                    IMatchExpression<IInstallableUnit> filter = req.getFilter();
                    if (filter != null) {
                        if (filter.isMatch(SOURCEC_CONTEXT)) {
                            // When dealing with a repository that contains published products, these products
                            // always include a dependency to 'tooling.source.default', which picks up all sources.
                            // And since https://github.com/eclipse-equinox/p2/pull/446 the target platform contains
                            // all the sources. They would be picked up unless we prevent this here.
                            // See https://github.com/eclipse-tycho/tycho/issues/3522
                            return false;
                        }
                    }
                }
```

but it does not work.

As outlined in the issue, the problematic IU is `tooling.source.default` which has an `IRequirement` on all source providers:
```java
		// Create a required capability on source providers
		IRequirement[] reqs = new IRequirement[] {MetadataFactory.createRequirement(NAMESPACE_ECLIPSE_TYPE, TYPE_ECLIPSE_SOURCE, VersionRange.emptyRange, null, true, true, false)};
		cu.setHost(reqs);
		Map<String, String> touchpointData = new HashMap<>();
```

https://github.com/eclipse-equinox/p2/blob/e0ce87e039a001c4aafcbb2dba6ca33455933097/bundles/org.eclipse.equinox.p2.publisher/src/org/eclipse/equinox/spi/p2/publisher/PublisherHelper.java#L169-L189


This requirement does not have any filter so the code above will not be able to filter anything here based on the context IU.

laeubi:
@merks yna you give a hint about the `tooling.source.default` IU?

merks:
One thing that might generally be helpful here is that this is an optional **non-greedy** requirement:
```
<unit id='tooling.source.default' version='1.0.0' singleton='false'>
  <hostRequirements size='1'>
    <required namespace='org.eclipse.equinox.p2.eclipse.type' name='source' range='0.0.0' optional='true' multiple='true' greedy='false'/>
  </hostRequirements>
  <properties size='1'>
    <property name='org.eclipse.equinox.p2.type.fragment' value='true'/>
  </properties>
  <provides size='2'>
    <provided namespace='org.eclipse.equinox.p2.iu' name='tooling.source.default' version='1.0.0'/>
    <provided namespace='org.eclipse.equinox.p2.flavor' name='tooling' version='1.0.0'/>
  </provides>
  <requires size='1'>
    <required namespace='org.eclipse.equinox.p2.eclipse.type' name='source' range='0.0.0' optional='true' multiple='true' greedy='false'/>
  </requires>
...
</unit>
```
As far as I understand it, at installation time, an optional non-greedy requirement **will not be installed even when it's available**, i.e., there must be some other non-optional requirements (in which case the required things must be available), or optional greedy requirement (in which case it's installed when available and quietly ignored when not).  

Of course the context here is mirroring and not installing.  I think in the filtering context here that generally it should be possible to ignore all optional non-greedy requirements.  I'm not sure if that should involve yet another configuration option...

laeubi:
As far as I know "inlcudeAllRequirements" makes everything greedy....

sratz:
Yes, we have `includeAllDependencies` and `includeAllSources`.

I think the intuitive expectation is that `includeAllDependencies` does not also imply `includeAllSources`.

And previously this expectation was (usually) met, because the sources typically weren't visible via the target platform in these kind of product-containing repository definitions.

So the current proposal of this PR basically ensures that this expectation is met deterministically.

sratz:
Instead of looking at the IU's name `tooling.source.default`, we could also just look at the `IRequirement` and do this:

```java
                if (!includeAllSource && req.getMin() == 0 && !req.isGreedy()
                        && (PublisherHelper.NAMESPACE_ECLIPSE_TYPE
                                .equals(RequiredCapability.extractNamespace(req.getMatches()))
                                && PublisherHelper.TYPE_ECLIPSE_SOURCE
                                        .equals(RequiredCapability.extractName(req.getMatches())))) {
                    return false;
                }
```

WDYT?

merks:
That's exactly what I was trying just now:

image

I was just making sure the tests still pass but was having problems with the 4.31 repository not loading for some unknown reason.

merks:
@sratz I suggest you make the change (yes, the test still passes) since I'm a bit incompetent at updating someone else's PR.

sratz:
> I was just making sure the tests still pass but was having problems with the 4.31 repository not loading for some unknown reason.

had the same, clearing`~.m2/cache` helped.

merks:
@sratz 

FYI, this test fails for me locally as well:

image

So I'm looking at that...

merks:
@sratz 

The test failure is unrelated and is already fixed in master:

https://github.com/eclipse-tycho/tycho/blob/d6366adf73b9028fb192a550e4bef2950700791a/tycho-core/src/test/java/org/eclipse/tycho/p2resolver/P2ResolverTest.java#L225-L226

The problem is that the test uses a moving target repository so the actual content can and apparently does change:

https://github.com/eclipse-tycho/tycho/blob/d6366adf73b9028fb192a550e4bef2950700791a/tycho-core/src/test/java/org/eclipse/tycho/p2resolver/P2ResolverTest.java#L75

Please ensure that you rebase your change on master.

laeubi (approved):
Looks good and is backed by test so I think its ready to go.

merks:
And a back port too.  That's great!  Thanks!!

laeubi:
@sratz many thanks for the fix, as you already have contributed to Tycho for a while I'd like to ask if you want to become a committer I just would need your email address of the eclipse account and can prepare the vote then!

sratz:
> @sratz many thanks for the fix, as you already have contributed to Tycho for a while I'd like to ask if you want to become a committer I just would need your email address of the eclipse account and can prepare the vote then!

Thanks for the offer! I pinged you via email.

PR #3663, opened 2024-03-14
Adjust resolving tests for 4.31

akurtakov:
With https://github.com/eclipse-equinox/p2/pull/446 and Tycho resolving all possibly needed bundles source bundle ends up there in case it's needed.


akurtakov:
plexus-xml dependency is in master already and this one fixes the tests thus merging.

PR #3674, opened 2024-03-16
ExpandedProduct.getFeatures(ROOT_FEATURES) returns over-qualified IDs

merks:
It returns IVersionedId instances where the ID is that of the installable unit, i.e., with an extra .feature.group suffix, rather than the ID of the feature itself.
ExpandedProduct.getFeatures(INCLUDED_FEATURES) returns the correct result. A problem arises from this now because with the 4.31 changes to org.eclipse.equinox.p2.publisher.eclipse.ProductAction to generate filtered requirements for the installMode="root" features, this inconsistency produces invalid requirements, i.e., a requirement on xyz.feature.group.feature.group, which does not exist, rather than on xyz.feature.group.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/p2/tools/publisher/ExpandedProduct.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/p2/tools/publisher/ExpandedProduct.java
@@ -46,7 +46,8 @@ class ExpandedProduct implements IProductDescriptor {
     private final String expandedVersion;
     private List<IVersionedId> expandedBundles = null;
     private List<IVersionedId> expandedFeatures = null;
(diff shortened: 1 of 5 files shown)


merks:
@laeubi 

The failing test suggests a good place to improve test for this change.   Now that the correct feature ID is used the test fails. I'm just not sure the best way to author the improved test.  The requirement is present now, as it should be,, but with a filter, so while we could just remove the not() call, we really should also test that the expected filter is present on the requirement.  I'm not sure how best to express that so I'm looking for advice.  Or maybe you can complete the changes?  I can't get this one test to run in debug mode for me.  :-(

merks:
@laeubi 

Two failing tests are updated to check for the requirement that was formerly absent now being present though with a filter, including one test that checks the exact value of the filter.

PR #3733, opened 2024-04-08
Add flag to fail API tools on resolution error

### LINKED ISSUE #3570: API Tools Plugin should fail the build if the baseline resolution fails
The current API Tools Plugin only prints a warning if the API baseline resolution fails. 

I don't think it should be expected that developers or code reviewers search though the build logs every time to verify the API checks actually ran. This should be considered a fatal error, or at least configurable to be if you want to keep the current behaviour as default.

PDE based tooling would fail the build on a API baseline resolution failure.

Caused by https://github.com/eclipse-tycho/tycho/blob/f4752edc0adea7b3b262cf332144cdb78d47b543/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisMojo.java#L163

Kummallinen:
Add failOnResolutionError to api-tools configuration, when true it will fail the build if the baseline repository does not correctly resolve. Default to false to maintain existing behaviour, though 'true' is probably the preferred behaviour for many users.

Also adds tests to tycho-its for this

Fixes #3570


Kummallinen:
Windows test failure seems unrelated to my change, looks like an unreliable test:
` TestPGPSigning.testSigningBinaries:251 Unexpected unsigned IUs. ==> expected: <[org.eclipse.equinox.common, org.eclipse.equinox.common.source, org.eclipse.osgi, org.eclipse.osgi.source]> but was: <[org.eclipse.equinox.common, org.eclipse.osgi, org.eclipse.osgi.source]>`

laeubi:
@Kummallinen thanks for providing this new option and testcases.

Kummallinen:
@laeubi thank you for the review, merge & backport

PR #3751, opened 2024-04-16
Fix validation of @dot.xml files

akurtakov:
As per https://www.eclipse.org/jdt/core/compiler_32_006.dtd problem tag has mandatory id attribute. Set it to the value passed as problemId.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/EcJLogFileEnhancer.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/EcJLogFileEnhancer.java
@@ -21,7 +21,6 @@ import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


akurtakov:
As I failed to find a list of acceptable values for id/problemId, this is blindly fixing the validation but doesn't improve the content as even before this patch the values are not really helpful , e.g.:
```
<problems errors="0" infos="0" problems="1" warnings="1">
<problem line="-1" severity="WARNING" charStart="-1" charEnd="-1" categoryID="99999" problemID="-473176368">
<message value="baseline and build for org/eclipse/e4/ui/tests/css/swt/ThemeTest.class have different contents"/>
</problem>
</problems>
```
in https://download.eclipse.org/eclipse/downloads/drops4/I20240415-1800/compilelogs/plugins/org.eclipse.e4.ui.tests.css.swt_0.12.600.v20240321-1245/@dot.xml

laeubi:
> As I failed to find a list of acceptable values for id/problemId, this is blindly fixing the validation but doesn't improve the content as even before this patch the values are not really helpful , e.g.:

Yes its also a bit unclear for me but hoped someone from JDT team maybe and could give a better answer :-)
It seems that this somehow matches marker ids but I haven't found a good source to explain this either.

PR #3785, opened 2024-04-27
Drop old java versions from the github builds

laeubi:


Bananeweizen:
Shouldn't this be merged even with the failing actions? If I get it right, every other pull request fails in the first some seconds of the MacOS build because there is no Java 8 available anymore.

laeubi:
@akurtakov has used an "older" image for mac as a workaround, so maybe that can apply here as well, this PR is more to aim to fix the dependencies to old java versions.

Bananeweizen (approved):

PR #3801, opened 2024-04-30
Add follow repository references

tivervac:
Adding support for the new PDE option `followRepositoryReferences` for `.target` files.

See https://github.com/eclipse-pde/eclipse.pde/pull/1253/files

### DIFF
--- a/tycho-baseline-plugin/.settings/org.eclipse.jdt.core.prefs
+++ b/tycho-baseline-plugin/.settings/org.eclipse.jdt.core.prefs
@@ -1,0 +1,8 @@
+eclipse.preferences.version=1
+org.eclipse.jdt.core.compiler.codegen.targetPlatform=17
+org.eclipse.jdt.core.compiler.compliance=17
+org.eclipse.jdt.core.compiler.problem.enablePreviewFeatures=disabled
+org.eclipse.jdt.core.compiler.problem.forbiddenReference=warning
+org.eclipse.jdt.core.compiler.problem.reportPreviewFeatures=ignore
+org.eclipse.jdt.core.compiler.release=disabled
+org.eclipse.jdt.core.compiler.source=17
--- a/tycho-core/src/main/java/org/eclipse/tycho/p2/resolver/URITargetDefinitionContent.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/p2/resolver/URITargetDefinitionContent.java
@@ -125,7 +125,7 @@ public class URITargetDefinitionContent implements TargetDefinitionContent {
             Collection<IRepositoryReference> references = repository.getReferences();
             subMonitor.setWorkRemaining(references.size());
             for (IRepositoryReference reference : references) {
-                if ((reference.getOptions() | IRepository.ENABLED) != 0) {
+                if ((reference.getOptions() & IRepository.ENABLED) != 0) {
                     URI location = reference.getLocation();
                     if (reference.getType() == IRepository.TYPE_METADATA) {
                         try {
--- a/tycho-targetplatform/src/main/java/org/eclipse/tycho/targetplatform/TargetDefinition.java
+++ b/tycho-targetplatform/src/main/java/org/eclipse/tycho/targetplatform/TargetDefinition.java
@@ -22,6 +22,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Properties;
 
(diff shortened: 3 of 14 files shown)


tycho-core/src/main/java/org/eclipse/tycho/p2/resolver/URITargetDefinitionContent.java
```diff
@@ -116,7 +116,7 @@ private static void loadMetadataRepositories(URI uri, String id, Map<URI, IMetad
                 Collection<IRepositoryReference> references = repository.getReferences();
                 subMonitor.setWorkRemaining(references.size());
                 for (IRepositoryReference reference : references) {
-                    if ((reference.getOptions() | IRepository.ENABLED) != 0) {
+                    if ((reference.getOptions() & IRepository.ENABLED) != 0) {
```
tivervac, line 119: @laeubi  Pretty sure this was an oversight

tycho-targetplatform/src/main/java/org/eclipse/tycho/targetplatform/TargetDefinition.java
```diff
@@ -69,6 +69,11 @@ public interface Location {
 	public interface InstallableUnitLocation extends Location {
 
 		public static String TYPE = "InstallableUnit";
+		
+		/**
+		 * Must match {@link org.eclipse.tycho.core.TargetPlatformConfiguration#referencedRepositoryMode}
+		 */
+		public static boolean FOLLOW_REPOSITORY_REFERENCES_DEFAULT = true;
```
tivervac, line 76: @laeubi In Tycho 4 `TargetPlatformConfiguration#referencedRepositoryMode` defaults to ignore. In PDE, it defaults to `include`. In Tycho5, it also defaults to `include`. Should I keep `true` here and change the behavior in Tycho 4, or should I make it `false` to maintain backward compatibility for Tycho 4 but remain inconsistent with PDE?

tycho-baseline-plugin/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,8 @@
+eclipse.preferences.version=1
```
tivervac, line 1: These jdt.core files were automatically added by importing the projects into the Eclipse SDK IDE. I can revert them if you want.

tycho-targetplatform/src/main/java/org/eclipse/tycho/targetplatform/TargetDefinition.java
laeubi: The TargetPlatform Configuration default should be false as it can influence other places as well (e.g. when using pom defined repository). For the Target Platform Location we can use `true` as default.

Just in case, for the "IncludeSourceMode" we even have multiple options that is either:

- ignore --> always ignore regardless of what is configured in the target file
- honor --> use whats configured in the target file
- force --> always enable regardless whats configured in the target file

tycho-baseline-plugin/.settings/org.eclipse.jdt.core.prefs
laeubi: I think thats fine.

tycho-targetplatform/src/main/java/org/eclipse/tycho/targetplatform/TargetDefinitionFile.java
```diff
@@ -447,6 +452,16 @@ public boolean includeAllEnvironments() {
...
             return includeSource;
         }
+
+        @Override
+        public boolean includeConfigurePhase() {
+            return includeConfigurePhase;
+        }
+
+        @Override
+        public Boolean followRepositoryReferences() {
```
laeubi, line 462: Maybe an enum would be better with value

- DEFAULT
- ENABLED
- DISABLED

Boolean objects tend to be causing NPE if devs are not carefull enough.

tivervac: Good idea, done

laeubi:
The Windows Test is a flapping test and the mac currently lacks a suitable java 8 JVM I'm currently looking into that here:
- https://github.com/eclipse-tycho/tycho/pull/3785

tivervac:
@laeubi In that case, seems like this MR is ready :D

laeubi:
Do you want to squash the changes or should I merge the individual commits?

Issue #3803, opened 2024-05-02
Unconsidered linked .classpath files and test source folders


HeikoKlare:
When using linked resources in an Eclipse project, they are not always properly considered by a Tycho build. This can particularly lead to missing consideration and/or execution of tests:
* A linked `.classpath` file is not considered, such that a JUnit test container specified in that file is not used for resolving JUnit, leading to compile failures of test classes.
* A linked test source folder is not considered, such that tests defined in that source folder are not compiled and executed.

The issue has been detected in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1203

I have provided a reproducer in #3804.

HeikoKlare:
The root cause of this issue seems to be that the `.classpath` file is assumed to exist as direct content of a project's basedir:
https://github.com/eclipse-tycho/tycho/blob/8728806c95948bbfe5a3538fc442a16bea01cc5d/tycho-metadata-model/src/main/java/org/eclipse/tycho/model/classpath/ClasspathParser.java#L40-L41

PR #3804, opened 2024-05-02
Add regression test for linked .classpath and test source folder

HeikoKlare:
When using linked resources for a project's `.classpath` file defining a JUnit classpath container or test source folders, they are not detected by Tycho. So the JUnit classpath container and the test source folder are not considered during the build and the tests are not executed.

This adds a test case demonstrating the integration of the `.classpath` file and a test source folder via linked resources in different ways:
* file relative to the project location
* folder relative the project location's parent and extracted into a variable

This serves as a reproducer for #3803.

I have used the project contents from another integration test, which is why I kept the copyright information that were already present in those files.


laeubi:
@HeikoKlare many thanks I'll now take a look at this!

laeubi:
Testcase and fix integrated here:
- https://github.com/eclipse-tycho/tycho/pull/3806

PR #3806, opened 2024-05-03
Pr 3804

laeubi:
FYI @HeikoKlare I have now used your testcase and added a fix on top of it.


HeikoKlare:
Thank you for fixing this, @laeubi! That was _very_ fast.

I've also validated the fix using the current Tycho 4.0.8 snapshot version for the SWT use case (https://github.com/eclipse-platform/eclipse.platform.swt/pull/1203). Everything works as expected 👍 

Do you think we could use the Tycho 4.0.8 snapshot version in the eclipse platform parent to have this fix available in the platform builds until 4.0.8 finally is released?

laeubi:
Here we go
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2007

PR #3813, opened 2024-05-04
fix build warnings and revert compiler workaround

Bananeweizen:
All commits except the last fix certain warnings that are shown during the build. See the commit messages for details.
The last commit is unrelated in the sense that it reverts a maven compiler workaround. However, sending it as separate PR would lead to a conflict, therefore it has been added here.


laeubi:
Seems one test is now failing

Bananeweizen:
I left away the "source/target to release" commit. But I must admit that was just guessing. If it still fails, I will have to run integration tests locally (which I try to avoid).

laeubi:
@Bananeweizen do you want to squash or merge the individual commits?

Bananeweizen:
There is nothing that would cause functional changes in Tycho itself, therefore this doesn't need to be ported. I tend to not squash independent changes, even though it means that all except the last have not been tested in isolation. If something comes up later, reverting one of the changes seems easier to me than reverse-engineering the original "splitting" of the squashed commit.

Issue #3824, opened 2024-05-06
[DS] Validation of transitive generated 'osgi.extender=osgi.component' requirement fails


HannesWell:
When the `dsVersion=1.4` is used the `tycho-ds-plugin` automatically adds the following requirement to a bundle providing declarative service components:
`Require-Capability: osgi.extender; filter:="(&(osgi.extender=osgi.component)(version>=1.4.0)(!(version>=2.0.0)))"`

In general this is good to ensure a corresponding service component runtime is available at runtime, but it fails the build of depending projects, probably if they don't contain any DS-component:
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.7:validate-classpath (default-validate-classpath) on project foo.bar2: Execution default-validate-classpath of goal org.eclipse.tycho:tycho-compiler-plugin:4.0.7:validate-classpath failed: org.osgi.framework.BundleException: Bundle foo.bar2 cannot be resolved:foo.bar2 [157]
[ERROR]   Unresolved requirement: Require-Bundle: foo.bar1; bundle-version="4.0.0"
[ERROR]     -> Bundle-SymbolicName: foo.bar1; bundle-version="1.0.0.202405060842"; singleton:="true"
[ERROR]        foo.bar1 [155]
[ERROR]          Unresolved requirement: Require-Capability: osgi.extender; filter:="(&(osgi.extender=osgi.component)(version>=1.4.0)(!(version>=2.0.0)))"
```

Adding the requirement manually/explicitly to the `MANIFEST.MF` in the project makes the build work.

laeubi:
Can you provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue?

HannesWell:
> Can you provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue?

I'm working on it.

I forgot the mention that adding the requirement manually/explicitly to the `MANIFEST.MF` in the project makes the build work. Updated the initial comment.

laeubi:
I think we should probably provide the capability here as a convenience, it has no real use at the compile stage.

mwgmnn:
[...]

PR #3832, opened 2024-05-07
Minimal reproducer test-case for #3824

HannesWell:
Add a minimal reproducer test-case for issue https://github.com/eclipse-tycho/tycho/issues/3824

regarding the validation failure of transitive generated 'osgi.extender=osgi.component' requirements in bundles containing DS components of version 1.4.

Adding `Require-Capability: osgi.extender; filter:="(&(osgi.extender=osgi.component)(version>=1.4.0)(!(version>=2.0.0)))"` explicitly to plugin.a makes the test pass.
But as mentioned in https://github.com/eclipse-tycho/tycho/issues/3824#issuecomment-2095585269, my uneducated guess is that the classpath validation does not handle generated requirements well.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -179,7 +179,7 @@
 			<dependency>
 				<groupId>org.apache.commons</groupId>
 				<artifactId>commons-lang3</artifactId>
-				<version>3.14.0</version>
+				<version>3.15.0</version>
 			</dependency>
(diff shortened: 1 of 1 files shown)


HannesWell:
Thank you Christoph for providing a fix. :)

PR #3841, opened 2024-05-12
Add mojo as a replacement for the rp2.emove-iu and task

laeubi:


laeubi:
This now sadly triggers a classlaoder problem. In the past it has helped to place the calling code inside e.g. tycho-core, but for now I put this back to draft state.

laeubi:
The problematic part seems to be the execute batch call, we might need to move that to the P2Repository manager, that way it can simply be a runnable (we really do not need to report progress here)

PR #3845, opened 2024-05-15
Add demo for pde.bnd integration with classic Eclipse product build

fipro78:
As discussed in https://github.com/eclipse-tycho/tycho/discussions/3823 this PR adds a demo for an Eclipse application project that mixes projects with the classical PDE layout with the new pde.bnd integration layout.


demo/runtime-tycho.demo.app.product/.metadata/.log
laeubi, line 1: Is this file required?

fipro78: Sorry, the whole folder shouldn't be part of the pr. I will update it later.

laeubi:
@fipro78 would you mind adding the demo here:

https://github.com/eclipse-tycho/tycho/blob/main/tycho-its/src/test/java/org/eclipse/tycho/test/DemoTest.java

that way it will be executed with each build, you can even add some assertions about the resulting product / jars ...

laeubi:
Maybe you want to add it to the global demo readme as well for easier discovery:

https://github.com/eclipse-tycho/tycho/blob/main/demo/README.md

Also a readme in the demo folder itself would make it easier to guess what the demo is about for example like this:

https://github.com/eclipse-tycho/tycho/blob/main/demo/bnd-pde-workspace/README.md

laeubi:
Maybe "pde-automatic-manifest" would be a better folder name for the demo, bnd is used technically but not everyone will probably will care/know about that.

PR #3860, opened 2024-05-19
Always use full-named o.e.equinox.launcher artifact

HannesWell:
For for Mac OS X there was no equinox-launcher fragment for CARBON with full name, i.e. with <ws>.<os>.<arch> suffix, but only a org.eclipse.equinox.launcher.carbon.macosx fragment. But since Eclipse 4.2, released 2012, CARBON is not supported anymore at all [1]. Therefore the code adjust to that special naming schema, which also imposed difficulties or new supported architectures on Macos should be removed.

See also https://github.com/eclipse-equinox/equinox/pull/622

[1] - https://download.eclipse.org/eclipse/updates/4.2/R-4.2-201206081400

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/AbstractArtifactDependencyWalker.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/AbstractArtifactDependencyWalker.java
@@ -28,7 +28,6 @@ import org.eclipse.tycho.ArtifactDescriptor;
 import org.eclipse.tycho.ArtifactKey;
 import org.eclipse.tycho.ArtifactType;
 import org.eclipse.tycho.DependencyArtifacts;
-import org.eclipse.tycho.PlatformPropertiesUtils;
 import org.eclipse.tycho.ReactorProject;
 import org.eclipse.tycho.TargetEnvironment;
 import org.eclipse.tycho.core.ArtifactDependencyVisitor;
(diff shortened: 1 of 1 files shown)


laeubi:
@HannesWell before being too much concerned you should check if the method is actually of any use today, as the DepedencyWalker is to be removed and beside one place should be unused anyways.

HannesWell:
From the callers it looks like only `BuildQualifierAggregatorMojo` uses it at:
https://github.com/eclipse-tycho/tycho/blob/cf559c6d76aed18e2d06a9c58643039120caab57/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/BuildQualifierAggregatorMojo.java#L85

While I'm not so much concerned about the qualifier of a feature, I wonder if it would still break a build if the short-named launcher fragment is removed one day? Because without this, when build a product for mac the computation of its qualifier will probably fail, given it uses dependency-walker/visitor:
https://github.com/eclipse-tycho/tycho/blob/cf559c6d76aed18e2d06a9c58643039120caab57/tycho-core/src/main/java/org/eclipse/tycho/core/ArtifactDependencyVisitor.java#L34-L37

But I'm not sure if that is really used.

HannesWell:
Btw. are the test failures related to this change? They looks somewhat unrelated.

HannesWell:
Thinking again about the current only usage (determine the qualifier of repositories/products/features), I think we should just remove the special treatment of the equinox-launcher fragments here.
Other parts of Tycho or P2 already determine what's contained in a product so we don't have to replicate that logic here again, IMHO.
@laeubi what do you think about just removing the following code entirely?
https://github.com/eclipse-tycho/tycho/blob/b36b02b25e39375ba2a9f21f74d472e5752003bf/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/AbstractArtifactDependencyWalker.java#L126-L153

HannesWell:
> @laeubi what do you think about just removing the following code entirely?
> This would also help if for example in the future no launcher fragments are required at all anymore.

See https://github.com/eclipse-tycho/tycho/pull/3862. If we decide for the other PR, this one can be closed.

PR #3862, opened 2024-05-20
Remove equinox-launcher fragments handling in AbstractArtifactDependencyWalker

HannesWell:
The AbstractArtifactDependencyWalker is only used to determine the build-qualifier of of features, products and repositories and for that should only consider what's contained in those assembly-artifacts. It therefore should not add any special fragments to the set of considered artifacts.

This also helps to simplify the situation for equinox-launcher fragments for Mac, where still a special treatment for CARBON API artifacts exists, although the  org.eclipse.equinox.launcher.carbon.macosx artifact is not supported anymore since 2012 [1].

Required for https://github.com/eclipse-equinox/equinox/pull/622

[1] - https://download.eclipse.org/eclipse/updates/4.2/R-4.2-201206081400


laeubi (approved):

HannesWell:
Test failures are the same as on main and unrelated.

PR #3863, opened 2024-05-20
Remove special handling of equinox-launcher fragments in

HannesWell:
Back-port of https://github.com/eclipse-tycho/tycho/pull/3862


HannesWell:
In tycho-4 the DependencyWalker is not only used to determine the build-timestamp of a features/repo/product (as desribed in https://github.com/eclipse-tycho/tycho/pull/3860#issuecomment-2120798598) but also to assemble the content of a `deployable-feature`. But since features cannot contain products and the changed code traverses products, this should not have an effect for deployable-features.

Issue #3887, opened 2024-05-29
Not possible to remove qualifier for release build from bundles/features and product when using CI friendly versions


jstroleny:
I struggle with qualifier a bit and how to get rid of it for release build, as described below in issues.
I'm using maven CI friendly versions in project (revision, changelist and qualifier) from this https://github.com/eclipse-tycho/tycho/commit/b9c5d7304593a05795787a9f072300324bb6c066
and based on related issues https://github.com/eclipse-tycho/tycho/issues/603 / https://github.com/eclipse-tycho/tycho/issues/611
where it was described how to have release build versions using maven CI friendly versions.

I run it with integration test provided in given commit where I just added product (with qualifier as well) and I run it following:
`./mvnw clean install -Dtycho.buildqualifier.format=\'my-snapshot-prefix-\'yyyy -Dchangelist= -Dqualifier= -DskipTests -Dtycho.strictVersions=false`
but result is that maven projects and bundles are build without qualifier correctly and installed in maven repo correctly, but when product build starts (materialize-products / archive-products / etc), it assigns qualifier anyway and it makes inconsistent result from build process from maven CI friendly version point of view. 
Is there some way how to remove qualifier also for product in similar way as it is done for each bundle separately for maven submodules?

I have tried also this:
`./mvnw clean install -Dtycho.buildqualifier.format= -Dchangelist= -Dqualifier= -DskipTests -Dtycho.strictVersions=false`
as I'm really not interested to format build qualifier anyhow for release build, but the build fails at first bundle with:

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-packaging-plugin:4.0.8:build-qualifier (default-build-qualifier) on project project.util: Invalid build qualifier '', it does not match the OSGi qualifier constraint ([0..9]|[a..zA..Z]|'_'|'-') -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho:tycho-packaging-plugin:4.0.8:build-qualifier (default-build-qualifier) on project project.util: Invalid build qualifier '', it does not match the OSGi qualifier constraint ([0..9]|[a..zA..Z]|'_'|'-')
Caused by: org.apache.maven.plugin.MojoFailureException: Invalid build qualifier '', it does not match the OSGi qualifier constraint ([0..9]|[a..zA..Z]|'_'|'-')
[...]

PR #3891, opened 2024-05-30
Fix qualifier validation when qualifier is empty for release build

jstroleny:
Fixes https://github.com/eclipse-tycho/tycho/issues/3887

The fix allows to skip qualifier validation when qualifier is empty for release build defined using maven CI friendly versions.
The provided integration test verifies that build is not failing on validation and proceeds further.

### DIFF
--- a/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/BuildQualifierMojo.java
+++ b/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/BuildQualifierMojo.java
@@ -175,7 +175,9 @@ public class BuildQualifierMojo extends AbstractVersionMojo {
         }
 		String qualifier = getDesiredQualifier(timestamp);
 
-        validateQualifier(qualifier);
+		if (!"".equals(qualifier)) {
+			validateQualifier(qualifier);
+		}
 
 		String pomOSGiVersion = getUnqualifiedVersion();
 		String suffix = "." + qualifier;
--- a/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/BuildQualifierMojo.java
+++ b/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/BuildQualifierMojo.java
@@ -188,26 +188,26 @@ public class BuildQualifierMojo extends AbstractVersionMojo {
 		return new TychoProjectVersion(pomOSGiVersion, qualifier);
     }
 
-	protected String getDesiredQualifier(String forceContextQualifier, Date timestamp) throws MojoExecutionException {
-		String qualifier = forceContextQualifier;
-		if (TychoConstants.QUALIFIER_NONE.equals(qualifier)) {
-			qualifier = "";
-		}
+    protected String getDesiredQualifier(String forceContextQualifier, Date timestamp) throws MojoExecutionException {
+        String qualifier = forceContextQualifier;
+        if (TychoConstants.QUALIFIER_NONE.equals(qualifier)) {
+            qualifier = "";
+        }
 
         if (qualifier == null) {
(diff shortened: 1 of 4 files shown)


tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/BuildQualifierMojo.java
```diff
@@ -175,7 +175,9 @@ private TychoProjectVersion calculateQualifiedVersion(Date timestamp)
         }
 		String qualifier = getDesiredQualifier(timestamp);
 
-        validateQualifier(qualifier);
+		if (!"".equals(qualifier)) {
```
laeubi, line 178: Can this check be performed in `validateQualifier`?

laeubi:
@jstroleny you mentioned in your issue the pde-build has `forceContextQualifier=none` as tycho already supports `forceContextQualifier` we probably should support that instead of allowing an empty qualifier here in all cases?

jstroleny:
@laeubi : thank you for your comments, I have question to your last point, how you envision to use `forceContextQualifier` together with maven CI friendly versions where property is named `classifier` and pass through directly to the pom here https://github.com/eclipse-tycho/tycho/blob/13489c62a9464a9c71d4e0a4fefbbc65feb25ae8/tycho-build/src/main/java/org/eclipse/tycho/build/TychoCiFriendlyVersions.java#L94 and here
https://github.com/eclipse-tycho/tycho/blob/13489c62a9464a9c71d4e0a4fefbbc65feb25ae8/tycho-build/src/main/java/org/eclipse/tycho/build/TychoGraphBuilder.java#L89

I'd like to have consistent versions in maven artifact as well in manifest that's why I just allow to skip validation in `BuildQualifierMojo::validateQualifier(String)` as seems most natural (for maven CI friendly versions as well). We're allowing to pass empty qualifieranyway as seen on above lines, it just hits on validation which is not counting with empty value.

Maybe you mean to support both? E.g. allow all of these:
`mvn clean install -Dqualifier= -DskipTests ` (maven artifact is CI friendly and changed and manifest version as well)
`mvn clean install -DforceContextQualifier=none -Dqualifier= -DskipTests`  (dtto)
and
`mvn clean install -DforceContextQualifier=none -DskipTests`  (this will force populate qualifier to be empty, so that maven artifact is changed as well as manifest)

Is my understanding correct?

laeubi:
If you look [here](https://github.com/eclipse-equinox/p2/blob/68bad3217eba1a65ac867bf1fcb39359b1e384d1/pom.xml#L32) the qualifier is used as 

`<qualifier>-SNAPSHOT</qualifier>`

for the "maven version" this will result in the maven artifact is named `1.0.0-SNAPSHOT.jar` and the OSGI version has ist qualifier `1.0.0.<buildtime>`.

Now if I build with mvn ... `-DforceContextQualifier=none` is would expect that:

1. `TychoCIFriendlyVersion` treats this as empty so the Maven version will be `1.0.0`
2. `BuildQualifierMojo` threats it as "use no qualifer" so the OSGi version becomes `1.0.0`

Maybe some other parts needs adjustments then, but at laest we have a clear expression of "don't use a qualifier" versus "it was empty because of an error".

tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/BuildQualifierMojo.java
jstroleny: I reworked according our discussion to use `forceContextQualifier=none` and moved checks inside, preserving flow in calculate method. Let me know if it looks fine to you, I can make needed adjustments if necessary.

laeubi (approved):
Looks good to me thanks for the improvement and test case

PR #3896, opened 2024-05-31
Add support for JUnit 'vintage' attribute on classpath

laeubi:
JDT supports a [new classpath attribute](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/886) on the JUnit5 container that disables the vintage engine. This is especially important if one wants to write plain JUnit5 test as it is otherwise easy to mix up Junit4/5.

This now adds a testcase that shows that such mixed usage would fail compilation.

FYI @HeikoKlare as you are interested in JUnit5 migration.


HeikoKlare:
Great feature! Thank you for notifying me.

PR #3899, opened 2024-05-31
Add test for build with updated bundle versions

fipro78:
This adds a test that 
- first builds the pde-automatic-manifest demo project
- then updates the version of the whole demo using the tycho-versions-plugin
- then checks if the build succeeded
- finally reverts the version update

### DIFF
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/DemoTest.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/DemoTest.java
@@ -49,6 +49,32 @@ public class DemoTest extends AbstractTychoIntegrationTest {
 	}
 
 	@Test
+	public void testAutomaticManifestUpdateVersion() throws Exception {
+		// run the initial build
+		Verifier verifier = runDemo("pde-automatic-manifest");
+
+		// now update the version
+		verifier.executeGoals(
+				List.of("org.eclipse.tycho:tycho-versions-plugin:set-version", "-DnewVersion=1.1.0.qualifier"));
+		verifier.verifyErrorFreeLog();
+
+		// build again
+		verifier.executeGoals(List.of("clean", "verify"));
+		verifier.verifyErrorFreeLog();
+
+		BundleDescription description = BundlesAction.createBundleDescription(Path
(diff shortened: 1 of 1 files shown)


tycho-its/src/test/java/org/eclipse/tycho/test/DemoTest.java
```diff
@@ -48,6 +48,37 @@ public void testAutomaticManifest() throws Exception {
 				.anyMatch(pkg -> "tycho.demo.service.api".equals(pkg.getName())));
 	}
 
+	@Test
+	public void testAutomaticManifestUpdateVersion() throws Exception {
+		// run the initial build
+		Verifier verifier = runDemo("pde-automatic-manifest");
+
+		// now update the version
+		Verifier updateVerifier = super.getVerifier("../../demo/pde-automatic-manifest", true, true);
```
laeubi, line 57: I think you don't need a new verifier here just reuse the one from above.

laeubi:
> finally reverts the version update

As you should get a copy for the test i don't think this is required?

laeubi:
@fipro78 for some reason this now has compile errors :-\

fipro78:
It somehow has changed the import statement. Can't explain why, but that seems to be the reason.

laeubi:
@fipro78 now all test succeed... can you maybe give some manual instructions and where /when you would it expect to fail with the given example? I'll then take a look at it.

fipro78:
I followed my tutorial after I updated to the automatic manifest generation. And everytime I change the version back and forth the issue appears.

https://github.com/fipro78/e4-cookbook-basic-recipe/blob/master/tutorials/Eclipse_RCP_Cookbook_p2.md

akurtakov:
Is there plan to finish this one? If not I plan to close it next week.

fipro78:
@akurtakov 
To be honest, I am not sure if there is something missing. I added the test back when I was trying the automatic manifest generation with my RCP Cookbook. The test seems to succeed if I understand the comment from @laeubi in the discussion. I can't tell if the issue still persists. To verify this I would need to update the whole cookbook repository with all steps to get to the status to verify it with the current Eclipse Platform, PDE and Tycho.

That said, if you think this test case is not useful enough now (it is more than 2 years old) you can close it from my point of view. 

@laeubi What do you think?

PR #3913, opened 2024-06-01
Use dependency metadata instead of MavenProjectDependencyProcessor

laeubi:
Currently full dependency closure is computed to determine if a unit is a reactor project. Instead one could/should use the dependency metadata.

FYI @ptziegler


ptziegler:
I remember some issues with the old implementation when calculating the dependencies of test fragments. I wonder if this is also fixed by this...

Issue #3911, opened 2024-06-01
[sbom] BOM is calculated multiple times for the same unit


laeubi:
While [debugging some problems with changes in Tycho](https://github.com/eclipse-tycho/tycho/pull/3897) I noticed the following message:

```
[INFO] --- cyclonedx:2.7.9:makeBom (default) @ repository.eclipse-repository ---
[INFO] CycloneDX: Resolving Dependencies
[INFO] Cannot use cached value for MavenProject: tycho-demo:repository.eclipse-repository:1.0.0-SNAPSHOT @ /home/christoph/git/laeubi/tycho/tycho-its/projects/sbom/repository/.polyglot.category.xml because project artifact has changed, update is forced
[WARNING] Unknown p2 repository for artifact: example.feature
[ERROR] Unable to calculate BOM for: org.eclipse.update.feature,example.feature,1.0.0.today
[WARNING] Unknown p2 repository for artifact: example.feature
[ERROR] Unable to calculate BOM for: org.eclipse.update.feature,example.feature,1.0.0.today
[INFO] CycloneDX: Creating BOM version 1.4 with 8 component(s)
[INFO] CycloneDX: Writing and validating BOM (XML): /home/christoph/git/laeubi/tycho/tycho-its/projects/sbom/repository/target/bom.xml
[INFO]            attaching as repository.eclipse-repository-1.0.0-SNAPSHOT-cyclonedx.xml
[INFO] CycloneDX: Writing and validating BOM (JSON): /home/christoph/git/laeubi/tycho/tycho-its/projects/sbom/repository/target/bom.json
[INFO]            attaching as repository.eclipse-repository-1.0.0-SNAPSHOT-cyclonedx.json
[INFO] ------------------------------------------------------------------------
```

As one can see error/warning printed twice, so it seems the bom is also computed twice for `org.eclipse.update.feature`!

@ptziegler do you like to take a look? It might not be a problem for small examples, but with larger products/sites it can easily become a performance bottleneck.

ptziegler:
Unless I'm missing something, you're talking about CycloneDX generating both an XML and a JSON file?

laeubi:
> Wait, you're probably talking about the PURL for example.feature being calculated twice...

yes :-)

ptziegler:
To briefly describe what's happening:

[...]

PR #3914, opened 2024-06-01
Calculate BOM representation for each IU only once

ptziegler:
The BOM representation is currently calculated twice for each artifact. Within a reactor build, such IUs should always produce the same BOM representation and should therefore be cached.

Resolves https://github.com/eclipse-tycho/tycho/issues/3911

### DIFF
--- a/tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
+++ b/tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
@@ -13,12 +13,14 @@
 package org.eclipse.tycho.sbom;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.concurrent.ConcurrentHashMap;
 
 import javax.inject.Inject;
 
@@ -53,6 +55,8 @@ public class TychoProjectDependenciesConverter extends DefaultProjectDependencie
 
 	@Inject
 	private P2DependencyTreeGenerator dependencyGenerator;
+
+	private final Map<IInstallableUnit, List<String>> bomRepresentations = new ConcurrentHashMap<>();
 
 	@Override
 	public void cleanupBomDependencies(Metadata metadata, Map<String, Component> components,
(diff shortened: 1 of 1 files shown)


tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
```diff
@@ -57,6 +58,8 @@ public class TychoProjectDependenciesConverter extends DefaultProjectDependencie
 	@Inject
 	private MavenProjectDependencyProcessor dependencyProcessor;
 
+	private final Map<IInstallableUnit, List<String>> bomRepresentations = new HashMap<>();
```
laeubi, line 61: Better use a `ConcurentHashMap` to support parallel builds

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
```diff
@@ -117,35 +120,37 @@ private void convertToDependency(DependencyTreeNode node, Set<Dependency> depend
...
+				if (bomRef == null) {
+					LOG.error("Unable to calculate BOM for: " + p2artifactKey);
+					continue;
+				}
+				bomRefs.add(bomRef);
 			}
-			bomRefs.add(bomRef);
-		}
-		return bomRefs; // mutable!
+			return bomRefs; // mutable!
```
laeubi, line 153: The APi says the returned list is mutable, so either drop the mutability (and use a read-only list) or return a copy of the cached value.

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
```diff
@@ -117,35 +120,37 @@ private void convertToDependency(DependencyTreeNode node, Set<Dependency> depend
...
-		// (I) IU describes local reactor project
-		try {
-			ProjectDependencyClosure dependencyClosure = dependencyProcessor
-					.computeProjectDependencyClosure(reactorProjects, mavenSession);
-			MavenProject iuProject = dependencyClosure.getProject(iu).orElse(null);
-			if (iuProject != null) {
-				String bomRef = modelConverter.generatePackageUrl(iuProject.getArtifact());
-				if (bomRef == null) {
-					LOG.error("Unable to calculate BOM for: " + iuProject);
+		return bomRepresentations.computeIfAbsent(iu, ignore -> {
```
laeubi, line 123: why not move session + reactor project inside the computation?

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
ptziegler: It's been a few months, but I believe mutability was required in an earlier implementation because CycloneDX would add element to this list. But now it's only used within loops, so I think dropping this property is the best way to go.

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
ptziegler: A fair point. I just double-checked and the map is indeed shared across the entire build.

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
ptziegler: No particular reason. I just overlooked that part.

ptziegler:
Apologies, I missclicked and deleted the branch by accident...

laeubi:
@ptziegler can you rebase / resolve conflict?

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
```diff
@@ -108,38 +111,39 @@ private void convertToDependency(DependencyTreeNode node, Set<Dependency> depend
...
+			// (I) IU describes local reactor project
+			for (MavenProject project : reactorProjects) {
+				ReactorProject reactorProject = DefaultReactorProject.adapt(project);
+				Set<IInstallableUnit> initalUnits = reactorProject.getDependencyMetadata(DependencyMetadataType.INITIAL);
+				Set<IInstallableUnit> seedUnits = reactorProject.getDependencyMetadata(DependencyMetadataType.SEED);
+				if (initalUnits.contains(iu) || seedUnits.contains(iu)) {
+					String bomRef = modelConverter.generatePackageUrl(project.getArtifact());
+					if (bomRef == null) {
+						LOG.error("Unable to calculate BOM for: " + project);
+						return List.copyOf(bomRefs);
```
laeubi, line 131:
```suggestion
						return List.of();
```

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
```diff
@@ -108,38 +111,39 @@ private void convertToDependency(DependencyTreeNode node, Set<Dependency> depend
...
+					}
+					bomRefs.add(bomRef);
+					return List.copyOf(bomRefs);
```
laeubi, line 134:
```suggestion
					return List.of(bomRef);
```

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
```diff
@@ -108,38 +111,39 @@ private void convertToDependency(DependencyTreeNode node, Set<Dependency> depend
...
+				ReactorProject reactorProject = DefaultReactorProject.adapt(project);
+				Set<IInstallableUnit> initalUnits = reactorProject.getDependencyMetadata(DependencyMetadataType.INITIAL);
+				Set<IInstallableUnit> seedUnits = reactorProject.getDependencyMetadata(DependencyMetadataType.SEED);
+				if (initalUnits.contains(iu) || seedUnits.contains(iu)) {
+					String bomRef = modelConverter.generatePackageUrl(project.getArtifact());
+					if (bomRef == null) {
+						LOG.error("Unable to calculate BOM for: " + project);
+						return List.copyOf(bomRefs);
+					}
+					bomRefs.add(bomRef);
```
laeubi, line 133:
```suggestion
```

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
```diff
@@ -108,38 +111,39 @@ private void convertToDependency(DependencyTreeNode node, Set<Dependency> depend
...
-		for (MavenProject project : reactorProjects) {
-			ReactorProject reactorProject = DefaultReactorProject.adapt(project);
-			Set<IInstallableUnit> initalUnits = reactorProject.getDependencyMetadata(DependencyMetadataType.INITIAL);
-			Set<IInstallableUnit> seedUnits = reactorProject.getDependencyMetadata(DependencyMetadataType.SEED);
-			if (initalUnits.contains(iu) || seedUnits.contains(iu)) {
-				String bomRef = modelConverter.generatePackageUrl(project.getArtifact());
+		return bomRepresentations.computeIfAbsent(iu, ignore -> {
+			final MavenSession mavenSession = legacySupport.getSession();
+			final List<MavenProject> reactorProjects = mavenSession.getAllProjects();
+			final List<String> bomRefs = new ArrayList<>();
```
laeubi, line 121:
```suggestion
```

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
```diff
@@ -108,38 +111,39 @@ private void convertToDependency(DependencyTreeNode node, Set<Dependency> depend
...
+					String bomRef = modelConverter.generatePackageUrl(project.getArtifact());
+					if (bomRef == null) {
+						LOG.error("Unable to calculate BOM for: " + project);
+						return List.copyOf(bomRefs);
+					}
+					bomRefs.add(bomRef);
+					return List.copyOf(bomRefs);
+				}
+			}
+			// (II) IU describes external artifact
```
laeubi, line 137:
```suggestion
			// (II) IU describes external artifact
			final List<String> bomRefs = new ArrayList<>();
```

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
ptziegler: This won't compile as `bomRefs` is already a list. But to avoid copying the array, I now use `Collections.unmodifiableList(...)`

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoProjectDependenciesConverter.java
ptziegler: I never really liked `List.copyOf()` and instead I use `Collections.emptyList()` to make it clear that this is an empty list.

PR #3916, opened 2024-06-01
Make Tycho SBOM mapper configurable w.r.t. the project packaging types

ptziegler:
Instead of being applied on all packaging types (even those where it
doesn't make sense like the target platform), the user is now able to
configure for which project types the Tycho model converter is used.

By default (if nothing else is specified), this converter is only used
for plugins and features, to avoid issues when the e.g. the target
definition is defined in its own module or the the reactor contains
non-Tycho packaging types (jar/pom/bundle/...).


ptziegler:
The test will likely fail because the current implementation can't handle multiple test fragments. But this should be fixed with https://github.com/eclipse-tycho/tycho/pull/3913.

laeubi:
@ptziegler do you plan to finish this one? There could probably be a new release coming at the end of the Platform release cycle.

ptziegler:
The tests work locally, let's hope Jenkins agrees as well.

ptziegler:
Hm... of course that's the problem; The log file is simply getting too big.
Because the test is executed on `debug` level, everything that is logged by CycloneDX will show up in the error log. And this is just too much for the test suite to handle.

ptziegler:
A log file with size 6.1GB... dear lord.

merks:
It’s probably rife with important and relevant details that you will wish to peruse carefully and fully.

ptziegler:
The remaining test failure seems unrelated. To keep things simple, I've removed the debug flag for the SBOM tests, which should drastically reduce the size of the build log.

laeubi:
@ptziegler should this be backported to Tycho 4 branch? If yes please add the label to PR. I have now rebased and enabled automerge.

ptziegler:
@laeubi 

If would be great if this could be back-ported, just to have a consistent behavior accross the board. But I fear I don't have permission to add any labels to this PR.
image

PR #3918, opened 2024-06-02
Fix OsgiCompilerTest for Java 21 JVM

Bananeweizen:
That test expects the byte code level to not be influenced by settings in POM or manifest. Therefore it must match the JVM level.

On my machine that is the only failing test when running on Java 21.


laeubi (approved):
Thanks for tackling this!

Issue #2875, opened 2023-10-02
fix-artifacts-metadata doesn't fix the md5.checksum


henryju:
Hi,

Using Tycho 4.0.2.

We are including some third-party dependencies in our p2 repository, and we are signing all artifacts with our own certificate. As a result, the checksums are incorrect, so we are [using](https://github.com/SonarSource/sonarlint-eclipse/blob/a37999db6a8de0e26732f001045eac1fb0c65458/org.sonarlint.eclipse.site/pom.xml#L83) the `fix-artifacts-metadata` goal to recompute all hashes.
A user [reported](https://github.com/SonarSource/sonarlint-eclipse/pull/482#issuecomment-1741052737) that the sh256 checksum is correct, while the md5 one is still the old one from the source (Orbit) repository.

I think I will simply exclude Orbit artifacts from our sign process, since they are already signed anyway, but maybe there is an issue to fix in Tycho as well?

laeubi:
@henryju I think the issue is that P2 has dropped md5 support, so I think Tycho must remove the md5 when mirroring things.

mdaloia:
@laeubi We came across this same issue using 4.0.7 (current latest).

It seems that the checksums are re-calculated but only for those that the current director version used by Tycho knows that it needs to be published.
The problem is that old checksums (like the md5) are left there with an outdated value and if you use a p2 director application that it is old (before md5 was disable) it fails the check. For example:

- Modifying a jar:

```terminal
[INFO] --- antrun:3.1.0:run (touch-jar) @ product ---
[INFO] Executing tasks
[WARNING]      [echo] File original MD5 = e88a1e99c0c9ee29b96b3159b54e419a
[WARNING]      [echo] File original SHA-1 = 0ef3b09190da2fb64d81985372e44c974ccd8468
[WARNING]      [echo] File original SHA-256 = c49f54f646587f98b9cbfd5239f34e3f52d3f85a1aa1f68b416192a563bca3ec
[WARNING]      [echo] File original SHA-512 = 22971289b7d28d47a102b339f7b896f3c4be517bc9bcb48d8ad2b5f298503cb0f57822c199a5961fb74fabecfefc22d4051dd531259301808dd024cfbf018e2d
[INFO]     [unzip] Expanding: /Users/mdaloia/tycho-issue/product/target/repository/plugins/org.eclipse.nebula.cwt_1.1.0.202402011804.jar into /Users/mdaloia/tycho-issue/product/target/tmp
[...]

PR #3921, opened 2024-06-03
Add Integration Test for issue #2875

mdaloia:
This integration test shows that `download.md5` and `download.checksum.md5` are present after running the `fix-artifacts-metadata` goal when they shouldn't.

If these are not removed and the content of the artifact changed then, after the recalculation done by this goal, new p2 libs will see an outdated value for these old checksum algorithms and will fail the installation due to a checksum mismatch.


laeubi:
This was already integrated in the main branch.

PR #3979, opened 2024-06-14
Check that the 34 test is seeing an older version

laeubi:
Exact version check complicates things as it has to be updated for every minor bump of equinox.

FYI @akurtakov  so we do not need to increment this on every equinox change!


merks:
Don’t let the license thing hold you up. It’s a glitch of some sort.

laeubi:
It currently breaks the auto-merge feature 😢

akurtakov:
Thanks, makes sense.

PR #4006, opened 2024-06-19
Use Maven 3.9.8 in gh actions

akurtakov:


akurtakov:
No need to wait for jenkins as the change is only in gh actions.

PR #4007, opened 2024-06-19
Switch to JSR330

akurtakov:
The following plugins and components are switched tycho-apitools-plugin, tycho-artifactcomparator, tycho-baseline-plugin, tycho-bnd-plugin, tycho-build are migrated.


laeubi:
I found that when one converts a component to JSR330 it might be necessary to use `@Inject` instead of `@Component` in Mojos that want to use that one.

PR #4067, opened 2024-07-30
Fix display of logo in readme after master->main branch change

Godin:
Here is how README is displayed in GitHub without this change:

Screenshot 2024-07-30 at 10 21 28


laeubi:
Thanks!

PR #4212, opened 2024-09-05
De-plexus Tycho

cstamas:
Applied https://cwiki.apache.org/confluence/display/MAVEN/Maven+Ecosystem+Cleanup

Changes:
* removed ol' Plexus
* and related classes like Logger, Disposable, etc


p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/DefaultRepositoryIdManager.java
```diff
@@ -19,27 +19,38 @@
...
  * Helper class for the Remote*RepositoryManagers taking care of mapping repository URLs to the
  * settings.xml-configured mirrors and setting passwords.
  */
-@Component(role = IRepositoryIdManager.class)
+@Singleton
+@Named
 public class DefaultRepositoryIdManager implements IRepositoryIdManager {
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    // TODO: wut? repositoryIds may CHANGE per project, as project may define new ones
```
laeubi, line 41: Maven wants server-id to get authenticaction / proxy infos, as P2 repository has no ID we let this component "learn" what URI was used in a given id context (e.g. pom or target file), so later we can map them back to ask if maven has credentials for that ID.

As credential IDs are global one can't use different credentials/mirrors/proxy per ID and even if in theory it could be different per project this actually won't work well

cstamas:
@laeubi updated with latest main

cstamas:
@laeubi you ported these or should I try to rebase?

laeubi:
@cstamas  I have ported *some* of the changes (but not all) and at least sometimes look at your changes to get an idea how to convert things, so if you want to take the effort a rebase would be useful.

What @akurtakov want to say is maybe its better to just "deplexus" one little part/component (so it can even be merged) instead of one big change but you can of course decide.

Issue #4214, opened 2024-09-07
Stay away from plexus-cipher 2.1.0


cstamas:
Maven is to stay on 2.0 as well. Basically Tycho with 2.1.0 is incompatible with Maven !

2.1.0 contains this commit:
https://github.com/codehaus-plexus/plexus-cipher/commit/176289fcc2780668208ea089fdcacc0c7145f7b1

PR message explains what is the problem:
https://github.com/codehaus-plexus/plexus-cipher/pull/23

In short, who/whatever moves to 2.1.0 needs to "reencrypt" everything. Moreover, as Maven uses 2.0, and Tycho would use 2.1.0, the two would not be compatible. Maven 3 probably remains on (flawed) 2.0 forever, as this "cipher" is anyway to be removed from future maven versions.

PR #4216, opened 2024-09-08
Do not explicitly include plexus-cipher

laeubi:
Lets see if we can work without it.

Fix https://github.com/eclipse-tycho/tycho/issues/4214

FYI @cstamas


laeubi:
Seems we not really need it for compiling (or running) so @cstamas @akurtakov any reason to keep it as an explicit managed dependency?!?

akurtakov:
Not that I know of.

PR #4230, opened 2024-09-09
Remove Mojo test cases

laeubi:
Tycho currently contains some so called "Mojo Testcases" that are ones that (partly) simulate the execution of a mojo, but these have some problems:

1) They currently depend on heavy outdated (and deprecated) maven-plugin-testharness
2) They often do not have a full test setup so we need to have some quirks in the code (e.g components are null)
3) They are often hard to understand as they are a mixture of (incomplete) maven setups with things setup in the code 4) We even have some constructors only used to support these cases 5) As they often test very specific internal behavior they often need adjustments.
6) For the Maven 4 transition one needs to adjust them to new maven4 api and components and executions.

On the other hand we have a very good integration suite that is made of actual executable examples and test, changing internal implementation details does not affect these or show clear use-cases that break.

Because of this and to prepare for the Maven 4 transition these tests are now deleted.

@akurtakov any objections?


akurtakov:
If you believe that what these test is covered by integration tests - go for it, otherwise we risk breaking functionality later on  if no test cover it anymore.

PR #4248, opened 2024-09-10
Adopt new o.e.e.p2.publisher.eclipse API for IMacOsBundleUrlType

sratz:
After #4246, we can now consume the new

org.eclipse.equinox.internal.p2.publisher.eclipse.IProductDescriptor.getMacOsBundleUrlTypes()

API in the Tycho wrapper
org.eclipse.tycho.p2.tools.publisher.ExpandedProduct.

This is necessary to properly support link handler registrations in product's launchers.

See
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1901 https://github.com/eclipse-equinox/p2/pull/533
for more information.


sratz:
@laeubi:
Can we get this to 4.0.x? That would require to
bump org.eclipse.platform:org.eclipse.equinox.p2.publisher.eclipse from 1.6.0 to 1.6.200
there, too. Do you see any problem with that?

laeubi:
Should we merge or is anything left?

PR #4249, opened 2024-09-10
[tycho-4.0.x] Adopt new o.e.e.p2.publisher.eclipse API for IMacOsBundleUrlType

sratz:
Backport of https://github.com/eclipse-tycho/tycho/pull/4246 + https://github.com/eclipse-tycho/tycho/pull/4248

### DIFF
--- a/p2-maven-plugin/pom.xml
+++ b/p2-maven-plugin/pom.xml
@@ -28,7 +28,7 @@
 		<dependency>
 			<groupId>org.eclipse.platform</groupId>
 			<artifactId>org.eclipse.equinox.p2.core</artifactId>
-			<version>2.11.0</version>
+			<version>2.12.100</version>
 		</dependency>
 		<dependency>
 			<groupId>org.eclipse.platform</groupId>
@@ -47,7 +47,7 @@
 		<dependency>
 			<groupId>org.eclipse.platform</groupId>
 			<artifactId>org.eclipse.equinox.p2.engine</artifactId>
-			<version>2.10.0</version>
+			<version>2.10.200</version>
 		</dependency>
 		<dependency>
 			<groupId>org.eclipse.platform</groupId>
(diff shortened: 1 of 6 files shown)


laeubi:
@sratz I think we last release skipped this update due to some incompatibility for Tycho 4, so one needs probably other updates/bugfixes as well. I usually try to compare some things from `main` where the update was performed to see if something is missing or can be adapted.

sratz:
> @sratz I think we last release skipped this update due to some incompatibility for Tycho 4, so one needs probably other updates/bugfixes as well. I usually try to compare some things from `main` where the update was performed to see if something is missing or can be adapted.

I am wondering where to get started, though.

I don't have any idea what to make of the generic

_org.apache.maven.it.VerificationException: Error in execution: [ERROR] [c8a6beaa-0a0d-4839-9687-0680bcd338fc][extension>org.eclipse.tycho:tycho-maven-plugin:4.0.9-SNAPSHOT] org.eclipse.equinox.p2.publisher.eclipse
	at org.apache.maven.it.Verifier.verifyErrorFreeLog(Verifier.java:312)_

error message that all tests are failing with :(

laeubi:
Yes that's a bit annoying as it hides the exception.

The easiest in such case is to do:

- `cd tycho-its`
- `mvn clean install -Pits -Dtest=<fqdn of test>`

That way you can run one test and hopefully have access to everything (e.g. logs).

In this particular case, one can maybe better run

- `cd tycho-its/projects/reactor.makeBehaviour`
- `mvn clean install -Dtycho-version=4.xx`

to see the real error that is printed.

sratz:
I updated all the P2 artifacts and their bouncycastle deps to be the same as in the `main` branch.

Looks pretty good and the one test failure on macOS seems unrelated.

Any concerns with this change?

PR #4289, opened 2024-09-19
Fix rendering markdowns in generated site

akurtakov:
Due to https://issues.apache.org/jira/browse/DOXIA-740 if a heading level is skipped the whole document is not rendered.


laeubi:
Many thanks! Waiting for the build is not required as only the publishing action will show it works.

PR #4295, opened 2024-09-24
Add swt verification build

laeubi:
This will hopefully catch some issues earlier in the future.

FYI @HeikoKlare

### DIFF
--- a/.github/workflows/verify-platform.yml
+++ b/.github/workflows/verify-platform.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Checkout platform code
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
-       path: 'platform'
+       path: 'aggregator'
        submodules: 'recursive'
        lfs: true
        repository: 'eclipse-platform/eclipse.platform.releng.aggregator'
@@ -60,7 +60,7 @@ jobs:
     - name: Run Platform Build
       env:
        JAVA_HOME: ${{ env.JAVA_HOME_17_X64 }}
-      working-directory: 'platform'
+      working-directory: 'aggregator'
       run: >- 
         mvn -U -V -e -B -ntp
         --global-toolchains ${{ github.workspace }}/tycho/.github/toolchains.xml
--- a/.github/workflows/verify-platform3.yml
+++ b/.github/workflows/verify-platform3.yml
@@ -1,0 +1,77 @@
+# This workflow will build a "Eclipse Platform Individual Bundles" using the eclipse.platform repository as it contains the largest set of inter-related bundles
+# For more information see: https://github.com/eclipse-platform/eclipse.platform
+
+name: Verify Eclipse Platform SWT Build
+concurrency: 
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
+on:
+  pull_request:
+    branches: 
+     - 'main'
+
+jobs:
+  build-swt:
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config: 
+          - { name: Linux, os: ubuntu-latest, native: gtk.linux.x86_64 }
+          - { name: Windows, os: windows-latest, native: win32.win32.x86_64 }
+          - { name: MacOS x86, os: macos-13, native: cocoa.macosx.x86_64 }
+          - { name: MacOS ARM, os: macos-latest, native: cocoa.macosx.aarch64 }
+    name: Verify ${{ matrix.config.name }}
+    steps:
+    - name: Checkout tycho code
+      uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      with:
+       path: 'tycho'
+    - name: Checkout platform code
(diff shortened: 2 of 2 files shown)


laeubi:
@HeikoKlare for some reason the SWT build succeeds here so do I need to enable something special?

HeikoKlare:
The build succeeds, but when looking at the logs it does not seem to do anything meaningful. It seem to fail at some early point of execution without making it the action fail, doesn't it? I cannot see any information about bundles being compiled. But I also cannot tell what's wrong with the actions specification so that the action is not executed properly.

Even if that was fixed, the build may succeed because it's only executed on Linux. For some reasons the GH actions Linux build for SWT currently succeeds. Only the MacOS builds are affected. As you guessed, this may be for reasons of ordering. So if we want to have that additional workflow fail for the current issue in SWT, we would probably have to make it a Matrix build also being executed on MacOS.

laeubi:
> The build succeeds, but when looking at the logs it does not seem to do anything meaningful.

I enabled the test and now something is borked, before it run without a problem :-\

> Even if that was fixed, the build may succeed because it's only executed on Linux. For some reasons the GH actions Linux build for SWT currently succeeds. Only the MacOS builds are affected.

Good point, it seems I need to setup a matrix-build for this case.

HeikoKlare:
I am not familiar with the existing verification builds, but taking a look at logs of the existing ones, they do also not provide reasonable information about the Maven build / indicate a proper execution of the Maven build. Is this intended/expected or may there be some issue with these verification builds overall?
For example, the [Eclipse Platform Individual Bundles Build](https://github.com/eclipse-tycho/tycho/actions/workflows/verify-platform2.yml) has similar logs. I took at look at, e.g., this build from some days ago: https://github.com/eclipse-tycho/tycho/actions/runs/10935901809/job/30358488371

laeubi:
> Is this intended/expected or may there be some issue with these verification builds overall?

Actually it should look like a usual maven build ... seems I need to take a look but can't spot anything obvious the log looks like the process just terminates somehow.

laeubi:
At least the ibuild looks fine, except the usual failure due to the CBI plugin:
- https://github.com/eclipse-tycho/tycho/actions/runs/11010139398/job/30571375964?pr=4295

HeikoKlare:
If I see correctly, the first verification build ([Verify Eclipse Platform Aggregator Build](https://github.com/eclipse-tycho/tycho/blob/main/.github/workflows/verify-platform.yml)) seems to work fine, while the existing second ([Verify Eclipse Platform Individual Bundles Build](https://github.com/eclipse-tycho/tycho/blob/main/.github/workflows/verify-platform2.yml)) and the one contributed in this PR do not run as expected. The only real difference I see is in the configuration of the toolchain and settings files for that build, though I am not sure whether them being missing may cause the issues:
https://github.com/eclipse-tycho/tycho/blob/6c80860d853832a08fe2339dc042a0d38565b051/.github/workflows/verify-platform.yml#L66
https://github.com/eclipse-tycho/tycho/blob/6c80860d853832a08fe2339dc042a0d38565b051/.github/workflows/verify-platform.yml#L68

akurtakov:
> At least the ibuild looks fine, except the usual failure [due to the CBI plugin](https://github.com/eclipse-cbi/org.eclipse.cbi/pull/542):
> 
>     * https://github.com/eclipse-tycho/tycho/actions/runs/11010139398/job/30571375964?pr=4295

This one is fine now.

laeubi:
I now added a matrix for the SWT test lets see what happens.

laeubi:
Okay now windows can't evaluate the shell expression to find the Maven version

`mvn -ntp --batch-mode -Pbuild-individual-bundles -Pbree-libs -DskipTests -Dtycho.version=$(mvn help:evaluate -f D:\a\tycho\tycho/tycho -Dexpression=project.version -q -DforceStdout) -T1C clean verify`

anyone has an idea for a cross-platform solution?!?

merks:
Are you expecting it to invoke powershell rather than sh.exe or bash.exe?

image

laeubi:
> Are you expecting it to invoke powershell rather than sh.exe or bash.exe?

Good point I tried to configure `bash`  as a shell explicitly now.

laeubi:
It is getting further but now path separators are messed up:

> `-Dtycho.version=$(mvn help:evaluate -f D:\a\tycho\tycho/tycho -Dexpression=project.version -q -DforceStdout)`

laeubi:
Interestingly the Windows build now works but the aggregator build shows the compile problem ...

laeubi:
I'll backport this manually ... if the action runs I think we can use it anyways it is better than having no cros--platform checks even though we not see the exact problem now.

laeubi:
@HeikoKlare I now see the SWT builds running with the Tycho snapshot, and even though they seem to succeed I would merge this or do see anything left here?

After that I'll backport this to tycho4.x branch

HeikoKlare:
> I now see the SWT builds running with the Tycho snapshot, and even though they seem to succeed I would merge this or do see anything left here?

It looks good to me.
It's interesting that the SWT compilation succeeds. I guess the behavior is non-deterministic: depending on the number of already completed compilations (the fragments are built in parallel), there may be multiple matching classes (i.e., with the same FQN) on the classpath. The compiler may just select any of them and in case it is the "correct" one, the compilation succeeds. But as said, that's just a guess.

PR #4296, opened 2024-09-24
Don't use repo system for standalone director

laeubi:


laeubi:
Superseded by
- https://github.com/eclipse-tycho/tycho/pull/5266

PR #4317, opened 2024-09-29
Support specifying the baseline artifact version range

HannesWell:
This can be used to select the desired baseline artifact if multiple versions of an artifact are available and to compare.

Instead of specifing the version range explicitly we could also define an enum similar to `ReportBehavior` to specify 'named version rules' for the baseline artifact, like `same-major` and `latest`.

### DIFF
--- a/tycho-extras/tycho-p2-extras-plugin/src/main/java/org/eclipse/tycho/plugins/p2/extras/CompareWithBaselineMojo.java
+++ b/tycho-extras/tycho-p2-extras-plugin/src/main/java/org/eclipse/tycho/plugins/p2/extras/CompareWithBaselineMojo.java
@@ -106,6 +106,17 @@ public class CompareWithBaselineMojo extends AbstractMojo {
     @Parameter(property = "onIllegalVersion", defaultValue = "fail", alias = "tycho.p2.baseline.onIllegalVersion")
     private ReportBehavior onIllegalVersion;
 
+    /**
+     * The version range of the reference artifact in the baseline.
+     * <p>
+     * This can be used to select the desired baseline artifact if multiple versions of an artifact
+     * are available and to compare. By default a local artifact is compared to the latest version
+     * in the baseline.
+     * </p>
+     */
+    @Parameter(defaultValue = "0.0.0", alias = "tycho.p2.baseline.versionRange")
+    private String baselineVersionRange = "0.0.0";
+
     @Component()
     P2ResolverFactory resolverFactory;
 
@@ -161,7 +172,7 @@ public class CompareWithBaselineMojo extends AbstractMojo {
             try {
                 String id = item.getId();
                 Version version = new Version(item.getVersion().toString());
-                P2ResolutionResult res = resolver.resolveInstallableUnit(baselineTP, id, "0.0.0");
+                P2ResolutionResult res = resolver.resolveInstallableUnit(baselineTP, id, baselineVersionRange);
 
                 for (Entry foundInBaseline : res.getArtifacts()) {
                     Version baselineVersion = new Version(foundInBaseline.getVersion());


tycho-extras/tycho-p2-extras-plugin/src/main/java/org/eclipse/tycho/plugins/p2/extras/CompareWithBaselineMojo.java
```diff
@@ -106,6 +106,17 @@ public static enum ReportBehavior {
...
+    /**
+     * The version range of the reference artifact in the baseline.
+     * <p>
+     * This can be used to select the desired baseline artifact if multiple versions of an artifact
+     * are available and to compare. By default a local artifact is compared to the latest version
+     * in the baseline.
+     * </p>
+     */
+    @Parameter(defaultValue = "0.0.0", alias = "tycho.p2.baseline.versionRange")
+    private String baselineVersionRange = "0.0.0";
```
laeubi, line 118: This seems hardly usable, as one needs to explicitly configure the version per project (also `0.0.0` is not really a version range).

In PDE features we have the [concept of a "match rule"](https://help.eclipse.org/latest/topic/org.eclipse.pde.doc.user/guide/tools/editors/feature_editor/feature_editor.htm?cp=4_3_2_1) that might be more sufficient here as we then can compute a version range based on the current project version.

laeubi (changes requested):
I thing an explicit range is not really suitable here.

Also it is a bit unclear how to use it, maybe provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the use-case would be helpful here also to prevent regressions in the future.

tycho-extras/tycho-p2-extras-plugin/src/main/java/org/eclipse/tycho/plugins/p2/extras/CompareWithBaselineMojo.java
HannesWell: Yes that's what I meant in my initial comment with:
> Instead of specifying the version range explicitly we could also define an enum similar to ReportBehavior to specify 'named version rules' for the baseline artifact, like `same-major` and `latest`.

So I think the main question is what rules to support?
We could start simple and just have something like `same-major` and `latest`. Using the exact set of rules from PDE doesn't look suitable to me, because `Perfect` and `Greater or Equal` doesn't make sense in this context, because this mojo also checks for backward steps.
So besides `same-major` and `latest` actually only `same-minor` might be of interest.
But the names of the rules could be more speaking.

For the implementation we maybe could re-use PDE's new:
https://github.com/eclipse-pde/eclipse.pde/blob/3f973e87fd7ede513388e8ed870d1b65de438947/ui/org.eclipse.pde.core/src/org/eclipse/pde/core/plugin/VersionMatchRule.java

HannesWell:
> Also it is a bit unclear how to use it, maybe provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the use-case would be helpful here also to prevent regressions in the future.

The idea was to use it for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3027, but the old version might vanish anyways soon.

tycho-extras/tycho-p2-extras-plugin/src/main/java/org/eclipse/tycho/plugins/p2/extras/CompareWithBaselineMojo.java
laeubi: > So I think the main question is what rules to support?

We can just support what is useful today and add more if required, I think thats how its already done on some other places in Tycho

> We could start simple and just have something like same-major and latest.

I would prefer to just use the same naming as for features as this is a P2 concept and we should not mix to much naming.

>  Using the exact set of rules from PDE doesn't look suitable to me, because Perfect and Greater or Equal doesn't make sense in this context, 

Greater or equal is actually what is the default today, the only one not so useful might be "Perfect" that actually would never match anything.

> For the implementation we maybe could re-use PDE's new:

Tycho has already `org.eclipse.tycho.model.Feature.getVersionRange(ImportRef)` we can simply generalize that method, so I don't think relying on PDE-UI in Tycho just to reuse an enum is very useful, if we ever do
- https://github.com/eclipse-pde/eclipse.pde/issues/34

it would of course be good to reuse as much as possible.

laeubi:
> The idea was to use it for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3027, but the old version might vanish anyways soon.

So do we still need this?

HannesWell:
> > The idea was to use it for [eclipse-jdt/eclipse.jdt.core#3027](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3027), but the old version might vanish anyways soon.
> 
> So do we still need this?

At the moment I don't have interest and time to continue this. Closing for now.

PR #4318, opened 2024-09-29
Revert "Add a flag to skip the service-component file check in packaging-plugin"

HannesWell:
Reverts eclipse-tycho/tycho#4280

For the intended use-case this is not necessary anymore and seems to be too specific for a general usage.

For the tycho 4.0.x-line I suggest to deprecate this for removal.

@laeubi Should we do the same for https://github.com/eclipse-tycho/tycho/pull/4279? It's also not necessary anymore for the intended use-case, but seems to be more general and might be useful as such?


laeubi:
I would be fine with just keeping the flag, it is maybe seldom used but does not contribute to the complexity of the mojo nor limits its implementation.

HannesWell:
> I would be fine with just keeping the flag, it is maybe seldom used but does not contribute to the complexity of the mojo nor limits its implementation.

Sure that's also fine for me. :)
 Then lets close this one.

PR #4360, opened 2024-10-16
Use records for inner data carrier classes in TargetDefinitionFile

HannesWell:
Using records saves a significant number of lines.

I assume the implicit implementation of `hashCode()/equals()` and `toString()` isn't a problem, is it?

### DIFF
--- a/tycho-targetplatform/src/main/java/org/eclipse/tycho/targetplatform/TargetDefinitionFile.java
+++ b/tycho-targetplatform/src/main/java/org/eclipse/tycho/targetplatform/TargetDefinitionFile.java
@@ -151,90 +151,26 @@ public final class TargetDefinitionFile implements TargetDefinition {
 
     }
 
-    private static class TargetRef implements TargetDefinition.TargetReferenceLocation {
-
-        private String uri;
-
-        public TargetRef(String uri) {
-            this.uri = uri;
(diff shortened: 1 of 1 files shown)


tycho-targetplatform/src/main/java/org/eclipse/tycho/targetplatform/TargetDefinitionFile.java
```diff
@@ -151,90 +151,22 @@ public String getVersion() {
...
-				String label) {
-            this.roots = roots;
-            this.includeDependencyScopes = includeDependencyScopes;
-            this.manifestStrategy = manifestStrategy;
-            this.includeSource = includeSource;
-            this.instructions = instructions;
-            this.dependencyDepth = dependencyDepth;
-            this.repositoryReferences = repositoryReferences;
-			this.label = label;
-            this.featureTemplate = featureTemplate == null ? null : (Element) featureTemplate.cloneNode(true);
```
laeubi, line 221: The element is no longer cloned as far as I can see.

HannesWell: Indeed. Thanks.
Fixed it.

PR #4374, opened 2024-10-21
Fix spelling in tycho-ds and make message more clear

kysmith-csg:


laeubi:
@merks @akurtakov @HannesWell what do you think? I'm not sure if the wording "new" is really appropriate here, generated code is usually always "new".

HannesWell:
I agree, especially in the context of CI builds where the repo is usually clean in the beginning and the missing components are always generated. 

The word `existing` could also be `preexisting` or `already existing`. But just existing is also fine for me.

merks:
Yes, `existing` is fine.

kysmith-csg:
Thanks all for the feedback. Updated to remove "new".

laeubi (approved):
Looks fine to me

PR #4379, opened 2024-10-22
Make JUnit 5 versions parameters

akurtakov:
This should reduce the number of PRs and commits on new JUnit versions


akurtakov:
I'm looking to have less of 
image
on the next update.

PR #4388, opened 2024-10-25
buffer file outputs

Bananeweizen:
This makes "Resolving MavenDependencyRoots..." output somewhat faster on my machine. I started from org.eclipse.tycho.core.resolver.MavenTargetDefinitionContent.generateSourceBundle(String, String, Manifest, File, IArtifactFacade), which took notable time in the profiler, and I then added buffered outputstreams and buffered writers to all FileOutputStreams.

I've run a local build of another project with this snapshot, but I have not run Tycho tests (my company machine fails too much with proxy stuff there and I can't get to my private machine for some days).


laeubi (approved):
Looks like a great improvement, too sad the OS is not capable of giving better performance by default.

PR #4408, opened 2024-11-04
Only add matching fragments to the classpath 

laeubi:
Supersedes #4293


HeikoKlare:
Awesome. Thank you, @laeubi! 🥇 

I have not taken a look at the change itself, but I have successfully tested whether it fixes the original issue (in the SWT build). To do so, I did the following on a MacOS laptop:
- Compiled and installed Tycho based on current `4.0.x` branch and ran the SWT build with that version. This results in the build failure (during `testCompile` of the Windows fragments) as reported here: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2360#issuecomment-2363386380
- Cherry-picked the fix of this PR to the `4.0.x` branch, compiled and installed Tycho version based on that state, and ran the SWT build with that version again. This succeeds without the build failure during `testCompile` of the Windows fragments.

I ran the builds multiple times to hopefully exclude any potential flaky behavior.

PR #4411, opened 2024-11-04
Adopt to new GPG Mojo

laeubi:


laeubi:
@merks @mickaelistria I just restored the current state here, but given that maven now support BC as well by default some of the code seems redundant now and we might be able to use the maven implementation directly.

If there is anything needed for that please let me know and I'll try to propose a change upstream as done with
- https://github.com/apache/maven-gpg-plugin/pull/112

I think its not so valuable and probably hindering future updates when we maintain our own BC implementation here.

Issue #2586, opened 2023-06-26
Tycho creating tar files with extended header keyword 'LIBARCHIVE.creationtime'


jonahgraham:
I recently updated to newer Tycho  (2.7.1 -> 3.0.5)  for creating EPP and now I get this warning for each file in the tar when I untar on Linux:

```
tar: Ignoring unknown extended header keyword 'LIBARCHIVE.creationtime'
```

Can the extended header keywords be removed as it generates warnings on multiple users machines even with recent gnu tar implementations. AFAIU this is a bsd tar extension.

xref: https://github.com/eclipse-platform/eclipse.platform.releng/issues/147

laeubi:
Tycho uses `org.apache.commons.compress.archivers.tar.TarArchiveOutputStream` so if you think anything is incalid there in general it might be best reported there, as far as I can see it only uses POSIX extensions here:

https://github.com/eclipse-tycho/tycho/blob/master/tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/tar/TarGzArchiver.java

so I can't see any immediate actions Tycho can perform here.

jonahgraham:
This is the change in Commons compress that added the header https://github.com/apache/commons-compress/pull/254

jonahgraham:
The issue report is here https://issues.apache.org/jira/projects/COMPRESS/issues/COMPRESS-612?filter=allissues

jonahgraham:
I'm trying to decide whether to invest more time on this.

Is no one else seeing these warnings when decompressing? I don't want to spend time answering support requests on this issue now that EPP will be distributing this.

laeubi:
> Is no one else seeing these warnings when decompressing? 

I usually not use the commandline to uncompress archives.

> I'm trying to decide whether to invest more time on this.

This is fair enough, but I get the feeling that the most valuable way would be to report it to the `tar` tool that:

1. The same warning should not be reported more than once
2. extended header 'LIBARCHIVE.creationtime' should either be supported or silently ignored (so it is not reported as unknown)

From Tycho side I don't see anything we can do as it seems it is a valid archive and the behavior itself is not configurable with the library we use here.

jonahgraham:
I'm going to refer back to this issue if anyone raises it, but I am not spending more time on this.

HannesWell:
[...]

PR #4427, opened 2024-11-10
Don't write LIBARCHIVE.creationtime attribute in tar.gz archive entries

HannesWell:
Fixes https://github.com/eclipse-tycho/tycho/issues/2586

### DIFF
--- a/tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/p2/director/ProductArchiverMojo.java
+++ b/tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/p2/director/ProductArchiverMojo.java
@@ -125,6 +125,15 @@ public final class ProductArchiverMojo extends AbstractProductMojo {
     @Parameter
     private boolean parallel;
 
+    /**
+     * Controls if for {@code .tar.gz} archives the creation-time is stored as
+     * {@code LIBARCHIVE.creationtime } attribute in each entry. Currently {@code GNU tar} does not
+     * support that attributes and emits warnings about the {@code unknown extended header keyword
+     * 'LIBARCHIVE.creationtime'} when extracting such archive.
+     */
+    @Parameter(defaultValue = "false")
+    private boolean storeCreationTime;
+
     @Component
     private MavenProjectHelper helper;
 
@@ -230,6 +239,7 @@ public final class ProductArchiverMojo extends AbstractProductMojo {
 
     private void createCommonsCompressTarGz(File productArchive, File sourceDir) throws IOException {
         TarGzArchiver archiver = new TarGzArchiver();
+        archiver.setStoreCreationTimeAttribute(storeCreationTime);
         archiver.setLog(getLog());
         archiver.addDirectory(sourceDir);
         archiver.setDestFile(productArchive);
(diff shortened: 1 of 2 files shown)


tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/tar/TarGzArchiver.java
```diff
@@ -129,6 +129,7 @@ private TarArchiveEntry createTarEntry(File tarRootDir, File source) throws IOEx
             tarEntry.setMode(FilePermissionHelper.toOctalFileMode(attrs.permissions()));
         }
         tarEntry.setModTime(source.lastModified());
+        tarEntry.setCreationTime(null); // tar cannot handle 'LIBARCHIVE.creationtime' attributes
```
laeubi, line 132: This is wrong, e.g. debian supports it: https://manpages.debian.org/unstable/libarchive-dev/tar.5.en.html

laeubi (changes requested):
There should be a config option to switch this behavior on/off

- for Tycho 4.x it must be on by default
- for Tycho 5.x we can decide what should be the default

HannesWell:
> There should be a config option to switch this behavior on/off
> 
>     * for Tycho 4.x it must be on by default
> 
>     * for Tycho 5.x we can decide what should be the default

If that's necessary get consensus I can add an option, but AFAIK an option was not added when this attribute was introduced, was it?

laeubi:
> when this attribute was introduced

This attribute was never (intentionally) introduced, we updated the library behind that and those maintainer choose for some reasons to support this "non standard" extension. So from a compatibility point of view it only adds something new that can be ignored by a tool if desired.

No we have the other way round, that *removing* it might break use-cases that have relied on this, so I better like to preserve this. It would have been different if we would have never added it before but now it is there (even not added by intention).

HannesWell:
> No we have the other way round, that _removing_ it might break use-cases that have relied on this, so I better like to preserve this. It would have been different if we would have never added it before but now it is there (even not added by intention) but disabling that by a config option is always possible of course.

Ok, then lets do it this way.
But I suggest to, for convenience to disable it by default for tycho-5 and keep the property by default for Tycho-4.
If one day tar can handle or ignore that attribute and the version that has the fix is widely adopted we can re-add it again.

tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/tar/TarGzArchiver.java
HannesWell: Understand. I refined it to `GNU tar`. I cannot provide the other details exhaustively.

HannesWell:
@laeubi are you fine with this now?

HannesWell:
Created the back-port with different default in
- https://github.com/eclipse-tycho/tycho/pull/4428

tycho-p2-director-plugin/src/main/java/org/eclipse/tycho/plugins/p2/director/ProductArchiverMojo.java
```diff
@@ -125,6 +125,15 @@ public final class ProductArchiverMojo extends AbstractProductMojo {
...
     private boolean parallel;
 
+    /**
+     * Controls if for {@code .tar.gz} archives the creation-time is stored as
+     * {@code LIBARCHIVE.creationtime } attribute in each entry. Currently {@code GNU tar} does not
+     * support that attributes and emits warnings about the {@code unknown extended header keyword
+     * 'LIBARCHIVE.creationtime'} when extracting such archive.
+     */
+    @Parameter(defaultValue = "false")
+    private boolean storeTarGZCreationTime;
```
laeubi, line 135:
```suggestion
    private boolean storeCreationTime;
```

should be enough.

laeubi (approved):
Looks good!

PR #4438, opened 2024-11-18
Replacing outdated wiki links with links to Github repos for

vogella:
PackagingTypes

The wiki is deprecated and we should link to the new content


laeubi:
By the way the best would be to add the information directly to the sitedocs so we don't need any external reference.

Issue #4451, opened 2024-11-26
Tycho 4.0.10: Stackoverflow when fetching from p2 repository that responds with 301


marc-christian-schulze:
Today the [issue 2358](https://github.com/eclipse-tycho/tycho/issues/2358) re-occurred with Tycho 4.0.10 when the p2 server respond with a "301 Moved Permanently".

This is the log out from tycho:
```
[INFO] Downloaded from p2: https://download.eclipse.org/releases/2024-12/202411221000/plugins/org.eclipse.core.filesystem_1.11.100.v20241022-0806.jar (77 KB at 1 MB/s)
[INFO] Downloading from p2: https://0ms.dev/mirrors/eclipse//releases/2024-12/202411221000/plugins/org.eclipse.core.filesystem.source_1.11.100.v20241022-0806.jar
[INFO] Downloaded from p2: https://0ms.dev/mirrors/eclipse//releases/2024-12/202411221000/plugins/org.eclipse.core.filesystem.source_1.11.100.v20241022-0806.jar (87 KB at 375 KB/s)
[INFO] Downloading from p2: https://ftp2.osuosl.org/pub/eclipse/releases/2024-12/202411221000/plugins/org.eclipse.core.jobs_3.15.400.v20240619-0602.jar
[INFO] ------------------------------------------------------------------------

... omitted ...

---------------------------------------------------
constituent[0]: file:/opt/hostedtoolcache/maven/3.9.6/x64/conf/logging/
constituent[1]: file:/opt/hostedtoolcache/maven/3.9.6/x64/lib/plexus-utils-3.5.1.jar
constituent[2]: file:/opt/hostedtoolcache/maven/3.9.6/x64/lib/wagon-http-3.5.3.jar
constituent[3]: file:/opt/hostedtoolcache/maven/3.9.6/x64/lib/maven-shared-utils-3.3.4.jar
constituent[4]: file:/opt/hostedtoolcache/maven/3.9.6/x64/lib/maven-settings-3.9.6.jar
constituent[5]: file:/opt/hostedtoolcache/maven/3.9.6/x64/lib/maven-builder-support-3.9.6.jar
constituent[6]: file:/opt/hostedtoolcache/maven/3.9.6/x64/lib/org.eclipse.sisu.inject-0.9.0.M2.jar
constituent[7]: file:/opt/hostedtoolcache/maven/3.9.6/x64/lib/aopalliance-1.0.jar
[...]

PR #4452, opened 2024-11-26
Don't await a StackOverflowError on a redirection loop

thahnen:
In case of a redirection loop for URI of a p2 repository, fail fast on a redirection loop. In this case provide a warning to the user and let the rest be handled by Tycho. Otherwise there might be a StackOverflowError due to the recursion used and no check for whether the base URI is the same as the redirected URI.

This links to #4451 but does not fix it, only raising visibility for such cases.

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/HttpCache.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/HttpCache.java
@@ -24,9 +24,9 @@ public interface HttpCache {
 	 * 
 	 * @param uri
 	 * @return
-	 * @throws FileNotFoundException
-	 *             if the URI is know to be not found
+	 * @throws FileNotFoundException    if the URI is know to be not found
+	 * @throws RedirectionLoopException if the URI redirects to itself
 	 */
-	CacheEntry getCacheEntry(URI uri, Logger logger) throws FileNotFoundException;
+	CacheEntry getCacheEntry(URI uri, Logger logger) throws FileNotFoundException, RedirectionLoopException;
 
 }
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RedirectionLoopException.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/RedirectionLoopException.java
@@ -1,0 +1,25 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Tobias Hahnen and others.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *    Tobias Hahnen - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.tycho.p2maven.transport;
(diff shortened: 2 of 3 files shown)


laeubi:
Any chance you can enhance the existing integration test to show the behavior?

thahnen:
@laeubi which one is the corresponding one?

I was thinking about creating a new one with a web server redirecting to itself to mimic this. But I haven't finished yet reading the documentation of Jetty ^^

laeubi:
If you search for [HttpServer](https://github.com/eclipse-tycho/tycho/blob/main/tycho-its/src/test/java/org/eclipse/tycho/test/util/HttpServer.java) references in the test you should find some test that already do something in that direction.

thahnen:
Hi @laeubi, sorry for the late response. I will take some time later today to create an IT or to update a fitting existing one.

thahnen:
Sadly, for whatever reason I wasn't able to run the ITs from within Eclipse itself due to the launch configuration `tycho-its - prepare test resources` failing with an exception from both the launcher as well as when running the test from the command line `mvn clean verify -f tycho-its/pom.xml -Dtest=P2RedirectLoopTest`: 
[prepare_test_resources.log](https://github.com/user-attachments/files/17936981/prepare_test_resources.log)

laeubi:
In eclipse you can simpel right click and choose to run as unit test. But due to the nature of tycho/maven it is usually better to use maven CLI with remote debugging. 

For this simply do

- `cd tycho-its`
- `mvn clean install -Pits -Dtest=<full class name of test>`

thahnen:
Hey @laeubi,

I have been trying for a few days now to create an IT for this behavior but with no success. Taking a look at all P2-related ITs, I could not really find a way to hook into one, so I created one on my own.

But even having a Jetty where the `p2.index` was reachable and linked to a `compositeContent.xml` and `compositeArtifacts.xml` on another server that would create the redirection loop did not work.
My initial idea of just having a target platform linking to a server that would redirect to itself for every file would not work because it failed earlier by not finding a p2 metadata repository :disappointed:

laeubi:
@thahnen can you maybe share what "did not work"? Maybe just upload the test and explain what you wanted to behave differently and we can probably find a solution.

thahnen:
Good idea! Currently OOF, will do so once I‘m back again.

laeubi:
@thahnen there will be anew Tycho release probably end of Feb do you like to complete this so we can include the fix?

thahnen:
Hey @laeubi, I sadly found no time lately. I will put this back on my TODO list to complete this. Thanks for the heads up!

laeubi:
@thahnen as a gentle reminder we prepare for a [final Tycho 4 release](https://github.com/eclipse-tycho/tycho/discussions/2455#discussioncomment-12584080) so if you want to see this feature there it would be good to finish this soon. Otherwise it could go into of course in the 5.x release without any problem.

thahnen:
Hey @laeubi, thanks for the reminder, I did not find the time to pick it up yet. But for the general implementation without reproducer tests I could do so today - I struggled the last time again heavely with trying to create a reproducer so I'm most certain that I will not do that here and stick with the implementation I currently have. Will resolve the merge conflicts tho.

akurtakov:
What is the plan for this one? If there is no interest in finishing this one I plan to close it.

thahnen:
Did not find the time to finish it, plus a shift in work-related priorities. I also have not encountered that issue since then and have not yet worked with Tycho 5 and found no time to dig into it. Closing it for now; will keep the patch around for later.

PR #4461, opened 2024-11-30
reduce file access

Bananeweizen:
At least on Windows the repeated invocation of file.exists takes a huge amount of time. The change removes several _minutes_ of runtime from a 1 hour product build: 
image

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/DefaultArtifactDescriptor.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/DefaultArtifactDescriptor.java
@@ -68,10 +68,13 @@ public class DefaultArtifactDescriptor implements ArtifactDescriptor {
         if (projectLocation != null) {
             return projectLocation;
         }
-        if (fetch && locationSupplier != null && (location == null || !location.exists())) {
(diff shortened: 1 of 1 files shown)


tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/DefaultArtifactDescriptor.java
```diff
@@ -68,10 +68,13 @@ public File getLocation(boolean fetch) {
         if (projectLocation != null) {
             return projectLocation;
         }
-        if (fetch && locationSupplier != null && (location == null || !location.exists())) {
+        if (fetch && locationSupplier != null && location == null) {
```
laeubi, line 71: The check was there to make sure the file was not deleted in the meantime (e.g. outside of control of the descriptor), this won't be detected anymore.

laeubi: > Also, if this is related to the "file is corrupt, trigger download again" functionality, then I think we should do something there were the download is re-triggered

The problem is that it is not that easy. We know that the file is corrupt (and can delete it) but there is no easy way to know here it was downloaded from.

laeubi:
lets give it a try

PR #4529, opened 2024-12-16
Add an option to print the java memory before/after mojo execution

laeubi:
To help identify possible leaks


merks:
Given that it appears the OS (Jenkins) is killing processes, might it be helpful to have some way to force garbage collection?

laeubi:
I fear for killing the agent a GC does not help because the JVM will not really "free" the memory on OS level, but this is mainly to understand if/where memory is used if we see OOM Exceptions in builds.

merks:
I see.

PR #4532, opened 2024-12-16
[bp] Add support for TargetDefinition#implicitDependencies

laeubi:
PDE has support for implicitDependencies defined in the target that Tycho currently ignores.

This adds the required support to Tycho to understand dependencies defined in this way.


laeubi:
This should probably wait unless
- https://github.com/eclipse-tycho/tycho/issues/4533

is resolved even though I don't expect any problems with the platform build with this change.

PR #4544, opened 2024-12-18
Rework site documentation

Bananeweizen:
* fix wording and typos
* enable syntax highlighting
* replace tabs in code blocks for unified layout
* upgrade skin (dependabot doesn't work for site.xml)

Be aware numbered lists don't work well here, if they have code blocks in between, because the index resets to 1 after each block. Therefore numbered lists have been replaced by bullet lists several times. An alternative might be converting to Asciidoc, which has far greater control over such things.

### DIFF
--- a/src/site/markdown/BndBuild.md
+++ b/src/site/markdown/BndBuild.md
@@ -1,11 +1,11 @@
 # BND Workspace Layout and Pomless Builds
 
-Tycho supports building projects that use the **BND Workspace Layout** as [described here](https://bndtools.org/concepts.html).
+Tycho supports building projects that use the [BND Workspace Layout](https://bndtools.org/concepts.html).
 
 ## BND Workspace Layout
 A BND Workspace layout build layout usually has the following structure:
 
-- `root folder` - this usually is the root of your project repository (e.g. git)
+- `root folder` - this usually is the root of your project repository
     - `cnf` - configuration folder for general setup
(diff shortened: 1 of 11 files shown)


src/site/markdown/BndBuild.md
```diff
@@ -15,37 +15,34 @@ A BND Workspace layout build layout usually has the following structure:
...
+      <version>${tycho-version}</version>
+   </extension>
 </extensions>
 ```
 
-3. create a file called `maven.config` in the `.mvn` folder with the following content (adjust the version accordingly!):
-
+- create a file called `maven.config` in the `.mvn` folder with the following content (adjust the version accordingly!):
+```properties
+-Dtycho-version=4.10.0
```
mlehmannm, line 39: 4.10.0?

laeubi:
@Bananeweizen can you share a screenshot of the new layout/skin?

Bananeweizen:
@laeubi The skin itself is the exact same. This shows left=PR, right=online Tycho documentation. I've marked some areas with notable differences.
grafik

laeubi:
@Bananeweizen Should this backported to Tycho 4?

Bananeweizen:
@laeubi might be useful to be backported so it goes live on the website. I guess a real 5.0 release is still far away, right? but I'm also okay with not backporting, if that creates additional effort.

Issue #4459, opened 2024-11-29
Mirror mojo fails due cache files being too long


rbioteau:
When mirroring a target platform containing the following repository

```
<plugin>
            <groupId>org.eclipse.tycho.extras</groupId>
            <artifactId>tycho-p2-extras-plugin</artifactId>
            <version>4.0.10</version>
            <executions>
              <execution>
                <phase>prepare-package</phase>
                <goals>
                  <goal>mirror</goal>
                </goals>
              </execution>
            </executions>
            <configuration>
              <source>
                <repository>
                  <url>https://groovy.jfrog.io/artifactory/plugins-release/e4.33</url>
                  <layout>p2</layout>
                </repository>
              </source>
              <ius>
                <iu>
                  <id>org.codehaus.groovy.compilerless.feature.feature.group</id>
                  <version>5.5.0.v202409302028-e2409</version>
                </iu>
                <iu>
                  <id>org.codehaus.groovy.m2eclipse.feature.feature.group</id>
                  <version>5.5.0.v202409302028-e2409</version>
                </iu>
                <iu>
                  <id>org.codehaus.groovy30.feature.feature.group</id>
                  <version>5.5.0.v202409302028-e2409</version>
                </iu>
                <iu>
                  <id>org.codehaus.groovy.eclipse.feature.feature.group</id>
                  <version>5.5.0.v202409302028-e2409</version>
                </iu>
              </ius>
            </configuration>
          </plugin>
```

The mojo fails with the following error:

```
[...]

PR #4547, opened 2024-12-20
Cache temporary redirections in the original location (#4459)

kysmith-csg:
Some p2 sites can temporarily redirect to another location. In the case of JFrog Artifactory, these redirection URLs are too long to make as a file path, so the build would fail in this case. For temporary redirects, we save the artifact in the original location instead, since it's likely the redirection might change in the future. Permanent redirects will get their own cache entry as before.

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/SharedHttpCacheStorage.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/SharedHttpCacheStorage.java
@@ -223,13 +223,13 @@ public class SharedHttpCacheStorage implements HttpCache {
 
 		public synchronized File fetchFile(URI uri, HttpTransportFactory transportFactory, Logger logger)
 				throws IOException {
-			boolean exits = file.isFile();
-			if (exits && !mustValidate()) {
+			boolean exists = file.isFile();
+			if (exists && !mustValidate()) {
 				return file;
 			}
 			HttpTransport transport = transportFactory.createTransport(uri);
 			Properties lastHeader = getHeader();
-			if (exits) {
+			if (exists) {
 				if (lastHeader.containsKey(Headers.ETAG_HEADER.toLowerCase())) {
 					transport.setHeader("If-None-Match", lastHeader.getProperty(Headers.ETAG_HEADER.toLowerCase()));
 				}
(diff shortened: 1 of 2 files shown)


laeubi:
This usually means something is messed up (e.g. redirection do not work anymore, mirrors not working runtime exception occurs and so on).

The best way then is to run 

- go into `tycho-its` directory
- run `mvn clean install -Pits -Dtycho.mvnDebug=8000 -Dtest=<fqdn of test>#<failingTest>`

this way you can debug one individual test.

Sometimes it also helps to change the test to dump the output to a file, then compare the output of `before` versus `after` to see if some log outputs change. One can also enable the debugging for the transport, just set [some of these setting](https://tycho.eclipseprojects.io/doc/main/SystemProperties.html) in the corresponding test verifier as system properties.

kysmith-csg:
@laeubi this test was actually for an invalid repository mirror. The test is verifying the error message, which after my changes I get a different one (but still the same reason - the repository is invalid). So I would say the test still "passes" but we just now get a different error message.

laeubi:
@kysmith-csg could you adjust the test then for the new message?

kysmith-csg:
@laeubi  this test failure doesn't seem related to my change. Even your latest commit failed because of this :)

kysmith-csg:
Thank you @laeubi  for handling this. Any chance of making a 4.0.11 release for this? It's currently impacting our organization and at least one other (the original author of the issue). We would greatly appreciate it.

laeubi:
@kysmith-csg thanks for the fix and patience! In general it is always useful to use 
try out the current [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) build to make sure everything would work before a release.

I think we have some important bugfixes already included so I will perform a release maybe next month but wanted to see some things settled, will this work for you?

kysmith-csg:
@laeubi I will try the snapshot build at some point and report back, thanks.

kysmith-csg:
@laeubi tested 4.0.11-SNAPSHOT (which I had to build locally because for some reason it wasn't using the snapshot repo) and confirmed it works.

A jar file download (https://my-site.com/files/foo.jar) that was redirected to a long url (https://my-very-very-long-site.com/qwertyuiop/files/foo.jar) site was cached in the directory for the original location from target-platform (`.cache/tycho/https/my-site.com/files/plugins/foo.jar`).

Confirmed by checking the `foo.jar.headers` file and the redirected location was there `location=https://my-very-very-long-site.com/qwertyuiop/files/foo.jar`.

PR #4548, opened 2024-12-20
Stop using simrel 2021-09 repo in tests

akurtakov:
It's archived and frequent usage leads to broken builds with: "org.eclipse.equinox.p2.core.ProvisionException: No repository found at https://download.eclipse.org/technology/epp/packages/2021-09/..." as archives are not mirrored and this leads to heavy load (and DoS) of archive.eclipse.org.


akurtakov:
Failure can be seen at https://github.com/eclipse-tycho/tycho/runs/34717856597 . It doesn't always fail as archive.eclipse.org is not overloaded constantly. This change should make it fine for some time but overall a better approach is to have specially crafted projects for such purposes that would:
* reduce network traffic and speed up tests
* reduce need for "updating" tests
* make tests more "obvious" - it took me some time to realize what exactly is being tested.

PR #4553, opened 2024-12-22
Only apply a better profile if allowed

laeubi:
Currently a configured profile is unconditionally replaced by a better one of the current JVM, but this is wrong if not explicitly allowed by the configuration.

This is now changed, to be only used if ignoring the BREE is actually enabled.

### DIFF
--- a/demo/bnd-pde-workspace/.mvn/maven.config
+++ b/demo/bnd-pde-workspace/.mvn/maven.config
@@ -1,3 +1,3 @@
--Dtycho-version=4.0.0-SNAPSHOT
--Dtarget-platform=https://download.eclipse.org/releases/2022-12/
+-Dtycho-version=5.0.0-SNAPSHOT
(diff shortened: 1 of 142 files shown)


laeubi:
Marking for Tycho 4 only as it is probably a breaking change, so lets see what all fails now.

laeubi:
I think the message is quite expected here:

> [ERROR] org.eclipse.core.jobs 3.15.400.v20240619-0602 requires Execution Environment that matches (&(osgi.ee=JavaSE)(version=17)) but the current resolution context uses [a.jre.javase 1.8.0]

so we probably should update the tests here and then apply the change even to Tycho 4

laeubi:
Update of the BREEs extracted here:
- https://github.com/eclipse-tycho/tycho/pull/4564

laeubi:
This should be manually be packported to tycho 4.x as we need to add a special note to the release notes about this possibly braking behavior compared to previous version.

PR #4564, opened 2024-12-27
Upgrade integration test bundles to BREE 17

laeubi:
Currently we have a lot of old integration tests that require different BREEs due to historical reasons even though we already depend on much newer dependencies.

This update the BREE of all test that do not explicitly test a specific feature related to the BREE to Java 17 to reflect the things we actually use.


laeubi:
This should result in the adjusted testcase to fail and show the bug described here:
- https://github.com/eclipse-tycho/tycho/pull/4549

If this is the case, we should merge this so it can be fixed in a follow-up PR.

PR #4546, opened 2024-12-20
Add reproducible timestamps in archives

Zlika:
Use the standard "project.build.outputTimestamp" maven property to enable reproducible timestamps in archives (cf. https://maven.apache.org/guides/mini/guide-reproducible-builds.html).

PR #4569, opened 2024-12-28
Use reproducible version qualifiers

Zlika:
Use reproducible version qualifiers when project.build.outputTimestamp property is set.
This is a follow-up of #4546.


laeubi:
This looks useful I just wondering if we probably should handle this in the default timestamp provider instead?

We would then just have the parameter defined as today but read its value here

https://github.com/eclipse-tycho/tycho/blob/af851639d020639fdea5625cdd3fcfb914ebdbc6/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/DefaultBuildTimestampProvider.java#L34

from the mojo execution configuration.

Zlika:
> This looks useful I just wondering if we probably should handle this in the default timestamp provider instead?
> 
> We would then just have the parameter defined as today but read its value here
> 
> https://github.com/eclipse-tycho/tycho/blob/af851639d020639fdea5625cdd3fcfb914ebdbc6/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/DefaultBuildTimestampProvider.java#L33-L35
> 
> from the mojo execution configuration.

I think you're right. I've refactored the PR to only change the DefaultBuildTimestampProvider. What do you think of this?

laeubi (approved):
Testfailure is unrelated

tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/BuildQualifierMojo.java
```diff
@@ -266,8 +266,8 @@ private String getUnqualifiedVersion() {
 	}
 
     protected Date getBuildTimestamp() throws MojoExecutionException {
-        String hint = timestampProvider != null ? timestampProvider : DefaultBuildTimestampProvider.ROLE_HINT;
-        BuildTimestampProvider provider = timestampProviders.get(hint);
+		String hint = (timestampProvider != null) ? timestampProvider : DefaultBuildTimestampProvider.ROLE_HINT;
```
laeubi, line 269: Maybe revert this formatting changes

laeubi:
I don't know why but Github indicates you have renamed your previous tests instead of adding a new test. Is this intentional?

tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/BuildQualifierMojo.java
Zlika: Done.

Zlika:
> I don't know why but Github indicates you have renamed your previous tests instead of adding a new test. Is this intentional?

I renamed the integration test class from ReproducibleArchiveTimestampTest to ReproducibleBuildTest because I will test here all the reproducibility problems that need to be fixed (other PR will be required for that). Here I added a check to assert that the build version qualifier is the one expected.

laeubi:
@Zlika this looks fine now, can you probably squash all commits into one?

Zlika:
> @Zlika this looks fine now, can you probably squash all commits into one?

Can't you squash the commits directly here when you merge the PR?

laeubi:
I can do it but this often leads to "strange" commits authored by github bot, so if you don't mind, just quash, rebase and force push and I think it can be merged (and backported) right away.

Zlika:
> I can do it but this often leads to "strange" commits authored by github bot, so if you don't mind, just quash, rebase and force push and I think it can be merged (and backported) right away.

Ok, done.Thanks.

laeubi (approved):
Thanks again!

Zlika:
Thanks @laeubi for the merge.
I want to propose another PR where I need to create a utility method that will be called from several plugins. I cannot find a good location for this utility method. I tried to put it in tycho-core, but this module is not used by all the plugins. Is there a good location somewhere in the code base for common code that can be used in all the plugins?

laeubi:
We have [tycho-api](https://github.com/eclipse-tycho/tycho/tree/main/tycho-api) module or [tycho-spi](https://github.com/eclipse-tycho/tycho/tree/main/tycho-spi), if there is no dependency to any of those it can usually easily be added.

PR #4572, opened 2024-12-29
Ceate ArtifactDownloadProvider that use maven repositories as mirror

laeubi:
In the meanwhile many artifacts are already published and consumed from maven central and we record the necessary metadata to locate them in P2.

This adds a new ArtifactDownloadProvider that is capable to use global configured maven repositories as a mirror for P2 artifacts.

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/MavenArtifactDownloadProvider.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/MavenArtifactDownloadProvider.java
@@ -1,0 +1,266 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Christoph Läubrich and others.
+ * This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/MavenArtifactDownloadProvider.java
```diff
@@ -0,0 +1,266 @@
...
+						break;
+					}
+					// checksum mismatch no need to further bother maven for this file
+					return Status.CANCEL_STATUS;
+				}
+			} catch (Exception e) {
+				// can't check this way...
+			}
+		}
+		// now we have some good certainty this P2 artifact is actually soured from
```
merks, line 167:
```suggestion
		// now we have some good certainty this P2 artifact is actually sourced from
```

merks:
Note the typo in the commit message.  Ceate -> Create

> **Creates** ArtifactDownloadProvider that **uses** maven repositories as mirror

PR #4573, opened 2024-12-29
[bp] Upgrade integration test bundles to BREE 17

laeubi:
Currently we have a lot of old integration tests that require different BREEs due to historical reasons even though we already depend on much newer dependencies.

This update the BREE of all test that do not explicitly test a specific feature related to the BREE to Java 17 to reflect the things we actually use.


laeubi:
Tycho show the same failure with the new test, so the same fix should be applied here.

PR #4582, opened 2025-01-01
Enable reproducible build of the tycho plugins

Zlika:
Cf. https://maven.apache.org/guides/mini/guide-reproducible-builds.html


akurtakov:
I don't envision harm from this one considering that Maven 4 will even enable it by default. Merging.

PR #4583, opened 2025-01-02
Write Properties files in a reproducible way

Zlika:
Java does not write java.util.Properties files in a reproducible way: there is a timestamp, the order of the lines is not reproducible, and the new line character used is system-dependent.
This PR fixes this problem by removing the timestamp, ordering the lines in the alphabetical order and using a system-independent new line character.
This is a follow-up of #4546 and #4569.

### DIFF
--- a/tycho-api/src/main/java/org/eclipse/tycho/ReproducibleUtils.java
+++ b/tycho-api/src/main/java/org/eclipse/tycho/ReproducibleUtils.java
@@ -1,0 +1,54 @@
+/*******************************************************************************
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.tycho;
+
+import java.io.BufferedOutputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Properties;
+import java.util.stream.Collectors;
+
+/**
+ * Utility methods for reproducible builds.
+ */
+public class ReproducibleUtils {
+    private ReproducibleUtils() {
+    }
+
+    /**
+     * Writes the property list to the output stream in a reproducible way. The java.util.Properties
+     * class writes the lines in a non-reproducible order, adds a non-reproducible timestamp and
+     * uses platform-dependent new line characters.
+     * 
+     * @param properties
+     *            the properties object to write to the file.
+     * @param file
+     *            the file to write to. All the missing parent directories are also created.
+     * @throws IOException
+     *             if writing the property list to the specified output stream throws an
+     *             IOException.
+     */
+    public static void storeProperties(Properties properties, Path file) throws IOException {
+        final Path folder = file.getParent();
+        if (folder != null) {
+            Files.createDirectories(folder);
+        }
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        properties.store(baos, null);
+        final String content = baos.toString(StandardCharsets.ISO_8859_1).lines().filter(line -> !line.startsWith("#"))
+                .sorted().collect(Collectors.joining("\n", "", "\n"));
+        try (OutputStream os = new BufferedOutputStream(Files.newOutputStream(file))) {
+            os.write(content.getBytes(StandardCharsets.ISO_8859_1));
+        }
+    }
+}
--- a/tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/MavenP2SiteMojo.java
+++ b/tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/MavenP2SiteMojo.java
@@ -72,6 +72,7 @@ import org.eclipse.equinox.app.IApplication;
 import org.eclipse.equinox.p2.core.IProvisioningAgent;
 import org.eclipse.equinox.p2.repository.artifact.IArtifactRepositoryManager;
 import org.eclipse.tycho.PackagingType;
+import org.eclipse.tycho.ReproducibleUtils;
 import org.eclipse.tycho.TychoConstants;
 import org.eclipse.tycho.core.PGPService;
 import org.eclipse.tycho.p2maven.tools.TychoFeaturesAndBundlesPublisherApplication;
@@ -484,9 +485,7 @@ public class MavenP2SiteMojo extends AbstractMojo {
             addProvidesAndProperty(properties, TychoConstants.PROP_EXTENSION, artifact.getType(), cnt++);
             addProvidesAndProperty(properties, TychoConstants.PROP_CLASSIFIER, artifact.getClassifier(), cnt++);
             addProvidesAndProperty(properties, "maven-scope", artifact.getScope(), cnt++);
-            try (OutputStream os = new BufferedOutputStream(new FileOutputStream(p2))) {
-                properties.store(os, null);
-            }
+            ReproducibleUtils.storeProperties(properties, p2.toPath());
             return p2;
         } catch (IOException e) {
             throw new MojoExecutionException("failed to generate p2.inf", e);
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/SharedHttpCacheStorage.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/SharedHttpCacheStorage.java
@@ -40,6 +40,7 @@ import org.codehaus.plexus.component.annotations.Component;
 import org.codehaus.plexus.component.annotations.Requirement;
 import org.codehaus.plexus.logging.Logger;
 import org.eclipse.equinox.internal.p2.repository.AuthenticationFailedException;
+import org.eclipse.tycho.ReproducibleUtils;
 
 @Component(role = HttpCache.class)
 public class SharedHttpCacheStorage implements HttpCache {
@@ -391,12 +392,9 @@ public class SharedHttpCacheStorage implements HttpCache {
 					header.put(key, value.stream().collect(Collectors.joining(",")));
 				}
 			}
-			FileUtils.forceMkdir(file.getParentFile());
-			try (OutputStream out = new BufferedOutputStream(new FileOutputStream(headerFile))) {
-				// we store the header here, this might be a 404 response or (permanent)
-				// redirect we probably need to work with later on
-				header.store(out, null);
-			}
+			// we store the header here, this might be a 404 response or (permanent)
+			// redirect we probably need to work with later on
+			ReproducibleUtils.storeProperties(header, headerFile.toPath());
 		}
 
 		private synchronized Date pareHttpDate(String input) {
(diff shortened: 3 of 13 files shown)


tycho-its/projects/resolver.split/org.eclipse.equinox.security/src/org/eclipse/equinox/internal/security/storage/SecurePreferencesRoot.java
```diff
@@ -196,7 +197,7 @@ else if (!response.booleanValue())
 		OutputStream stream = null;
 		try {
 			stream = StorageUtils.getOutputStream(location);
-			properties.store(stream, description);
+			ReproduciblePropertiesUtils.store(properties, stream, description);
```
laeubi, line 200: I don't think this makes much sense in a test-case....

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/MavenP2SiteMojo.java
```diff
@@ -485,7 +486,7 @@ protected File createMavenAdvice(Artifact artifact) throws MojoExecutionExceptio
             addProvidesAndProperty(properties, TychoConstants.PROP_CLASSIFIER, artifact.getClassifier(), cnt++);
             addProvidesAndProperty(properties, "maven-scope", artifact.getScope(), cnt++);
             try (OutputStream os = new BufferedOutputStream(new FileOutputStream(p2))) {
-                properties.store(os, null);
+                ReproducibleUtils.storeProperties(properties, os, null);
```
laeubi, line 489: As it seems always the same boiler-plate, maybe we can have a utils method that accepts a file ans creates the stream and alike 

`ReproducibleUtils.storeProperties(properties, p2);`

tycho-api/src/main/java/org/eclipse/tycho/ReproducibleUtils.java
```diff
@@ -0,0 +1,54 @@
...
+     *            the properties object to write to the output stream.
+     * @param out
+     *            an output stream.
+     * @param comments
+     *            a description of the property list.
+     * @throws IOException
+     *             if writing the property list to the specified output stream throws an
+     *             IOException.
+     */
+    public static void storeProperties(Properties properties, OutputStream out, String comments) throws IOException {
```
laeubi, line 41: 1. Do we really use the comments anywhere? If not just drop it.
2. Is there maybe any maven utility or library we can reuse instead of came up with something our self?

laeubi:
> and the new line character used is system-dependent.

Last time I followed a discussion with the maven folks, they said that newlines are always a problem for reproducible build and one should rebuild on the same system architecture, so maybe not a big deal but probably also don't harm much.

tycho-its/projects/resolver.split/org.eclipse.equinox.security/src/org/eclipse/equinox/internal/security/storage/SecurePreferencesRoot.java
Zlika: Sure, but it's easier to apply this fix everywhere to be sure.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/MavenP2SiteMojo.java
Zlika: Yes, why not.

tycho-api/src/main/java/org/eclipse/tycho/ReproducibleUtils.java
Zlika: The comments seem to be used only once, in OsgiSourceMojo. When it creates the pom.properties file, maven-archiver simply drops all the comments (cf. https://github.com/apache/maven-archiver/blob/master/src/main/java/org/apache/maven/shared/archiver/PomPropertiesUtil.java). If it's ok for you I can also drop all the comments, it could greatly simplify the code. I don't know any maven utility to do that, maven-archiver uses a private method for that so I cannot reuse it.

Zlika:
> > and the new line character used is system-dependent.
> 
> Last time I followed a discussion with the maven folks, they said that newlines are always a problem for reproducible build and one should rebuild on the same system architecture, so maybe not a big deal but probably also don't harm much.

That's true, but when maven-archiver fixes the pom.properties file (in https://github.com/apache/maven-archiver/blob/master/src/main/java/org/apache/maven/shared/archiver/PomPropertiesUtil.java) it also uses a system-independent newline character. Anyway, trying to reuse the existing newline characters would make the code more complex, so it's simpler to use a system-independent newline character here. I would be dangerous to try to blindly fix all the text files to use a system-independent newline character, but when we need to rewrite a file it makes sens to take the opportunity to do it.

tycho-its/projects/resolver.split/org.eclipse.equinox.security/src/org/eclipse/equinox/internal/security/storage/SecurePreferencesRoot.java
laeubi: But this is a testcase that has nothing to do with tycho (and fails to compile because it has no dependency to tycho codebase at all)

tycho-api/src/main/java/org/eclipse/tycho/ReproducibleUtils.java
laeubi: I think this comment is not very useful and can be omitted, so I would suggest to go ahead and just drop any comment lines.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/MavenP2SiteMojo.java
laeubi: Too bad that this is private: https://github.com/apache/maven-archiver/blob/ffca18fc302c9f19cf2e5bfe7aad4a89a5d83fc6/src/main/java/org/apache/maven/shared/archiver/PomPropertiesUtil.java#L51

Zlika:
Hi @laeubi, I took into account all your remarks. Thanks.

tycho-api/src/main/java/org/eclipse/tycho/ReproducibleUtils.java
```diff
@@ -0,0 +1,50 @@
...
+     * @throws IOException
+     *             if writing the property list to the specified output stream throws an
+     *             IOException.
+     */
+    public static void storeProperties(Properties properties, File file) throws IOException {
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        properties.store(baos, null);
+        final String content = baos.toString(StandardCharsets.ISO_8859_1).lines().filter(line -> !line.startsWith("#"))
+                .sorted().collect(Collectors.joining("\n", "", "\n"));
+        try (OutputStream os = new BufferedOutputStream(new FileOutputStream(file))) {
```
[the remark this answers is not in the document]
Zlika: Done.

tycho-api/src/main/java/org/eclipse/tycho/ReproducibleUtils.java
```diff
@@ -0,0 +1,55 @@
...
+     * 
+     * @param properties
+     *            the properties object to write to the file.
+     * @param file
+     *            the file to write to. All the missing parent directories are also created.
+     * @throws IOException
+     *             if writing the property list to the specified output stream throws an
+     *             IOException.
+     */
+    public static void storeProperties(Properties properties, File file) throws IOException {
```
laeubi, line 42:
```suggestion
    public static void storeProperties(Properties properties, Path file) throws IOException {
```

Just a nitpick here, but for new API better always use `Path` as `File` can trivially be converted into a `Path`.

merks:
Perhaps the same conclusion as in https://github.com/eclipse-equinox/p2/pull/589#issuecomment-2571671377 applies here,  i.e., with Java 21 the properties are sorted and you can set a fixed reproducible timestamp via `java.properties.date`...

tycho-api/src/main/java/org/eclipse/tycho/ReproducibleUtils.java
Zlika: We have a File object in all the call sites, so using a Path will just add code for nothing. Also I prefer to use File because the name is more clear and File objects are simpler to create. Path does not supersede File, I don't think it's necessarily better to use it on new code.

laeubi: Maven4 is migrating all `File` > `Path` as well and as you already converted the file into path in this method it seems suitable here as well. I also don't see how `new File("...")` is simpler than `Path.of("...")`... and path is usually more flexible as it is not limited to local files.

Zlika:
> Perhaps the same conclusion as in [eclipse-equinox/p2#589 (comment)](https://github.com/eclipse-equinox/p2/pull/589#issuecomment-2571671377) applies here, i.e., with Java 21 the properties are sorted and you can set a fixed reproducible timestamp via `java.properties.date`...

Using a system property to set the date is quite messy, but it's an interesting workaround when tycho calls external tools like p2. It's a pity that this feature was only added in Java 18.

laeubi:
@Zlika can you adjust the util method to use `Path` API then I think this can be merged so we have also the support for java 21 at laest in the core Tycho.

Zlika:
> @Zlika can you adjust the util method to use `Path` API then I think this can be merged so we have also the support for java 21 at laest in the core Tycho.

Done.

tycho-api/src/main/java/org/eclipse/tycho/ReproducibleUtils.java
```diff
@@ -0,0 +1,55 @@
...
+     * @throws IOException
+     *             if writing the property list to the specified output stream throws an
+     *             IOException.
+     */
+    public static void storeProperties(Properties properties, Path file) throws IOException {
+        final Path folder = file.getParent();
+        if (folder != null) {
+            Files.createDirectories(folder);
+        }
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
laeubi, line 47: This should be surround by try/with/resources (just to make compilers happy)

tycho-api/src/main/java/org/eclipse/tycho/ReproducibleUtils.java
```diff
@@ -0,0 +1,55 @@
...
+        properties.store(baos, null);
+        final String content = baos.toString(StandardCharsets.ISO_8859_1).lines().filter(line -> !line.startsWith("#"))
+                .sorted().collect(Collectors.joining("\n", "", "\n"));
+        try (OutputStream os = new BufferedOutputStream(new FileOutputStream(file.toFile()))) {
```
laeubi, line 51:
```suggestion
        try (OutputStream os = new BufferedOutputStream(Files.newOutputStream((file))) {
```

laeubi (commented):
Just a few adjustments and I think we are done!

tycho-api/src/main/java/org/eclipse/tycho/ReproducibleUtils.java
Zlika: I don't understand: Files.createDirectories() can throw IOExceptions, just like the storeProperties() method, I don't see any try with resources to add. Also the code compiles without error, I don't even see a compiler warning.

laeubi: `ByteArrayOutputStream` is never closed if any exception (runtime or IO) happens after that line.

tycho-api/src/main/java/org/eclipse/tycho/ReproducibleUtils.java
laeubi: I think this is the last task remaining then....

laeubi (approved):
This all looks fine and consistent to me now.

PR #4671, opened 2025-02-01
Use the main project to let bnd compute the output file

laeubi:
Currently we compute the name of the subjar manually, but this can be overridden by the user so it is better to let bnd calculate the file itself.

FYI @chrisrueger


chrisrueger:
I can confirm that this PR  fixes our build which has set:

`target-dir=generated${if;${driver;eclipse};;/${driver}}` 

(see related https://bnd.bndtools.org/chapters/150-build.html#avoiding-target-dir-conflicts-between-different-build-tools )

Only thing I noticed is that it generates a folder `generated/unset/`

I believe `unset` is from the `${driver}` above which does seem to be populated with 'unset'. Not sure that is intended or correct. It seems that Eclipse identifies itself as `eclipse`. Not sure if it is possible that tycho populates it with `eclipsetycho`?

The idea behind this is that each build tool writes into its own folder so that different build tools (drivers) do not step on eachothers toes.

laeubi:
The driver name is now init with "tycho-maven-build"

PR #4673, opened 2025-02-01
Propagate configured folders from bnd to maven model

laeubi:
The user can define custom folders for output and sources, these are now propagated from the bnd config to the maven model.

FYI @chrisrueger


chrisrueger:
I retested this PR with our build. still successfull ✅  

Our folders in `cnf/build.bnd` are configured like this:

```
target-dir=generated${if;${driver;eclipse};;/${driver}}
bin=${target-dir}/classes
testbin=${target-dir}/test-classes
```

chrisrueger:
Question: One thing I noticed is that mvn clean does not remove the target-dir defined via `cnf/build-bnd`.

This is how it looks after a clean. 

image

laeubi:
> Question: One thing I noticed is that mvn clean does not remove the target-dir defined via cnf/build-bnd

Should be fixed now!

PR #4675, opened 2025-02-01
Extend  demo bnd-workspace for creation of executable jar via .bndrun

chrisrueger:
- added a .bndrun in the demo.impl bundle which is used to build the executable jar (comparable to building the product)
- added another consumer bundle which
- installed Felix Gogo Shell, so that you can even start the application and type 'hello' in the console to get an output from the service. this is basically now a very simple OSGi application
- changed to Java17
- removed all warnings in Eclipse

```
mvn clean verify -Dtycho-version=5.0.0-SNAPSHOT

[INFO] bnd-parent ......................................... SUCCESS [  0.038 s]
[INFO] tycho.demo.api ..................................... SUCCESS [  1.409 s]
[INFO] tycho.demo.consumer ................................ SUCCESS [  1.127 s]
[INFO] tycho.demo.markdown ................................ SUCCESS [  1.469 s]
[INFO] tycho.demo.impl .................................... SUCCESS [  1.219 s]
[INFO] BUILD SUCCESS

```

In Eclipse you can start the .bndrun and also Export it as an executable jar like  this:

image

The .bndrun can also be created and run for test and comparison purposes via the bnd command line: 

```
cd demo/bnd-workspace
bnd export tycho.demo.impl/tycho.demo.app.bndrun

```
or just run it as a application:

`bnd run tycho.demo.impl/tycho.demo.app.bndrun
`
which starts the app and shows the Felix Gogo Shell.

```
____________________________
Welcome to Apache Felix Gogo

g! Please type 'hello' into the console and press Enter, and magic will happen.
```

Then type 'hello'.

```
hello
Hello BND Workspace!
Render some markdown to HTML: <h2>H2 Headline</h2>
```


demo/bnd-workspace/cnf/build.bnd
```diff
@@ -1,3 +1,7 @@
 Bundle-Version: 1.0.0-SNAPSHOT
--runfw:         org.eclipse.osgi
```
laeubi, line 2: Without this I get

```
Exception in thread "main" java.lang.NoClassDefFoundError: org/osgi/framework/ServiceListener
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1027)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
	at java.base/java.net.URLClassLoader.defineClass(URLClassLoader.java:524)
	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:427)
	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:421)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:714)
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:420)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:593)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
	at aQute.launcher.pre.EmbeddedLauncher.executeWithRunPath(EmbeddedLauncher.java:160)
	at aQute.launcher.pre.EmbeddedLauncher.findAndExecute(EmbeddedLauncher.java:118)
	at aQute.launcher.pre.EmbeddedLauncher.main(EmbeddedLauncher.java:51)
Caused by: java.lang.ClassNotFoundException: org.osgi.framework.ServiceListener
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:445)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:593)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
```

when launching the produced export.

laeubi (commented):
For the upcomming support you need to add

`-Dbndrun.exports=tycho.demo.app`

to the `maven.config` file.

Also we should not check in the `.project`/`.classpath` files as this is a headless demo.

It would also be good the enhance this testcase:

https://github.com/eclipse-tycho/tycho/blob/9330061417e71eaddd4f7f99c8165cf6839d5b16/tycho-its/src/test/java/org/eclipse/tycho/test/DemoTest.java#L79-L82

to assert that the output file (will be `target/tycho.demo.app.zip` in this case) is actually produced, this should fail for now and should succeed after we have added support for this.

demo/bnd-workspace/cnf/build.bnd
chrisrueger: Hmm. I thought this belongs to `.bndrun`, as it it about "running" I thought. And I could have e.g. different .bndrun files using different frameworks (felix, equinox).

I found this sentence

> ...This means the bnd.bnd file must have the necessary -runfw and -runbundles to support the test bundles built by the project. The check task depends on the testOSGi task.

in
https://github.com/bndtools/bnd/blob/master/gradle-plugins/README.md

and an issue created by you: https://github.com/bndtools/bnd/issues/5504

Is this related?

I can put it back of course, if it is needed. 
I try  to reproduce it locally and maybe compare what bnd CLI is doing with / without it.

chrisrueger:
> Also we should not check in the `.project`/`.classpath` files as this is a headless demo.

ok, but then it gets hard to extend it. Without Eclipse I would not have been able to develop it... e.g. I need to resolve button to resolve the `-runbundles` in `.bndrun`. 
Maybe that is an indicator we also need something like  bnd-resolver-maven-plugin https://github.com/bndtools/bnd/blob/master/maven-plugins/bnd-resolver-maven-plugin/README.md too in the future.

I guess this is the "thinking" behind the current bnd-maven plugins to be able to develop completely headless. Ideally this would allow a development workflow independent of Eclipse IDE and also use VSCode or whatever.

demo/bnd-workspace/cnf/build.bnd
laeubi: I was also surprised a bit and can only tell that it worked but I have not investigated why :-)

chrisrueger:
@laeubi build and DemoTest works now. I verified with your PR https://github.com/eclipse-tycho/tycho/pull/4676

laeubi:
> ok, but then it gets hard to extend it. Without Eclipse I would not have been able to develop it... e.g. I need to resolve button to resolve the `-runbundles` in `.bndrun`.

Eclipse has so called [Project Configurator](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/extension-points/org_eclipse_ui_ide_projectConfigurators.html?cp=2_1_1_179) that allows for a feature called "Smart Import" so if you choose this here

grafik

and select the `tycho/demo/bnd-workspace` bndtools *should* offer you to import the projects and generate automatically some `.project` and other files.

chrisrueger:
> @laeubi build and DemoTest works now. I verified with your PR #4676\n\nBUT: see #4676 (comment) the content of the exported jar does not seem correct yet

Sorry for being not clear. 
- a jar is exported : ✅ 
- the content of the jar seems not correct yet

I just realized that after that comment

demo/bnd-workspace/cnf/build.bnd
laeubi: This works now without it!

demo/bnd-workspace/.gitignore
```diff
@@ -0,0 +1 @@
+generated
```
laeubi, line 1:
```suggestion
generated/
target/
```

tycho-its/src/test/java/org/eclipse/tycho/test/DemoTest.java
```diff
@@ -77,7 +77,10 @@ public void testTychoBndDemo() throws Exception {
 
 	@Test
 	public void testTychoBndWorkspaceDemo() throws Exception {
-		runDemo("bnd-workspace");
+		Verifier verifier = runDemo("bnd-workspace");
+		File exportedJar = Path.of(verifier.getBasedir(), "tycho.demo.impl/target/export/tycho.demo.app.jar")
```
laeubi, line 81:
```suggestion
		File exportedJar = Path.of(verifier.getBasedir(), "tycho.demo.impl/target/executable/tycho.demo.app.jar")
```

demo/bnd-workspace/tycho.demo.impl/tycho.demo.app.bndrun
```diff
@@ -0,0 +1,23 @@
+-runrequires: \
+	bnd.identity;id='tycho.demo.impl',\
+	bnd.identity;id='org.apache.felix.gogo.shell',\
+	bnd.identity;id='org.apache.felix.gogo.command',\
+	bnd.identity;id='tycho.demo.consumer',\
+	bnd.identity;id='tycho.demo.markdown.impl'
+-runee: JavaSE-17
+-runfw: org.eclipse.osgi;version='[3.21.0.v20240717-2103,3.21.0.v20240717-2103]'
```
laeubi, line 8: Do we need such a strict version range here?

demo/bnd-workspace/tycho.demo.impl/tycho.demo.app.bndrun
```diff
@@ -0,0 +1,23 @@
...
+	org.apache.felix.gogo.command;version='[1.1.2,1.1.3)',\
+	org.apache.felix.scr;version='[2.2.12,2.2.13)',\
+	org.osgi.service.component;version='[1.5.1,1.5.2)',\
+	org.osgi.util.function;version='[1.1.0,1.1.1)',\
+	org.osgi.util.promise;version='[1.3.0,1.3.1)',\
+	tycho.demo.api;version=snapshot,\
+	tycho.demo.impl;version=snapshot,\
+	tycho.demo.markdown.api;version=snapshot,\
+	tycho.demo.consumer;version=snapshot,\
+	tycho.demo.markdown.impl;version=snapshot
```
laeubi, line 21:
```suggestion
```

Instead we can set `-Dbndrun.resolve=true` now

laeubi (commented):
Thanks for the example it looks quite good now I added some comments how we possibly can simplify it a bit and because of changed implementation.

laeubi:
@chrisrueger can you also extend the demo readme here:

https://github.com/eclipse-tycho/tycho/tree/main/demo/bnd-workspace

mentioning the application e.g. something similar like:

> After building with `mvn clean install` you can run the demo application with `java -jar tycho.demo.impl/target/executable/tycho.demo.app.jar` then type ... and you will see ...
> ... 

Also currently your assert failed with:

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.eclipse.tycho.test.DemoTest.testTychoBndWorkspaceDemo(DemoTest.java:83)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
```

I think it would be good to add a little message like "Application was not build" or similar.

chrisrueger:
@laeubi I think I have everything now. let's wait if the build passes.

Regarding the `-runbundles` and `-Dbndrun.resolve=true` I like to discuss. I think it is a bit too much magic and maybe should be done in another demo / test

laeubi:
> Regarding the -runbundles and -Dbndrun.resolve=true I like to discuss. I think it is a bit too much magic and maybe should be done in another demo / test

Its is more an idea to show that one don't need to maintain the list and to test the resolve feature as well, but it is not mandatory!

chrisrueger:
> > Regarding the -runbundles and -Dbndrun.resolve=true I like to discuss. I think it is a bit too much magic and maybe should be done in another demo / test
> 
> Its is more an idea to show that one don't need to maintain the list and to test the resolve feature as well, but it is not mandatory!

Ok let's maybe have a chat about it. In eclipe when I press the "Resolve" button, it physically updates the `.bndrun` file and updates the `-runbundles`. 

I actually like that because I can "see" it with my eyes, and commit it to git and from then on it is written in stone. No surprises later. 

I expected something similar to happen... but it didn't. 
Not sure how the bnd-maven-resolve-plugin works. So maybe give it a bit more time.

laeubi:
> I expected something similar to happen... but it didn't.
> Not sure how the bnd-maven-resolve-plugin works. So maybe give it a bit more time.

Currently it is not written back but displayed to the console to not "dirty" the workingtree. I could write it back but usually a maven build should never modify your inputs directly (everything transient goes to the `target`) but of course one still can add such option.

chrisrueger:
> Currently it is not written back but displayed to the console to not "dirty" the workingtree. I could write it back but usually a maven build should never modify your inputs directly (everything transient goes to the `target`) but of course one still can add such option.

It's not that I am against it. It just depends on the development workflow you are used to and at which point in time you "resolve". 
We currently "resolve" at development time in Eclipse. So the .bndrun is our "set in stone" definition of the final product. 
The build then just "builds" this .bndrun file which was produced earlier. 

I have not thought about other ways of doing it. Maybe let's ask some more people.

laeubi:
> I have not thought about other ways of doing it. Maybe let's ask some more people.

As you provided the example I let you decide what to choose (of course in a real setup one has always both choices anyways).

PR #4676, opened 2025-02-02
Add support for exporting bndruns in bndworkspace projects

laeubi:
This adds support for "building" bndrun files by exporting them into a runnable jar.

This successfully builds the extended demo project supplied here:
- https://github.com/eclipse-tycho/tycho/pull/4675

FYI @chrisrueger

### DIFF
--- a/tycho-bnd-plugin/pom.xml
+++ b/tycho-bnd-plugin/pom.xml
@@ -48,6 +48,11 @@
 			<artifactId>biz.aQute.bnd.embedded-repo</artifactId>
 			<version>${bnd.version}</version>
 		</dependency>
+		<dependency>
+			<groupId>biz.aQute.bnd</groupId>
+			<artifactId>biz.aQute.resolve</artifactId>
+			<version>${bnd.version}</version>
+		</dependency>
 	</dependencies>
 	<build>
 		<plugins>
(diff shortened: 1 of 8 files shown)


chrisrueger:
@laeubi since the export should export an "executable jar":

Could you change the extension to .jar instead of .zip? e.g. in the demo bnd-workspace:  `target/tycho.demo.app.jar` instead of `target/tycho.demo.app.zip`

Then it would be in sync with current bnd export behavior (gradle, bnd cli and probably also bnd-maven-plugin).

laeubi:
> Could you change the extension to .jar instead of .zip? 

Done that and also slightly modified the path to be `tycho.demo.impl/target/export/tycho.demo.app.jar` so it does not collide with the main jar produced by the project

chrisrueger:
Hmm something is strange. The content of the exported jar looks different than when I create it with Eclipse or bnd cli:

with tycho export:

image

- There is a bnd-workspace.jar
- and when I run java -jar tycho.demo.app.jar nothing happens

And here the executable jar exported with Eclipse (Export button in .bndrun editor)

image

@laeubi

laeubi:
You now can specify

`-Dbndrun.resolve=true`

to perform a resolve operation, but currently it seems to do "nothing" (maybe because everything is already resolved).

chrisrueger:
> You now can specify
> 
> `-Dbndrun.resolve=true`
> 
> to perform a resolve operation, but currently it seems to do "nothing" (maybe because everything is already resolved).

Cool. You could test it by removing one of the -runbundles  in the demo

I will test it later today. On the road now

laeubi:
> Hmm something is different. The content of the exported jar looks different than when I create it with Eclipse or bnd cli

This is fixed now I accidentally passed the wrong file to the bndrun exporter and therefore it was empty :-\

laeubi:
Everything should now work as expected!

laeubi:
I slightly adjusted this to behave similar to `gradle` build where it exports into a `executable` subfolder.

So it is now `tycho.demo.impl/target/executable/tycho.demo.app.jar`

PR #4681, opened 2025-02-03
make synchronization of classpath calculation more fine-grained

andreas-schwarz-vector:
Currently, the calculation of the classpath for a project is effectively a singleton because the BundleProject is used as monitor (synchronization). This leads to a poor performance if executing a tycho build with  many threads, many projects and complex dependencies.

This change replaces the global monitor with a fine-grained monitor the given project the classpath should be calculated for. This speeds up the parallel build.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
@@ -320,13 +320,8 @@ public class OsgiBundleProject extends AbstractTychoProject implements BundlePro
         return getBundleClassPath(project).getExtraBootClasspathAccessRules();
     }
 
-    public synchronized BundleClassPath getBundleClassPath(ReactorProject project) {
-        if (project.getContextValue(CTX_CLASSPATH) instanceof BundleClassPath bundleClassPath) {
-            return bundleClassPath;
-        }
-        BundleClassPath cp = resolveClassPath(getMavenSession(project), getMavenProject(project));
-        project.setContextValue(CTX_CLASSPATH, cp);
-        return cp;
+    public BundleClassPath getBundleClassPath(ReactorProject project) {
+        return project.computeContextValue(CTX_CLASSPATH, () -> resolveClassPath(getMavenSession(project), getMavenProject(project)));
     }
 
     /**


tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
```diff
@@ -320,13 +320,15 @@ public List<ClasspathEntry.AccessRule> getBootClasspathExtraAccessRules(ReactorP
         return getBundleClassPath(project).getExtraBootClasspathAccessRules();
     }
 
-    public synchronized BundleClassPath getBundleClassPath(ReactorProject project) {
-        if (project.getContextValue(CTX_CLASSPATH) instanceof BundleClassPath bundleClassPath) {
-            return bundleClassPath;
+    public BundleClassPath getBundleClassPath(ReactorProject project) {
+        synchronized(project) {
```
laeubi, line 324: Good catch! 

I think one can even use `ReactorPoject#computeContextValue(String, Supplier<T>)` here

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
```diff
@@ -320,13 +320,10 @@ public List<ClasspathEntry.AccessRule> getBootClasspathExtraAccessRules(ReactorP
...
```
laeubi, line 324: You don't need to synchronize when using `computeContextValue` as it is internally using a concurrent map.

andreas-schwarz-vector: Thanks for the hint. Updated it.

andreas-schwarz-vector: Should i adjust the line formating also?

laeubi: @andreas-schwarz-vector at best only include the line that contains the real change and revert other formatting changes, beside from that it looks good now!

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
```diff
@@ -299,7 +299,7 @@ public EclipsePluginProject getEclipsePluginProject(ReactorProject otherProject)
     /**
      * Add to maven project some properties storing info about the plugin project model, for easier
      * reuse in further mojos (eg combining with eclipse-run and PDE API Tools).
-     * 
+     *
```
laeubi, line 302: Please revert formatting only changes here

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
```diff
@@ -521,7 +516,7 @@ public Filter getTargetEnvironmentFilter(MavenProject project) {
 
     @Override
     public void readExecutionEnvironmentConfiguration(ReactorProject project, MavenSession mavenSession,
-            ExecutionEnvironmentConfiguration sink) {
+           ExecutionEnvironmentConfiguration sink) {
```
laeubi, line 519: Please revert formatting only changes here

laeubi (approved):
Thanks for spotting this and the fix

HeikoKlare:
Thanks for your improvement propsals and for quickly processing this! 🙂

PR #4682, opened 2025-02-03
Consider runbundles when computing dependencies

laeubi:
If we have a bndrun activated then the runbundles become dependencies of the maven project.

This now calculates the enabled bndruns and return the bundle names so they can be used to add required reactor projects.

FYI @chrisrueger


chrisrueger:
I confirm this fixes our build and does not require the -testpath workaround anymore : 🎉

PR #4683, opened 2025-02-03
IT test for Issue 4653 reproducer

zladdi:
- External dependency `org.apache.commons:commons-lang3-${commons-lang3.version}.jar` is downloaded and placed into `target.eagerResolver/dependee/lib` at build time using `maven-dependency-plugin`.
- Please also review the naming of the project, naming and placement of the test class, and naming and placement of the other files and let me know if adjustments are requested.


laeubi:
@zladdi best is to always create PR on master, it can then be backported automatically.

zladdi:
> @zladdi best is to always create PR on master, it can then be backported automatically.

Ok, will redo that then. Thanks for the hint.

zladdi:
Closed, will redo pull request based on main branch

PR #4692, opened 2025-02-04
Move the readme into correct place, add an example with configurator pom

laeubi:
This moves the readme from .mvn into the root folder and also adds a new template that includes a configurator

WDYT @chrisrueger  ?


chrisrueger (commented):
LGTM. Should I also add this template to bnd? 
see https://github.com/bndtools/workspace-templates/pull/2/files and feel free to suggest what I should add here (naming, description) :)

chrisrueger (approved):

PR #4702, opened 2025-02-09
Remove signatures from wrapped source bundles

laeubi:
Currently when the source bundle is signed, wrapping it into a source bundle will break the signature and make the jar invalid.

This now removes the signature from source bundles as they are transformed.


merks:
Cool. Sounds great 👍

PR #4742, opened 2025-02-23
Use the new MavenDependencyCollector shared with m2e

laeubi:
Currently Tycho uses the usual maven resolver to find transitive dependencies but this works differently than what we want here. Instead we want all dependencies and their dependencies and then filter for the given scopes provided if any match.As this is a non trivial operation, it seems useful to make the code also sharable with Tycho so we ensure there is always the same result in Tycho as in m2e.

This now uses the new MavenDependencyCollector that is shared with Tm2e that traverses the dependency tree in a way we need here and collects everything along the way. This also has the advantage, that if we choose to only collect direct dependencies, not the whole tree has to be traversed as before.

### DIFF
--- a/tycho-api/src/main/java/org/eclipse/tycho/IArtifactFacade.java
+++ b/tycho-api/src/main/java/org/eclipse/tycho/IArtifactFacade.java
@@ -14,8 +14,6 @@
 package org.eclipse.tycho;
 
 import java.io.File;
-import java.util.Collections;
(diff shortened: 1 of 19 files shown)


laeubi:
There are still some deviations (e.g. how features and source bundles are generated) that should be unified/shared between Tycho+m2e. This also acounts for the required migration from "Maven Artifacts" > "Aether/Resolver".

PR #4743, opened 2025-02-23
Consider inter-fragment dependencies during compilation

HannesWell:
If a fragment depends on another fragment, e.g. by importing a package or requiring a capability, this is currently not considered during compilation, leading to an incomplete classpath and compilation failures.

Required for
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638

This is currently a draft that maybe needs some more fine-tuning.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/DependencyComputer.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/DependencyComputer.java
@@ -28,6 +28,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.stream.Collectors;
 
 import org.codehaus.plexus.component.annotations.Component;
 import org.eclipse.osgi.container.ModuleCapability;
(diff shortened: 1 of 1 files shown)


laeubi (changes requested):
> If a fragment depends on another fragment, e.g. by importing a package or requiring a capability

I think this is conceptually wrong. A fragment can't depend on another fragment as in the runtime fragments are merged with their host, so the host gets the req/cap and they share a linear class space.

> this is currently not considered during compilation, leading to an incomplete classpath and compilation failures

Currently Tycho already adds all fragments for compilation so maybe you can add an example/test that shows the problem.

As a workaround it might be possible to simply express the intend by adding an `jar.extra.classpath` entry to the build.properties.

HannesWell:
> > If a fragment depends on another fragment, e.g. by importing a package or requiring a capability
> 
> A fragment can't depend on another fragment as in the runtime fragments are merged with their host, so the host gets the req/cap and they share a linear class space.

Exactly. But unfortunately the shared, linear classpath isn't reflected during compilation at build-time.

> > this is currently not considered during compilation, leading to an incomplete classpath and compilation failures
> 
> Currently Tycho already adds all fragments for compilation so maybe you can add an example/test that shows the problem.
> 
That might be the case when one requires the host-bundle but it definitively doesn't happen if one fragment has to be compiled against the code of another fragment of the same host.
One example and the reason I created this is
- https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638

> As a workaround it might be possible to simply express the intend by adding an `jar.extra.classpath` entry to the build.properties.

That doesn't work in case of SWT as we would have to add all fragments for all platforms (which is a bit inconvenient, but that's acceptable) in order to have the one for the current platform available, but then all but the actually matching one fail to resolve and cannot be found, which fails the build.
I tried a lot of different configurations with all the things that are already possible, but nothing succeeded. So something probably has to be enhanced to support this case.

We could use the `jar.extra.classpath` approach if entries could be marked as optional (greedy), then we could list all native fragments. But this means we would have to maintain the list of available configurations in the build.properties of the new fragments. Therefore this isn't my favorite solution.

An alternative is to continue the path you mentioned that all fragments of a host are added for compilation and apply this to other fragments of the same host too. This would then even be a bit simpler than this suggestion, but I wonder if this could cause problems because then for example test-fragments are added to the compile classpath too?

HannesWell:
> > One example and the reason I created this is
> 
> Can you provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue? What exact error do we see and how does the classpath looks like?

Not sure if I have the time to assemble an integration test.
Without any special configuration the error is that the compilation classpath of one fragment (i.e. the one build second) does not contain the other (resolved) fragment of the same host (that is build first).

In case of SWT it means that e.g. if one builds on Windows x86 that the `o.e.swt.win32.win32.x86_64` fragment is not on the compile-classpath if the new `o.e.swt.svg` fragment is compiled.

> still I think for such a very special situation as in SWT it might be better to use special configuration that try to solve the situation in general what easily can break expectations.

If would be totally fine with that. But even with special configuration and tricks I knew I didn't succeed.

laeubi:
That's why I think we need a testcase first, otherwise its to risky and would probably break in the future.

HannesWell:
With the changed configuration in https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638 and the smaller adjustment in Tycho done via https://github.com/eclipse-tycho/tycho/pull/4771, this is now obsolete.

PR #4752, opened 2025-02-24
Only support 4.7+

laeubi:
This is not for Tycho 4.x


akurtakov:
Junit 5.x runs junit 4 tests via the vintage engine so it might be even possible to have only JUnit 5.

laeubi:
> Junit 5.x runs junit 4 tests via the vintage engine so it might be even possible to have only JUnit 5.

Maybe yes, but lets first start by reducing things to **one** major version, I don't know if there are any legacy involved in using "real" JUnit 4, at least maven-surefire still has this distinction so there might be a reason. Maybe to simply disallow using of JUnit5 features...

Also Eclipse itself has JUnit3 (emulated through JUNit4 jar), JUnit4 and JUnit 5 ... so I think if we want to go that path, the best would be if Platform takes the lead (as Tycho is mostly IDE first approach), and removing JUnit 3/4 as an option and only have JUnit 5 with the option to enable vintage or not.

See also:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1433

PR #4761, opened 2025-02-26
Only support latest JUnit 5

laeubi:
Historically Tycho has multiple providers for JUnit 5 to support different version. This has different drawbacks like maintenance burden and actually if one is coding against the JUnit 4 API there should be no immediate need to use older versions. Also as Tycho is loading providers from the maven classpath users that really require an older version can always choose to stay on a previous release (of Tycho) for that part like it is already possible with JDT compiler.

Because of this we should from now on only ever support one (preferably the latest) JUnit 5 release and make it automatic updateable.

**Only for Tycho 5.x**


laeubi:
This still by intention do not upgrade to latest JUnit 5.12 now. I plan to have a separate commit for the changes as this one is already massive change and we can better track any problems that might arise here.

So this just deletes all previous providers and rename `junit47` -> `junit4` and `junit59` -> `junit5`

laeubi:
This also removes support for the "legacy orbit names" of JUnit bundles.

PR #4762, opened 2025-02-26
Add support for using latest JUnit releases

laeubi:
This now upgrades to the latest JUnit 5.12 release and also makes it
auto-updateable by dependabot PRs (independent from junit used in Tycho
itself).

As we no longer statically know what version of junit is actually used,
this also adds a new way for the provider to determine the version
dynamically from the classpath of the mojo. If no version info is found,
it falls back to 5.0.0 as the reported version.

**For Tycho 5.x only**


laeubi:
The `LocalRepositoryCrosstalkTest` succeeds locally for me and also seem to suceed on one Github runner. I'll now try to update it to a more recent version to see if that makes a difference.

PR #4771, opened 2025-03-01
Resolve jars and library extra.classpath entries optionally

HannesWell:
If a bundle is referenced in an 'jars.extra.classpath' entry cannot be resolved (e.g. because it's a fragment) then the resolution of the referencing project should not fail. This is also in accordance with PDE where entries that cannot be resolved are simply ignored.

A reproducer is https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638, where we have the following entry in the `build.properties` of a new `o.e.swt.svg` fragment:
```
jars.extra.classpath = platform:/plugin/org.eclipse.swt.cocoa.macosx.aarch64,\
                       platform:/plugin/org.eclipse.swt.cocoa.macosx.x86_64,\
                       platform:/plugin/org.eclipse.swt.gtk.linux.aarch64,\
                       platform:/plugin/org.eclipse.swt.gtk.linux.ppc64le,\
                       platform:/plugin/org.eclipse.swt.gtk.linux.riscv64,\
                       platform:/plugin/org.eclipse.swt.gtk.linux.x86_64,\
                       platform:/plugin/org.eclipse.swt.win32.win32.aarch64,\
                       platform:/plugin/org.eclipse.swt.win32.win32.x86_64
```
That currently fails with the following error:
```
[ERROR] Cannot resolve dependencies of project eclipse.platform.swt:org.eclipse.swt.svg:eclipse-plugin:1.0.0-SNAPSHOT
 [ERROR]  with context {osgi.os=linux, org.eclipse.update.install.features=true, org.eclipse.swt.buildtime=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true, osgi.ws=gtk, org.eclipse.jdt.buildtime=true}
 [ERROR]   Software being installed: org.eclipse.swt.svg 1.0.0.qualifier
 [ERROR]   Missing requirement: org.eclipse.swt.svg 1.0.0.qualifier requires 'org.eclipse.equinox.p2.iu; org.eclipse.swt.win32.win32.x86_64 0.0.0' but it could not be found: See log for details
```
It does so, even if one only specifies only the fragment of the current platform.

If necessary we can also make this new behavior configurable.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/p2/publisher/AbstractMetadataGenerator.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/p2/publisher/AbstractMetadataGenerator.java
@@ -38,10 +38,10 @@ import org.eclipse.tycho.BuildProperties;
 import org.eclipse.tycho.BuildPropertiesParser;
 import org.eclipse.tycho.IArtifactFacade;
 import org.eclipse.tycho.IDependencyMetadata.DependencyMetadataType;
-import org.eclipse.tycho.core.shared.StatusTool;
 import org.eclipse.tycho.OptionalResolutionAction;
 import org.eclipse.tycho.TargetEnvironment;
 import org.eclipse.tycho.TychoConstants;
+import org.eclipse.tycho.core.shared.StatusTool;
 import org.eclipse.tycho.p2.metadata.PublisherOptions;
 
 public abstract class AbstractMetadataGenerator {
@@ -110,7 +110,7 @@ public abstract class AbstractMetadataGenerator {
         Matcher m = TychoConstants.PLATFORM_URL_PATTERN.matcher(url);
         if (m.matches())
             result.add(MetadataFactory.createRequirement(IInstallableUnit.NAMESPACE_IU_ID, m.group(2),
-                    VersionRange.emptyRange, null, false, false));
+                    VersionRange.emptyRange, null, true, false));
     }
 
     private DependencyMetadata publish(IPublisherInfo publisherInfo, List<IPublisherAction> actions) {


tycho-core/src/main/java/org/eclipse/tycho/p2/publisher/AbstractMetadataGenerator.java
```diff
@@ -110,7 +110,7 @@ private void createRequirementFromPlatformURL(ArrayList<IRequirement> result, St
         Matcher m = TychoConstants.PLATFORM_URL_PATTERN.matcher(url);
         if (m.matches())
             result.add(MetadataFactory.createRequirement(IInstallableUnit.NAMESPACE_IU_ID, m.group(2),
-                    VersionRange.emptyRange, null, false, false));
+                    VersionRange.emptyRange, null, true, false));
```
laeubi, line 113:
```suggestion
                    VersionRange.emptyRange, null, true, true));
```
If we make it optional it should be at least greedy.

laeubi (commented):
Making it configurable sounds a good idea (especially for backward compatibility) bet through https://tycho.eclipseprojects.io/doc/latest/target-platform-configuration/target-platform-configuration-mojo.html#dependency-resolution what already has a flag for `optionalDependencies` so adding a new option `classpathDependencies` with values of

- require
- optional
- ignore

would be useful.

tycho-core/src/main/java/org/eclipse/tycho/p2/publisher/AbstractMetadataGenerator.java
HannesWell: I considered this as well, but according to the doc the second flag doesn't handle greediness but multiplicity:

https://github.com/eclipse-equinox/p2/blob/5b57f58f9ad96c332dcf99218b24dbc5a5cfb705/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/p2/metadata/MetadataFactory.java#L535-L537

HannesWell:
> Making it configurable sounds a good idea (especially for backward compatibility) bet through https://tycho.eclipseprojects.io/doc/latest/target-platform-configuration/target-platform-configuration-mojo.html#dependency-resolution what already has a flag for `optionalDependencies` so adding a new option `classpathDependencies` with values of

Yes, I already had a look at that and it seems suitable. It's just that we have to pull this through a few places and I first wanted to see if making it simple would also work.
But I update this PR.

tycho-core/src/main/java/org/eclipse/tycho/p2/publisher/AbstractMetadataGenerator.java
laeubi: You need then this method: https://github.com/eclipse-equinox/p2/blob/5b57f58f9ad96c332dcf99218b24dbc5a5cfb705/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/p2/metadata/MetadataFactory.java#L565

laeubi:
I think this is only for the metadata generator itself it not needs the configuration an can simply be greedy (so it is included if available), but later when we build the classpath of the project the jar.extra is evaluated again, that would be the right place and should have the target configuration at hand.

tycho-core/src/main/java/org/eclipse/tycho/p2/publisher/AbstractMetadataGenerator.java
HannesWell: In fact the currently used method also makes the created requirement greedy:
https://github.com/eclipse-equinox/p2/blob/5b57f58f9ad96c332dcf99218b24dbc5a5cfb705/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/p2/metadata/MetadataFactory.java#L543C39-L543C43

So this is already as desired.

HannesWell:
> I think this is only for the metadata generator itself it not needs the configuration an can simply be greedy (so it is included if available), but later when we build the classpath of the project the jar.extra is evaluated again, that would be the right place and should have the target configuration at hand.

Yes, but then I think this is already in the desired state, because it seems that the addition to the compilation-classpath done later already ignores missing bundles by default. And since I now don't have a need to make it required, I don't want to extend the work on this more than necessary. And it would also add an API, that maybe none uses.

laeubi:
> because it seems that the addition to the compilation-classpath done later already ignores missing bundles by default

SO maybe that is actually desired, I just though people would maybe get a notice if something is not found.

HannesWell:
> > because it seems that the addition to the compilation-classpath done later already ignores missing bundles by default
> 
> SO maybe that is actually desired, I just though people would maybe get a notice if something is not found.

In same cases for sure, but in other cases (like the one in SWT) not really. At the moment I would just leave it as it is.

I also tried to create a minimal test-case that reproduces the scenario from SWT, but I failed to reproduce the exact failure from SWT. I assume that the p2-instructions defined in SWT also play a role here.
Since the SWT build is also executed as part of the Tycho verification build suite future regression should be avoided.
Therefore I suggest to submit this now as it is so that we can make progress in SWT soon.

PR #4772, opened 2025-03-01
Clean-up integration tests and improve re-use of pomless projects

HannesWell:
If tycho-pomless is used in (integration) test-resources/projects the entire project structure, especially including the '.mvn' folder has to be copied into the clean test-environment, even if just a sub-project is executed.
This extends AbstractTychoIntegrationTest.getVerifier() to add overloads that allow to specify the project-structure root and the specific sub-folder that is eventually executed.

There are probably more integration that could be cleaned-up with this, but as a first step I have only used it in one place.


HannesWell:
Just just rebased this.
But as this currently contains two commits with back and forth I would like to know first if the latest state works. If yes, I'll squash this and it should be ready.

PR #4863, opened 2025-03-13
Update to Java 21 and cleanup actions

laeubi:
Now that platform more and more requires Java 21 it becomes time for Tycho 5.x to start requiring Java 21 as well.

This also updates some of the workflows to remove no longer needed options and using defaults offered by these actions.


laeubi:
We have one strange error here from JDT:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3794

beside that I have adjusted some of the test. **This does not look correct to me** as it means changing the build JVM somehow break existing projects what is not really desirable.

laeubi:
This has now split into individual PRs

PR #4790, opened 2025-03-06
Setversion updaterange on major version change

phermsdorf:
Followup on the issue described in discussion #4788 .

This PR adds an integration test demonstrating the described issue.

Please adjust as you see fit.

Thanks for looking into the problem.

bye Peter

PR #4882, opened 2025-03-15
Add support for consumer version range updates

laeubi:
FY @phermsdorf 

This includes the the testcase from
- https://github.com/eclipse-tycho/tycho/pull/4790

### DIFF
--- a/tycho-its/projects/tycho-version-plugin/set-version/version_ranges_major_version/consumer.bundle/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/tycho-version-plugin/set-version/version_ranges_major_version/consumer.bundle/META-INF/MANIFEST.MF
@@ -1,0 +1,7 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Test Bundle
+Bundle-SymbolicName: consumer.bundle
+Bundle-Version: 1.1.3.qualifier
+Require-Bundle: provider;bundle-version="[1.0.0,2.0.0)"
+
--- a/tycho-its/projects/tycho-version-plugin/set-version/version_ranges_major_version/consumer.bundle/pom.xml
+++ b/tycho-its/projects/tycho-version-plugin/set-version/version_ranges_major_version/consumer.bundle/pom.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/maven-v4_0_0.xsd">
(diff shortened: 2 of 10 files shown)


laeubi:
@phermsdorf the test currently fails with:

```
Import-Package felix.bundle.exported;version="[1.0.0,2.0.0)": expected version range = [11.0.0,12.0.0) but actual version range = [1.0.0,2.0.0)
java.lang.AssertionError: Import-Package felix.bundle.exported;version="[1.0.0,2.0.0)": expected version range = [11.0.0,12.0.0) but actual version range = [1.0.0,2.0.0)
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.eclipse.tycho.test.versionsplugin.TychoVersionsPluginTest.assertVersionRange(TychoVersionsPluginTest.java:428)
	at org.eclipse.tycho.test.versionsplugin.TychoVersionsPluginTest.updateVersionRangesMajorVersionChanges(TychoVersionsPluginTest.java:265)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
```

As this is a generated manifest I actually think from executing the update mojo itself it should not change (only the `pom.xml`) and the expectation of the test is wrong here.

phermsdorf:
Mhm, that's correct.

I tried to replicate our use case as close as possible, but still the testcase behaved a bit different.
I tried adding a verify goal before the set-version to trigger the MANIFEST.MF generation (which basically works), but still the set-version call does not seem to update the manifest version.

What i wanted to show is my expected behaviour of that set-version in a mixed reactor (which maybe is a litte beyond the initial problem with the correct generation of the bounds): so changing a version of a plain maven module which in turn get's a generated MANIFEST.MF including that changed version can have an impact on osgi artifacts in the same reactor (eg. changed bounds in import package or require bundle ...)

I need to investigate further ... maybe the behaviour inside the ide is a bit different

phermsdorf:
yes, please do that and i'll have a look hwo we can handle the problem in this mixed reactor use case

laeubi:
@phermsdorf I checked again and my initial analysis was wrong, it is actually correct that the test fails currently as the import-package is not updated (even though the require bundle is correctly found) I need do check if this is a general problem or maybe because of the generated manifest is not detected correctly (what might be tricky).

laeubi:
I now fixed up the build that it actually is possible to build it, but one problem remains:

As the `maven-bundle-plugin` is a "pom-only" then Tycho do not know that an exported package has changed its version. Therefore there is no way for the update mojo to really know it should touch the package import at all. It seem to work for the require bundle because we always assume that artifactid == bundle id what is also not always correct.

laeubi:
@phermsdorf my current approach would be to convert the testcase to not using automatic manifest generation. This seems to be not supported at all at the moment and is a bigger task.

If this is then merged and fixed, it would be good to add another test-case explicit for this case of automatic manifest as well, but this is independent from the case of updating major versions.

phermsdorf:
I'm fine with that.

PR #4866, opened 2025-03-14
testcase for contextclassloader problem

cdietrich:

PR #4883, opened 2025-03-15
Delegate resource loads from SurefireLoader

laeubi:
FYI @cdietrich 

this includes the testcase from:
- https://github.com/eclipse-tycho/tycho/pull/4866


cdietrich:
Many thanks

PR #4897, opened 2025-03-18
tycho-apitools-plugin:verify: Add option to write XML report

sratz:
The XML format is the based on the API Tooling ANT task [1,2], with the following differences / improvements:
- all categories are considered (the ant task only serializes the 'compatibility', 'usage', and 'version' categories).
- the human-readable names of the categories (written as 'value' attributes) are taken from ApiProblem.getProblemCategory(category) instead of from local hard-coded values.
- a 'path' argument to an <api_problem> element is added.
- empty categories are omitted from the XML (ant task always created an empty XML structure).

[1] org.eclipse.pde.api.tools.internal.tasks
[2] https://help.eclipse.org/latest/topic/org.eclipse.pde.doc.user/reference/api-tooling/ant-tasks/analysis-ant-task.htm

### DIFF
--- a/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisMojo.java
+++ b/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisMojo.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2023 Christoph Läubrich and others.
+ * Copyright (c) 2023, 2025 Christoph Läubrich and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -9,6 +9,7 @@
  *
  * Contributors:
  *    Christoph Läubrich - initial API and implementation
+ *    SAP SE - add option to write an XML report
  *******************************************************************************/
 package org.eclipse.tycho.apitools;
 
@@ -27,6 +28,9 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.TransformerException;
+
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Repository;
@@ -40,6 +44,7 @@ import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
 import org.apache.maven.project.MavenProject;
+import org.eclipse.core.runtime.CoreException;
 import org.eclipse.pde.api.tools.internal.IApiCoreConstants;
 import org.eclipse.pde.api.tools.internal.provisional.ApiPlugin;
 import org.eclipse.pde.api.tools.internal.provisional.problems.IApiProblem;
@@ -61,6 +66,7 @@ import org.eclipse.tycho.osgi.framework.EclipseWorkspace;
 import org.eclipse.tycho.osgi.framework.EclipseWorkspaceManager;
 import org.osgi.framework.BundleException;
 import org.osgi.framework.Version;
+import org.w3c.dom.DOMException;
 
 /**
  * Performs a PDE-API Tools analysis of this project.
@@ -140,6 +146,24 @@ public class ApiAnalysisMojo extends AbstractMojo {
 
 	@Parameter(defaultValue = "false", property = "tycho.apitools.enhanceLogs")
 	private boolean enhanceLogs;
+
+	/**
+	 * The mojo by default also creates an XML report of the API problems. The
+	 * format is understood by the <a href=
+	 * "https://github.com/jenkinsci/warnings-ng-plugin/blob/main/doc/Documentation.md#export-your-issues-into-a-supported-format">warnings-ng
+	 * Jenkins plugin</a> using the native format.
+	 * <p>
+	 * If you make use of this, consider setting
+	 * {@code <failOnError>false</failOnError>} in order to not fail the Maven build
+	 * on the first bundle with API problems, so that all the reports of all the
+	 * bundles can be analyzed subsequently by external means.
+	 * <p>
+	 * This may be set to an empty string to skip creation of the XML report.
+	 *
+	 * @since 4.0.13
+	 */
+	@Parameter(defaultValue = "${project.build.directory}/apianalysis/report.xml", property = "tycho.apitools.report")
+	private File report;
 
 	/**
 	 * Configures if the API Analysis should run as a workspace job, this ensure
@@ -235,6 +259,16 @@ public class ApiAnalysisMojo extends AbstractMojo {
 					}
 				} catch (IOException e) {
 					log.warn("Can't enhance logs in directory " + logDirectory);
+				}
+			}
+			if (report != null && !report.getPath().isEmpty()) {
+				try {
+					new ApiAnalysisXmlGenerator(
+							projectManager.getTychoProject(project).get()
+									.getArtifactKey(DefaultReactorProject.adapt(project)).getId(),
+							analysisResult, project.getBasedir().toPath(), report.toPath()).writeReport();
+				} catch (DOMException | IOException | ParserConfigurationException | TransformerException e) {
+					throw new MojoExecutionException(e);
 				}
 			}
 			if (printSummary) {
--- a/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisXmlGenerator.java
+++ b/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisXmlGenerator.java
@@ -1,0 +1,167 @@
+/*******************************************************************************
+ * Copyright (c) 2025 SAP SE and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     SAP SE - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.tycho.apitools;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import javax.xml.parsers.DocumentBuilder;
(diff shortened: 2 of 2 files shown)


tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisMojo.java
```diff
@@ -141,6 +145,30 @@ public class ApiAnalysisMojo extends AbstractMojo {
...
+	private boolean writeXmlReport;
+
+	/**
+	 * If {@link #writeXmlReport} is enabled, the directory where to write the
+	 * {@code report.xml} to.
+	 *
+	 * @since 4.0.13
+	 */
+	@Parameter(defaultValue = "${project.build.directory}/apianalysis")
+	private File xmlReportDirectory;
```
laeubi, line 170:
```suggestion
	@Parameter(defaultValue = "${project.build.directory}/apianalysis/report.xml", property = "tycho.apitools.report")
	private File xmlReport;
```
Instead of a directory I would simply allow configuration of the whole file, also we should have a property for this.

tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisMojo.java
```diff
@@ -141,6 +145,30 @@ public class ApiAnalysisMojo extends AbstractMojo {
...
+	 * Tools ANT task</a>.
+	 * <p>
+	 * Typically, this is best used in combination with
+	 * {@code <failOnError>false</failOnError>} in order to not fail the Maven build
+	 * and allow subsequent analysis by external means.
+	 *
+	 * @since 4.0.13
+	 */
+	@Parameter(defaultValue = "false", property = "tycho.apitools.writeXmlReport")
+	private boolean writeXmlReport;
```
laeubi, line 161: I don't think we need an extra boolean here, the usual default with maven is that one writes a report and then it might not get used, given that I assume its quite fast to generate the xml.

We might skip writing the report if the user sets the result file to an explicit empty string.

laeubi (commented):
> @laeubi WDYT? I was trying to migrate from the ANT task to a more standard in-build approach using Tycho, but was missing the option to generate an XML report.

Thanks for this its an interesting approach and I haven't know that such format exits, I'm wondering if there are any tools that read this report?

tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisMojo.java
sratz: I'm fine with either way. The only important thing to consider is that if `<failOnError>true</failOnError>` is enabled and the build fails immediately after any bundle has a problem the XML files are not all that useful, since you won't get results from all the other bundles. Hence also the javadoc comment.

laeubi: Can we maybe just write the report before the check for errors (and fail the mojo) then one should get a report for the current module (what is usual maven behavior with failing tests as well)

laeubi:
> I wrote a custom parser for https://plugins.jenkins.io/warnings-ng/ to visualize them on our Jenkins.

Do you think that can be contributed to the warnings-ng plugin directly? That would really be useful, at the moment we use the option to [weave this into the compiler logfile](https://tycho.eclipseprojects.io/doc/latest/tycho-apitools-plugin/verify-mojo.html#enhanceLogs) but having a dedicated entry in the Jenkins would of course be great!

sratz:
> Do you think that can be contributed to the warnings-ng plugin directly? That would really be useful, at the moment we use the option to [weave this into the compiler logfile](https://tycho.eclipseprojects.io/doc/latest/tycho-apitools-plugin/verify-mojo.html#enhanceLogs) but having a dedicated entry in the Jenkins would of course be great!

It's just this in the pipeline right now, nothing fancy, also no proper XML parser:
```groovy
recordIssues enabledForFailure: true, qualityGates: [[threshold: 1, type: 'TOTAL', unstable: true]], tools: [taskScanner(
    includePattern: 'git/META/staticchecks/apichecks/apitools/target/reports/**/report.xml',
    highTags: '^.*<(api_problem).*message="(.*)".*severity="2".*$',
    normalTags: '^.*<(api_problem).*message="(.*)".*severity="1".*$',
    ignoreCase: true,
    isRegularExpression: true,
    id: 'broken-apis',
    name: 'Broken APIs'
)]
```

laeubi:
I don't know how it exactly works, but there is already a parser for [Eclipse](https://github.com/jenkinsci/warnings-ng-plugin/blob/main/plugin/src/main/java/io/jenkins/plugins/analysis/warnings/Eclipse.java) and of course its much nicer to setup than copy a snippet that probably needs to be maintained.

This is how it looks like e.g. for Eclipse platform:

https://ci.eclipse.org/platform/job/eclipse.platform/job/master/1089/eclipse/
https://github.com/eclipse-platform/eclipse.platform/blob/c0d9a0f95a4526288df3ab1f93f7002f4e65c70a/Jenkinsfile#L35

sratz:
> I don't know how it exactly works, but there is already a parser for [Eclipse](https://github.com/jenkinsci/warnings-ng-plugin/blob/main/plugin/src/main/java/io/jenkins/plugins/analysis/warnings/Eclipse.java) 

Hmm, interesting. So `enhanceLogs` picky-bags on the compiler log.

The parser code for that is here:
https://github.com/jenkinsci/analysis-model/blob/2a4a293cc9c253d0d4ee6c40eaddd35f3911625f/src/main/java/edu/hm/hafner/analysis/parser/EclipseXmlParser.java

> This is how it looks like e.g. for Eclipse platform:
> 
> https://ci.eclipse.org/platform/job/eclipse.platform/job/master/1089/eclipse/

Problem is that we lose much of the semantics of the API warnings, e.g. they all end up in the 'OTHERS' category:
https://github.com/jenkinsci/analysis-model/blob/2a4a293cc9c253d0d4ee6c40eaddd35f3911625f/src/main/java/edu/hm/hafner/analysis/parser/EclipseXmlParser.java#L139

> [...] and of course its much nicer to setup than copy a snippet that probably needs to be maintained.

Hence my proposal to instead of picky-bagging a foreign format, we explicitly produce a format that warnings-ng understands out-of-the-box and in which we are able to convey all the semantics of the API warnings.
Then, we could simply change in
> https://github.com/eclipse-platform/eclipse.platform/blob/c0d9a0f95a4526288df3ab1f93f7002f4e65c70a/Jenkinsfile#L35

```
recordIssues(publishAllIssues:false, ignoreQualityGate:true,
		tool: eclipse(name: 'Compiler and API Tools', pattern: '**/target/compilelogs/*.xml'),
```
to
```
recordIssues(publishAllIssues:false, ignoreQualityGate:true,
		tools: [
			eclipse(name: 'Compiler', pattern: '**/target/compilelogs/*.xml'),
			issues(name: 'API Warnings'. pattern: '**/target/apianalysis/*.xml')
		],
```

`issues` is the default warnings-ng format, and no non-portable parser code would have to be maintained in the Jenkinsfile.

> I think an important part is to be able to jump to the actual source file.

This would be supported by `issues` as well.

laeubi:
> Problem is that we lose much of the semantics of the API warnings, e.g. they all end up in the 'OTHERS' category:

Yes that true and it is more a workaround.

> `issues` is the default warnings-ng format, and no non-portable parser code would have to be maintained in the Jenkinsfile.

What do you think about creating only warnings-ng format? Given that it seems there are no tool out there I'm not sure if it is worth to support the ant-style format. Once implemented, we can then even remove the compiler log extension in Tycho 5.x so we only have one way.

sratz:
> What do you think about creating only warnings-ng format? Given that it seems there are no tool out there I'm not sure if it is worth to support the ant-style format. Once implemented, we can then even remove the compiler log extension in Tycho 5.x so we only have one way.

Sounds like a plan. That way we also don't need to concern ourselves with keeping the ANT tooling and Tycho XML generator in sync. I'll keep working on this as soon as I find some free time.

sratz:
Updated for analysis-ng format.
Tested locally by removing a constant from `org.eclipse.core.runtime.Platform`:

```xml
<report>
    <issue>
        <moduleName>org.eclipse.core.runtime</moduleName>
        <message>The field org.eclipse.core.runtime.Platform.PT_PREFERENCES has been removed</message>
        <category>COMPATIBILITY</category>
        <type>REMOVED</type>
        <severity>ERROR</severity>
        <lineStart>58</lineStart>
        <lineEnd>58</lineEnd>
        <columnStart>2569</columnStart>
        <columnEnd>2577</columnEnd>
        <fileName>/home/jenkins/jenkins/workspace/test_apiwarnings/git/eclipse.platform/runtime/bundles/org.eclipse.core.runtime/src/org/eclipse/core/runtime/Platform.java</fileName>
        <packageName>org.eclipse.core.runtime</packageName>
        <additionalAttributes>id=338755678</additionalAttributes>
    </issue>
    <issue>
        <moduleName>org.eclipse.core.runtime</moduleName>
        <message>The major version for the package 'org.eclipse.core.runtime' should be incremented to version 4.7.0, since API breakage occurred since version 3.7.0</message>
        <category>VERSION</category>
        <type>UNKNOWN_KIND</type>
        <severity>ERROR</severity>
        <lineStart>9</lineStart>
        <lineEnd>9</lineEnd>
        <columnStart>312</columnStart>
        <columnEnd>312</columnEnd>
        <fileName>/home/jenkins/jenkins/workspace/test_apiwarnings/git/eclipse.platform/runtime/bundles/org.eclipse.core.runtime/META-INF/MANIFEST.MF</fileName>
        <additionalAttributes>id=934281281</additionalAttributes>
    </issue>
    <issue>
        <moduleName>org.eclipse.core.runtime</moduleName>
        <message>The major version should be incremented in version 3.33.0, since API breakage occurred since version 3.33.0</message>
        <category>VERSION</category>
        <type>MAJOR_VERSION_CHANGE</type>
        <severity>ERROR</severity>
        <lineStart>4</lineStart>
        <lineEnd>4</lineEnd>
        <columnStart>89</columnStart>
        <columnEnd>105</columnEnd>
        <fileName>/home/jenkins/jenkins/workspace/test_apiwarnings/git/eclipse.platform/runtime/bundles/org.eclipse.core.runtime/META-INF/MANIFEST.MF</fileName>
        <additionalAttributes>id=923795461</additionalAttributes>
    </issue>
</report>
```

Jenkins was configured with default format:

```groovy
post {
    always {
        recordIssues tools: [issues(id: 'apichecks', name: 'API', pattern: '**/apianalysis/report.xml')]
    }
}
```
image
image

laeubi (approved):
Many thanks, if you like you might want to add a follow-up to deprecate the parameter `enhanceLogs` in favor of this new option. Also one this is available we should make use of this in platform itself!

laeubi:
@sratz you probably also want to mention this in the release notes!

Issue #4905, opened 2025-03-23
Classloader leak in API analysis mojo


ptziegler:
This is a profiler snapshot when executing the Eclipse Platform build on Windows and with following command line instructions:

> mvn --batch-mode -V -U -e -ntp -D compare-version-with-baselines.skip=false -P bree-libs -P api-check --fail-at-end clean verify

![Image](https://github.com/user-attachments/assets/d438f266-f33f-4d2d-b3d4-47ddfed1f232)

As opposed to how the same build looks like on Linux:

![Image](https://github.com/user-attachments/assets/e5079fa9-cc92-4e86-8707-4a0114211a44)

For each module, the API analysis loads a set of classes which are unable to be garbage collected by the JVM, leading an increasing memory demand and inevitably to an OOM error.

The reasons for this problem are dangling references to the ApiModelCache and the SoftClassesCache of the API tools and JDT respectively, which are primarily seen on Windows due to the use of native I/O hooks, which are not properly disposed after the Mojo has been executed.

![Image](https://github.com/user-attachments/assets/00e78406-8816-4acb-8c61-07a0e04a6d93)

ptziegler:
PR will follow

PR #4906, opened 2025-03-23
Dispose API analysis objects after the mojo has been executed

ptziegler:
Closes https://github.com/eclipse-tycho/tycho/issues/4905

### DIFF
--- a/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysis.java
+++ b/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysis.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2023 Christoph Läubrich and others.
+ * Copyright (c) 2023, 2025 Christoph Läubrich and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -45,6 +45,7 @@ import org.eclipse.core.resources.IWorkspaceDescription;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.resources.WorkspaceJob;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.ILogListener;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
@@ -128,10 +129,10 @@ public class ApiAnalysis implements Serializable, Callable<ApiAnalysisResult> {
 
 	@Override
 	public ApiAnalysisResult call() throws Exception {
-
-		Platform.addLogListener((status, plugin) -> debug(status.toString()));
+		ILogListener logListener = (status, plugin) -> debug(status.toString());
+		Platform.addLogListener(logListener);
 		IJobManager jobManager = Job.getJobManager();
-		jobManager.addJobChangeListener(new IJobChangeListener() {
+		IJobChangeListener jobListener = new IJobChangeListener() {
 
 			@Override
 			public void sleeping(IJobChangeEvent event) {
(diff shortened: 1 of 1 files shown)


tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysis.java
```diff
@@ -174,20 +175,24 @@ public void aboutToRun(IJobChangeEvent event) {
...
+					}
+					throw error;
 				}
-				throw error;
+				return result;
 			}
-			return result;
+			throw exception;
+		} finally {
+			Platform.removeLogListener(logListener);
```
ptziegler, line 194: The log listener keeps an I/O reference if not removed afterwards.

image

tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysis.java
```diff
@@ -251,6 +256,7 @@ private IStatus performAPIAnalysis(IProject project, IPath projectPath, ApiAnaly
 					debug(String.valueOf(problem));
 				}
 			} finally {
+				baseline.dispose();
```
ptziegler, line 259: The baseline needs to be explicitly disposed. Otherwise the instance doesn't de-register itself from JDT.

image

tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysis.java
```diff
@@ -174,20 +176,25 @@ public void aboutToRun(IJobChangeEvent event) {
...
+					}
+					throw error;
 				}
-				throw error;
+				return result;
 			}
-			return result;
+			throw exception;
+		} finally {
+			jobManager.removeJobChangeListener(jobListener);
```
ptziegler, line 195: Cleaning up the job listener is probably also a good idea.

image

laeubi (approved):
Interesting findings @ptziegler just let me know when you are fine with your PR and I'll enable it for automerge and backport.

laeubi:
@ptziegler anything left or should we merge?

ptziegler:
> @ptziegler anything left or should we merge?

Sadly, it's not done yet. It seems like just creating the project also causes a leak that can't be closed, even by deleting the project afterwards. And I haven't figured out a way to solve that problem yet.

https://github.com/eclipse-tycho/tycho/blob/1f8aeaccf8b9f49f54a7a9f04872b8e3d91a8335/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysis.java#L317-L320

laeubi:
@ptziegler If you think parts of this should already be merged let me know!

ptziegler:
@laeubi I went through everything again and I believe that this is as much as what can be improved, at least from the Tycho side.

Generally, I get the impression that neither JDT nore PDE were ever designed for a use-case where one spawns and destroys an Equinox framework over and over again. Or to rephrase: The plugins don't clean up properly after themselves, after they have been shut down.

As an example, the [WorkspaceModelManager](https://github.com/eclipse-pde/eclipse.pde/blob/master/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/WorkspaceModelManager.java)

This method is called when the bundle is started.

```java
protected void addListeners() {
    IWorkspace workspace = PDECore.getWorkspace();
    workspace.addResourceChangeListener(this, IResourceChangeEvent.PRE_CLOSE | IResourceChangeEvent.POST_CHANGE);
    if (bundleRootChangedListener != null) {
        Arrays.stream(workspace.getRoot().getProjects()).forEach(this::addBundleRootChangedListener);
    }
}
```

This method is called when the bundle is stopped.

```java
protected void removeListeners() {
    PDECore.getWorkspace().removeResourceChangeListener(this);
    super.removeListeners();
}
```

Notice how the "bundleRootChangedListener" is never removed again? Which, unsurprisingly, is a leak...
image

laeubi:
> Generally, I get the impression that neither JDT nore PDE were ever designed for a use-case where one spawns and destroys an Equinox framework over and over again. Or to rephrase: The plugins don't clean up properly after themselves, after they have been shut down.

Yes this is sadly an outstanding issue and requires us to restart Eclipse after installing updates (among other things not handling dynamics very well).

> As an example, the [WorkspaceModelManager](https://github.com/eclipse-pde/eclipse.pde/blob/master/ui/org.eclipse.pde.core/src/org/eclipse/pde/internal/core/WorkspaceModelManager.java)

Good catch, beside that it seems ResourcePlugin also does not free dangling references on shutdown, it should probably clear all remaining listener it is holding references.

PR #4907, opened 2025-03-24
ApiAnalysisMojo: Remove <enhanceLogs> + add release notes for new way

sratz:


laeubi (approved):

laeubi:
The parameter should be deprecated in 4.x line, this is only for 5.x

PR #4909, opened 2025-03-25
EclipseProject meta model: Support custom variables in linked resources

### LINKED ISSUE #3820: Parsing custom variables in linked project resources
### Current Behavior
The `ProjectParser` is able to resolve linked resources and also supports some pre-defined variable names supported by the Eclipse platform (such as `PROJECT_LOC` and `PARENT-X-PROJECT_LOC`). It does, however, not yet support custom variables that can be defined in the `variablesList` of a project description file to share paths between multiple linked resources. When using custom variables in the location URIs of linked resources, they are considered as ordinary path segments. 

### Example
The binary fragments in SWT define [a variable for the path to the host project](https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/binaries/org.eclipse.swt.win32.win32.x86_64/.project):
```
<variableList>
	<variable>
		<name>SWT_HOST_PLUGIN</name>
		<value>$%7BPARENT-2-PROJECT_LOC%7D/bundles/org.eclipse.swt</value>
	</variable>
</variableList>
```
This is then used in locationURIs for linked resources, such as:
```
<link>
	<name>Eclipse SWT WebKit</name>
	<type>2</type>
	<locationURI>SWT_HOST_PLUGIN/Eclipse%20SWT%20WebKit</locationURI>
</link>
```
In this case, Tycho considers `SWT_HOST_PLUGIN` as an ordinary path segment as it is not aware of custom variables.

### Expected Behavior
Tycho should resolve custom variables just like the project description resolver of the Eclipse platform does. 

I do not consider this an urgent topic, as those variables are only a means to avoid repetitions. Still it would by nice to have in order to be consistent with the capabilities of project description files in the Eclipse platform.

sratz:
* Support resolving of custom variables.

* Refactor code.

* Handle certain (invalid) edge cases.

* `<link>` elements may either contain a `<locationUri>` with variable expansion or a literal `<location>` with an absolute path. Also add support for the latter.

* Add unit tests.

Resolves #3820.

### DIFF
--- a/tycho-metadata-model/src/main/java/org/eclipse/tycho/model/project/ProjectParser.java
+++ b/tycho-metadata-model/src/main/java/org/eclipse/tycho/model/project/ProjectParser.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2023 Christoph Läubrich and others.
+ * Copyright (c) 2023, 2025 Christoph Läubrich and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -9,6 +9,7 @@
  *
  * Contributors:
  *    Christoph Läubrich - initial API and implementation
+ *    SAP SE - support custom variables
  *******************************************************************************/
 package org.eclipse.tycho.model.project;
 
@@ -21,10 +22,8 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
-import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
@@ -42,6 +41,8 @@ import org.xml.sax.SAXException;
 
 class ProjectParser {
 
+    private static final String PROJECT_LOC = "PROJECT_LOC";
+    private static final Pattern DOLLAR_VARIABLE_PATTERN = Pattern.compile("\\$\\{([^\\$]+)\\}");
     private static final Pattern PARENT_PROJECT_PATTERN = Pattern.compile("PARENT-(\\d+)-PROJECT_LOC");
 
     public static EclipseProject parse(Path path) throws IOException {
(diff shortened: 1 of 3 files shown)


tycho-metadata-model/src/main/java/org/eclipse/tycho/model/project/ProjectParser.java
```diff
@@ -173,34 +162,144 @@ public Collection<ProjectVariable> getVariables() {
...
+            }
+        } else {
+            return null;
+        }
+        if (link.type() == LinkDescription.FILE) {
+            //the path must actually match each others as it is a file!
+            if (path.getNameCount() > 1) {
+                return null;
+            }
+        }
```
HannesWell, line 198: @sratz or anybody else, can you tell what this exactly aims at?
As far as I can tell this prevents for example individually linked preference files in a projects `.settings` folder to be considered by the tycho-compiler plugin.
E.g if one has this section in a .projects file:
```
	<linkedResources>
		<link>
			<name>.settings/org.eclipse.core.resources.prefs</name>
			<type>1</type>
			<locationURI>PARENT-2-PROJECT_LOC/.settings/org.eclipse.core.resources.prefs</locationURI>
		</link>
		<link>
			<name>.settings/org.eclipse.jdt.core.prefs</name>
			<type>1</type>
			<locationURI>PARENT-2-PROJECT_LOC/.settings/org.eclipse.jdt.core21.prefs</locationURI>
		</link>
		<link>
			<name>.settings/org.eclipse.pde.ds.annotations.prefs</name>
			<type>1</type>
			<locationURI>PARENT-2-PROJECT_LOC/.settings/org.eclipse.pde.ds.annotations.prefs</locationURI>
		</link>
	</linkedResources>
```

laeubi: @HannesWell I would suggest to enhance the test with your use-case and then possibly adapt the code as needed. I don't think the idea is to forbid anything that is valid in Eclipse here.

HannesWell: Yes, done in
- https://github.com/eclipse-tycho/tycho/pull/5780

PR #4910, opened 2025-03-25
Apianalysis fixes

sratz:
- Do not add unknown line/column numbers
- Resolve path to linked resources

### DIFF
--- a/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisMojo.java
+++ b/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisMojo.java
@@ -238,7 +238,7 @@ public class ApiAnalysisMojo extends AbstractMojo {
 					new ApiAnalysisXmlGenerator(
 							projectManager.getTychoProject(project).get()
 									.getArtifactKey(DefaultReactorProject.adapt(project)).getId(),
-							analysisResult, project.getBasedir().toPath(), report.toPath()).writeReport();
+							analysisResult, report.toPath()).writeReport();
 				} catch (DOMException | IOException | ParserConfigurationException | TransformerException e) {
 					throw new MojoExecutionException(e);
 				}
--- a/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisResult.java
+++ b/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysisResult.java
@@ -23,7 +23,7 @@ import org.eclipse.pde.api.tools.internal.provisional.problems.IApiProblem;
 
 public class ApiAnalysisResult implements Serializable {
 
-	private List<IApiProblem> problems = new ArrayList<>();
+	private List<ApiProblemDTO> problems = new ArrayList<>();
 	private List<ResolverError> resolveError = new ArrayList<>();
 	private String version;
 
@@ -31,7 +31,7 @@ public class ApiAnalysisResult implements Serializable {
 		this.version = version;
 	}
 
-	public Stream<IApiProblem> problems() {
+	public Stream<ApiProblemDTO> problems() {
 		return problems.stream();
 	}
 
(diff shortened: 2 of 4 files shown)


laeubi:
@sratz I'm wondering if we should resolve the path already in [ApiAnalysis](https://github.com/eclipse-tycho/tycho/blob/main/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysis.java) as this is running inside Eclipse itself we should have al features there to resolve an absolute system path.

sratz:
> @sratz I'm wondering if we should resolve the path already in [ApiAnalysis](https://github.com/eclipse-tycho/tycho/blob/main/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiAnalysis.java) as this is running inside Eclipse itself we should have al features there to resolve an absolute system path.

Hmm, interesting point.

`IApiProblem` does not have setters, so we'd have to do some post-processing and wrap those somehow - or add some additional API in side the Eclipse layer.

I think using this approach of using the improved `org.eclipse.tycho.model.project.EclipseProject.getFile(String)` is fine for now. it should have the same performance and this also gives the variable resolution logic some real-world usage and testing :)

laeubi:
> `IApiProblem` does not have setters, so we'd have to do some post-processing and wrap those somehow - or add some additional API in side the Eclipse layer.

There is already [ApiProblemDTO](https://github.com/eclipse-tycho/tycho/blob/main/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiProblemDTO.java) that handles the case of transferring the object from one side to the other.

There we currently pass the path 1:1 but it also has access to the `IProject`:
https://github.com/eclipse-tycho/tycho/blob/1f8aeaccf8b9f49f54a7a9f04872b8e3d91a8335/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiProblemDTO.java#L44

sratz:
> > `IApiProblem` does not have setters, so we'd have to do some post-processing and wrap those somehow - or add some additional API in side the Eclipse layer.
> 
> There is already [ApiProblemDTO](https://github.com/eclipse-tycho/tycho/blob/main/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiProblemDTO.java) that handles the case of transferring the object from one side to the other.
> 
> There we currently pass the path 1:1 but it also has access to the `IProject`:
> 
> https://github.com/eclipse-tycho/tycho/blob/1f8aeaccf8b9f49f54a7a9f04872b8e3d91a8335/tycho-apitools-plugin/src/main/java/org/eclipse/tycho/apitools/ApiProblemDTO.java#L44


Ah, I see.

Yes, this works as well:

```java
public ApiProblemDTO(IApiProblem problem, IProject project) {
	if (path != null) {
		IPath absolutePath = project.getFile(new Path(problem.getResourcePath())).getLocation();
		if (absolutePath != null) {
			path = absolutePath.toOSString();
		}
	}
	resourcePath = path;

```

But I don't really like that `ApiProblemDTO` implements `IApiProblem` again, and violates its contract, which clearly states that

```java
	/**
	 * Returns the project relative path to the resource this problem was found
	 * in, or <code>null</code> if none.
	 *
	 * @return the project relative path to the resource the problem was found
	 *         in, or null if none
	 */
	public String getResourcePath();
```
returns a project **relative** path.

Do you think it would make sense to just remove the `implements IApiProblem` from `ApiProblemDTO` and use that class inside the Tycho layer directly?

sratz:
> `ApiProblemDTO` is really ever only used internally in Tycho so it is not a big problem and we can do whatever we like to, technically I think it must not implement `IApiProblem` it was more to ensure that we implement as much as possible of the data that needs to be transferred.

I implemented the suggestion and added a comment to the `ApiProblemDTO.getResourcePath()` that we deviate in this aspect.

sratz:
> > `ApiProblemDTO` is really ever only used internally in Tycho so it is not a big problem and we can do whatever we like to, technically I think it must not implement `IApiProblem` it was more to ensure that we implement as much as possible of the data that needs to be transferred.
> 
> I implemented the suggestion and added a comment to the `ApiProblemDTO.getResourcePath()` that we deviate in this aspect.

Ok, one more iteration: Adding a separate method for the absolute path to `ApiProblemDTO`, keeping the `implements IApiProblem`, using `ApiProblemDTO` class internally.

laeubi (approved):
Looks good to me, you might want to squash commits into one and merge as you find suitable!

PR #4951, opened 2025-04-24
Fix case where a non jar dependency is in the chain for wrapping

laeubi:
Currently when there is a non jar in the dependency chain it fails to resolve the target because it can not read as a jar.

Needs to be backported to m2e as well!


laeubi:
Needs also be fixed in m2e!

laeubi:
The fix for this issue was gently sponsored by [Katalon](https://katalon.com/).

PR #4955, opened 2025-05-01
Build Jenkins with Java 21

laeubi:


laeubi:
There is only one test that needs explicit adjustment when upgrading to Java 21 (because `-release` and a `-add-export` is not valid), the rest is also extracted in dedicated PRs, so once these are merged this should be rebased and Jenkins should succeed with Java 21.

PR #4960, opened 2025-05-02
Update Xtend maven plugin to support Java 21

laeubi:
Currently used Xtend version do not support Java 21

See
- https://github.com/eclipse-tycho/tycho/pull/4955


merks:
Thanks for proactively addressing this. I know people who would be very grateful.

PR #4970, opened 2025-05-07
Add artifact to feature dependendencies

MrIngelborn:
Fixes issue resolving version for included features that are pom dependencies and not included in the current reactor.

## Issue description
In `tycho-packaging-plugin` when not using the default timestamp provider `BuildQualifierAggregatorMojo` creates a dependency walker to resolve the build qualifier of all dependencies. 

When visiting a feature with an included feature that is a maven dependency, declared in the pom.xml file, but but present in the current reactor, `AbstractArtifactDependencyWalker#traverseFeature` cannot find the included artifact.

This is caused by `P2ResolverImpl#addUnit` not adding the artifact due to no artifact being linked to the `InstallableUnit`.

The included bundles are working without any issues.

## Solution
Comparing the execution of `FeatureDependencyAction` and `BundleAction` I found that `BundleAction` is setting the artifacts explicitly during the IU creation.

Applying this strategy to `FeatureDependencyAction` solved the issue.

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/actions/AbstractDependenciesAction.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/actions/AbstractDependenciesAction.java
@@ -19,6 +19,7 @@ import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.equinox.internal.p2.metadata.InstallableUnit;
+import org.eclipse.equinox.p2.metadata.IArtifactKey;
 import org.eclipse.equinox.p2.metadata.IInstallableUnit;
 import org.eclipse.equinox.p2.metadata.IProvidedCapability;
 import org.eclipse.equinox.p2.metadata.IRequirement;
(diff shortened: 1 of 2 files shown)


laeubi:
@MrIngelborn thanks for the contribution, you need to sign the ECA so we can accept your contribution. Also can you 
provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the issue so we are not running in the problem later again?

MrIngelborn:
Thanks for you r quick reply. I have been unable to reproduce the issue in an integration test. Will try again later, but maybe it's just our setup that is badly configured somehow.

laeubi:
@MrIngelborn we will have a Tycho 4 release son, can you please sign the ECA so we can include this?

MrIngelborn:
Done!

laeubi:
This fails the aggregator build with:

> Caused by: java.lang.IllegalStateException: Inconsistent artifact with key eclipse-feature:org.eclipse.equinox.p2.user.ui.source:2.4.2800.v20250513-1051 classifier is different

MrIngelborn:
Seems this change caused other problems at our site. And I wasn't able to reproduce the issue in a IT. Perhaps a misconfiguration at our site. I need to find another solution

PR #4973, opened 2025-05-10
Update to 2025-03

akurtakov:

PR #4974, opened 2025-05-12
Use Java 21 EE as default

akurtakov:
As Eclipse Platform 2025-03 requires it, default should be changed accordingly in order to keep working with it by default.


laeubi:
I think basically having a fixed default value here is bad anyways. We should remove the default and if the parameter is null use the one derived from the current runtime version.

akurtakov:
This is a spin off of https://github.com/eclipse-tycho/tycho/pull/4973 . I wouldn't have the time to dig into it more thus I would prefer to push it that way and if/when someone has the time to improve more they can remove it.

laeubi:
You can look at `copied.org.apache.maven.plugin.AbstractCompilerMojo.DEFAULT_SOURCE_VERSION` for example there we already derive it by default from the running environment.

akurtakov:
Opened task to not forget it https://github.com/eclipse-tycho/tycho/issues/4975 .

PR #4976, opened 2025-05-13
Add flag to skip execution of tycho-source-plugin

HannesWell:
This is similar to the [maven-source-plugin#skipSource](https://maven.apache.org/plugins/maven-source-plugin/jar-mojo.html#skipSource).


laeubi (approved):
Looks good!

Issue #2482, opened 2023-06-11
Performance problem in classpath validation


Bananeweizen:
Just noticed that for many of my plugins the classpath validation takes longer than the actual compilation, as can be seen in the time measurements per goal in maven debug mode (where times are wallclock time):
![image](https://github.com/eclipse-tycho/tycho/assets/406876/364f6af6-abcf-40ef-9b5e-591497c70431)

Running the same build under a profiler, it confirms that (with the times being a mixture of CPU time and wallclock time):
![image](https://github.com/eclipse-tycho/tycho/assets/406876/62d3d108-f0ad-4e5c-8b34-065575106bbd)

Be aware this is Windows and a virus scanner is enabled, so you may see much smaller numbers without these constraints. The [location.exists() check](https://github.com/eclipse-tycho/tycho/blob/67e87d873a8bf264d3cf40ece7fda28a14589439/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/DefaultArtifactDescriptor.java#L61) was introduced in 67e87d873a8bf264d3cf40ece7fda28a14589439. While that check is reasonable by itself (with the goal of re-trying failed downloads), we should try to have that file existence check at some other place (like moving it to all places where the locationResolver is used, and _not_ setting location if the file is missing afterwards).

laeubi:
@Bananeweizen there was some performance improvements here:

- https://github.com/eclipse-tycho/tycho/pull/4995

can you check if it already makes the situation better for you?

laeubi:
Closing as no further feedback was given, if there are still problems please let me know.

PR #4991, opened 2025-05-21
Bug: 2482 - Fixed performance issue with respect to classpath resolution

hendricc:
This PR relates to bug #2482 

Improves the efficiency of classpath lookups by fetching artifacts on demand (instead of always). In larger projects this bug could consume more than half of the entire build time.

With this fix applied, our largest project went from **55** to **25** minutes build time (without unit test execution).

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/targetplatform/ArtifactCollection.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/targetplatform/ArtifactCollection.java
@@ -316,10 +316,10 @@ public class ArtifactCollection {
      *         default artifact)
      */
     public Map<String, ArtifactDescriptor> getArtifact(File location) {
-        artifacts.values().forEach(artifact -> artifact.getLocation(true));
         File normalized = normalizeLocation(location);
         Map<String, ArtifactDescriptor> map = artifactsWithKnownLocation.get(normalized);
         if (map == null) {
+            artifacts.values().forEach(artifact -> artifact.getLocation(true));
             LinkedHashMap<String, ArtifactDescriptor> hashMap = new LinkedHashMap<>();
             for (ArtifactDescriptor descriptor : artifacts.values()) {
                 ReactorProject mavenProject = descriptor.getMavenProject();


hendricc:
yeah - I will set this up today

If there are more sophisticated solutions (which I might not be able to contribute for), the better ;-)
You can consider this as the bare minimum solution that lowers the build time significantly.

A potential backport to 4.0.x would be appreciated if this or any other fix for #2482 gets approved.

laeubi:
I have checked this now and even though it does not breaks a test there is some unpleasant side-effect here that might break if we change the code here.

When artifacts are fetched first, they might already initialize the map with some values, so further down the call to get the map might not return `null` anymore. If we move the code down, it means that probably the map was initialized and now we overwrite it with the reactor map only.

The good thing here is that the method itself is only called from one place (`OsgiBundleProject.getArtifact(DependencyArtifacts, File, String)`) so it is quite clear what callers might expect here.

laeubi:
I have now added a first cleanup here that simply removes all unused methods that reference the cache:

- https://github.com/eclipse-tycho/tycho/pull/4992

now we only have two places where it is used, once in the `getArtifact` and then in the `registerArtifactLocation` call, this should make it more clear and safe to change things.

laeubi:
Further looking into the callstack I get the impression that this mainly works for a simple reason:

When this method is called, the map is simply never null. The reason is simple: How would one gather the location for the lookup if not already have required to fetch its location before? So somewhere in the chain wee loose the mapping between a location fetched and an artifact key and need to recover it here.

@hendricc as you seem to debugged it already for your case did you ever hit a case where the map actually was `null`?

laeubi:
@hendricc I have now created a "light" variant for Tycho 4.x here:
- https://github.com/eclipse-tycho/tycho/pull/4995

For Tycho 5.x I go for a more sophisticated approach removing the need for this code path at all:
- https://github.com/eclipse-tycho/tycho/pull/4993

I think we can close this and proceed in the other PRs, thank you for investigation and proposing an initial solution. 

Can you please verify that these fixes the issue you observed?

hendricc:
Hi @laeubi, wow, lot's of input

So I debugged the ArtifactCollection.getArtifact(...) method on our largest project (with hundreds of reactor projects and even more third party dependencies) and to my surprise the map 'artifactsWithKnownLocation' always returned a non-null value for the given key, therefore the if-condition was never met. This was observable with the original 4.0.12 and the patched one. 
This implies that the 'artifactsWithKnownLocation' map was only modified (in terms of 'put') within the method 'registerArtifactLocation'.

Therefore your fix for tycho-4.0.x will work like my original one.
I couldn't test the 5.x solution so far, but will try it.

In general reactor projects and dependencies shouldn't be fetched all the time since these files do not change during the build once they are available.
Thank you for all the input and the provided PR's

PR #5048, opened 2025-06-10
Issue 5047 reproducer

jan-pesta:


akurtakov:
This PR looks totally wrong. Please submit only what you have changed/added.

jan-pesta:
I just reworked it

jan-pesta:
those test failures indikates the issue with versioning. What next is needed here ?

laeubi:
> those test failures indikates the issue with versioning

Thanks for the testcase!

> What next is needed here ?

Someone needs time to pick this up and provide a fix for it.

jan-pesta:
I can create fix for it.
I would like to open discussion because this code has been added with #611.
I am not aware why this part of code has been added, because I did not find a reason for such behavior.

I can imagine to keep it backward compatible to restrict fix only if new configuration is specified. e.q.

```
    @Parameter(name = "stripQualifierIfDuplicates", defaultValue = "true", property = "tycho.buildqualifier.stripduplicatedqualifier")
    protected Boolean stripQualifierIfDuplicates;
```

laeubi:
My recommendation would be to add a proposed fix to this PR and then we will see if anything fails and then one might understand it better or even notice that it was unnecessary or ... as we currently preparing a new major release also no immediate need for staying backward compatible.

akurtakov:
This slipped through the cracks. Please sign your ECA and rebase on master.

akurtakov:
@laeubi As you have worked on this one, would you please review it?

laeubi (changes requested):
@jan-pesta sorry for the delay, but I think this is also a bit cause by the PR having no useful title and description. So if you are still interested I would suggest to change that.

Next this introduces a new system property to skip the code in question what is bad, we don't want to have a system property for each case in the code, it should work in general, so if you suspect a bug we should fix the cause and not just skip the (possibly buggy) code path.

Further this seems to be an interaction with a plain maven bundle build correct? I think its worth to make this more clear and try to describe the setup and interaction a bit more in detail, currently it is quite hard to understand the rationale here.

akurtakov:
If there is no interest in this anymore I plan to close it next week.

PR #5138, opened 2025-06-12
Deprecate source features mojo and mark for removal

laeubi:
See corresponding change in
- https://github.com/eclipse-tycho/tycho/pull/5137

### DIFF
--- a/REMOVAL.MD
+++ b/REMOVAL.MD
@@ -1,18 +1,6 @@
-## List of removals for next major version of tycho
+# List of removals for next major version of Tycho
 
-# the plugin 'tycho-extras/tycho-source-feature-plugin' should be removed on next major release
+## the mojo 'tycho-source-plugin:feature-source' should be removed on next major release
 
-replaced by core tycho-source
-
-# org.eclipse.tycho.PackagingType.TYPE_ECLIPSE_REPOSITORY and org.eclipse.tycho.PackagingType.TYPE_ECLIPSE_UPDATE_SITE and corresponding mojos
-
-The packagetypes ECLIPSE_REPOSITORY+ECLIPSE_UPDATE_SITE are deprecated for a long time and should be removed, corresponding mojos and code should be dropped also
-
-# drop "fixSWT"
-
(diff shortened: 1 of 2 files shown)


mickaelistria:
Does this end up with a visible notification in the build log?
Also, I would IMO be good if the mojo could also log as a warning a link to https://github.com/eclipse-tycho/tycho/issues/3458 which explains the rationale for the removal and the possible alternatives so users can understand and accept things more easily.

laeubi:
> Does this end up with a visible notification in the build log?

As far as I know it will print an warning if you use deprecated mojos, but we can of course add a explicit warning as well.

laeubi:
@akurtakov @merks Do we have any general concerns about this for platform in general or can we merge this (with or without and additional message) for Tycho 5 to get a warning in the build?

merks:
I did send this email and no one came here to complain:

https://www.eclipse.org/lists/simrel-dev/msg00072.html

I still have a general concerns that removing things is disruptive.  It's already cost me the better part of two days to deal with the impact of removal in EPP.  I still have a bunch of projects that I'd needed to fix (which can be done later).  And we're not actually done yet with the Platform:

image

So maybe its best we finish with the Platform such that it's possible to build the platform with the latest Tycho.

akurtakov:
My preferred workflow would be that migration in Platform is done before it is deprecated. 
On the other side, as soon as there is plan for 5.0 release deprecation should go in so it can be actually removed in v6.

laeubi:
The plan is to release Tycho 5 at end of M3 so we can go into final phase with a released version instead of snapshots.

akurtakov:
> The plan is to release Tycho 5 at end of M3 so we can go into final phase with a released version instead of snapshots.

In that case I would say land the deprecation in so there is some time to see it in action.

laeubi:
I now added a warning that will point to the REMOVAL.MD if we want any further links or explanation I would suggest to enhance that file after this is merged here, that way we can adapt the description even after a release.

PR #5197, opened 2025-07-18
Parse P2 advice-file with expanded project version if available

HannesWell:
Using the expanded version ensures that the $version$ parameter is substituted with the that expanded version and does not end with literally `.qualifier`.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetPlatformFactoryImpl.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/p2resolver/TargetPlatformFactoryImpl.java
@@ -64,6 +64,7 @@ import org.eclipse.equinox.p2.repository.artifact.IArtifactRepository;
 import org.eclipse.equinox.p2.repository.artifact.IArtifactRepositoryManager;
 import org.eclipse.equinox.p2.repository.metadata.IMetadataRepository;
 import org.eclipse.equinox.p2.repository.metadata.IMetadataRepositoryManager;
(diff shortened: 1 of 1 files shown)


laeubi:
I think I tried similar in the past but this introduced some kind of regressions on other place, maybe @merks remembers? At least we should be a bit careful and watch out the results.

HannesWell:
> I think I tried similar in the past but this introduced some kind of regressions on other place, maybe @merks remembers?

I would expect this to be relatively save as it falls back to the project version as before, if the expanded version is not available (e.g. during TP resolution). But of course the devil is in the details :)

Also it seems that for bundles the `$version$` is already substituted, as it is for example used in
https://github.com/eclipse-platform/eclipse.platform.swt/blob/1a1f0c22b89d8c99ff9ad58c2bbcf82147852e5a/bundles/org.eclipse.swt/META-INF/p2.inf#L10

and the generated P2 repo content contains the desired expanded version.

But when using it with a `feature.xml` the version in the metadata ends with `.qualifier` at the moment.
So maybe the artifact-key is updated for bundle-projects?

> At least we should be a bit careful and watch out the results.

So should this be submitted if all tests (and the SDK build) pass or do you want to check something else?

laeubi:
> So should this be submitted if all tests (and the SDK build) pass or do you want to check something else?

we can try it but it would also be interesting to see why it works for bundles and not for features (where features are maybe a bit uncommon). The problem is more when some metadata slips into the build that is inconsistent but I also have improved that on some place so maybe no longer an issue.

PR #5208, opened 2025-07-26
Make a context project optional for EclipseRunMojo

HannesWell:
This might be useful if one wants to run Eclipse from Maven without a context project.

### DIFF
--- a/tycho-eclipse-plugin/src/main/java/org/eclipse/tycho/eclipserun/EclipseRunMojo.java
+++ b/tycho-eclipse-plugin/src/main/java/org/eclipse/tycho/eclipserun/EclipseRunMojo.java
@@ -67,7 +67,7 @@ import org.eclipse.tycho.p2.target.facade.TargetPlatformFactory;
  * Launch an eclipse process with arbitrary commandline arguments. The eclipse
  * installation is defined by the dependencies to bundles specified.
  */
-@Mojo(name = "eclipse-run", threadSafe = true)
+@Mojo(name = "eclipse-run", threadSafe = true, requiresProject = false)
 public class EclipseRunMojo extends AbstractMojo {
 
 	/**
@@ -105,7 +105,7 @@ public class EclipseRunMojo extends AbstractMojo {
 	@Parameter(defaultValue = "true")
 	private boolean clearWorkspaceBeforeLaunch;
 
-	@Parameter(property = "project")
+	@Parameter(property = "project", readonly = true)
 	private MavenProject project;
 
 	/**
@@ -446,7 +446,7 @@ public class EclipseRunMojo extends AbstractMojo {
 					+ ". Current Java runtime will be used");
 		}
 		cli.setJvmExecutable(executable);
-		cli.setWorkingDirectory(project.getBasedir());
+		cli.setWorkingDirectory(project != null ? project.getBasedir() : new File("").getAbsoluteFile());
 
 		cli.addVMArguments(splitArgLine(argLine));
 		if (jvmArgs != null) {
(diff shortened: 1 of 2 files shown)


tycho-eclipse-plugin/src/main/java/org/eclipse/tycho/eclipserun/EclipseRunMojo.java
```diff
@@ -105,7 +105,7 @@ public class EclipseRunMojo extends AbstractMojo {
 	@Parameter(defaultValue = "true")
 	private boolean clearWorkspaceBeforeLaunch;
 
-	@Parameter(property = "project")
+	@Parameter(property = "project", readonly = true)
```
laeubi:
```suggestion
	@Component
```

Is actually the Maven3 way even though it is claimed to be deperated.

The next (JSR330) way should also work


```suggestion
	@Inject
```

tycho-eclipse-plugin/src/main/java/org/eclipse/tycho/eclipserun/EclipseRunMojo.java
```diff
@@ -446,7 +446,7 @@ public LaunchConfiguration createCommandLine(EquinoxInstallation runtime) throws
 					+ ". Current Java runtime will be used");
 		}
 		cli.setJvmExecutable(executable);
-		cli.setWorkingDirectory(project.getBasedir());
+		cli.setWorkingDirectory(project != null ? project.getBasedir() : new File("").getAbsoluteFile());
```
laeubi: If I remember correctly maven always create an empty project for you when you execute it from within a directory without a pom.

Beside that a temp directory might be better to use?

tycho-eclipse-plugin/src/main/java/org/eclipse/tycho/eclipserun/EclipseRunMojo.java
HannesWell: The latter looks like a good way forward, but there are many occasion of this pattern and I think it's better to fix them in a separate PR. Revert it for this one. And do the migration in:
- https://github.com/eclipse-tycho/tycho/pull/5224

tycho-eclipse-plugin/src/main/java/org/eclipse/tycho/eclipserun/EclipseRunMojo.java
HannesWell: > If I remember correctly maven always create an empty project for you when you execute it from within a directory without a pom.

I don't know that, but I assume the empty project, will then have the current working directory as directory? So basically the same result.

> Beside that a temp directory might be better to use?

The idea was that the current working directory where the direct CLI invocation is launched is probably also interesting was working directory of the eclipse run (just 'forwarding' the shell's workding directory to the Eclipse run).

HannesWell:
I did now make some experiments and there is indeed an empty project injected if no pom is available, but its `basedir` is null. So I adjusted that.
But this experiment also showed that it doesn't work that simply as one has to define a (p2-) repository to resolve the artifacts building the runtime.
And as a repository is a [complex type](https://tycho.eclipseprojects.io/doc/4.0.12/tycho-eclipse-plugin/eclipse-run-mojo.html#repositories), I'm not aware of a way to define it on the CLI, or do you know one?

laeubi:
What cam into my mind is to define an additional parameter (e.g. `repo-url`) that can be used as a fallback for CLI invocation, we can then use an ID as `eclipse-run-cli` for example so users can still configure that.

PR #5209, opened 2025-07-29
Add mojo to modify P2-repository metadata to tycho-p2-repository-plugin

HannesWell:
This adds a mojo to modify the metadata (currently only name and properties) of a P2-repository as suggested in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3204#issuecomment-3121650948


tycho-core/src/main/java/org/eclipse/tycho/p2/tools/RepositoryModifier.java
```diff
@@ -0,0 +1,182 @@
...
+import org.eclipse.tycho.p2tools.copiedfromp2.RepositoryDescriptor;
+
+@Named
+@Singleton
+public class RepositoryModifier {
+
+    public record ModifiedRepositoryDescriptor(RepositoryDescriptor sourceRepository, Path destinationLocation,
+            boolean compress, boolean xzCompress, boolean keepNonXzIndexFiles) {
+
+        public static ModifiedRepositoryDescriptor create(URI repositoryLocation, Path destinationLocation,
```
laeubi: Can this method be moved to the upper type `RepositoryModifier`?

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Location of the <em>source</em> repository to modify.<br>
+     * May be the {@code Path} of a local file or {@code URL}/{@code URI} to a remote
+     * repository.<br>
+     * If this is set to a remote URL the {@link #outputLocation} must be set to a local path as
+     * remote locations cannot be written.
+     */
+    @Parameter(property = "p2.repository.location")
+    private URI repositoryLocation;
```
laeubi: We should use a `Repository` here as we do at other places as well, this is a bit more verbose but allows to specify an ID what is important for proxy, authentication and mirroring of repositories.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
```diff
@@ -0,0 +1,159 @@
...
+    private Path outputLocation;
+
+    /**
+     * The kind of modified repository: Either {@code artifact} or {@code metadata}<br>
+     * If {@code artifact} is specified, only the {@code artifact.xml/jar/xml.xz} files are
+     * modified, if {@code metadata} only {@code content.xml/jar/xml.xz}. If blank (the default)
+     * both are modified.
+     */
+    @Parameter(property = "p2.repository.kind", defaultValue = "")
+    private String repositoryKind;
```
laeubi: can we then use an enum here if only a limited set of values is allowed?

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
```diff
@@ -0,0 +1,159 @@
...
+
+    /** Optional repository name to set. An existing name is overwritten. */
+    @Parameter
```
laeubi: Do we want a user property here?
Same for the below parameters.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
```diff
@@ -0,0 +1,159 @@
...
+     * If {@link #xzCompress} is <code>true</code>, whether jar or xml index files should be kept in
+     * addition to XZ-compressed index files. This fallback provides backwards compatibility for
+     * pre-Mars p2 clients which cannot read XZ-compressed index files.
+     * </p>
+     */
+    @Parameter(defaultValue = "true")
+    private boolean keepNonXzIndexFiles;
+
+    @Inject
+    private RepositoryModifier modifier;
```
laeubi: I must confess the name looks odd here, a modifier is usually something that modifies the behavior (e.g. a flag), what about RepositoryUpdater?

tycho-core/src/main/java/org/eclipse/tycho/p2/tools/RepositoryModifier.java
```diff
@@ -0,0 +1,182 @@
...
+import org.eclipse.equinox.p2.repository.artifact.ArtifactKeyQuery;
+import org.eclipse.equinox.p2.repository.artifact.IArtifactRepository;
+import org.eclipse.equinox.p2.repository.artifact.IArtifactRepositoryManager;
+import org.eclipse.equinox.p2.repository.metadata.IMetadataRepository;
+import org.eclipse.equinox.p2.repository.metadata.IMetadataRepositoryManager;
+import org.eclipse.tycho.p2tools.copiedfromp2.RepositoryDescriptor;
+
+@Named
+@Singleton
+public class RepositoryModifier {
```
laeubi: From past experience the JSR330 does not work well with classes, but requires an interface.

Also I wonder if we maybe should reuse the `RepositoryManager` from the P2 plugin here for loading the repositories instead.

tycho-core/src/main/java/org/eclipse/tycho/p2/tools/RepositoryModifier.java
HannesWell: Sure. I considered it as a factory of the nested type, but it can also be moved to the top-level class.
In general I encapsulated this logic as I will re-use it for a `modify-composite` mojo I'm working on as well and want to contribute after this.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: In this case it modifies the content of a P2 repository. It also does not really update a repository as it does general modifications, which might add something new but could also just remove something.

What about `RepositoryTransformer`, `RepositoryMutator` or `RepositoryProcessor`?
I'm currently leaning towards the first one. But in the end it's internal and can be changed at any time.

tycho-core/src/main/java/org/eclipse/tycho/p2/tools/RepositoryModifier.java
HannesWell: > From past experience the JSR330 does not work well with classes, but requires an interface.

I can only say, that in the use-case here it works as expected.
But what are you suggesting specifically as an alternative?

> Also I wonder if we maybe should reuse the `RepositoryManager` from the P2 plugin here for loading the repositories instead.

I don't understand exactly what you mean by that?

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: I can add them, but I assume that one cannot configure the mojo via a property for a `Map` parameter, can one?

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: I assume you mean `org.eclipse.tycho.plugins.p2.Repository`?
But this would mean that a dependency from `tycho-p2-repository-plugin` to `tycho-p2-plugin` is necessary.

> allows to specify an ID what is important for proxy, authentication and mirroring of repositories.

That makes sense, but where exactly is in the presented code the id is then considered?
Also consider that this is intended to work with local repositories (with `file` URI/URL) too.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: We can, but it makes the allowed values much more strict. Before every string starting with `A` (lower or upper-case) was considered an artifact repository, the same for `M` and metadata-repositories and blank was a repository for both.
But probably that's a good thing.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
laeubi: It works for lists at least:
https://maven.apache.org/guides/mini/guide-configuring-plugins.html#Mapping_Collections_and_Arrays

> Since Maven 3.3.9 ([MNG-5440](https://issues.apache.org/jira/browse/MNG-5440)), you can list individual items alternatively as comma-separated list in the XML value of animals directly. This approach is also used if configuring collection/array parameters via command line

So for maps one would need to add support, or we allow another parameter as a list of `key=value` pairs.

HannesWell: A little bit further down this guides it says:
> In contrast to value objects and collections/arrays there is no string coercion defined for maps, i.e. you cannot give parameters of that type via CLI argument.

> or we allow another parameter as a list of key=value pairs.

That would be possible, but I suggest to save this part for now as I assume that one will provide specific executions for intended tasks in the `pluginManagement` section anyways in a pom (and we need a pom at least to define the tycho-version or we have to hard-code it into a using pipeline).

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
laeubi: > But this would mean that a dependency from tycho-p2-repository-plugin to tycho-p2-plugin is necessary.

Yes this is usually no problem, just the `tycho-p2-plugin` should not depend on anything else (e.g. Tycho core). My long term goal would be to move all functionality related to P2 into that plugin so we have:

- tycho-spi/api generic data definition and sharing of types
- tycho-core - Tycho itself without dependency to p2
- tycho-p2(-plugin) everything related to handling p2 stuff

> That makes sense, but where exactly is in the presented code the id is then considered?

For example `org.eclipse.tycho.eclipsebuild.AbstractEclipseBuildMojo.baselines` and `org.eclipse.tycho.eclipsebuild.AbstractEclipseBuildMojo.eclipseRepository`

If you look at `P2RepositoryManager` and search for references to these methods you should find other places as well (some still use `MavenRepositoryLocation` but this is actually interchangeable)


> Also consider that this is intended to work with local repositories (with file URI/URL) too.

You can always specify a `file:/` repo everywhere. We might want to even enhance it to consider plain path in the `P2RepositoryManager`

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
laeubi: The point is more that I would suspect such mojo to be most useful as a CLI invocation but that's not mandatory of course. As an alterative I think one can use PlexusXML as a parameter type then you get the raw XML (or string) and can do whatever format you want but that maybe something for a different task.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
laeubi: As I first read `RepositoryModifier` I though it would be something like `READONLY` or similar flags :-)

We also have many "Helper" named components (e.g. MavenPropertyHelper, ProjectHelper and so on ...) so I would even not mind name it `RepositoryHelper` and moving in into the `p2-maven-plugin` to make it generally available.

tycho-core/src/main/java/org/eclipse/tycho/p2/tools/RepositoryModifier.java
laeubi: > But what are you suggesting specifically as an alternative?

The usual "pattern" with JSR330 is to have an interface (e.g. `RepositoryHelper`) and then a Default Implementation (e.g. `DefaultRepositoryHelper`), this also ensures one can have overrides for components easily.

> I don't understand exactly what you mean by that?

Currently you are making your own use of factories to load repositories. But we already have 

- `org.eclipse.tycho.p2maven.repository.P2RepositoryManager.getArtifactRepository(Repository)`
- `org.eclipse.tycho.p2maven.repository.P2RepositoryManager.getMetadataRepository(Repository)`

as loading repos is sometimes brittle (classloading and handling of mirrors, authentication and logging, you will for example see call to `org.eclipse.tycho.IRepositoryIdManager.addMapping(String, URI)` in that code) I try to route this throug central dedicated places (in this case `P2RepositoryManager`) instead of doing it on multiple places.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: > > But this would mean that a dependency from tycho-p2-repository-plugin to tycho-p2-plugin is necessary.
> 
> Yes this is usually no problem, just the `tycho-p2-plugin` should not depend on anything else (e.g. Tycho core).

But `tycho-p2-plugin` already depends on `tycho-core`. Or do you mean it the other way round?

> For example `org.eclipse.tycho.eclipsebuild.AbstractEclipseBuildMojo.baselines` and `org.eclipse.tycho.eclipsebuild.AbstractEclipseBuildMojo.eclipseRepository`
> 
> If you look at `P2RepositoryManager` and search for references to these methods you should find other places as well (some still use `MavenRepositoryLocation` but this is actually interchangeable)

These methods use `org.apache.maven.model.Repository` respectively `org.eclipse.tycho.MavenRepositoryLocation` and not the `org.eclipse.tycho.plugins.p2.Repository` I mentioned before, so I assume you mean those?
And while the latter resides in `tycho-p2-plugin` the former reside in maven-model respectively `tycho-api`.
On the other hand `P2RepositoryManager` resides in `p2-maven-plugin`.
All suggestion made are from the different artifacts, but I assume all are fine except for the one in `tycho-p2-plugin` as I think it's not ideal to have one Maven-Plugin to depend on another (just for a shared type that's eventually wrapped into `MavenRepositoryLocation` anyways).
And when choosing between `MavenRepositoryLocation` and `org.apache.maven.model.Repository` I think it's better to use the former as a P2 repository is no Maven repository and some of its properties don't exist for P2-repos. Furthermore, the former uses a `URI` type to model the URL instead of a plain string, which captures malformed URIs earlier.

> If you look at P2RepositoryManager and search for references to these methods you should find other places as well (some still use MavenRepositoryLocation but this is actually interchangeable)

Thanks for that hint, that looks promising. I'll look into that.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: > The point is more that I would suspect such mojo to be most useful as a CLI invocation but that's not mandatory of course.

In general yes, but I believe that it's often requires configuration in the pom anyways.

> but that maybe something for a different task.

Yes, lets keep that for a follow-up.
In general I think it would be ideal if Maven would provide a standard way to define a map via CLI. Either the way you described (of course one would have to handle `=` in values (but one could just read them as value up until the next comma, assuming the keys don't contain an equal symbol).
Alternatively one could think of using the property as a prefix and add different key-suffixes in multiple property definitions. E.g. `-DmyProp.key1=value1 -DmyProp.key2=value2`.
But that's something else, too.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: Naming it `RepositoryHelper` sounds too generic too me. But I'll think about it more. Maybe it could also be `P2RepositoryMetadataManager` or just `P2RepositoryMetadata`.

Moving it to `p2-maven-plugin` could fit well.

tycho-core/src/main/java/org/eclipse/tycho/p2/tools/RepositoryModifier.java
HannesWell: > > But what are you suggesting specifically as an alternative?
> 
> The usual "pattern" with JSR330 is to have an interface (e.g. `RepositoryHelper`) and then a Default Implementation (e.g. `DefaultRepositoryHelper`), this also ensures one can have overrides for components easily.

As it currently works as expected, I would like to keep it simple for now and keep it as it is. Splitting this up, can still be done at any time, if there are use-cases for it.

> Currently you are making your own use of factories to load repositories. But we already have
> 
>     * `org.eclipse.tycho.p2maven.repository.P2RepositoryManager.getArtifactRepository(Repository)`
> 
>     * `org.eclipse.tycho.p2maven.repository.P2RepositoryManager.getMetadataRepository(Repository)`
> 
> 
> as loading repos is sometimes brittle (classloading and handling of mirrors, authentication and logging,

Will use that, thanks.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
laeubi: Some things are due to historic reasons and undergone some refactorings so things are maybe not aligned well. From the past we have many "facadetypes" so `org.eclipse.tycho.MavenRepositoryLocation` was a facade for `org.apache.maven.model.Repository` example. In the end it does not matter much as they have similar signatures.

And P2 repositories are actually represented as Maven repositories, as you can haven them in `<repository>` section of the pom and we handle mirroring and authentication for them like maven repos. They just don'T use the `default` layout but the `p2` layout.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
laeubi: I don't think this thing "manages" something, in any case we can later rename it if we find another name more suitable, this is nothing that faces the user visible API.

tycho-core/src/main/java/org/eclipse/tycho/p2/tools/RepositoryModifier.java
laeubi: > As it currently works as expected, I would like to keep it simple for now and keep it as it is.

I can only tell that sometimes it just don't work, e.g. if you try to use the component in a maven mojo what makes it quite confusing and annoying to migrate to JSR330.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: > was a facade for `org.apache.maven.model.Repository`

Ok then let's use this.
But I assume if we use here a complex type, i.e. anything except URI we cannot define the repository as property in the CLI as discussed below, at least not without any further configuration in a pom.xml.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
```diff
@@ -0,0 +1,193 @@
...
+			throw new IllegalStateException(e);
+		}
+	}
+
+	private static void assertNoError(IStatus status, Path location) {
+		if (!status.isOK() && status.matches(IStatus.ERROR)) {
+			throw new IllegalStateException("Failed to create output repository at: " + location + "\n" + status,
+					status.getException());
+		}
+	}
```
HannesWell: This could be moved to the P2RepositoryManager as well, as this might be relevant for other tasks too.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: I now choose `P2RepositoryDataManipulator` as this should also fit for a `modify-composite` (repository) mojo I intend for a follow-up.
But yes, it can be renamed at any time.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
```diff
@@ -0,0 +1,193 @@
...
+				repository.query(ArtifactKeyQuery.ALL_KEYS, null).stream()
+						.map(a -> repository.getArtifactDescriptors(a)).forEach(a -> mirror.addDescriptors(a, null));
+				// Copy (potentially customized) mapping rules
+				if (repository instanceof SimpleArtifactRepository simpleSource
+						&& mirror instanceof SimpleArtifactRepository simpleTarget) {
+					simpleTarget.setRules(simpleSource.getRules()); // Copy artifact mapping rules
+				}
+			}, null);
+			assertNoError(result, targetLocation);
+			// TODO: remove repository from manager again?
```
HannesWell: Should we make sure the source repository is again removed from the manager? or at least a potential (local) output/target repository?
I saw this in multiple occasions and think that it could make sense to keep the manager clean of such modified/worked on repos?

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
```diff
@@ -101,6 +102,12 @@ public IArtifactRepository getCompositeArtifactRepository(Collection<Repository>
 		return new ListCompositeArtifactRepository(childs, agent);
 	}
 
+	public IArtifactRepository createLocalArtifactRepository(Path location, String name, Map<String, String> properties)
+			throws ProvisionException {
+		return getArtifactRepositoryManager().createRepository(location.toUri(), name,
```
laeubi: I won't use the repository manager here but `SimpleArtifactFactory` directly. this will also avoid the need to remove the repos later from the manager because it will never be added.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
laeubi: See my comment above, when we can avoid adding it to the set of managed repos at the first place it would be better than removing them later on.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
laeubi: We already have the [StatusTool](https://github.com/eclipse-tycho/tycho/blob/main/tycho-core/src/main/java/org/eclipse/tycho/core/shared/StatusTool.java) that should better be used for giving good error messages.

HannesWell: Unfortunately this cannot be used here as it resides in `tycho-core` which depends on `p2-maven-plugin`. So it would require a cycle dependency.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
HannesWell: Done.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
HannesWell: Yes. Initially I also thought (and implemented it locally) that even the source repository is removed again, if it's a local repo, a remote repo is IMO still reasonable to cache.
But for the sake of simplicity I have now dropped that part as at least the use case in the platform-releng only involves a remote repo to modify.
If a cached local source repo makes trouble, it can still be enhanced later.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
laeubi: we can move it probably to `tycho-spi` or `tycho-api` module.

HannesWell: Moved it to `tycho-spi `, but still had to add a method that asserts the severity of a status is below a certain level.
But this is now reusable.

laeubi: > Moved it to tycho-spi , but still had to add a method that asserts the severity of a status is below a certain level.

Can we have this as a separate PR to keep this one easier to review? I think that move can then be merged right away (if not tests fail).

HannesWell: Done with:
- https://github.com/eclipse-tycho/tycho/pull/5225

Auto-merge is enabled.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryKind.java
```diff
@@ -0,0 +1,18 @@
+package org.eclipse.tycho.p2maven.repository;
```
laeubi, line 13: Copyright header missing

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
```diff
@@ -0,0 +1,151 @@
...
+ * Modifies metadata of a P2-repository like its name or properties.
+ * <p>
+ * This can be used to customize metadata of a just assembled repository or to modify an already
+ * existing (remote) repository. The modified repository metadata files
+ * ({@code artifact.xml/jar/xml.xz} or {@code content.xml/jar/xml.xz}) are always written to a local
+ * file. For a remote repository a suitable other tool (like SSH) must be used to copy these files
+ * subsequently back to the remote server.
+ * </p>
+ */
+@Mojo(name = "modify-metadata", threadSafe = true)
```
laeubi: Should we better name it


```suggestion
@Mojo(name = "modify-p2-metadata", threadSafe = true)
```

?

tycho-spi/src/main/java/org/eclipse/tycho/helper/StatusTool.java
```diff
@@ -141,6 +144,13 @@ public static Throwable findException(IStatus status) {
         return null;
     }
 
+    public static void assertNoProblemOf(int severity, IStatus status, Supplier<String> message) {
```
laeubi: I'm not sure this method really does fit its purpose.

I would suggest to use something like


```suggestion
    public static boolen checkStatus(int severity, IStatus status)
```

if one really wants to save a one liner check.

The caller should the use 

if (StatusTool.checkStatus(....)) {

} else {
  //print log message and/or throw a suitable exception e.g CoreException or MojoFailureException or ...
}

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
```diff
@@ -0,0 +1,151 @@
...
+        }
+        if (!propertiesToAdd.isEmpty()) {
+            logger.info("  Set properties:");
+            propertiesToAdd.forEach((key, value) -> {
+                logger.info("   " + key + " = " + value);
+            });
+        }
+        try {
+            manipulator.modifyRepositoryMetadata(descriptor, repositoryName, propertiesToRemove, propertiesToAdd);
+        } catch (ProvisionException | IOException | URISyntaxException e) {
```
laeubi: `ProvisionException` should be handled separately here by

```
getLog().error(StatusTool.toLogMessage(e.getStatus()), e);
throw new MojoExecutionException("Failed to modify repository " + location+": "+StatusTool.collectProblems(e.getStatus), e);
```

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
```diff
@@ -0,0 +1,136 @@
...
+	public record ModifiedRepositoryDescriptor(Repository sourceRepository, P2RepositoryKind repositoryKind,
+			Path outputLocation, boolean compress, boolean xzCompress, boolean keepNonXzIndexFiles) {
+	}
+
+	@Inject
+	private P2RepositoryManager manager;
+
+	public void modifyRepositoryMetadata(ModifiedRepositoryDescriptor repository, String repositoryName,
+			List<String> propertiesToRemove, Map<String, String> propertiesToAdd)
+			throws ProvisionException, URISyntaxException, IOException {
```
laeubi:
```suggestion
			throws ProvisionException, IOException {
```

I would catch `URISyntaxException` inside the method and convert it into an `IOException`. As an alternative don't pass in any data that requires paring of URIs but hand over proper URI objects in the first place.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
```diff
@@ -0,0 +1,136 @@
...
+					Files.createDirectories(outputLocation);
+				} else {
+					try (Stream<Path> files = Files.list(outputLocation);) {
+						if (files.findAny().isPresent()) {
+							throw new IllegalStateException("Output location is not empty: " + outputLocation);
+						}
+					}
+				}
+				repositoryLocation.setUrl(location.toString()); // reset (potentially) normalized URL
+			} catch (IOException e) {
```
laeubi: I won't catch the exceptions here but rethrow them so the caller can possibly handle them.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
```diff
@@ -101,6 +111,39 @@ public IArtifactRepository getCompositeArtifactRepository(Collection<Repository>
...
+				// Copy (potentially customized) mapping rules
+				if (repository instanceof SimpleArtifactRepository simpleSource
+						&& mirror instanceof SimpleArtifactRepository simpleTarget) {
+					simpleTarget.setRules(simpleSource.getRules()); // Copy artifact mapping rules
+				}
+			}, null);
+			StatusTool.assertNoProblemOf(IStatus.ERROR, result,
+					() -> "Failed to create output repository at: " + targetLocation);
+			return mirror;
+		} catch (ProvisionException e) {
```
laeubi: rethrow `ProvisionException`

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
```diff
@@ -101,6 +111,39 @@ public IArtifactRepository getCompositeArtifactRepository(Collection<Repository>
...
+				// Copy artifacts to modifiable output repository
+				repository.query(ArtifactKeyQuery.ALL_KEYS, null).stream()
+						.map(a -> repository.getArtifactDescriptors(a)).forEach(a -> mirror.addDescriptors(a, null));
+				// Copy (potentially customized) mapping rules
+				if (repository instanceof SimpleArtifactRepository simpleSource
+						&& mirror instanceof SimpleArtifactRepository simpleTarget) {
+					simpleTarget.setRules(simpleSource.getRules()); // Copy artifact mapping rules
+				}
+			}, null);
+			StatusTool.assertNoProblemOf(IStatus.ERROR, result,
```
laeubi: Throw `ProvisionException` (or `CoreEception`) instead

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
```diff
@@ -136,11 +179,34 @@ public IQueryable<IInstallableUnit> getCompositeMetadataRepository(Collection<Re
...
+	/** Creates a local mirror of the given {@link IMetadataRepository}. */
+	public IMetadataRepository mirrorMetadataRepository(IMetadataRepository repository, Path targetLocation) {
+		try {
+			IMetadataRepository mirror = createLocalMetadataRepository(targetLocation, repository.getName(),
+					repository.getProperties());
+			IStatus result = mirror.executeBatch(m -> {
+				// Copy units to modifiable output repository
+				repository.query(QueryUtil.ALL_UNITS, null).stream().map(List::of).forEach(mirror::addInstallableUnits);
+			}, null);
+			StatusTool.assertNoProblemOf(IStatus.ERROR, result,
```
laeubi: Same here throw `ProvisionException`

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: The mojo is in `tycho-p2-repository-plugin` and other mojos operating on p2-repos don't mention p2 explicitly.
On the other hand, this mojo is about modifying the metadata of the repository itself (and the repository can be a `metadata` repository). So mentioning `p2-metadata` could help to avoid confusion with a mojo that would modify just a metadata repository (which this doesn't).
So I renamed it as you suggested.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: Ok, done.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
HannesWell: Makes sense.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
HannesWell: That's the biggest problem I see when using the maven Repository to model repositories as it models the location `URI` as String. This delays the surfacing of errors and requires to re-parse the String again and again (depending on the use-case). So ideally we would wrap the `Repository` immediately that only uses an URI or use another type (like `MavenRepositoryLocation`), but we discussed that already in
- https://github.com/eclipse-tycho/tycho/pull/5209#discussion_r2260505150

For now I'll keep using `Repository` for the user API and use `MavenRepositoryLocation` internally, this can then be unified in
- https://github.com/eclipse-tycho/tycho/pull/5222

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
HannesWell: This makes it more complicated to use this method as it's called in a lambda BiFunction, but I can work-around that in the caller.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
HannesWell: As above.

tycho-spi/src/main/java/org/eclipse/tycho/helper/StatusTool.java
HannesWell: As there is already `IStatus.matches(int)` a `checkStatus` helper method doesn't seem to add much value here.
If the generalized method doesn't fit here, lets just return to the specialization defined in `P2RepositoryManager`.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
HannesWell: Done.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: On the other hand, with that in mind, it might be even better to name it `modify-repository-metadata`?

laeubi: > On the other hand, with that in mind, it might be even better to name it `modify-repository-metadata`?

Why not simply `modify-repository` ?

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
```diff
@@ -153,11 +211,18 @@ private IMetadataRepository getMetadataRepositor(URI location, String id) throws
 		return metadataManager.loadRepository(location, new LoggerProgressMonitor(logger));
 	}
 
-	public void downloadArtifact(IInstallableUnit iu, IArtifactRepository artifactRepository,
-			OutputStream outputStream) throws IOException {
+	public static void assertNoError(IStatus status, Path targetLocation) throws ProvisionException {
+		if (!status.isOK() && status.matches(IStatus.ERROR)) {
+			String msg = "Failed to create output repository at: " + targetLocation + "\n";
+			throw new ProvisionException(msg + StatusTool.toLogMessage(status), StatusTool.findException(status));
```
laeubi: This is still wrong, `toLogMessage` (as the name suggest) is when output it as a log message not using it as an exception message.

I would also not do to much unwrapping here, so why not simply create an multi error status and attach the original status as a child? Or even throw a ProvisionException with the status itself, I don't think the message here buys us much as the status should already include why it failed.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
```diff
@@ -0,0 +1,158 @@
...
+            }
+            if (!propertiesToAdd.isEmpty()) {
+                log.info("  Add properties:");
+                propertiesToAdd.forEach((key, value) -> {
+                    log.info("   " + key + " = " + value);
+                });
+            }
+
+            manipulator.modifyRepositoryMetadata(descriptor, repositoryName, propertiesToRemove, propertiesToAdd);
+        } catch (IOException | URISyntaxException e) {
```
laeubi: `URISyntaxException` more feels like it should be a `MojoFailureException` (as it indicates a user problem).

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: This sounds very general. And while we can for sure extend the possible modifications in the future, I'm not sure if it will and should evolve into a mojo that's so general that such general name is really suitable.

laeubi: As you mentioned there is artifact and metadata repository. and as both are metadata each moidification of a repo include the metadata itself.

So I think it is okay here if we really want it to modify both types, otherwise an abstract base mojo and two specializations as

- `modify-metadata-repository`
- `modify-artifact-repository`

would be more suitable.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: Changed it so that it's catched at it's origin and wrapped into a special `RepositoryLocationURISyntaxException`, which is then rethrown as cause of a `MojoFailureException`.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
HannesWell: > Or even throw a ProvisionException with the status itself, I don't think the message here buys us much as the status should already include why it failed.

Indeed that's much simpler. Done.

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/ModifyRepositoryMetadataMojo.java
HannesWell: As discussed in a private chat, I changed the name now to `modify-repository-properties` as that's exactly what the mojo is doing (and the name is also a property).

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryKind.java
```diff
@@ -0,0 +1,30 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Hannes Wellmann - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.tycho.p2maven.repository;
+
+public enum P2RepositoryKind {
+    /** Kind representing a repository with both, metadata and artifacts. */
+    both,
```
laeubi: Just one thing that came into my mind would be possibly name it better `colocated`? At least that is how it is named in case the same url is used for artifact and metadata

laeubi (approved):
Looks good to me, naming it `modify-repository-properties` seems a good choice here to distinguish it from other operations.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryKind.java
HannesWell: At least in P2's RepositoryDescriptor class this type/kind is named `both` if `artifact` and `metadata` are at the same location.
I cannot tell how public P2 API names this case (if it's named there at all) or how the doc names it.
Have to do some research on that.

laeubi: It was just something that came into my mind, so your decision here!

HannesWell: In the Eclipse help of P2 I only found this that mentions any of the terms:
https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Fguide%2Fp2_repositorytasks.htm&cp%3D2_0_20_3_3

And it talks about _both_ for that case, although there's no 'kind' literal, but it's considered as _both_ if no kind is specified. We can also reflect this and remove the `both` literal and assume null means both/colocated.

The term `colocated` doesn't occur at all in the entire doc.

HannesWell: As discussed (in a private chat), now the absence of a kind respectively `null`, is treated as both and the explicit `both` literal is removed.

PR #5222, opened 2025-08-08
Unify modeling of P2-Repositories

HannesWell:
Remove all usages of org.eclipse.tycho.plugins.p2.Repository (and remove that class) and reduce usages of MavenRepositoryLocation. Both are replaced by org.apache.maven.model.Repository.

Additionally migrate touched plexus components to use JSR-330 javax.inject annotations.


HannesWell:
Replacing `MavenRepositoryLocation` by Maven's repository entirely requires even more changes and is left for later.
Maybe then one could also find a better handling of the conversion of Strings to `URIs` as the maven Repository models each `URI` as `String`... Ideally any malformation of an URI is detected as early as possible.

HannesWell:
The verification builds using this change currently fail with the following error
```
Error:  Failed to execute goal org.eclipse.tycho:tycho-p2-plugin:6.0.0-SNAPSHOT:p2-metadata-default (default-p2-metadata-default) on project org.eclipse.swt: Execution default-p2-metadata-default of goal org.eclipse.tycho:tycho-p2-plugin:6.0.0-SNAPSHOT:p2-metadata-default failed: Unable to load the mojo 'p2-metadata-default' (or one of its required components) from the plugin 'org.eclipse.tycho:tycho-p2-plugin:6.0.0-SNAPSHOT': com.google.inject.ProvisionException: Unable to provision, see the following errors:
Error:  
Error:  1) [Guice/NullInjectedIntoNonNullable]: null returned by binding at LocatorWiring
Error:   but BaselineValidator.baselineService(BaselineValidator.java:56) is not @Nullable
Error:    at LocatorWiring
Error:    at BaselineValidator.baselineService(BaselineValidator.java:56)
Error:        \_ for field baselineService
Error:    at LocatorWiring
Error:    while locating BaselineValidator annotated with interface TypeArguments$Implicit
Error:    at LocatorWiring
Error:    at P2MetadataMojo.baselineValidator(P2MetadataMojo.java:47)
Error:        \_ for field baselineValidator
Error:    while locating P2MetadataDefaultMojo
Error:    at ClassRealm[plugin>org.eclipse.tycho:tycho-p2-plugin:6.0.0-SNAPSHOT, parent: ClassLoaders$AppClassLoader@c387f44]
Error:        \_ installed by: WireModule -> PlexusBindingModule
Error:    while locating Mojo annotated with @Named("org.eclipse.tycho:tycho-p2-plugin:6.0.0-SNAPSHOT:p2-metadata-default")
Error:  
Error:  Learn more:
Error:    https://github.com/google/guice/wiki/NULL_INJECTED_INTO_NON_NULLABLE
Error: [ERROR] 
Error: [ERROR] 1 error
Error:  
Error:  ======================
Error:  Full classname legend:
Error:  ======================
Error:  BaselineValidator:           "org.eclipse.tycho.plugins.p2.BaselineValidator"
Error:  ClassLoaders$AppClassLoader: "jdk.internal.loader.ClassLoaders$AppClassLoader"
Error:  LocatorWiring:               "org.eclipse.sisu.wire.LocatorWiring"
Error:  Mojo:                        "org.apache.maven.plugin.Mojo"
Error:  Named:                       "com.google.inject.name.Named"
Error:  P2MetadataDefaultMojo:       "org.eclipse.tycho.plugins.p2.P2MetadataDefaultMojo"
Error:  P2MetadataMojo:              "org.eclipse.tycho.plugins.p2.P2MetadataMojo"
Error:  PlexusBindingModule:         "org.eclipse.sisu.plexus.PlexusBindingModule"
Error:  TypeArguments$Implicit:      "org.eclipse.sisu.inject.TypeArguments$Implicit"
Error:  WireModule:                  "org.eclipse.sisu.wire.WireModule"
Error:  ========================
Error:  End of classname legend:
Error:  ========================
Error:  
Error:        role: org.apache.maven.plugin.Mojo
Error:    roleHint: org.eclipse.tycho:tycho-p2-plugin:6.0.0-SNAPSHOT:p2-metadata-default
```
Looks like the migration to javax-annotation is not correct. @laeubi do you have any immediate hint?

HannesWell:
It seems like the definition of `BaselineServiceImpl` does make it discoverable for `BaselineService` fields annotated with just `@Inject`:
```
@Singleton
public class BaselineServiceImpl implements BaselineService {
...
}
```
I did some changes locally that seem to confirm that this is the mainly problematic point.

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/BaselineServiceImpl.java
```diff
@@ -43,44 +45,39 @@
...
-import org.eclipse.tycho.MavenRepositoryLocation;
+import org.eclipse.equinox.p2.repository.artifact.IProcessingStepDescriptor;
 import org.eclipse.tycho.core.shared.StatusTool;
 import org.eclipse.tycho.p2.metadata.IP2Artifact;
 import org.eclipse.tycho.p2.publisher.P2Artifact;
-import org.eclipse.tycho.p2maven.ListQueryable;
 import org.eclipse.tycho.p2maven.repository.P2RepositoryManager;
 
-@Component(role = BaselineService.class)
+@Singleton
```
laeubi:
```suggestion
@Singleton
@Named
```

laeubi:
> It seems like the definition of `BaselineServiceImpl` does make it discoverable for `BaselineService` fields annotated with just `@Inject`:
> 
> ```
> @Singleton
> public class BaselineServiceImpl implements BaselineService {
> ...
> }
> ```
> 
> I did some changes locally that seem to confirm that this is the mainly problematic point.

Yes that is quite annoying you should consider this points

1. Only ever change one component at a time and make it into the codebase
2. JSR330 generation is not enabled for all modules so be sure its in the build section and produces the desired output in the final jar.
3. Use an interface for the service (seems the case here) with a default impl that has the annotations

HannesWell:
> 2\. JSR330 generation is not enabled for all modules so be sure its in the build section and produces the desired output in the final jar.

I believe that's the missing piece here, because the change you suggested (alone) also didn't work.
Were do I enable that?

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/BaselineServiceImpl.java
HannesWell: I wonder when I exactly it's necessary to have an empty `@Named` annotation and when not? I haven't found a doc explaining it well and easy for Maven.
And from other changes I have the impression it also works in case where no annotation is present at all?

laeubi:
> I believe that's the missing piece here, because the change you suggested (alone) also didn't work.
> Were do I enable that?

See for example here:
https://github.com/eclipse-tycho/tycho/commit/111abc9c06e5588d9a91a28ecc30dcb3419c7326

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/BaselineServiceImpl.java
laeubi: Maybe `@Singleton` somehow includes `@Named` but as  I never seen it without in the maven source I would add it. The named annotation is used later on to generate the metadata as well, it tags the component as being the default component for this interface.

HannesWell:
Locally this worked now. Thanks for the hints. The solution seem to be the combination of both: Annotating an implementation with `@Named` and running `org.eclipse.sisu:sisu-maven-plugin`.
It looks like `org.eclipse.sisu:sisu-maven-plugin` only recognizes components when they are annotated with `@Named` and only then adds them to the `META-INF/sisu/javax.inject.Named` file. But it also seems that components in the same jar/artifact/realm are found even when not recorded in that file. So for 'internal' use it seems to be found even without any annotation (at least when the class is used directly and is not referenced over the implemented interface.
As all the `javax.inject/jakarta.inject/com.google.inject` annotations have retention-policy runtime, all further configuration is probably read from the classes at runtime.

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/BaselineServiceImpl.java
HannesWell: It's relevant, see my answer below:
- https://github.com/eclipse-tycho/tycho/pull/5222#issuecomment-3170495224

HannesWell:
> Additionally migrate touched plexus components to use JSR-330 javax.inject annotations.

Reverted all these parts for this PR. It can be done separately and hopefully unblocks those PRs depending on this one.

PR #5223, opened 2025-08-08
Replace ListQueryable by P2's CompoundQueryable

HannesWell:
This requires
- https://github.com/eclipse-tycho/tycho/pull/5222


tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/BaselineServiceImpl.java
```diff
@@ -73,39 +75,40 @@ public Map<String, IP2Artifact> getProjectBaseline(Collection<Repository> baseli
...
+                artifactRepositories.add(repositoryManager.getArtifactRepository(location));
             } catch (ProvisionException | URISyntaxException e) {
                 // baseline repository may not exist yet
                 logger.warn(e.getMessage(), e);
             }
         }
+        IQueryable<IInstallableUnit> baselineUnits = metadataRepositories.size() == 1 ? metadataRepositories.getFirst()
+                : new CompoundQueryable<>(metadataRepositories);
+        IArtifactRepository baselineArtifacts = artifactRepositories.size() == 1 ? artifactRepositories.getFirst()
+                : new ListCompositeArtifactRepository(artifactRepositories, null);
```
HannesWell: Do we need to pass a real `IProvisioningAgent` here or is actually unused and `null` is sufficient?
At least locally all tests passed with `null`.

laeubi: It all depends on how it is used. Code can call `repository.getAgent() `and then might run into a NPE

HannesWell: Yes. I was mainly wondering if we can say with certainty that this method is never used.
But I assume we can't, so I changed it to pass a real agent, just in case.

PR #5224, opened 2025-08-08
Replace readonly @Parameter annotations for components with @Inject

HannesWell:
This modernizes common old ways of configurations like
```
	@Parameter(property = "session", required = true)
	private MavenSession session;
	@Parameter(property = "project", readonly = true)
	private MavenProject project;
```
to just
```
	@Inject
	private MavenSession session;
	@Inject
	private MavenProject project;
```
which should also remove these parameters from the Mojo's help page, which can actually not be configured.

And migrate some plain `@Component` annotations by `@Inject`.


laeubi:
@HannesWell do you plan to continue or can we close this?

HannesWell:
> do you plan to continue or can we close this?

Generally yes. But I currently don't know why the tests failed.
I've rebased the change now. Lets see if something else fixed it in the meantime.
Maybe the `org.eclipse.sisu:sisu-maven-plugin` execution was missing which was added in the meantime by another change.

PR #5228, opened 2025-08-11
Add mojo to modify a composite P2-repository

HannesWell:
This adds a mojo to modify and create composite P2-repositories, similar to https://github.com/eclipse-tycho/tycho/pull/5209.

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
@@ -35,6 +35,7 @@ import org.eclipse.equinox.p2.core.IProvisioningAgent;
 import org.eclipse.equinox.p2.core.ProvisionException;
 import org.eclipse.equinox.p2.metadata.IArtifactKey;
 import org.eclipse.equinox.p2.metadata.IInstallableUnit;
+import org.eclipse.equinox.p2.query.IQueryResult;
 import org.eclipse.equinox.p2.query.IQueryable;
 import org.eclipse.equinox.p2.query.QueryUtil;
 import org.eclipse.equinox.p2.repository.artifact.ArtifactKeyQuery;
@@ -118,6 +119,23 @@ public class P2RepositoryManager {
 	}
 
 	/**
+	 * Returns all artifacts contained in the {@link IArtifactRepository} at the
+	 * given location. This method does NOT check the type of the repository! <br>
+	 * This can be useful to obtain overview information and can serve as existence
+	 * check for a repository.
+	 * 
+	 * @param location the location of the repository to load
+	 * @param id       the optional id of the repository (e.g. used for
+	 *                 authentication), may be {@code null}
+	 * @return the number of contained artifacts
+	 * @throws ProvisionException if loading the repository failed
+	 */
+	public IQueryResult<IArtifactKey> allArtifacts(URI location, String id) throws ProvisionException {
+		IArtifactRepository repository = getArtifactRepository(location, id);
+		return repository.query(ArtifactKeyQuery.ALL_KEYS, null);
+	}
+
+	/**
 	 * Creates a local mirror only of the <b>data</b> of the given
 	 * {@link IMetadataRepository}, i.e. its {@code artifacts.xml/jar/xml.xz} files.
 	 */
@@ -183,6 +201,23 @@ public class P2RepositoryManager {
 		MetadataRepositoryFactory factory = new SimpleMetadataRepositoryFactory();
 		factory.setAgent(agent);
 		return factory.create(location.toUri(), name, null, properties);
+	}
+
+	/**
+	 * Returns all units contained in the {@link IMetadataRepository} at the given
+	 * location. This method does NOT check the type of the repository! <br>
+	 * This can be useful to obtain overview information and can serve as existence
+	 * check for a repository.
+	 * 
+	 * @param location the location of the repository to load
+	 * @param id       the optional id of the repository (e.g. used for
+	 *                 authentication), may be {@code null}
+	 * @return the number of contained units
+	 * @throws ProvisionException if loading the repository failed
+	 */
+	public IQueryResult<IInstallableUnit> allMetadataUnits(URI location, String id) throws ProvisionException {
+		IMetadataRepository repository = getMetadataRepositor(location, id);
+		return repository.query(QueryUtil.ALL_UNITS, null);
 	}
 
 	/** Creates a local mirror of the given {@link IMetadataRepository}. */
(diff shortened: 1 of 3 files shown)


p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
```diff
@@ -142,4 +159,123 @@ private void xzCompress(ModifiedRepositoryDescriptor repository, Path outputLoca
...
+			for (URI child : childrenToAdd) {
+				children.add(child);
+			}
+			if (childLimit > 0) {
+				while (children.size() > childLimit) {
+					children.removeFirst();
+				}
+			}
+			repo.setChildren(children.toArray(URI[]::new));
+			// TODO: check current Jenkins job!
```
laeubi: ?

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
```diff
@@ -117,6 +117,24 @@ public IArtifactRepository createLocalArtifactRepository(Path location, String n
...
+	 * This can be useful to obtain overview information and as existence check for
+	 * a repository.
+	 * 
+	 * @param location the location of the repository to load
+	 * @param id       the optional id of the repository (e.g. used for
+	 *                 authentication), may be {@code null}
+	 * @return the number of contained artifacts
+	 * @throws ProvisionException if loading the repository failed
+	 */
+	public long countArtifacts(URI location, String id) throws ProvisionException {
```
laeubi: Should we not better make it return all Artifacts and callers can then count, query, do whatever?

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
```diff
@@ -185,6 +203,24 @@ public IMetadataRepository createLocalMetadataRepository(Path location, String n
...
+	 * This can be useful to obtain overview information and as existence check for
+	 * a repository.
+	 * 
+	 * @param location the location of the repository to load
+	 * @param id       the optional id of the repository (e.g. used for
+	 *                 authentication), may be {@code null}
+	 * @return the number of contained units
+	 * @throws ProvisionException if loading the repository failed
+	 */
+	public long countMetadataUnits(URI location, String id) throws ProvisionException {
```
laeubi: same here, just return the result of the query?

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
```diff
@@ -67,12 +82,15 @@ public static ModifiedRepositoryDescriptor createDescriptor(Repository repositor
...
-				repositoryKind, outputLocation, compress, xzCompress, keepNonXzIndexFiles);
+		boolean isArtifact = repositoryKind == null || repositoryKind == P2RepositoryKind.artifact;
+		boolean isMetadata = repositoryKind == null || repositoryKind == P2RepositoryKind.metadata;
+		return new ModifiedRepositoryDescriptor(new MavenRepositoryLocation(repository.getId(), location), isArtifact,
+				isMetadata, outputLocation, compress, xzCompress, keepNonXzIndexFiles);
 	}
 
-	public record ModifiedRepositoryDescriptor(MavenRepositoryLocation repository, P2RepositoryKind repositoryKind,
-			Path outputLocation, boolean compress, boolean xzCompress, boolean keepNonXzIndexFiles) {
+	public record ModifiedRepositoryDescriptor(MavenRepositoryLocation repository, boolean isArtifact,
```
laeubi: why now two booleans? Could we not maybe better add a method named `isMetadata` / `isArtifact` that contains the check. This would make the constructor less hard to read/use and has the benefit we do not loose information here.

laeubi (commented):
@HannesWell do we like to finish this as release is due on end of month?

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
HannesWell: That was referring to the eclipse-platform RelEng Jenkins job to modify a composite P2-repo, which currently uses P2 ANT-tasks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/5a8118d7624f3ee5d7e2e31e17de8c735e6d1afc/JenkinsJobs/Releng/modifyP2CompositeRepository.jenkinsfile

But that was just a note for me and I just removed it.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
HannesWell: That's more flexible yes. Done.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/P2RepositoryManager.java
HannesWell: Done.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/tools/P2RepositoryDataManipulator.java
HannesWell: Yes, that works as well and prevents further increasing the already large number of arguments.

HannesWell:
@laeubi My local testing was successful, thus I consider this ready for submission.
I'll create a follow-up with release notes for this and https://github.com/eclipse-tycho/tycho/pull/5209. AFAIK release notes could even be enhanced after the release.

LorenzoBettini:
@HannesWell thanks for this nice addition!
I was working on a similar project, but this makes it useless ;)
I see no test in the PR, so is there some kind of example about how to use this?

HannesWell:
> thanks for this nice addition!
> I was working on a similar project, but this makes it useless ;)

You're welcome. :)
Yes I recalled your announcement from a while ago an believe I therefore tagged you in one of the PR/issues of this multi-step effort.

> I see no test in the PR, so is there some kind of example about how to use this?


Indeed. One example (and the main reason I created it) is the this Jenkins job used in Eclipse-platform:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/d532c874f13031e455d243ae05b90a77272aa091/JenkinsJobs/Releng/modifyP2CompositeRepository.jenkinsfile#L42-L49

in conjunction with this configuration
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/d532c874f13031e455d243ae05b90a77272aa091/eclipse-platform-parent/pom.xml#L1121-L1139

Please also note the discussion about
- https://github.com/eclipse-tycho/tycho/discussions/5429

If you want to 'resolve' it a contribution would be welcome (maybe ask the other guy before if he's still interested).

LorenzoBettini:
@HannesWell thanks I'll have a look.

I was wondering why you're not using the standard p2 composite application to manage (i.e., modify) the composite repository. That takes care of most of the work and the implementation code would be much smaller (without re-inventing the wheel) and more maintainable. (That's what I do in my own plugin for managing composite sites).

Issue #5241, opened 2025-08-20
SurefireLoader discovers JUnit 5 TestEngine service resources in unstable order


trancexpress:
While investigating https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2423, I found weird behavior in `tycho` and in JUnit 5.

For the latter I opened a discussion: https://github.com/junit-team/junit-framework/discussions/4855

I'm still debugging the why, but I've had two different `mvn verify` sessions for JDT UI tests, in which `org.eclipse.tycho.surefire.osgibooter.SurefireLoader.findResources(String)` will find bundle resources in different order. I.e. when JUnit 5 is looking for `org.junit.platform.engine.TestEngine`, `org.eclipse.tycho.surefire.osgibooter.SurefireLoader` will provide those services in instable order. This then leads to instable test order, which then leads to issues like https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2423.

In essence a committer added a test which didn't properly clean up, CI tests were green, after merging CI tests became flaky. A while after (after someone found time to open an issue for the flaky tests, after someone else found time to look into the flaky tests), we discovered the problematic tests. This was not simple, since test order was not stable.

So we should look into why `SurefireLoader` is not loading the JUnit 5 services in a stable order.

trancexpress:
I'll be debugging this, but if someone already knows the reason it would be great to comment, to save some time.

trancexpress:
Here is the stack trace of service discovery:

```
"main" #1 [1536573] prio=6 os_prio=0 cpu=4549.07ms elapsed=660.86s tid=0x00007fd39c04f7c0 nid=1536573 at breakpoint [0x00007fd3a1921000]
   java.lang.Thread.State: RUNNABLE
        at java.util.ServiceLoader$LazyClassPathLookupIterator.parseLine(java.base@21.0.8/ServiceLoader.java:1158)
        at java.util.ServiceLoader$LazyClassPathLookupIterator.parse(java.base@21.0.8/ServiceLoader.java:1177)
        at java.util.ServiceLoader$LazyClassPathLookupIterator.nextProviderClass(java.base@21.0.8/ServiceLoader.java:1213)
        at java.util.ServiceLoader$LazyClassPathLookupIterator.hasNextService(java.base@21.0.8/ServiceLoader.java:1228)
        at java.util.ServiceLoader$LazyClassPathLookupIterator.hasNext(java.base@21.0.8/ServiceLoader.java:1273)
[...]

PR #5242, opened 2025-08-20
Sort fragments URLs passed to SurefireLoader

trancexpress:
The result of OsgiSurefireBooter.getFragments(Bundle) depends on iterating over HashMap containers, i.e. it doesn't have a stable order.

SurefireLoader will then load service resources in an unstable order, causing JUnit 5 to run tests an unstable order - if the tests have a mix of test suites, JUnit 4 and JUnit 5 tests. With this unstable order, the JUnit 5 vintage test engine can be listed before or after the suite and JUnit 5 test engines.

This change ensures fragments of the bundle
org.eclipse.tycho.surefire.osgibooter (junit5, junit5 vintage) are sorted, ensuring test engines are discovered by JUnit 5 in a stable order.

Fixes: #5241

### DIFF
--- a/tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/java/org/eclipse/tycho/surefire/osgibooter/OsgiSurefireBooter.java
+++ b/tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/java/org/eclipse/tycho/surefire/osgibooter/OsgiSurefireBooter.java
@@ -29,6 +29,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
@@ -401,7 +402,7 @@ public class OsgiSurefireBooter {
             return new Bundle[0];
         }
         return hostWires.stream().map(wire -> wire.getRequirer().getBundle()).filter(Objects::nonNull)
-                .toArray(Bundle[]::new);
+                .sorted(Comparator.comparing(Bundle::getSymbolicName)).toArray(Bundle[]::new);
     }
 
 }


trancexpress:
@akurtakov any idea who can take a look here?

Also any idea if I can run JDT UI tests without merging this PR, but with using the change from it?

laeubi:
> SurefireLoader will then load service resources in an unstable order, causing JUnit 5 to run tests an unstable order - if the tests have a mix of test suites, JUnit 4 and JUnit 5 tests. With this unstable order, the JUnit 5 vintage test engine can be listed before or after the suite and JUnit 5 test engines.

@trancexpress thanks for this PR, I think we should just clarify the on the "unstable order":

- The order of test execution should not matter, so so if we have classes `A`, `B`, `C` all containing tests, they might be executed in any order as long as it is always the same order
- If we have a mixture of Junit4+5 it should not matter what tests are executed first as long as its always the same order
- So in general we do not guarantee any specific order one could use to make tests depend on each other, but we ensure that two consecutive executions always result in the same order.

laeubi:
> Also any idea if I can run JDT UI tests without merging this PR, but with using the change from it?

You can simply run `mvn clean install -T1C -DskipTests` then run run the local maven command of your build.

trancexpress:
> I think we should just clarify the on the "unstable order":
> 
>     * The order of test execution should not matter, so so if we have classes `A`, `B`, `C` all containing tests, they might be executed in any order as long as it is always the same order
>     * If we have a mixture of Junit4+5 it should not matter what tests are executed first as long as its always the same order
>     * So in general we do not guarantee any specific order one could use to make tests depend on each other, but we ensure that two consecutive executions always result in the same order.

Yes, this is my expectation too. There should be a stable order, even if its not well defined what exactly that order is (at least not without using the ordering annotations).

tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/java/org/eclipse/tycho/surefire/osgibooter/OsgiSurefireBooter.java
```diff
@@ -106,6 +106,7 @@ public static int run(String[] args, Properties testProps) throws Exception {
         Bundle bundle = FrameworkUtil.getBundle(OsgiSurefireBooter.class);
         urls.add(getURL(bundle));
         Bundle[] fragments = getFragments(bundle);
+        Arrays.sort(fragments);
```
laeubi: Instead of sorting, can we better use a `LinkedHashMap` in `getFragments`?
The reason is that sorting would imply some kind of precedence, while all we want here is that the order is the same between two runs.

laeubi: Okay that was misleading.. as you described the Map is not created by Tycho (what would need to be changed in Equinox). But as we already have a stream there, we could simply use 

hostWires.stream().map(wire -> wire.getRequirer().getBundle()).filter(Objects::nonNull)**.sorted()**
                .toArray(Bundle[]::new);

trancexpress: I don't know, the information from `getFragments()` comes from other hash maps... I gave up on trying to follow that since the bundle wiring (that is used there) has tons of initializations. If you insist I can continue debugging, it didn't seem trivial to find the initial hash map (and there is at least one more hash map on the way).

laeubi: No I don't think its useful, and even doubt the OSGi specification mandates any special ordering here, so just performing the sorting in Tycho seems fine, I just would move it to the `getFragments` method instead of performing it afterwards on the final array.

trancexpress:
> mvn clean install -T1C -DskipTests

I assume I must bump some version in JDT UI, since I don't see the newer code when I debug. But where is that version?

```
$ git grep tycho.version
org.eclipse.jdt.text.tests/pom.xml:        <version>${tycho.version}</version>
org.eclipse.jdt.ui.tests.refactoring/pom.xml:        <version>${tycho.version}</version>
org.eclipse.jdt.ui.tests/pom.xml:        <version>${tycho.version}</version>
```

Sorry I have next to no knowledge in this area.

trancexpress:
OK, thank you @laeubi !

Still though, how do I tell JDT UI to use the newer version here? Do I need to bump some versions also in `tycho`?

Or things will get picked up automatically due to some already-done effort for that?

laeubi:
> I assume I must bump some version in JDT UI, since I don't see the newer code when I debug. But where is that version?

A version bump is not required, but platform is using Tycho 5 (SNAPSHOT) while the main branch is at 6 (SNAPSHOT), you could simply append `-Dtycho.version=6.0.0-SNAPSHOT` to your comandline when building jdt-ui

You should now see something like 

```
[INFO] Tycho Version:  6.0.0-SNAPSHOT (<your commit hash>)
[INFO] Tycho Mode:     extension
[INFO] Tycho Builder:  maven
[INFO] Build Threads:  1
```

in the build log.

tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/java/org/eclipse/tycho/surefire/osgibooter/OsgiSurefireBooter.java
```diff
@@ -400,7 +400,7 @@ private static Bundle[] getFragments(Bundle bundle) {
         if (hostWires == null) {
             return new Bundle[0];
         }
-        return hostWires.stream().map(wire -> wire.getRequirer().getBundle()).filter(Objects::nonNull)
+        return hostWires.stream().map(wire -> wire.getRequirer().getBundle()).filter(Objects::nonNull).sorted()
```
laeubi:
```suggestion
        return hostWires.stream().map(wire -> wire.getRequirer().getBundle()).filter(Objects::nonNull)
                .sorted(Comparator.comparing(Bundle::getSymbolicName))
```

I think it is more safe to use the symbolic name here instead of the default (what is the bundle id assigned by the framework), that way install order should not matter.

trancexpress:
Fail is:

```
Error:  Failures: 
Error:    TestPGPSigning.testSigningSkipIfJarSignedAndAnchored:230 Unexpected unsigned IUs. ==> expected: <[org.eclipse.equinox.common, org.eclipse.equinox.common.source, org.eclipse.osgi, org.eclipse.osgi.source, org.eclipse.platform_root]> but was: <[org.eclipse.equinox.common.source, org.eclipse.osgi, org.eclipse.osgi.source, org.eclipse.platform_root]>
[INFO] 
Error:  Tests run: 450, Failures: 1, Errors: 0, Skipped: 8
```
```
Unexpected unsigned IUs. ==> expected: <[org.eclipse.equinox.common, org.eclipse.equinox.common.source, org.eclipse.osgi, org.eclipse.osgi.source, org.eclipse.platform_root]> but was: <[org.eclipse.equinox.common.source, org.eclipse.osgi, org.eclipse.osgi.source, org.eclipse.platform_root]>
org.opentest4j.AssertionFailedError: Unexpected unsigned IUs. ==> expected: <[org.eclipse.equinox.common, org.eclipse.equinox.common.source, org.eclipse.osgi, org.eclipse.osgi.source, org.eclipse.platform_root]> but was: <[org.eclipse.equinox.common.source, org.eclipse.osgi, org.eclipse.osgi.source, org.eclipse.platform_root]>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1156)
	at org.eclipse.tycho.test.pgp.TestPGPSigning.testSigningSkipIfJarSignedAndAnchored(TestPGPSigning.java:230)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
```

```
 <[org.eclipse.equinox.common, org.eclipse.equinox.common.source, org.eclipse.osgi, org.eclipse.osgi.source, org.eclipse.platform_root]> 
 <[                            org.eclipse.equinox.common.source, org.eclipse.osgi, org.eclipse.osgi.source, org.eclipse.platform_root]>
```

`org.eclipse.equinox.common` is missing from the list. I doubt this would come from sorting. Is this a known failure? I didn't find an issue for it, though I didn't search for long.

trancexpress:
Unfortunately I still see an instable order in JDT UI tests, I'm debugging again.

Even with the test engine order being stable, there are still differences in test order. So there must be more problems.

Lets keep this PR open until I figure out how to make the order stable, I now know how to change code in `tycho` and use the changed code in JDT UI tests.

trancexpress:
Thank you for reviewing and merging @laeubi !

PR #5258, opened 2025-08-27
Use Maven#removeTestCompileSourceRoot instead of Iterator#remove

laeubi:
FYI @hd42 @akurtakov 

Fixes warnings:

```
[WARNING] Direct modification of testCompileSourceRoots through iterator.remove() is deprecated
 and will not work in Maven 4.0.0. Please use the add/remove methods instead.
 If you're using a plugin that causes this warning, please upgrade to the latest version and report an issue if the 
warning persists. To disable these warnings, set -Dmaven.project.sourceRoots.warningsDisabled=true
on the command line, in the .mvn/maven.config file, or in project POM properties.
```


laeubi:
This will require Maven 3.9.11, so I'd like to delay it until after the Tycho 5.0.0 release.

tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/OsgiBundleProject.java
```diff
@@ -241,20 +240,19 @@ private void addPDESourceRoots(MavenProject project) {
...
                 project.addCompileSourceRoot(sourceFolder.getAbsolutePath());
             }
         }
     }
 
-    private void removeDuplicateTestCompileRoot(File sourceFolder, List<String> testCompileSourceRoots) {
-        for (Iterator<String> iterator = testCompileSourceRoots.iterator(); iterator.hasNext();) {
-            String testCompileRoot = iterator.next();
+    private void removeDuplicateTestCompileRoot(File sourceFolder, MavenProject project) {
+        for (String testCompileRoot : project.getTestCompileSourceRoots()) {
```
sratz, line 250: I think we must first copy the list, the implicit iterator of the for-each loop does not play well together with calling remove() on the same list instance.

```diff
-for (String testCompileRoot : project.getTestCompileSourceRoots()) {
+for (String testCompileRoot : new ArrayList<>(project.getTestCompileSourceRoots())) {
```

laeubi: @sratz while this is true in general and I had the same feeling before when adapting the code, we immediately return when finding a match here so it is okay in this case.

Issue #5270, opened 2025-09-01
org.eclipse.tycho.extras:tycho-p2-extras-plugin mirror goal with mvn: protocol not working with Tycho 5.0.0


merks:
The following builds successfully with `4.0.13` but not with `5.0.0`
```
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>

  <groupId>org.eclipse.oomph.jetty.mirror</groupId>
  <artifactId>org.eclipse.oomph.jetty.mirror</artifactId>
  <version>1.0.0-SNAPSHOT</version>
  <packaging>pom</packaging>

  <properties>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    <tycho-version>4.0.13</tycho-version>
    <jetty-version>10.0.26</jetty-version>
  </properties>

  <pluginRepositories>
    <pluginRepository>
      <id>eclipse-maven-releases</id>
      <url>https://repo.eclipse.org/content/repositories/releases</url>
    </pluginRepository>

    <pluginRepository>
      <id>eclipse-cbi-releases</id>
      <url>https://repo.eclipse.org/content/repositories/cbi-releases</url>
    </pluginRepository>
  </pluginRepositories>

  <build>
    <plugins>
      <plugin>
        <groupId>org.eclipse.tycho</groupId>
        <artifactId>tycho-p2-publisher-plugin</artifactId>
        <version>${tycho-version}</version>
        <configuration>
          <skip>true</skip>
        </configuration>
      </plugin>
      <plugin>
        <groupId>org.eclipse.tycho.extras</groupId>
        <artifactId>tycho-p2-extras-plugin</artifactId>
        <version>${tycho-version}</version>
        <executions>
          <execution>
            <phase>generate-resources</phase>
            <goals>
              <goal>mirror</goal>
            </goals>
          </execution>
        </executions>
        <configuration>
          <source>
            <repository>
              <id>jetty</id>
              <layout>p2</layout>
              <url>mvn:org.eclipse.jetty:jetty-p2:${jetty-version}:zip:p2site</url>
            </repository>
          </source>
          <compress>true</compress>
[...]

PR #5271, opened 2025-09-01
Create integration test for mirror goal with mvn protocol

merks:
https://github.com/eclipse-tycho/tycho/issues/5270


laeubi:
@merks this gives me

> Could not find artifact org.eclipse.jetty.http3:http3-server:jar:10.0.26

while in https://github.com/eclipse-tycho/tycho/issues/5270 you say it fails with

> download from mvn:org.eclipse.jetty:jetty-maven-plugin:10.0.26:maven-plugin: failed

is this intentional? In any case I can confirm it works with 4.0.13 anyways...

**EDIT**: I now get the same error, it seems to depend on if the local maven repo is empty or not what error one gets! That's a bit surprising.

merks:
Yes, I was about to say it might depend on what's already cached.  Indeed it's kind of odd...

laeubi:
This is now part of
- https://github.com/eclipse-tycho/tycho/pull/5272

PR #5272, opened 2025-09-01
Resolve extension for type / do not add local repo as remote repository

laeubi:
Currently it can happen that a _type_ instead of an _extension_ is
passed to the mvn protocol and then it fails to download. For example
maven-plugin must map to jar (same for eclipse-plugin). Also adding the
local repository will confuse the maven resolver.

This now do the following:
1) resolve the type into an extension before passing it as a resolver
artifact
2) do not add the local repository to the repositories of the request
3) simplify testcase / add default tycho-version property to not confuse
m2e (will be overridden anyways)


merks:
Thanks for the improvements and research.

Issue #5284, opened 2025-09-06
Small Reproducible Builds issues for release 5.0.0


hboutemy:
see https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/eclipse/tycho/README.md

hboutemy:
@laeubi FYI

laeubi:
@hboutemy do you like to propose a PR to fix the issue? I'm not completely sure what the exact issue is but its good to know that Tycho mostly passes the check :-)

hboutemy:
as you can see from https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/eclipse/tycho/tycho-5.0.0.diffoscope
- the first issue is to keep Git branch = something that works during release, but not when rebuilding
- second is settings properties with your local directory: I don't know how/why
- the third one is caused by https://github.com/eclipse-tycho/tycho/blob/main/tycho-repository-plugin/pom.xml#L93 = full path instead of relative path

first analysis done, I'll try to create a PR if you don't beat me at it

laeubi:
PR is merged now, thanks for making us aware and providing a fix!
This will shortly show up in the current [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) builds.

hboutemy:
I'm convinced 2 out of the 3 issues are fixed: we'll see on the next release :)

on the 3rd issue (the settings), I'm not clear if/how it is solved: not a big problem, let's go step by step, just sharing

hboutemy:
@laeubi ok, first 2 issues fixed, one remaining on 5.0.2
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/eclipse/tycho/README.md

I don't know why full path is stored, but it means sharing on Maven Central does not really make sense
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/eclipse/tycho/tycho-5.0.2.diffoscope

laeubi:
> I don't know why full path is stored, but it means sharing on Maven Central does not really make sense

If I remember right this is sued somehow in executing the tests locally in an IDE context. From what its wort, I think the [tycho-testing-harness](https://github.com/eclipse-tycho/tycho/tree/main/tycho-testing-harness) part of Tycho can be omitted safely from the deployment entirely, it is not useful as a general purpose thing and obviously not needed for its purpose of being a maven plugin.

[...]

PR #5285, opened 2025-09-07
fix Reproducible Builds issues #5284

hboutemy:
2 issues fixed on 3
on the third one, need to decide how to avoid publishing current directory in interpolated file https://github.com/eclipse-tycho/tycho/blob/main/tycho-testing-harness/src/main/resources/org/eclipse/tycho/testing/settings.properties


laeubi:
This seem to consistently break `org.eclipse.tycho.test.DemoTest#testOsgiMavenRepositoryDemo`.
`TYCHO308importSystemPackage.ImportSystemPackagesTest` seems to be an issue on MacOS only.

tycho-repository-plugin/pom.xml
```diff
@@ -90,7 +90,7 @@
 			<resource>
 				<directory>src/main/resources</directory>
 				<filtering>true</filtering>
-				<targetPath>${project.build.directory}/filtered</targetPath>
+				<targetPath>target/filtered</targetPath>
```
laeubi: Maven javadoc says:

> Describe the resource target path. The path is relative to the target/classes directory (i.e. ${project.build.outputDirectory})

but we are using `${project.build.directory}` before (what only points to target) so I assume it must be something like


```suggestion
				<targetPath>../filtered</targetPath>
```

maybe?

tycho-spi/src/main/resources/org/eclipse/tycho/version/version.properties
```diff
@@ -3,5 +3,4 @@ bnd=${bnd.version}
 scm=${env.GIT_COMMIT}
 url=${env.GIT_URL}
 branch=${env.GIT_BRANCH}
-scmBranch=${scmBranch}
```
laeubi, line 6: It looks like this is no longer in use so it seems safe to remove but I wonder why it affects reproducibility?

laeubi:
> on the third one, need to decide how to avoid publishing current directory in interpolated file https://github.com/eclipse-tycho/tycho/blob/main/tycho-testing-harness/src/main/resources/org/eclipse/tycho/testing/settings.properties

Looking through the code I think we can simply [delete the file it](https://github.com/eclipse-tycho/tycho/blob/main/tycho-testing-harness/src/main/resources/org/eclipse/tycho/testing/settings.properties) I think this was used a while back to support local execution of integration tests with a prebuild (and installed) local copy. But I can't find any reference to the file or property anywhere in the source code.

tycho-spi/src/main/resources/org/eclipse/tycho/version/version.properties
hboutemy: when you `git checkout` a git tag (to be able to rebuild the release today), what is the current branch?

laeubi: Good point, as mentioned I think its safe to remove anyways as its unused. I think we mainly used it when one has a certain snapshot jar to see where it was build from (e.g. main vs. maintenance branch).

laeubi:
@hboutemy thanks for fixing this!

PR #5415, opened 2025-10-01
Add inner jars parameter

laeubi:
This pull request adds support for a new configuration option in the `GeneratorMojo` class to control whether the bundle classpath should be processed. This includes introducing a new parameter and updating the command-line arguments accordingly.

New feature: Bundle classpath processing

* Added a new boolean parameter `process-bundle-classpath` to the `GeneratorMojo` class, allowing users to enable or disable processing of the bundle classpath.
* Updated the `execute` method to append the `-process-bundle-classpath` argument when the new parameter is set.


laeubi:
@copilote can you resolve the merge conflict?

PR #5452, opened 2025-10-09
Add JUnit 6 provider support

laeubi:
Basic structure derived from
- https://github.com/eclipse-tycho/tycho/pull/5413

but there is a lot more to do. e.g. currently the test does not really uses JUnit 6 ... we need a target with JUnit 6 and providers need to use proper ranges excluding each other.

FYI @akurtakov

### DIFF
--- a/tycho-its/projects/tycho-surefire-plugin/junit6/basic/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/tycho-surefire-plugin/junit6/basic/META-INF/MANIFEST.MF
@@ -1,0 +1,9 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: JUnit6 Test Plug-in
+Bundle-SymbolicName: bundle.test.junit6
+Bundle-Version: 1.0.0
+Bundle-RequiredExecutionEnvironment: JavaSE-17
+Import-Package: org.junit.jupiter.api;version="[6.0, 7.0)",
+ org.junit.jupiter.params;version="[6.0, 7.0)",
+ org.junit.jupiter.params.provider;version="[6.0, 7.0)"
--- a/tycho-its/projects/tycho-surefire-plugin/junit6/basic/build.properties
+++ b/tycho-its/projects/tycho-surefire-plugin/junit6/basic/build.properties
@@ -1,0 +1,3 @@
+source.. = src/
+bin.includes = META-INF/,\
+               .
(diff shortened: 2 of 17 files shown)


akurtakov:
This seems to be working .

tycho-its/projects/tycho-surefire-plugin/junit6/basic/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,9 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: JUnit6 Test Plug-in
+Bundle-SymbolicName: bundle.test.junit6
+Bundle-Version: 1.0.0
+Bundle-RequiredExecutionEnvironment: JavaSE-17
+Import-Package: org.junit.jupiter.api;version="6.0",
+ org.junit.jupiter.params;version="6.0",
+ org.junit.jupiter.params.provider;version="6.0"
```
laeubi: We really should use version ranges here.

akurtakov: Do you think of [6.0, 6.1) or [6.0, 7.0) range?

merks: The latter i hope or TP updates will be challenging.

akurtakov: As this is for tycho there are no TP updates. JUnit 5 versions have a number of cases where a few minor  versions are grouped together thus none of them look future proof to me - one would be too limited the other too broad.

merks: Sorry, iphones have such a small screen.  ☹️

akurtakov: I will go with [6.0, 7.0) as that is the semantically correct one IMO.

laeubi: > I will go with [6.0, 7.0) as that is the semantically correct one IMO.

Yes this is correct (as for now). In the past we have had one test for each minor release, because junit features / bindings sometimes differed and we have had different providers for that.

If that's the case for JUnit 6 as well we maybe later will need to adapt, but for now it seems a sufficient range.

akurtakov:
MacOS failure is due to macos-13 builders gone on github. Platform failure is due to yet another network failure. Merging.

PR #5535, opened 2025-10-26
Add class-level Javadoc to WrapMojo

chrisrueger:
Introduced a Javadoc comment describing the purpose and advantages of the WrapMojo class, including its ability to create OSGi jars with bnd instructions and support for publishing both plain and OSGi-fied artifacts.
Added a link to [bnd tutorial for wrapping jars](https://bnd.bndtools.org/chapters/160-jars.html)

Hope that this then appears on https://tycho.eclipseprojects.io/doc/main/tycho-wrap-plugin/wrap-mojo.html

Wording taken from https://github.com/eclipse-tycho/tycho/blob/tycho-5.0.0/RELEASE_NOTES.md#new-wrap-mojo


laeubi:
@chrisrueger thanks for improving that, yes it will automatically appear at the doc pages!

PR #5780, opened 2026-01-20
Support resource links to nested files and folders

HannesWell:
Generally support the resolution of links from file-system resources (files and folders) located within any folder, not only to top-level resources.
Also ensure links are processed in the order recorded in the `.project` file to allow more specific links to be listed and considered first.


HannesWell:
Test failures are unrelated (look like infrastructure hiccups)

PR #5790, opened 2026-01-26
Consider linked preference files in tycho-ds-plugin

HannesWell:


tycho-ds-plugin/src/main/java/org/eclipse/tycho/ds/DefaultDeclarativeServiceConfigurationReader.java
```diff
@@ -85,12 +88,14 @@ public String getPath() {
...
-            Logger logger) throws FileNotFoundException, IOException {
+	private Properties getProjectSettings(MavenProject mavenProject) throws FileNotFoundException, IOException {
+		Properties mojoProperties = getMojoSettings(mavenProject, logger);
         Properties properties = new Properties(mojoProperties);
-        File prefs = new File(basedir, PDE_DS_ANNOTATIONS_PREFS);
-        if (prefs.exists()) {
-            try (FileInputStream stream = new FileInputStream(prefs)) {
+		Path prefs = projectManager.getEclipseProject(mavenProject)
+				.map(eclipse -> eclipse.getFile(PDE_DS_ANNOTATIONS_PREFS))
+				.orElseGet(() -> mavenProject.getBasedir().toPath().resolve(PDE_DS_ANNOTATIONS_PREFS));
```
HannesWell, line 96: I wonder if there is a relevant case where `projectManager.getEclipseProject()` will return an empty Optional?

laeubi: I'm not sure what you mean with "relevant case" but if users choose to not check in their `.project` files (for whatever reasons) this will of course always be empty.

And as you can enable the ds plugin with mojo configuration you don't even need a settings file at all.

HannesWell: > I'm not sure what you mean with "relevant case" but if users choose to not check in their `.project` files (for whatever reasons) this will of course always be empty.

Makes sense.

> And as you can enable the ds plugin with mojo configuration you don't even need a settings file at all.

That's right and the work-around I use at the moment. But using a shared/linked settings file in the build as well allows to have one central place to configure DS-processing in the IDE and the build.

PR #5798, opened 2026-01-27
Add skipAssembling parameter to assemble-repository goal

HannesWell:
A skipAssembling parameter for the assemble-repository goal of the tycho-p2-repository-plugin can be useful when building products, whose repository is not used further.

See https://tycho.eclipseprojects.io/doc/latest/tycho-p2-repository-plugin/assemble-repository-mojo.html

### DIFF
--- a/tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
+++ b/tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
@@ -302,6 +302,12 @@ public class AssembleRepositoryMojo extends AbstractRepositoryMojo {
     @Parameter(defaultValue = "repository.xml")
     private String repositoryFileName;
 
+    /**
+     * Whether or not to skip assembling the repository. False by default.
+     */
+    @Parameter(property = "p2.repository.assemble.skip", defaultValue = "false")
+    private boolean skip;
+
     @Component
     private RepositoryReferenceTool repositoryReferenceTool;
 
@@ -319,6 +325,9 @@ public class AssembleRepositoryMojo extends AbstractRepositoryMojo {
 
     @Override
     public void execute() throws MojoExecutionException, MojoFailureException {
+        if (skip) {
+            return;
+        }
         File destination = getAssemblyRepositoryLocation();
         try (var locking = fileLockService.lockVirtually(destination)) {
             destination.mkdirs();


tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
```diff
@@ -302,6 +302,12 @@ public static class RepositoryReferenceFilter {
     @Parameter(defaultValue = "repository.xml")
     private String repositoryFileName;
 
+    /**
+     * Whether or not to skip assembling the repository. False by default.
+     */
+    @Parameter(defaultValue = "false")
+    private boolean skipAssembling;
```
laeubi: For consistency this should be 

```suggestion
    private boolean skip;
```

(and maybe assign a property like `assemble.repository.skip`)

laeubi:
Even though a skip parameter can be generally useful, if you just want to prevent the mojo from being run you can also configure its default execution phase to `<none>` (what is a bit verbose of course).

tycho-p2-repository-plugin/src/main/java/org/eclipse/tycho/plugins/p2/repository/AssembleRepositoryMojo.java
HannesWell: Done, although I'm a bit torn on it. Consistency is good, but if multiple mojos of a plugin share this parameter and one wants to skip them individually (or not skip them), then it makes the configuration more complex since one has to configure each execution. But of course the property can also help in this case.
And as far as I can tell, this plugin doesn't have another skip parameter. Only the archive-repository mojo has `skipArchive`. So there is already an distinction possible at the moment.

I added `p2.repository.assemble.skip` to be consistent to the properties of other Mojos.

HannesWell:
> Even though a skip parameter can be generally useful, if you just want to prevent the mojo from being run you can also configure its default execution phase to `<none>` (what is a bit verbose of course).

Yes, that's right. But this parameter aims add exactly this convenience.
Especially if one has an inheritance structure this verbosity may repeat. And the need to configure an execution explicitly can also prevent the use of tycho-pomless since one can only inject properties through `build.properties` but cannot configure executions.

laeubi:
> And the need to configure an execution explicitly can also prevent the use of tycho-pomless since one can only inject properties through `build.properties` but cannot configure executions.

This is only partly true. e.g. a common technique is to configure a profile in the parent that triggers on a specific packaging-type, the existence of a file or similar and where you configure it for the pom-less project.

Also I would assume that one usually wont have a mixed setup in one project where some products should use that setting and others not.

sratz:
> Also I would assume that one usually wont have a mixed setup in one project where some products should use that setting and others not.

I would assume it to be quite common actually. E.g., we have one (end-user) product where we use the generated repository as an update site and another product which produces a minimal, stand-alone language server binary that does not have any update mechanism.

On the other hand such a (global) property affecting all products can also be useful, e.g. to speed up CI builds.

laeubi:
I must confess until now I though one can't skip this step because Tycho require it. In general I won't recommend using the produced repo as an update site except for some special cases because it can have unintended side-effects and instead build a dedicated one with true updatable features.

Generally, if this step is really optional, we should aim for Tycho 6 to remove it from the default life-cycle and let the user explicitly configure this as an execution as we already do for `materialize-products` and `archive-repository`

HannesWell:
> I must confess until now I though one can't skip this step because Tycho require it.

At least in my experiments it seems to work if the  [product's source](https://tycho.eclipseprojects.io/doc/5.0.2/tycho-p2-director-plugin/materialize-products-mojo.html#source) is the `targetPlatform`, which is the default. If the source is `repository` of course it doesn't work.

> In general I won't recommend using the produced repo as an update site except for some special cases because it can have unintended side-effects and instead build a dedicated one with true updatable features.

I think it depends: If one just wants to provide an update-site for that product it should be fine to use it (at least I haven't experienced problems with that).
If one wants to provide more, of course a dedicated repository is necessary (which could include the product as `iu`, like it's for example done with the platform repository).

> Generally, if this step is really optional, we should aim for Tycho 6 to remove it from the default life-cycle and let the user explicitly configure this as an execution as we already do for `materialize-products` and `archive-repository`

You mean `archive-product`, don't you?
But I agree that it would be good to make the decision more explicit respectively users more aware what's happening and what could be optimized. On the other hand, it's also not always trivial. As we saw above some other settings can have implications that make the repository necessary even if it's not used as an artifact of the build.
We also have to consider what's generated for a pomless product and have to think about how thinks are ideally enabled or disabled there.
We could, for example, keep the lifecycle for the `eclipse-repository` packaging type as it is, but inject the skip properties for assembling and archiving the repositories with the desired values. This could then also try to be smart and consider some of the implications mentioned above. Of course if a users overwrites it, that should take precedence.

But all that being said, I think this addition is still useful. Any objection to submit it (and have it back-ported)?

PR #5833, opened 2026-02-17
added capability to override existing manifest in bundles

andreas-schwarz-vector:
The purpose of this change is to add the capability to override existing manifest files. Therefore a new attribute in the maven location was introduced. Added 2 integration tests to ensure that the functionality works as expected.

This RP is related to https://github.com/eclipse-m2e/m2e-core/pull/2115

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/m2e/pde/target/shared/MavenBundleWrapper.java
+++ b/tycho-core/src/main/java/org/eclipse/m2e/pde/target/shared/MavenBundleWrapper.java
@@ -103,6 +103,8 @@ public class MavenBundleWrapper {
      * @param syncContextFactory
      *            the sync context factory to acquire exclusive access to the wrapped artifact and
      *            its dependencies
+     * @param ignoreExistingMetadata
+     *            if true, the manifest file is forced to be overridden
      * @return the wrapped artifact
      * @throws Exception
      *             if wrapping the artifact fails for any reason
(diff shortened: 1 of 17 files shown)


merks:
@laeubi 

I thought you objected to this approach conceptually.  Or?

laeubi (changes requested):
For now blocked unless we have concluded on the m2e parts.

andreas-schwarz-vector:
I have rebased my changes on origin/main and applied the necessary changes from the related PR. Is there anything left todo @laeubi?

tycho-core/src/main/java/org/eclipse/m2e/pde/target/shared/MavenBundleWrapper.java
```diff
@@ -206,7 +208,7 @@ private static WrappedBundle getWrappedNode(DependencyNode node,
         List<DependencyNode> children = node.getChildren();
         List<WrappedBundle> depends = new ArrayList<>();
         for (DependencyNode child : children) {
-            depends.add(getWrappedNode(child, instructionsLookup, visited));
+            depends.add(getWrappedNode(child, instructionsLookup, visited, ignoreExistingMetadata));
```
laeubi: It seems this class what not synced with the m2e changes (where we do not rewrite dependencies).

laeubi (changes requested):
Please make sure to sync the "shared" part with m2e (also the tests).

tycho-core/src/main/java/org/eclipse/m2e/pde/target/shared/MavenBundleWrapper.java
andreas-schwarz-vector: It seems this slipped through. Thanks for pointing this out!

andreas-schwarz-vector:
> Please make sure to sync the "shared" part with m2e (also the tests).

I compared the "shared" test with the linked PR, and they seem to be identical except for the packages. 
Furthermore, the MavenBundleWrapper should now be identical as well. 
Some of the changes had already been made by you and were included via rebase.

laeubi (approved):
Thanks it now looks fine to me.

laeubi:
@andreas-schwarz-vector It seems I totally missed to backport this so we sadly have this not included yet in a release :-(

andreas-schwarz-vector:
I noticed this as well and was wondering a week ago whether I had forgotten anything.

Thanks for merging it now! Could you let me know when we can expect the next release?

laeubi:
I do releases mostly before the Eclipse Releases if there is something interesting to include. The best is always to 
try out the current [tycho snapshot](https://github.com/eclipse/tycho/wiki#getting-tycho-snapshots) builds so things like this surface before.

PR #5887, opened 2026-03-09
Fix for 5886

mlehmannm:
Please backport to tycho-5.0.x as well.


laeubi:
@mlehmannm Version bumps are for bundles- not packages... so maybe you want to explain a bit more context.
Then it would be good to provide a integration test similar to the existing ones to be sure it actually works (and not break in the future).

mlehmannm:
We develop our software in feature branches which takes sometimes weeks or months. In these feature branches we do not bump versions at all. This happens after merging the feature branches back to develop and right before opening a release branch.

We have a script that calls `bump-versions` (together with the baseline check) and loops until the build succeed.

With tycho/5.0.2 we get logs like this

> 11:47:56 [ERROR] /path/to/bundle [0:0]: Baseline problems found! Project version: 2.1.0.v20260306-1005, baseline version: 2.0.0.v20250813-0651, suggested version: 2.1.0

This looks like bogus, because the bundle version has been changed already, but if you look closer/earlier in the log, one can see that bnd suggests changing some package version(s) as well.

After [5567](https://github.com/eclipse-tycho/tycho/pull/5567) we now have to run `bump-versions` and manually sync the bundle-version to the version of exported packages and start the script again. This process as expensive (while manual) and error-prone.

This change only restores (via configuration) the previous behaviour.

Is there another way to create the necessary changes to the manifests to make the baseline-check happy?

laeubi:
That's tricky - I think here there is a direct suggestion to update a package (what is different to what https://github.com/eclipse-tycho/tycho/pull/5567 aimed for).

My expectation would be that this works without adding special configuration - do you think you can try to came up with a reproducer of a problem so we can maybe fix it instead of having a workaround here?

mlehmannm:
Unfortunately our software is closed-source.

But I can provide an excerpt from the build log. See below.

The first delta is already the problem. Bnd suggest raising the version for the package to 1.2.0. Tycho seems the treat this version as new bundle-version.

> 11:46:37 [INFO] --- tycho-baseline:5.0.2:verify (baseline-check) @ com.aaaaaaa.bbbbbbb.cccc.model ---
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] ?Delta                     ?Type                      ?Name                      ?Project Version          ?Baseline Version         ?Suggested Version        ?
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] ?MINOR                     ?PACKAGE                   ?com.aaaaaaa.bbbbbbb.cccc.m?1.1.5                    ?1.1.5                    ?1.2.0                    ?
> 11:46:37 [ERROR] ?                          ?                          ?odel                      ?                         ?                         ?                         ?
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] ?..MINOR                   ?CLASS                     ?com.aaaaaaa.bbbbbbb.cccc.model.SomeImpl                                                                 ?
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] ?....ADDED                 ?METHOD                    ?isWhatever()                                                                                            ?
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] ?....ADDED                 ?METHOD                    ?setWhatever(boolean)                                                                                    ?
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] ?MICRO                     ?MANIFEST                  ?com.aaaaaaa.bbbbbbb.cccc.m?1.2.0                    ?1.1.5                    ?1.1.6                    ?
> 11:46:37 [ERROR] ?                          ?                          ?odel-1.2.0-SNAPSHOT.jar   ?                         ?                         ?                         ?
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] ?..REMOVED                 ?HEADER                    ?Bundle-Version:1.1.5                                                                                    ?
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] ?..ADDED                   ?HEADER                    ?Bundle-Version:1.2.0                                                                                    ?
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] ?..CHANGED                 ?HEADER                    ?Import-Package                                                                                          ?
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] ?MICRO                     ?RESOURCES                 ?com.aaaaaaa.bbbbbbb.cccc.m?1.2.0                    ?1.1.5                    ?1.1.6                    ?
> 11:46:37 [ERROR] ?                          ?                          ?odel-1.2.0-SNAPSHOT.jar   ?                         ?                         ?                         ?
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] ?..CHANGED                 ?CLASS                     ?com/aaaaaaa/bbbbbbb/cccc/model/SomeImpl.class                                                           ?
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] ?..CHANGED                 ?CLASS                     ?com/aaaaaaa/bbbbbbb/cccc/model/gson/SomeImplAdapter.class                                               ?
> 11:46:37 [ERROR] ????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????????
> 11:46:37 [ERROR] /path/to/bundle/com.aaaaaaa.bbbbbbb.cccc.model [0:0]: Baseline problems found! Project version: 1.2.0.v20260306-1005, baseline version: 1.1.5.v20250320-1315, suggested version: 1.2.0

mlehmannm:
Is there any chance to merge this simply PR in the near future and to release this as part of tycho 5.0.3?

laeubi:
I can try to take a look but can't promise anything.

Beside that, if it is crucial to your business you can find dedicated support options for Tycho here:
https://github.com/eclipse-tycho/tycho#getting-support

As mentioned I don't think adding such parameter is useful or should be necessary - and a testcase would still be required in any case.

mlehmannm:
The VersionEngine already has a test for whether updating packages or not. The PR only makes the option available in the VersionBumpMojo. The change is equally written as changing the increment in the same mojo!

laeubi:
> The VersionEngine already has a test for whether updating packages or not. The PR only makes the option available in the VersionBumpMojo. The change is equally written as changing the increment in the same mojo!

All test for Tycho currently pass without your change, so whatever issue you see it is obviously not covered by tests.

laeubi:
@mlehmannm I now created
- https://github.com/eclipse-tycho/tycho/pull/6011

can you confirm this closely enough resembles your described use-case?

> The first delta is already the problem. Bnd suggest raising the version for the package to 1.2.0. Tycho seems the treat this version as new bundle-version.

I think this is the crucial part here anyways, currently we only see "version bump required" but we do not distinguish between *what* actually needs to be bumped resulting in the actual problem that right now only a *bundle version bump* is assumed (and therefore skipping package updates is correct).

But it might perfectly be the case that the bundle version is already bumped (but not the package).

mlehmannm:
Thanks for your effort. I'll try to test your changes until end of week.
Is there a backport planned to 5.0.x?

laeubi:
Actually yes but ti seems not working automatically for some reason need to check...

PR #5933, opened 2026-03-19
Refactor o.e.t.versions.engine.Versions to o.e.t.helper.VersionTool

merks:
- Replace SourcesBundleDependencyMetadataGenerator.toCanonicalVersion with VersionTool.toCanonicalVersion and fix it to ensure that valid Maven versions such as 1-SNAPSHOT and 1.0-SNAPSHOT are handled just like 1.0.0-SNAPSHOT to produce 1.0.0.qualifier, i.e., ensure that it returns a valid OSGi version.

### DIFF
--- a/tycho-spi/src/main/java/org/eclipse/tycho/helper/VersionTool.java
+++ b/tycho-spi/src/main/java/org/eclipse/tycho/helper/VersionTool.java
@@ -11,24 +11,27 @@
  *    Sonatype Inc. - initial API and implementation
  *    Marco Lehmann-Mörz - issue #2877 - tycho-versions-plugin:bump-versions does not honor SNAPSHOT suffix
  *******************************************************************************/
-package org.eclipse.tycho.versions.engine;
+package org.eclipse.tycho.helper;
 
 import java.io.File;
 
 import org.eclipse.tycho.TychoConstants;
 import org.osgi.framework.Version;
 
-public class Versions {
+public class VersionTool {
+
     public static String toCanonicalVersion(String version) {
         if (version == null) {
             return null;
         }
-
         if (version.endsWith(TychoConstants.SUFFIX_SNAPSHOT)) {
-            return version.substring(0, version.length() - TychoConstants.SUFFIX_SNAPSHOT.length())
-                    + TychoConstants.SUFFIX_QUALIFIER;
+            String unqualifiedVersion = version.substring(0,
+                    version.length() - TychoConstants.SUFFIX_SNAPSHOT.length());
+            // Ensure that valid Maven versions such as 1-SNAPSHOT, 1.0-SNAPSHOT, and 1.0.0-SNAPSHOT all produce a valid OSGi version
+            // by padding missing minor and micro segments with 0.
+            // For all 3 cases it returns 1.0.0.qualifier.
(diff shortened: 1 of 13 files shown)


tycho-spi/src/main/java/org/eclipse/tycho/helper/VersionTool.java
```diff
@@ -11,24 +11,27 @@
...
-
         if (version.endsWith(TychoConstants.SUFFIX_SNAPSHOT)) {
-            return version.substring(0, version.length() - TychoConstants.SUFFIX_SNAPSHOT.length())
-                    + TychoConstants.SUFFIX_QUALIFIER;
+            String unqualifiedVersion = version.substring(0,
+                    version.length() - TychoConstants.SUFFIX_SNAPSHOT.length());
+            // Ensure that valid Maven versions such as 1-SNAPSHOT, 1.0-SNAPSHOT, and 1.0.0-SNAPSHOT all produce a valid OSGi version
+            // by padding missing minor and micro segments with 0.
+            // For all 3 cases it returns 1.0.0.qualifier.
+            return Version.valueOf(unqualifiedVersion) + TychoConstants.SUFFIX_QUALIFIER;
```
HannesWell, line 33: This change breaks the preparation of a new development cycle of Eclipse respectively the `tycho-versions:set-version` mojo in conjunction projects that reference a variable from a parent pom in its version definition:
That's currently the case for
https://github.com/eclipse-equinox/p2/blob/e3b8d670634ce7c93dc5f60cd2d697fdb9b219d5/pom.xml#L25

merks: It’s not clear how that’s  the case. What value is passed in?  Does it end with ‘-SNAPSHOT’?

HannesWell: To provide more context, in the case mentioned above the version read from the model at

https://github.com/eclipse-tycho/tycho/blob/030869a94697c2037c5b76f87be8728284687aa3/tycho-versions-plugin/src/main/java/org/eclipse/tycho/versions/manipulation/PomManipulator.java#L86

does not have all variables resolved, i.e. those obtained from the parent model. This then leads to Strings that are not proper versions:

grafik

merks: I see.  That’s unexpected.  It might even be the case that the qualifier expression too isn’t expanded and then no replacement would be done. Maybe it wouldn’t happen later then either?

laeubi: I'm not 100% sure that a CI friendly version even is allowed to reference variables from the parent. It seems we need a testcase for this and check if it works in maven at all.
In any case I think this was meant to fix cases where the version is "to short" and as the previous approach has worked maybe we just need to enhance the version replacement to allow both cases?

merks: Yes one might look for a $ or catch the exception and do the string thing.  But there should be test cases.

Issue #4517, opened 2024-12-07
Add ability to define a "top level" category when mirroring repos using tycho-p2-extras-plugin


andrew-tram:
A feature request based off of this - https://github.com/eclipse-tycho/tycho/discussions/4422

Currently when mirroring repos, using `tycho-p2-extras-plugin:mirror`, if available, the `org.eclipse.equinox.p2.name` from the respective repositories are used and displayed at the top level.

I'd like the option to define my own "top-level" category that then expands to the repos that are being mirrored.

Example:

- My Eclipse projects
--- Buildship Extra
--- Build Extras - Incubating
--- Buildship: Eclipse Plugins for Gradle
--- etc etc etc

andrew-tram:
Making another push for this feature.

laeubi:
@andrew-tram do you like to propose a PR for this feature?

Beside that, if it is crucial to your business you can find dedicated support options for Tycho here:
https://github.com/eclipse-tycho/tycho#getting-support

andrew-tram:
@laeubi , sure thing. I'll look into implementing this.

laeubi:
FYI @HannesWell who has worked on improving udpate sites recently.

PR #5936, opened 2026-03-25
Add ability to define a "top level" category when mirroring repos using tycho-p2-extras-plugin

andrew-tram:
PR to address - https://github.com/eclipse-tycho/tycho/issues/4517

When mirroring multiple p2 repositories into a single destination using the `mirror` goal, each source repository's category IUs appear as separate top-level entries in Eclipse's "Install New Software" dialog. This change adds a new optional `<categoryName>` parameter that injects a synthetic root category IU into the destination repository, causing all mirrored source categories to appear as children of a single named entry instead.


andrew-tram:
Oops, I think I was supposed to created the PR for main. Would a maintainer be able to change that? I dont seem to see that option on my end.

merks:
I'm not sure what's correct. 

I think you can click the little pencil:

image

And then you can change the branch like this:

image

laeubi:
@andrew-tram I changed it to main but I think you need to rebase the branch.

laeubi:
@andrew-tram do you plan on continuing on this? It might be easier to start fresh by creating a new branch from main, then copy your changes over and create a new PR. We can then just back-port it automatically (hopefully) to tycho-5.

I plan to make a release soon so if we can complete this in time the improvement can be included.

andrew-tram:
@laeubi , thanks for the reminder. I'll try to get this going in the next day or two.

PR #5971, opened 2026-04-20
Distinguish cache hits from real downloads in p2 transport log

vogella:
Tycho's p2 transport previously logged `Downloading from <url>` for every resource access, even when the response was served entirely from the local p2 bundle cache, making warm-cache builds look like they were downloading gigabytes and hiding real network activity in CI logs. Logging now distinguishes cache hits (`Fetched from cache`, DEBUG), 304 revalidations (`Up-to-date`, DEBUG), and real downloads (`Downloading from`, INFO), and the misleading `Downloaded from ... at X/s` summary is suppressed when nothing was actually fetched. Transport behavior is otherwise unchanged; this is a logging-only fix that makes warm-cache builds quiet and real downloads easy to spot.


laeubi:
> Tycho's p2 transport currently logs Downloading from <url> whenever a resource is requested, even when the response is served entirely from the local p2 bundle cache

This is actually desired to make it mostly similar how Maven works. As you can see you can even suppress these with `-B` (non interactive mode) if you like.

> and makes it hard to spot real download activity in a CI log

This logging is not meant as a debug facility of the caching framework (for what we have system properties available).

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/TychoRepositoryTransport.java
```diff
@@ -108,18 +111,18 @@ public IStatus downloadArtifact(URI source, OutputStream target, IArtifactDescri
 			}
 		}
 		String id = "p2"; // TODO we might compute the id from the IRepositoryIdManager based on the URI?
-		if (cacheConfig.isInteractive()) {
-			logger.info("Downloading from " + id + ": " + source);
-		}
```
laeubi, line 113: This log message should remain to stay consistent with how maven works and to know that the download has started.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/TychoRepositoryTransport.java
```diff
@@ -80,6 +80,9 @@ public Thread newThread(Runnable r) {
 	@Requirement
 	TransportCacheConfig cacheConfig;
 
+	@Requirement
+	HttpCache httpCache;
```
laeubi, line 84: We should not bind directly to a specific cache implementation there might be other caching techniques see comment below.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/TychoRepositoryTransport.java
```diff
@@ -108,18 +111,18 @@ public IStatus downloadArtifact(URI source, OutputStream target, IArtifactDescri
...
 		try {
 			DownloadStatusOutputStream statusOutputStream = new DownloadStatusOutputStream(target,
 					"Download of " + source);
 			stream(source, monitor).transferTo(statusOutputStream);
 			DownloadStatus downloadStatus = statusOutputStream.getStatus();
 			if (cacheConfig.isInteractive()) {
-				logger.info("Downloaded from " + id + ": " + source + " ("
-						+ FileUtils.byteCountToDisplaySize(downloadStatus.getFileSize()) + " at "
-						+ FileUtils.byteCountToDisplaySize(downloadStatus.getTransferRate()) + "/s)");
+				CacheState state = httpCache.getLastCacheState(source);
```
laeubi, line 120: Instead of using a side-car approach here, the information should be transported through the `DownloadStatusOutputStream`

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/TychoRepositoryTransport.java
```diff
@@ -108,18 +111,18 @@ public IStatus downloadArtifact(URI source, OutputStream target, IArtifactDescri
...
+				if (state == CacheState.DOWNLOADED || state == CacheState.UNKNOWN) {
+					logger.info("Downloaded from " + id + ": " + source + " ("
```
laeubi, line 122: The info should always be logged and we should just replace the transfer rate by a status e.g. 

Downloaded from " + id + ": " + source + " ("
						+ FileUtils.byteCountToDisplaySize(downloadStatus.getFileSize()) + " (from cache)"

p2-maven-plugin/src/test/java/org/eclipse/tycho/p2maven/transport/SharedHttpCacheStorageLoggingTest.java
```diff
@@ -0,0 +1,196 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+/**
+ * Verifies that {@link SharedHttpCacheStorage} emits the correct log message
+ * for the three possible outcomes of a cache access: pure cache hit, 304
+ * revalidation, and real download.
+ */
+class SharedHttpCacheStorageLoggingTest {
```
laeubi, line 44: It has shown that mocking does not work well on the long run in Tycho as it always tests internal behavior what is hard to maintain.
Instead please add a real integration-test that uses e.g. a java 21 jdk server (we also have some examples using jetty already).
This proves it works in real world use-cases end-to-.end

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/CacheState.java
```diff
@@ -0,0 +1,26 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.tycho.p2maven.transport;
+
+/**
+ * Describes the outcome of a {@link HttpCache} access, so that callers can
+ * produce accurate log output (e.g. distinguish "Downloading" from "Fetched
+ * from cache").
+ */
+public enum CacheState {
```
laeubi, line 17: Please think about better naming here that is independent from caching and agnostic from the transport technique. e.g. we also have ftp transports that have no meaning of (http) status codes.

laeubi (changes requested):
See comments

laeubi:
@vogella do you plan to further work on this? I want to make a new Tycho release likely end of month and it could then be included if we finish before.

vogella:
Your initial comment indicates that this does not fit to Maven.

So you agree with the change now? I can finish it in this case

vogella:
PR has been updated

merks:
Please squash and rebase.

laeubi:
> So you agree with the change now?

I need to take a deeper look, using a thread local does not looks really nice here, we likely want to enhance some method signatures to directly pass the destination so it can simply do an instanceof check!

laeubi:
@vogella I have now take a deeper look and think we need to do it more systematically:

1. `org.eclipse.tycho.transport.TransportProtocolHandler.getFile(URI)` currently returns a `File` what is the root of all evil here. We should change it to return a record like `FileState(Path file, DownloadState state)` where `DownloadState` must represent all possible case we get on our path e.g. it could be `OFFLINE_FROM_CACHE`, `FROM_CACHE_UP_TO_DATE`, `NOT_MODIFIED`, ..., `DOWNLOAD_TO_CACHE`
2. Instead of `stream(source, monitor).transferTo(statusOutputStream);` we should have an internal method `createStream(source)` that returns a record `StreamState(InputStream stream, DownloadState state)` that can be called by `stream` and `downloadArtifact` so the caller can get this state infos passed through the chain.

This way we need no thread locals and alike and can directly access necessary information (and even give better feedback about what kind of cache hit we have).

Further the `SharedHttpCacheStorageIntegrationTest` is still not an integration test, see provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) for details how it should look like and where it needs to be placed.

vogella:
The ThreadLocal was a workaround precisely because the state had nowhere to live in the signatures.                                                

I follow your suggestion in the next ps.
                                                                                                                                                                                                                                                                                                          The "IntegrationTest" name was wrong; it's a unit test. I'll rename it to SharedHttpCacheStorageTest.

laeubi:
> The "IntegrationTest" name was wrong; it's a unit test. I'll rename it to SharedHttpCacheStorageTest.

Please make it a proper integration test (as requested!) 

> Further the `SharedHttpCacheStorageIntegrationTest` is still not an integration test, see provide an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) for details how it should look like and where it needs to be placed.

Testing internal behavior (and especially mocking) tightly couples tests to the implementation and we better want to test user observable behavior (e.g. the user does not care *how* the messages are generated).

tycho-spi/src/main/java/org/eclipse/tycho/transport/FileState.java
laeubi, line 1: As this file seems nowhere referenced in the SPI it could be moved to the p2 transport package (same for `StreamState`).

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/DownloadStatusOutputStream.java
laeubi, line 1: Seems to contain only whitespace changes

laeubi (commented):
I think this is quite close - a proper integration test and some minor cleanups are missing.

vogella:
Updated as requested

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/TychoRepositoryTransport.java
```diff
@@ -139,6 +146,23 @@ public IStatus downloadArtifact(URI source, OutputStream target, IArtifactDescri
...
+			}
+			if (e instanceof AuthenticationFailedException afe) {
+				throw afe;
+			}
+			throw new CoreException(statusWithCode("download from %s failed", ProvisionException.REPOSITORY_FAILED_READ,
+					toDownload, e));
+		}
+	}
+
+	InputStream openStream(URI toDownload, IProgressMonitor monitor, Consumer<DownloadState> stateSink)
```
laeubi: Should return a record of (InputStream, FileState) (that can implement AutoClosable for convenience) instead of using a consumer. Then we do not need an own `DownloadState` and no changes to the `DownloadStatusOutputStream`

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/TychoRepositoryTransport.java
```diff
@@ -139,6 +146,23 @@ public IStatus downloadArtifact(URI source, OutputStream target, IArtifactDescri
 	@Override
 	public InputStream stream(URI toDownload, IProgressMonitor monitor)
 			throws FileNotFoundException, CoreException, AuthenticationFailedException {
+		try {
+			return openStream(toDownload, monitor, state -> {
+			});
+		} catch (IOException e) {
```
laeubi: Previously the stream method already wrapped everything into core exception why do we need another catch here? one would expect this method to be a simple delegate returning the stream only.

laeubi (commented):
Integration test looks fine but the code can still need some improvements.

vogella:
Updated

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/DownloadStatusOutputStream.java
laeubi, line 1: Please revert whitespcae only change

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/HttpCache.java
laeubi, line 1: Whitespace only changes - please revert.

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/FileTransportProtocolHandler.java
```diff
@@ -17,21 +17,23 @@
...
+		return getFile(uri).file().toFile().lastModified();
 	}
 
 
 	@Override
-	public File getFile(URI remoteFile) throws IOException {
+	public FileState getFile(URI remoteFile) throws IOException {
 		try {
-			return new File(remoteFile);
+			return new FileState(new File(remoteFile).toPath(), DownloadState.FROM_CACHE);
```
laeubi:
```suggestion
			return new FileState(new File(remoteFile).toPath(), DownloadState.LOCAL_FILE);
```

laeubi (commented):
Some smaller nit-picks...

vogella:
Updated

laeubi (changes requested):
@vogella it would make things easier if you review the changes made before pushing if you are using the AI, or give at least rationale if you changed things intentionally as this PR is going back and forth. Why do we now have `TychoTransportProtocolHandler` and the FileState / DownloadState moved from spi to the maven-plugin + instanceof checks?

vogella:
> @vogella it would make things easier if you review the changes made before pushing if you are using the AI, or give at least rationale if you changed things intentionally as this PR is going back and forth. Why do we now have `TychoTransportProtocolHandler` and the FileState / DownloadState moved from spi to the maven-plugin + instanceof checks?

I think that happened when I was wrestling with the integration test. Once I find time I move it back to the spi API.

vogella:
I've reworked it the way you suggested:

- FileState and DownloadState are back in tycho-spi
- TransportProtocolHandler.getFile now returns FileState directly, so the TychoTransportProtocolHandler sub-interface and the instanceof checks are gone. All handlers implement the SPI interface directly again.
- StreamState stays in p2-maven-plugin, since it's internal to TychoRepositoryTransport and not referenced by the SPI

p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/transport/TychoRepositoryTransport.java
```diff
@@ -111,15 +113,17 @@ public IStatus downloadArtifact(URI source, OutputStream target, IArtifactDescri
...
 			if (cacheConfig.isInteractive()) {
+				String suffix = streamState.fileState().state().isFromCache()
+						? " (from cache)"
+						: " at " + FileUtils.byteCountToDisplaySize(downloadStatus.getTransferRate()) + "/s";
 				logger.info("Downloaded from " + id + ": " + source + " ("
-						+ FileUtils.byteCountToDisplaySize(downloadStatus.getFileSize()) + " at "
-						+ FileUtils.byteCountToDisplaySize(downloadStatus.getTransferRate()) + "/s)");
+						+ FileUtils.byteCountToDisplaySize(downloadStatus.getFileSize()) + suffix + ")");
 			}
 			return reportStatus(downloadStatus, target);
```
laeubi, line 133: One thing that came into my mind: We should also likely report the download rate as 0 her to not push mirrors up that where served from the cache.

vogella: Took me a time to figure out how to do that, but I agree, makes sense, I update the PR soon

vogella: Update available.

vogella:
Update available.

laeubi (approved):
This now all looks sane to me

PR #6016, opened 2026-05-18
materialize-products: deleteP2Cache on macOS Foo.app/Contents/Eclipse

sratz:
`deleteP2Cache` was not considering the macOS-specific substructure `<productRoot>/Foo.app/Contents/Eclipse`.


laeubi (approved):
Looks good to me, thanks for spotting!

PR #6034, opened 2026-05-19
Add ability to define a "top level" category when mirroring repos using tycho-p2-extras-plugin

andrew-tram:
PR to address - https://github.com/eclipse-tycho/tycho/issues/4517

When mirroring multiple p2 repositories into a single destination using the mirror goal, each source repository's category IUs appear as separate top-level entries in Eclipse's "Install New Software" dialog. This change adds a new optional <categoryName> parameter that injects a synthetic root category IU into the destination repository, causing all mirrored source categories to appear as children of a single named entry instead.

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2023 SAP SE and others.
+ * Copyright (c) 2010, 2026 SAP SE and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
@@ -19,11 +19,13 @@ import static java.util.stream.Collectors.mapping;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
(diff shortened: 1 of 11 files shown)


laeubi:
@andrew-tram This looks good in general can you transfor your unit test into an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) here as these are usually easier to maintain.

tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
```diff
@@ -405,6 +416,36 @@ private Stream<IInstallableUnit> getRepositoryContent(URI uri, IRepository<IInst
         return Stream.empty();
     }
 
+    private static void injectParentCategory(String categoryName, IMetadataRepository destRepo, Logger logger) {
+        IQueryResult<IInstallableUnit> allUnits = destRepo.query(QueryUtil.ALL_UNITS, null);
```
HannesWell: Couldn't the query be adjusted to just return categories respectively units with the type-category property being true? Then the check in the loop could be avoided and maybe P2 internally can find them more efficiently.

HannesWell (commented):
Sounds good to me, I just have one questions, suggestion below.

tycho-core/src/main/java/org/eclipse/tycho/p2tools/TychoMirrorApplication.java
andrew-tram: Oh that's true. Just pushed - https://github.com/eclipse-tycho/tycho/pull/6034/commits/0a9a91c8255907e852d902124f9f827818719ced

andrew-tram:
> @andrew-tram This looks good in general can you transfor your unit test into an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) here as these are usually easier to maintain.

https://github.com/eclipse-tycho/tycho/pull/6034/changes/6d937b2fce6609b0d98e3b3f22279935f52aadbb - I hope that looks better

laeubi (approved):
Looks good to me - can you squash all commits into one with a descriptive commit message?

andrew-tram:
I believe it should be good to go now

HannesWell (commented):
Looks sound to me too (in a brief quick review).

Issue #6041, opened 2026-05-24
Features fail to build with CI-Friendly versions


basilevs:
# Test steps
- Unzip [repro.zip](https://github.com/user-attachments/files/28190573/repro.zip) (minimized reproducer)
- Ensure it builds with `mvn clean verify`
- Try `mvn clean verify -Drevision=0.0.7-SNAPSHOT`

# Expected result
Build success

# Observed result
```
[INFO] --- tycho-packaging:5.0.2:validate-version (default-validate-version) @ org.basilevs.jstackfilter.eclipse.feature ---
[ERROR] /Users/agent/git/jstackfilter/feature.xml [0:0]: Unqualified OSGi version 0.0.6.qualifier must match unqualified Maven version 0.0.7-SNAPSHOT for SNAPSHOT builds
```

ZIP contains:
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">

	<modelVersion>4.0.0</modelVersion>
	<groupId>org.basilevs.jstackfilter</groupId>
	<artifactId>org.basilevs.jstackfilter.eclipse.feature</artifactId>
	<version>${revision}</version>
	<packaging>eclipse-feature</packaging>

	<properties>
		<revision>0.0.6-SNAPSHOT</revision>
		<tycho.version>5.0.2</tycho.version>
	</properties>

	<build>
		<plugins>
			<plugin>
				<groupId>org.eclipse.tycho</groupId>
				<artifactId>tycho-maven-plugin</artifactId>
				<version>${tycho.version}</version>
				<extensions>true</extensions>
			</plugin>
		</plugins>
	</build>
</project>
```
```xml
<?xml version="1.0" encoding="UTF-8"?>
<feature
      id="org.basilevs.jstackfilter.eclipse.feature"
      label="Jstackfilter"
      version="0.0.6.qualifier">

   <description url="https://github.com/basilevs/jstackfilter">
      Do not show idle threads in Debug view.
   </description>

</feature>
```

laeubi:
I don't think this can play well. If you really want the artifact to have a different maven-version you need to disable the version check.

basilevs:
> a different maven-version

Could you clarify this? Different from the version on-disk? Or different between Maven and P2?

[...]

PR #6042, opened 2026-05-24
Add integration test for CI-friendly version validation

basilevs:
Reproduces the validate-version failure when using CI-friendly ${revision} property and overriding it on the command line with a different version than what is in feature.xml.

The test verifies that building an eclipse-feature project with -Drevision=0.0.7-SNAPSHOT succeeds even though feature.xml contains a previous version 0.0.6.qualifier.

See #6041

Existing tests mostly verified qualifiers. This one touches release cycle.

Used https://github.com/eclipse-tycho/tycho/blob/5387e2f9f8aa9f213fc4252280ad6c0fa2e79b9b/.github/skills/tycho-its/SKILL.md

Got strange copyright header, otherwise ok.

### DIFF
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/buildextension/CiFriendlyVersionsTest.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/buildextension/CiFriendlyVersionsTest.java
@@ -17,6 +17,8 @@ import static org.junit.Assert.assertTrue;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.time.format.DateTimeFormatter;
 import java.util.Calendar;
 import java.util.List;
@@ -24,6 +26,7 @@ import java.util.TimeZone;
 import java.util.jar.JarFile;
 
 import org.apache.maven.it.Verifier;
+import org.eclipse.tycho.model.Feature;
 import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
 import org.junit.Test;
 import org.osgi.framework.Constants;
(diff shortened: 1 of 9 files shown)


laeubi:
@basilevs it would be good to add the testcase to the existing `CiFriendlyVersionsTest` - also you claimed its working for bundles--- it would therefore be good to have a bundle + a feature in the test set so one can see the outcome - e.g one might use `-pl` to first test the bundle builds, then the feature builds and then they build all together.

tycho-its/src/test/java/org/eclipse/tycho/test/buildextension/CiFriendlyVersionsTest.java
laeubi, line 1: To make it explicit (and prevent from surprise) one would best verify the p2 version an maven version from the jars manifest/feature file and the pom.properties.

basilevs: Not sure if got you right. Please check the latest commit.

basilevs:
> also you claimed its working for bundles--- it would therefore be good to have a bundle + a feature in the test set so one can see the outcome

Turns out, it only "works" for bundles in a sense that a pomless bundle build passes. The generated bundle has incorrect version without regard for CI-Friendly override.

Pomful builds fail, so "working" bundles were a red herring.

tycho-its/src/test/java/org/eclipse/tycho/test/buildextension/CiFriendlyVersionsTest.java
```diff
@@ -116,4 +119,101 @@ private static void checkManifestVersion(File file, String expectedVersion) thro
...
+		Path bundleJar = basedir
+				.resolve("bundles/org.example.bundle/target/org.example.bundle-0.0.6-SNAPSHOT.jar");
+		assertTrue("Bundle jar not found: " + bundleJar, Files.isRegularFile(bundleJar));
+		try (JarFile jar = new JarFile(bundleJar.toFile())) {
+			String bundleVersion = jar.getManifest().getMainAttributes().getValue(Constants.BUNDLE_VERSION);
+			assertTrue("Bundle-Version should start with 0.0.6: " + bundleVersion,
+					bundleVersion.startsWith("0.0.6."));
+		}
+
+		// Post-build: verify built feature jar has expanded version
```
basilevs, line 149: reword: original version

basilevs:
Moved to https://github.com/eclipse-tycho/tycho/pull/6056

PR #6052, opened 2026-05-27
Correct central search property name in SBOM.md

DieterMai:
Property name of central search was wrong in SBOM.md.

see [ticket](https://github.com/eclipse-tycho/eclipse-tycho.github.io/issues/3)


laeubi (approved):
Thanks for spotting

PR #6062, opened 2026-05-29
Deduplicate versions by major.minor.micro in EclipseIndexArtifactVersionProvider

akurtakov:
Retain only the highest-qualifier entry per major.minor.micro before streaming results to the checkers. Checking multiple bundles that differ only by qualifier is a total loss of time as requirements can not contain the qualifier.
Speeds up significantly in e.g. cases containing org.eclipse.swt as Orbit index contains I/Y-builds too.

### DIFF
--- a/tycho-baseline-plugin/src/main/java/org/eclipse/tycho/baseline/provider/EclipseIndexArtifactVersionProvider.java
+++ b/tycho-baseline-plugin/src/main/java/org/eclipse/tycho/baseline/provider/EclipseIndexArtifactVersionProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2025 Christoph Läubrich and others.
+ * Copyright (c) 2025, 2026 Christoph Läubrich and others.
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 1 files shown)


merks:
If someone specified in their *.target an exact version that's older than some other available version, would such a change break that?  It was very hard to get rid of the logging bundle with the qualifier to replace it with the one without a qualifier, which only worked by specifying explicitly in the target platform the one without the qualifier so I'm a bit concerned about an assumption here...

akurtakov:
This is used only by https://tycho.eclipseprojects.io/doc/main/tycho-baseline-plugin/check-dependencies-mojo.html aka a post build validation check that is way too slow in some cases. It should not affect anything else.

merks:
Then I think it makes good sense!

laeubi:
The main problem is that the eclipse index seem to also contain I/Y/whatever builds. Even though qualifier only changes should not change API we can not really enforce it, in general it might be better and speed up things even more if we can skip certain updatesites by a pattern.

akurtakov:
I like your proposal as extra step to speed it up even more in addition to this one.

akurtakov:
@laeubi So what do you say - should we merge this one or close the PR?

akurtakov:
If there is no objection I plan to merge on Tuesday.

PR #6138, opened 2026-06-27
Add debugPort parameter to bnd-test Mojo

HannesWell:
Similar to the existing `debugPort` parameter of the other `test`/`verify` mojo of this `tycho-surefire-plugin`.

### DIFF
--- a/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/BndTestMojo.java
+++ b/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/BndTestMojo.java
@@ -192,6 +192,13 @@ public class BndTestMojo extends AbstractTestMojo {
     @Parameter(defaultValue = ENGINES_DEFAULT, required = true)
     private String testEngines;
 
+    /**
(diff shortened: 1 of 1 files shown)


tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/BndTestMojo.java
```diff
@@ -192,6 +192,13 @@ public class BndTestMojo extends AbstractTestMojo {
     @Parameter(defaultValue = ENGINES_DEFAULT, required = true)
     private String testEngines;
 
+    /**
+     * Set this parameter to suspend the test JVM waiting for a client to open a remote debug
+     * session on the specified port.
+     */
+    @Parameter(property = "debugPort")
+    private int debugPort;
```
HannesWell, line 200: In general we could also make this more flexible and convert it to a string, which is then passed as it is to `-runjdb`: https://bnd.bndtools.org/instructions/runjdb.html

But then we should also explain that using only a port is less secure and that one has to prepend a `+` to suspend the test JVM until the debugger is attached.
And I assume usually one wants the debuggee to wait for a debugger.

PR #6144, opened 2026-07-01
Make p2 to Maven metadata handling configurable (#6145)

andrew-tram:
https://github.com/eclipse-tycho/tycho/issues/6145

Controls how Tycho maps p2 units to their Maven coordinates:
 
- `validate` (default): map and resolve each artifact against the configured Maven repositories
- `inject`: map coordinates but do not resolve them remotely
- `ignore`: do not map p2 units to Maven coordinates
Setting `inject` or `ignore` avoids a remote Maven lookup per artifact (and the accompanying `Mapping P2 > Maven Coordinates failed` warnings), which can significantly speed up builds with large target platforms.
 
```xml
<plugin>
  <artifactId>target-platform-configuration</artifactId>
  <configuration>
    <p2MavenMetadataHandling>inject</p2MavenMetadataHandling>
  </configuration>
</plugin>
```


merks:
This sounds useful.

merks:
This looks good to me and I will appreciating have this option!

andrew-tram:
@laeubi , shall I update branch for review?

tycho-core/src/test/java/org/eclipse/tycho/core/resolver/DefaultTargetPlatformConfigurationReaderTest.java
laeubi, line 1: Please make this a proper [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) see also https://github.com/eclipse-tycho/tycho/blob/main/.github/skills/tycho-its/SKILL.md

This also allows to see how the feature might be used. If you think the already added one is sufficient we can just remove this, there is no need to test internal behavior.

laeubi (changes requested):

tycho-its/src/test/java/org/eclipse/tycho/test/target/P2MavenMetadataHandlingTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
+import org.eclipse.tycho.test.util.ResourceUtil.P2Repositories;
+import org.junit.Test;
+
+public class P2MavenMetadataHandlingTest extends AbstractTychoIntegrationTest {
+
+	@Test
+	public void ignoreSkipsMapping() throws Exception {
+		Verifier verifier = getVerifier("target.p2MavenMetadataHandling", false);
+		verifier.addCliOption("-Dp2.repo=" + P2Repositories.ECLIPSE_LATEST.toString().replace("/", "//"));
```
laeubi: Why not use set target platform = true is it reuired to have an own property for this?

andrew-tram: True! I addressed this.

tycho-core/src/test/java/org/eclipse/tycho/core/resolver/DefaultTargetPlatformConfigurationReaderTest.java
andrew-tram: Reworked this as well.

tycho-core/src/main/java/org/eclipse/tycho/core/resolver/DefaultTargetPlatformConfigurationReader.java
```diff
@@ -275,9 +292,9 @@ private void setClasspathDependencies(TargetPlatformConfiguration result, Xpp3Do
         try {
             result.setClasspathDependenciesAction(ClasspathDependenciesAction.valueOf(value.toUpperCase()));
         } catch (IllegalArgumentException e) {
-            throw new BuildFailureException("Illegal value of <" + CLASSPATH_DEPENDENCIES
-                    + "> dependency resolution parameter: " + value + ", allowed values are: "
-                    + Arrays.toString(ClasspathDependenciesAction.values()));
+            throw new BuildFailureException(
+                    "Illegal value of <" + CLASSPATH_DEPENDENCIES + "> dependency resolution parameter: " + value
+                            + ", allowed values are: " + Arrays.toString(ClasspathDependenciesAction.values()));
         }
```
laeubi: This looks like a formatting only change

tycho-its/src/test/java/org/eclipse/tycho/test/target/P2MavenMetadataHandlingTest.java
```diff
@@ -0,0 +1,40 @@
...
+public class P2MavenMetadataHandlingTest extends AbstractTychoIntegrationTest {
+
+	@Test
+	public void injectMapsCoordinatesWithoutRemoteResolution() throws Exception {
+		buildWithHandling("inject");
+	}
+
+	@Test
+	public void ignoreSkipsMapping() throws Exception {
+		buildWithHandling("ignore");
```
laeubi: You should assert something here - e.g. something has to be *not* there if ignore is used...

tycho-its/src/test/java/org/eclipse/tycho/test/target/P2MavenMetadataHandlingTest.java
```diff
@@ -0,0 +1,40 @@
...
+ * Demonstrates the {@code <p2MavenMetadataHandling>} target-platform-configuration option, which
+ * controls whether p2 units are mapped to (and resolved against) Maven coordinates. Building a
+ * bundle against a real p2 target platform with {@code inject} or {@code ignore} lets Tycho skip the
+ * per-artifact remote Maven lookup that {@code validate} (the default) performs.
+ */
+public class P2MavenMetadataHandlingTest extends AbstractTychoIntegrationTest {
+
+	@Test
+	public void injectMapsCoordinatesWithoutRemoteResolution() throws Exception {
+		buildWithHandling("inject");
```
laeubi: You should assert something here, e.g. something has to be present if *inject* is used.

tycho-core/src/test/java/org/eclipse/tycho/core/resolver/DefaultTargetPlatformConfigurationReaderTest.java
laeubi: @andrew-tram I can't really see what changed here or is reworked, this is still a bit bogus junit test that mostly checks some setters.

It feels like the test can be removed as such thing is already covered by the P2MavenMetadataHandler integration test (if we add some assertion there se previous comments).

If you are working with an AI tool, please make sure to read and follow https://docs.github.com/en/copilot/tutorials/review-ai-generated-code ... this tends to go in the same direction here as another PR where the AI has "addressed" review comments by changing code back-and-forth (so ignoring previous comments/state) what makes it quite annoying for reviewers if users do not check the outcome is actually addressing it in a consistent way.

laeubi (changes requested):
See comments below.

andrew-tram:
Updates made. I did this myself and used Claude to add comments and make things kosher. Will check out that link.

laeubi:
> Updates made. I did this myself and used Claude to add comments and make things kosher. Will check out that link.

Now **every test is deleted** instead of keeing the valuable (and required) integration test and enhance it. I can really only suggest **NOT** allowing claude to modify your PR and act on your behalf this is giving clear negative results here!

Instead apply changes locally, review them if they make sense and only apply them to the PR if you are confident and cross checked they addressed the reviewers comments/intends. This is exactly the same hassle we had with https://github.com/eclipse-tycho/tycho/pull/5971 what ended up with almost 40 comments where 90% where not constructive as the AI totally messed around back and forth...

andrew-tram:
Yes, I deleted it on purpose because I noticed something odd last night. Anyway, I'll just put a pin in this PR until I have more time again in the future.

laeubi:
> Yes, I deleted it on purpose because I noticed something odd last night. Anyway, I'll just put a pin in this PR until I have more time again in the future.

It would be good to share your findings so we can help here (e.g. in form of a class comment) instead of removing such file entirely because in general it looked sound and useful.

As we likely will do a release for Tycho in coordination with the upcoming Eclipse release, it would be good to decide if this is still important enough for you to complete in the meanwhile.

PR #6148, opened 2026-07-03
Only require IDE application bundle for UI harness default application

vogella:
When `useUIHarness` is enabled, tycho-surefire-plugin unconditionally adds `org.eclipse.ui.ide.application` to the test runtime requirements, even when a custom `application` is configured. This makes it impossible to run UI tests against RCP applications whose target platform does not contain the Eclipse IDE bundles, because dependency resolution fails before the configured application is ever considered.

With this change the IDE application bundle is only required when no `application` is configured, i.e. when the harness can actually fall back to the default `org.eclipse.ui.ide.workbench`. For a custom application only `org.eclipse.ui.workbench` is required, which the UI test harness itself needs. This allows RCP products to run UI tests without pulling the Eclipse IDE into their target platform.

The integration test `surefire.uiHarness.customApplication` demonstrates this: its target platform filters out `org.eclipse.ui.ide.application` and it configures a custom application. Without the fix the build already fails while resolving the test runtime; with it the runtime resolves, Equinox starts and reports `org.eclipse.ui.workbench` but no IDE bundle, and the launch only stops at the intentionally unknown application id. The test is headless, since actually running tests in a custom RCP application would need a real workbench and therefore an X display, which the ITS job does not provide.

### DIFF
--- a/tycho-surefire/tycho-surefire-plugin/src/test/java/org/eclipse/tycho/surefire/TestMojoTest.java
+++ b/tycho-surefire/tycho-surefire-plugin/src/test/java/org/eclipse/tycho/surefire/TestMojoTest.java
@@ -33,6 +33,7 @@ import org.codehaus.plexus.util.ReflectionUtils;
 import org.eclipse.sisu.equinox.launching.DefaultEquinoxInstallationDescription;
 import org.eclipse.sisu.equinox.launching.internal.DefaultEquinoxInstallation;
 import org.eclipse.sisu.equinox.launching.internal.EquinoxLaunchConfiguration;
+import org.eclipse.tycho.ArtifactKey;
 import org.junit.Test;
 
 public class TestMojoTest {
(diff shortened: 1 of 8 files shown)


tycho-surefire/tycho-surefire-plugin/src/test/java/org/eclipse/tycho/surefire/TestMojoTest.java
laeubi, line 1: Please add an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) to demonstrate the feature, the unit test are only quite weak, easy to break and could not be executed as a standalone maven build what makes them often not very useful.

e.g in this case you only test that the dependencies contain a certain bundle but this does not prove that the test runtime can start up successful.

For AI agents the repository even [contains a skill](https://github.com/eclipse-tycho/tycho/blob/main/.github/skills/tycho-its/SKILL.md) to create such tests.

laeubi (changes requested):

laeubi:
@vogella do you plan to continue on this one?

vogella:
Yes, but still afk for the next two weeks

tycho-surefire/tycho-surefire-plugin/src/test/java/org/eclipse/tycho/surefire/TestMojoTest.java
vogella: Thanks, IT test added. FYI - your skill now located in .agents/skills/tycho-its/SKILL.md still suggests to use JUnit 4, while the existing IT tests use JUnit5.

vogella:
IT tests added

Issue #1494, opened 2022-10-09
Replace Plexus Components by plain JSR 330 Components


kwin:
Plexus Components are deprecated (https://codehaus-plexus.github.io/plexus-containers/) and should be replaced by plain JSR 330 components (https://github.com/eclipse/sisu.plexus/wiki/Plexus-to-JSR330).

Here are some examples of JSR 330 components:

1. lifecycle mapping: https://github.com/apache/maven-rar-plugin/blob/master/src/main/java/org/apache/maven/plugins/rar/internal/RarLifecycleMappingProvider.java
2. artifact handler: https://github.com/apache/maven-rar-plugin/blob/master/src/main/java/org/apache/maven/plugins/rar/internal/RarArtifactHandler.java

(there is no component descriptor for JSR 330 as in https://codehaus-plexus.github.io/plexus-containers/plexus-container-default/plexus-components.html, therefore all `component.xml` files need to be replaced by Java code)

laeubi:
@kwin Currently Maven 3.8.x contain JSR 330 implementation that is not compatible so we have to wait for Maven 3.9.x release here.

HannesWell:
> @kwin Currently Maven 3.8.x contain JSR 330 implementation that is not compatible with Java 17 so we have to wait for Maven 3.9.x release here.

Are you referring to [Eclipse Sisu-Inject 0.3.5](https://github.com/eclipse/sisu.inject/releases/tag/releases%2F0.3.5) as implementation? It looks like support for later Java-versions was already added, but it has not yet been released.
So unless this happens before Maven 3.9.x, which then has to be included, we have to wait longer.

kwin:
@HannesWell https://issues.apache.org/jira/browse/MNG-7502 is required for Java 17 support. Not sure if https://github.com/eclipse/sisu.inject/commit/4790d3e28987ee4c2472d576e544c07028a85f42 is required as well (the latter would require a new sisu.inject release….

HannesWell:
> @HannesWell https://issues.apache.org/jira/browse/MNG-7502 is required for Java 17 support. Not sure if [eclipse/sisu.inject@4790d3e](https://github.com/eclipse/sisu.inject/commit/4790d3e28987ee4c2472d576e544c07028a85f42) is required as well (the latter would require a new sisu.inject release….

[...]

PR #6191, opened 2026-07-25
Migrate Plexus components to JSR330 annotations in tycho-sbom

HannesWell:
Part of
- https://github.com/eclipse-tycho/tycho/issues/1494

### DIFF
--- a/tycho-its/projects/sbom/pom.xml
+++ b/tycho-its/projects/sbom/pom.xml
@@ -68,7 +68,7 @@
 			<plugin>
 				<groupId>org.cyclonedx</groupId>
 				<artifactId>cyclonedx-maven-plugin</artifactId>
-				<version>2.7.9</version>
+				<version>2.9.3</version>
 				<configuration>
 					<includeTestScope>true</includeTestScope>
 				</configuration>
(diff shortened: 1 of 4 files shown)


HannesWell:
Is it possible that the cyclonedx-maven-plugin has to migrate off the plexus component annotations first before this can work?
https://github.com/CycloneDX/cyclonedx-maven-plugin/tree/master/src/main/java/org/cyclonedx/maven

HannesWell:
With https://github.com/CycloneDX/cyclonedx-maven-plugin/pull/674 being submitted, this change should work and be ready as soon as a new release of the cyclondx-maven-plugin is available (see also https://github.com/eclipse-tycho/tycho/pull/6195#issuecomment-5097308580).

The only change necessary here was to use `javax.annotation.Priority` instead of `org.eclipse.sisu.Priority`, with the latter the tycho overrides where not sorted before the default components.

HannesWell:
This is now finally working and ready for submission.

I had set `@Priority(999)` because when debugging the relevant code paths in SiSu I noticed that default components get ascending priority values. As far as I can tell, the default priority is therefore not predictable/stable and the number should be higher than the number of all available default components in the Maven session.
@cstamas or @laeubi, do you think it would make sense (and work) to set the priority of the default components in the cyclonedx maven plugin to a fixed value like `1`?

PR #6195, opened 2026-07-26
Override CycloneDX components

cstamas:
This is just a draft, interested in ITs.

Rework of https://github.com/eclipse-tycho/tycho/pull/6191

ITs has added category of `SBOM` so for quick turnaround this works:
```
mvn -f tycho-its/ clean verify -Dgroups=org.eclipse.tycho.test.categories.SBOM
```

### DIFF
--- a/tycho-build/src/main/resources/META-INF/maven/extension.xml
+++ b/tycho-build/src/main/resources/META-INF/maven/extension.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<extension>
+  <exportedPackages>
+    <exportedPackage>com.google.inject.*</exportedPackage>
+    <exportedPackage>com.google.inject.binder.*</exportedPackage>
+    <exportedPackage>com.google.inject.matcher.*</exportedPackage>
+    <exportedPackage>com.google.inject.name.*</exportedPackage>
+    <exportedPackage>com.google.inject.spi.*</exportedPackage>
+    <exportedPackage>com.google.inject.util.*</exportedPackage>
+    <exportedPackage>com.google.inject.internal.*</exportedPackage>
+  </exportedPackages>
+</extension>
--- a/tycho-its/projects/sbom/.mvn/maven.config
+++ b/tycho-its/projects/sbom/.mvn/maven.config
@@ -1,1 +1,1 @@
--Dtycho-version=5.0.0-SNAPSHOT
+-Dtycho-version=6.0.0-SNAPSHOT
--- a/tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoModelConverter.java
+++ b/tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoModelConverter.java
@@ -21,14 +21,13 @@ import java.util.List;
 import java.util.function.Supplier;
 
 import javax.inject.Inject;
+import javax.inject.Provider;
 
 import org.apache.maven.RepositoryUtils;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.plugin.LegacySupport;
 import org.apache.maven.project.MavenProject;
-import org.codehaus.plexus.component.annotations.Component;
 import org.cyclonedx.maven.DefaultModelConverter;
-import org.cyclonedx.maven.ModelConverter;
 import org.eclipse.aether.artifact.Artifact;
 import org.eclipse.equinox.p2.core.ProvisionException;
 import org.eclipse.equinox.p2.metadata.IArtifactKey;
@@ -52,14 +51,14 @@ import org.slf4j.LoggerFactory;
 /**
  * Custom implementation of the CycloneDX model converter with support for both
  * Maven and p2 artifacts. The generated PURL is usually of the form:
- * 
+ *
  * <pre>
  * pkg:/p2/&lt;id&gt;@&lt;version&gt;?classifier=&lt;classifier&gt;&amp;location=&lt;download-url&gt;
  * </pre>
- * 
+ *
  * This converter can be used with the {@code cyclonedx-maven-plugin} by adding
  * it as a dependency as follows:
- * 
+ *
  * <pre>
  * &lt;plugin>
  *   &lt;groupId&gt;org.cyclonedx&lt;/groupId&gt;
@@ -73,7 +72,6 @@ import org.slf4j.LoggerFactory;
  * &lt;/plugin&gt;
  * </pre>
  */
-@Component(role = ModelConverter.class)
 public class TychoModelConverter extends DefaultModelConverter {
 	private static final String KEY_CONTEXT = TychoSBOMConfiguration.class.toString();
 	private static final Logger LOG = LoggerFactory.getLogger(TychoModelConverter.class);
@@ -82,10 +80,10 @@ public class TychoModelConverter extends DefaultModelConverter {
 	private P2RepositoryManager repositoryManager;
 
 	@Inject
-	private TychoProjectManager projectManager;
+	private Provider<TychoProjectManager> projectManagerProvider;
 
 	@Inject
-	private TychoReactorReader reactorReader;
+	private Provider<TychoReactorReader> reactorReaderProvider;
 
 	@Inject
 	private LegacySupport legacySupport;
@@ -134,6 +132,7 @@ public class TychoModelConverter extends DefaultModelConverter {
 	private String generatePackageUrl(Artifact artifact, boolean withVersion, boolean withClassifier,
 			Supplier<String> fallback) {
 		TychoSBOMConfiguration sbomConfig = getOrCreateCurrentProjectConfiguration();
+		TychoReactorReader reactorReader = reactorReaderProvider.get();
 		if (sbomConfig.getIncludedPackagingTypes().contains(reactorReader.getPackagingType(artifact))) {
 			ArtifactKey artifactKey = getQualifiedArtifactKey(artifact);
 			IArtifactKey p2artifactKey = ArtifactTypeHelper.toP2ArtifactKey(artifactKey);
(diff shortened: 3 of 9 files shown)


tycho-build/src/main/resources/META-INF/maven/extension.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cstamas, line 1: Make Tycho build extension export Guice guts (will be needed below).

Also, see https://eclipse.dev/sisu/org.eclipse.sisu.plexus/conversion-to-jsr330.html#Custom_Bindings

tycho-its/projects/sbom/.mvn/maven.config
```diff
@@ -1 +1 @@
--Dtycho-version=5.0.0-SNAPSHOT
+-Dtycho-version=6.0.0-SNAPSHOT
```
cstamas, line 1: Unsure what this is, but I was building 6.0.0-SNAPSHOT

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoModelConverter.java
```diff
@@ -82,10 +80,10 @@ public class TychoModelConverter extends DefaultModelConverter {
 	private P2RepositoryManager repositoryManager;
 
 	@Inject
-	private TychoProjectManager projectManager;
+	private Provider<TychoProjectManager> projectManagerProvider;
```
cstamas, line 83: session scoped

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoModelConverter.java
```diff
@@ -82,10 +80,10 @@ public class TychoModelConverter extends DefaultModelConverter {
...
 
 	@Inject
-	private TychoReactorReader reactorReader;
+	private Provider<TychoReactorReader> reactorReaderProvider;
```
cstamas, line 86: session scoped

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoSBOMOverrides.java
```diff
@@ -0,0 +1,33 @@
...
+ * without any hint or annotation) is to blindly follow Plexus XML.
+ *
+ * Hence, we need to <em>bind these components exactly the same way as Plexus Shim would do</em>. This is why
+ * we do this "dance" with Guice binds.
+ *
+ * If CycloneDX project would abstain of using Maven Plugin API {@code @Component}, but would use standard {@link @Inject},
+ * none of this would be needed.
+ */
+@Named
+public class TychoSBOMOverrides extends AbstractModule {
```
cstamas, line 27: here we override but "a la plexus" keyed.
note: modelconverter is singleton, but projectdepconverter was NOT! (see CycloneDX)

tycho-build/src/main/resources/META-INF/maven/extension.xml
laeubi: Should extensions really export such maven internals?

tycho-its/projects/sbom/.mvn/maven.config
laeubi: The integration test redefine `tycho-version` by default so one can easily use the same test for different Tycho versions and we do not need to update all of them every time the version changes. So most of the time these more document what Tycho version introduced the test, but it of course does not harm to update them.

tycho-sbom/src/main/java/org/eclipse/tycho/sbom/TychoSBOMOverrides.java
laeubi: This really sounds odd to me and feels like Maven should handle it in the first place. It completely contradicts the idea of DI if one has to manually "model" things and I would then rather not convert this component to JSR330 at all rather than such workarounds especially as SBOM plugin is already a bit special as its an extension to the cyclone-dx plugin so more bridge between two worlds.

tycho-build/src/main/resources/META-INF/maven/extension.xml
cstamas: Per Sisu documentation, yes.

cstamas:
The real problem is the third annotation type (used in CycloneDX Mojo):

The three annotation types create an asymmetry:

| Annotation | Purpose | Resolution |
|---|---|---|
| @javax.inject.Named | Definition | Sisu scan → Guice → BeanLocator |
| @plexus.Component | Definition | components.xml → Guice binding |
| @maven.plugins.annotations.Component | Injection point | plugin.xml → Plexus lookup() |

The third one is the troublemaker. It's not a real DI annotation — it generates a descriptor in plugin.xml that Maven's plugin manager resolves through Plexus lookup(), completely bypassing `@Inject`. A pure JSR330 `@Named` override in an extension doesn't create a binding with the Plexus-compatible key shape, so the lookup doesn't find or prefer it.

Why the Guice module works: `TychoSBOMOverrides` creates explicit bindings with `Key.get(ModelConverter.class, Names.named("default"))` — matching exactly what Plexus lookup expects. Extension realm bindings rank higher than plugin realm bindings in the BeanLocator, so they win.

The fix is on CycloneDX's side — if they used `@Inject` instead of `@Component` on their Mojo fields, JSR330 overrides would work naturally. The workaround is needed because the injection point lives in a world that's neither JSR330 nor Plexus.

Thanks to @gnodet

cstamas:
This is one of the reasons why this annotation is deprecated: https://github.com/apache/maven-plugin-tools/blob/master/maven-plugin-annotations/src/main/java/org/apache/maven/plugins/annotations/Component.java

tycho-build/src/main/resources/META-INF/maven/extension.xml
laeubi: We already have had some issues in the past where different extension exported the same "thing" so I'm a bot concerned to open something here not really needed/used by Tycho but looks like an Maven/Sisu internal instead. So if this is the way to go, why does maven not already export/expose it by default?

cstamas: Several issues are conflicting here:
* CycloneDX uses deprecated `@Component` annotation from plugin API (see below for analysis)
* Tycho extends an existing component -- this is bad practice and something not recommended
* Maybe Tycho and CycloneDX should agree to create something _reusable_ instead, than this messy solution. Accumulating bad practices, like depending on a maven-plugin, extending already existing component, are all things that should be avoided.

Basically, we battle multiple bad practices here, but in fact, everything should work as in @HannesWell original PR, only if everyone would "play by the book".

laeubi:
> The fix is on CycloneDX's side — if they used `@Inject` instead of `@Component` on their Mojo fields, JSR330 overrides would work naturally. The workaround is needed because the injection point lives in a world that's neither JSR330 nor Plexus.

So I think then we should likely wait for cyclone-dx to migrate before changing things in Tycho that requires such deep coupling with DI internals. Apart from that of course it would be good if maven where supporting its annotation better even though it is now deprecated as this already produced some trouble in the past.

tycho-build/src/main/resources/META-INF/maven/extension.xml
laeubi: The core problem is that maven does not allow a good way for plugins to offer "extensibility" apart from add something to the classpath and hope its gets picked up. Tycho uses [PluginRealmHelper](https://github.com/eclipse-tycho/tycho/blob/main/tycho-spi/src/main/java/org/eclipse/tycho/helper/PluginRealmHelper.java) to allow others to provide dedicated extension points but it comes as a cost (requires plugin to be extension and export the artifacts) and is not baked into maven core so hard to agree on.
If a native way would exits (e.g. a jar could declare itself as being an SPI or a plain plugin declare it similar to how maven extensions already works declarativly) that would help, as it is usually hardly possible to convince others to add something for Tycho or any one else (e.g. similar problem exits for years at maven-jar plugin with different workarounds a quirks to enhance manifest with additional entries).

Second bad practice" often change over time... what was good practice in the past (`@plexus.Component`, `@maven.plugins.annotations.Component`) suddenly comes "bad" with a promised migration that then requires a coordinate rewrite of different code-bases.

cstamas: Maven does allow good way for plugins to offer extensibility, if you play by the "best practices". 

Sisu JSR330 components are picked up, just look at Resolver.

And about second best practice, just like `StringBuffer` or `Vector` was good practice long time ago, today is not. In our profession, things do change, especially in 15 years span, as Plexus XML was in 2000s, while (runtime kept) annotations are with us since 2010s.

cstamas:
Agreed with @laeubi I don't consider this PR as "how thing should be done", it was merely "it can be done". But is very involved and brittle. Best would be if the two project (Tycho and CycloneDX) both agreed on intent, communicate requirements, and solve the reusability issue of CycloneDX together.

HannesWell:
> Best would be if the two project (Tycho and CycloneDX) both agreed on intent, communicate requirements, and solve the reusability issue of CycloneDX together.

Absolutely.
Nevertheless, can you assess if just migrating the CycloneDX mojo would already fix this problem (although not optimally)?
I think that should be feasible without too much effort and I can try that.

In general there seem to have been a larger discussion about the integration in
- https://github.com/CycloneDX/cyclonedx-maven-plugin/issues/137

cstamas:
> Absolutely. Nevertheless, can you assess if just migrating the CycloneDX mojo would already fix this problem (although not optimally)? I think that should be feasible without too much effort and I can try that.

Yes, all they would need to do is stop using (deprecated) `@Component` and replace it with `@Inject`. In that moment, your original PR should just work.

HannesWell:
> Yes, all they would need to do is stop using (deprecated) `@Component` and replace it with `@Inject`. In that moment, your original PR should just work.

That sounds good.
So I tried the easy way and proposed to migrate the cyclondx maven plugin:
- https://github.com/CycloneDX/cyclonedx-maven-plugin/pull/674

Let's hope for the best :)
Thanks for the insights.

HannesWell:
With https://github.com/eclipse-tycho/tycho/pull/6191 being submitted, I think this can be closed.
Nevertheless, thanks a lot @cstamas for your help and insights.

PR #6198, opened 2026-07-27
Bug: 6196 Fix exclusion of Plexus defaults on useDefaultExcludes

### LINKED ISSUE #6196: Tycho 5.0.3 no longer keeps .gitignore files in jars
Since Tycho 5.0.3, the `tycho-packaging-plugin` silently drops all `**/.gitignore` files from
eclipse-plugin jars, even when the plugin is explicitly configured with:

```xml
<plugin>
      <artifactId>tycho-packaging-plugin</artifactId>
      <configuration>
              <useDefaultExcludes>false</useDefaultExcludes>
      </configuration>
</plugin>
```

With Tycho 5.0.2, the same configuration keeps the `.gitignore` files in the jar.
In our case, the bundle intentionally ships test projects whose `.gitignore`
files are part of the content, and regression tests started failing after the 5.0.2 → 5.0.3
bump because the files vanished from the built bundle.

### Reproducing

Follow the todos in the pom of the example project.

[tycho-gitignore-repro.zip](https://github.com/user-attachments/files/30408493/tycho-gitignore-repro.zip)

### Analysis

[AbstractTychoPackagingMojo.getFileSet(...)](https://github.com/eclipse-tycho/tycho/blob/tycho-5.0.3/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/packaging/AbstractTychoPackagingMojo.java#L116-L142) only applies `useDefaultExcludes` to the
exclude patterns *Tycho itself* puts on the `DefaultFileSet`:

The returned `DefaultFileSet` however maintains its own
`usingDefaultExcludes` flag at its default of `true` (`getFileSet` never calls
`fileSet.setUsingDefaultExcludes(false)`). So when `PackagePluginMojo` hands the file set to
the archiver, plexus-archiver *additionally* applies `AbstractScanner.DEFAULTEXCLUDES` from
whatever plexus-utils is on the classpath, regardless of Tycho's `useDefaultExcludes` setting.

In the older plexus-utils 3.4.x–3.5.1, `**/.gitignore` is not listed in
`DEFAULTEXCLUDES`, so this didn't create any issues. Since Tycho 5.0.3 bumped Plexus to 4.0.3, this surfaced the issue.

### Suggested fix

Propagate the mojo parameter to the plexus file set in
`AbstractTychoPackagingMojo.getFileSet(...)`:

```java
fileSet.setUsingDefaultExcludes(useDefaultExcludes);
```

### Workaround

Re-adding the files through `additionalFileSets` works, because those are passed to the
archiver as-is and their `usingDefaultExcludes` flag is honored:

```xml
<additionalFileSets>
      <fileSet>
              <directory>${project.basedir}</directory>
              <includes>
                      <include>**/.gitignore</include>
              </includes>
              <usingDefaultExcludes>false</usingDefaultExcludes>
      </fileSet>
</additionalFileSets>
```

tivervac:
Fixes #6196

### DIFF
--- a/tycho-its/projects/packaging.useDefaultExcludes/disabled/META-INF/MANIFEST.MF
+++ b/tycho-its/projects/packaging.useDefaultExcludes/disabled/META-INF/MANIFEST.MF
@@ -1,0 +1,5 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Use Default Excludes Disabled
+Bundle-SymbolicName: useDefaultExcludes.disabled
+Bundle-Version: 1.0.0.qualifier
--- a/tycho-its/projects/packaging.useDefaultExcludes/disabled/build.properties
+++ b/tycho-its/projects/packaging.useDefaultExcludes/disabled/build.properties
@@ -1,0 +1,2 @@
+bin.includes = META-INF/,\
+               testdata/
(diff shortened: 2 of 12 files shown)


laeubi (approved):
looks good

PR #6215, opened 2026-08-06
Add support for filtering OSGI Capabilities at build time

laeubi:
This pull request introduces a mechanism to break direct dependency cycles caused by `Require-Capability` requirements in OSGi bundles during the Maven/Tycho reactor build. It does this by allowing specific capability requirements to be disabled for a project build via profile properties, while keeping them active for consumers and in published metadata. The PR also adds a comprehensive integration test to verify this behavior using a simulated real-world scenario.

This change enables Tycho to build projects that would otherwise be blocked by direct dependency cycles in OSGi metadata, by allowing targeted disabling of capability requirements during the build, without affecting runtime or consumers.

See also
- https://github.com/eclipse-equinox/p2/pull/1073

### DIFF
--- a/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
+++ b/tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
@@ -132,8 +132,10 @@ public class EquinoxResolver implements DependenciesResolver {
             EquinoxResolverConfiguration config, Map<Module, ArtifactDescriptor> descriptorLookup)
             throws BundleException {
         Properties properties = getPlatformProperties(project, mavenSession, ee);
-        ModuleContainer container = newState(artifacts, properties, mavenSession, executorService, config,
-                descriptorLookup);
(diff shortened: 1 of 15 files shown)


tycho-core/src/main/java/org/eclipse/tycho/core/osgitools/EquinoxResolver.java
laeubi, line 1: I need to check if this change is actually needed/useful at all, but for now lets see what the integration builds tell us here.

laeubi: I now come to the conclusion that we can at least generalize this to just remove all required capabilities as they are never producing any usefull wires anyways... lets see if anything breaks.

HannesWell: What about test runtimes? They can rely on required capabilities when assembling runtime.
Or is this part not used for assembling test runtimes?

PR #6221, opened 2026-08-07
Reduce Helios p2 repo usage in tests

akurtakov:
Switch the source bundle integration fixtures from the old Helios-era p2 repository to the build target platform. Update the remote source-bundle expectations to the latest org.junit and org.hamcrest source artifacts and use the target-platform verifier path in the tests.


akurtakov:
Initiated due to EF archive failed to deliver helios artifacts which broke a build https://ci.eclipse.org/tycho/job/Tycho%20CI%20Build/job/PR-6220/1/

PR #6222, opened 2026-08-07
Pass the test method filter to the test provider

### LINKED ISSUE #6217: tycho-surefire drops the `#method` suffix of `-Dtest` before any provider sees it
## Environment

- Tycho 5.0.2, Maven 3.9.11, Java 25
- An `eclipse-test-plugin` module whose JUnit 4 tests run through the `junit6` provider with a platform-discoverable vintage engine; class-level selection works (`-Dtest=<Class>` runs exactly that class)

## Observed

`-Dtest=Class#method` silently runs the entire class. The suffix is discarded at two independent layers:

- `AbstractTestMojo.scanForTests` collapses `-Dtest` into a class-file include pattern for its directory scanner (https://github.com/eclipse-tycho/tycho/blob/9f65dc09a72ee08f29cc8a4d5381a45b33dc7144/tycho-surefire/tycho-surefire-plugin/src/main/java/org/eclipse/tycho/surefire/AbstractTestMojo.java#L247-L270); the TODO there cites Eclipse bug 495353 and states that method selection would require handing `TestListResolver` the unparsed string.
- `OsgiSurefireBooter.invokeSureFire` constructs the provider's `TestRequest` with `TestListResolver.getEmptyTestListResolver()` (https://github.com/eclipse-tycho/tycho/blob/9f65dc09a72ee08f29cc8a4d5381a45b33dc7144/tycho-surefire/org.eclipse.tycho.surefire.osgibooter/src/main/java/org/eclipse/tycho/surefire/osgibooter/OsgiSurefireBooter.java#L176-L177), so even a forwarded filter would be dropped there.

Because the truncation is mojo- and booter-side, no provider choice can honor a method filter.

## Expected

Either the method part of `-Dtest` reaches the provider as a populated `TestListResolver` (maven-surefire's own semantics for the property), or the mojo rejects the suffix with a diagnostic instead of silently widening the selection to the whole class.

## Cross-Refs

- Eclipse bug 495353 (pre-migration report of the same limitation): https://bugs.eclipse.org/bugs/show_bug.cgi?id=495353
- Consumer-side diagnosis: https://github.com/ponder-lab/Hybridize-Functions-Refactoring/issues/749#issuecomment-5208820604

khatchad:
Fixes #6217.

The test mojo accepts `-Dtest=Class#method`, and the scan resolves the class, but `OsgiSurefireBooter` builds the `TestRequest` with an unconditional `TestListResolver.getEmptyTestListResolver()`, so the provider never sees the method part and runs the whole class.

This change forwards the filter through the surefire properties and constructs a real `TestListResolver` from it in the booter. The filter only travels when it carries a `#` method suffix: plain class selection is already narrowed by the scan, and forwarding it unconditionally would change the established pattern semantics, so this stays the minimal slice discussed on the issue.

Validation:

- New IT case `RunSingleTestTest.testSingleMethod` selects `bundle.MixedTest#working` in a fixture whose other method throws, and additionally checks `Tests run: 1` in the log. It passes with the change and fails without it (the throwing method runs), so the case is two-sided.
- Verified end to end on the PDE reactor where #6217 was found: with this change, `-Dtest=HybridizeFunctionRefactoringProcessorTest#testSetInferInputSignatures` runs exactly that one test under the junit4 provider instead of the whole class.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AM8pvSeP5UqDyyTn9h5E2P

### DIFF
--- a/tycho-its/projects/surefire.testSelection/src/bundle/MixedTest.java
+++ b/tycho-its/projects/surefire.testSelection/src/bundle/MixedTest.java
@@ -1,0 +1,26 @@
+/*******************************************************************************
+ * Copyright (c) 2026 others.
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package bundle;
+
+import org.junit.Test;
+
+public class MixedTest
+{
+    @Test
+    public void working()
+    {
+    }
+
+    @Test
+    public void broken()
+    {
+        throw new RuntimeException();
+    }
+}
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/surefire/RunSingleTestTest.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/surefire/RunSingleTestTest.java
@@ -26,4 +26,17 @@ public class RunSingleTestTest extends AbstractTychoIntegrationTest {
 		verifier.executeGoal("integration-test");
 		verifier.verifyErrorFreeLog();
 	}
+
+	@Test
+	public void testSingleMethod() throws Exception {
+		Verifier verifier = getVerifier("surefire.testSelection");
+
+		// the method suffix must reach the test provider, see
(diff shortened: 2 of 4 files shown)


laeubi (approved):
The change looks sound and minimal.

laeubi:
@khatchad thanks for the fix. If you think the trimming of whitespaces is important as suggested by the copilot review please don't hesitate to open a follow up.

PR #6226, opened 2026-08-09
Improve the documentation of o.e.tycho.plugins.p2.extras.P2ManagerMojo

merks:
- Provide links to the JustJ tools documentation for each parameter.


merks:
@laeubi 

I think with this additional change to specify the ECF provider with an appropriate default that the mojo is ready for production use.  I can retest SNAPSHOT builds when they are available, but so far so good!

laeubi:
@merks this seems like a useful addition here!

PR #6232, opened 2026-08-10
Assert without a JUnit dependency in tycho-testing-harness, migrate tycho-packaging-plugin

akurtakov:


laeubi (commented):
The TychoPlexusTestCase is already a hack, I'm not sure we should add more on top of it and we likely need to migrate everything during maven 4 migration, so I'm not completely sure what is the intend here?

Given that many of the tests are more there for legacy reasons anyways I wonder if its actually worth to migrate them to JUnit5 at all.

akurtakov:
This is smth I see as temporary thing to move all the test infra to JUnit 5/6. Once migration is finished the JUnit 4 part/hack will be removed. 
This effort is not preventing actual review of JUnit tests whether they are not needed and changed/removed/etc. but right now running any tests has the warning:
```
Aug 07, 2026 9:50:54 AM org.junit.platform.launcher.core.DiscoveryIssueNotifier logIssues
INFO: TestEngine with ID 'junit-vintage' encountered a non-critical issue during test discovery:

(1) [INFO] The JUnit Vintage engine is deprecated and should only be used temporarily while migrating tests to JUnit Jupiter or another testing framework with native JUnit Platform support.

```
and I am working towards get rid of this for now.

laeubi:
This looks a bit strange to me. So actually nothing is broken and the warning clearly state one should migrate.
So why not just perform the migration instead of installing a temporary workaround to suppress this warning that then should be removed afterwards after the migration (but then no one will ever notice)?

akurtakov:
Nothing suppresses a warning, it allows gradual migration of tests by supporting the 2 different engines.
The second commit in this PR is doing converstion of one module https://github.com/eclipse-tycho/tycho/pull/6232/changes/0d714016c22db1e84a688d5893ec8447a8563803 . There are more to be converted thus the workaround is needed till all the tests are converted.

laeubi:
This looks totally wrong from a JUnit 5 perspective to me.

In a JUnit 5 world a test should never need to extend `TychoPlexusTestCase` but instead gets added the `@ExtendWith(TychoPlexusExtension.class)` instead (what then requires it to contain something useful, maybe delegate to a shared static helper method in the meanwhile).

This would then make the `TychoPlexusTestCase` become unused some day and therefore deleted and does not transfer tis bad pattern over to the JUnit 5 migration.

akurtakov:
Changed to not use @ExtendWith(TychoPlexusExtension.class) at all.

laeubi (approved):
This looks much more JUnit 5 like.

Issue #3504, opened 2024-02-13
Allow to generate the top list of modules (based on a pattern)


vogella:
Tychos build extension allows to include modules based on common directories in https://github.com/eclipse-tycho/tycho/blob/c9e3fe9eeb83efb46891247b697fcfe3ad96fc40/tycho-extras/tycho-pomless/src/main/java/org/eclipse/tycho/pomless/TychoAggregatorMapping.java#L46

If we have a flat structure I still have to add the modules manually.  

To simplify this, Tycho could allow to specify another directory in which another pom file is generated in which all modules are listed which belong to the main directory. This would allow to configure only one module in the parent pom and all other. 

(Maybe this could also be combined with the build.properties settings to define directory patterns which should be include or not).

laeubi:
> If we have a flat structure I still have to add the modules manually.

Tycho pomless works best with a [structured layout](https://tycho.eclipseprojects.io/doc/latest/StructuredBuild.html) ...

> To simplify this, Tycho could allow to specify another directory in which another pom file is generated in which all modules are listed which belong to the main directory. This would allow to configure only one module in the parent pom and all other.

One could simply use a structured layout as well instead of adding some kind of indirection ... the main issue with such approaches is that they need to be smart for little benefit and heavily complicate things, e.g if it is a sub directory, it also needs to use a different parent in the flat modules (because the default for a parent is `../pom.xml`), then in a flat hierarchy often people have also an extra parent so we need to account for this and so on... 

So I would recommend to use a [structured layout](https://tycho.eclipseprojects.io/doc/latest/StructuredBuild.html) if you want to have it easy and convenient.

vogella:
Using maven.config and pom.tycho this is actually already possible. 

For an example see https://github.com/vogellacompany/codeexamples-ide

Specific files for this setup: 

https://github.com/vogellacompany/codeexamples-ide/blob/main/.mvn/maven.config 
https://github.com/vogellacompany/codeexamples-ide/blob/main/pom.xml
https://github.com/vogellacompany/codeexamples-ide/blob/main/build/pom.tycho
https://github.com/vogellacompany/codeexamples-ide/blob/main/lsp/pom.tycho

PR #6227, opened 2026-08-09
Add contribution guidelines for AI/GenAI agents (AGENTS.md, CONTRIBUTING.md, issue/PR templates)

laeubi:
## What does this PR do?

Adds an explicit contribution policy for AI/GenAI coding agents acting on behalf of users, since we've
seen recurring problems in issues/PRs/discussions (vague reports missing key details submitted
piecemeal, verbose generated prose, and agents losing track of review-thread history leading to
frustrating back-and-forth).

- `AGENTS.md`: new "Rules for AI Agents" section — disclosure, user accountability (per Eclipse's
  GenAI committer guidelines), no chat-in-tracker behavior, concise/test-first reporting, duplicate
  search, self-review gate, ask-before-submit, full-thread-history discipline, one submission at a
  time, respect for maintainer time.
- `CONTRIBUTING.md`: human-facing mirror of the same expectations, linked from a new TOC entry.
- `.github/ISSUE_TEMPLATE/{bug_report,feature_request,config}.yml`: structured issue forms with a
  required duplicate-check box and an AI-disclosure field.
- `.github/PULL_REQUEST_TEMPLATE.md`: AI-disclosure section + reviewer-confirmation checkboxes.

## How was this tested?

Docs/templates only — no build changes. Verified the issue forms render correctly by inspecting the
YAML structure against GitHub's issue-forms schema; no automated test applicable.

## AI/GenAI disclosure

- Tool/agent + model/version + mode: GitHub Copilot CLI, claude-sonnet-5, interactive
- [x] I have personally reviewed all text in this PR and understand and stand behind every change,
      per the Eclipse GenAI contribution guidelines.
- [x] I will personally review and confirm any AI-assisted follow-up changes made in response to
      review comments before pushing them.

## Checklist

- [x] I have searched existing PRs/issues and this is not a duplicate.
- [x] N/A — documentation/template only change, no code to build/test.
- [x] Commits are small and self-contained.

PR #6256, opened 2026-08-20
Document how pomless modules find their parent POM

vogella:
The `tycho.pomless.parent` property is currently only discoverable by reading `AbstractTychoMapping` or a few integration test projects. This documents it in the structured build page: it can be set per module in `build.properties` or globally as a system property, and an absolute value works as well.

The useful part is the combination with `${maven.multiModuleProjectDirectory}`, which makes every module of a build resolve the root POM as its parent regardless of how deeply it is nested. Flat layouts and repositories whose modules come from Git submodules then need no per module configuration at all, which is the parent resolution concern raised in #3504.

The added integration test covers exactly that setup and fails without the property ("No parent pom file found"), so the behaviour stays covered.

Disclaimer: Created and tested with the help of Qwen3.8 27B DenseHetzner Inference

### DIFF
--- a/src/site/markdown/StructuredBuild.md
+++ b/src/site/markdown/StructuredBuild.md
@@ -88,3 +88,33 @@ Tycho however can derive most of the information from other already existing fil
 ```
 
 - You can now run your build with `mvn verify`.
+
+## Where a pomless module looks for its parent
+
+A pomless module has no `pom.xml` in which a `<parent>` could be declared, so Tycho derives the parent from the file system.
+By default it uses the directory one level up (`..`), which is what a structured layout provides.
+
+If the parent is somewhere else, set the property `tycho.pomless.parent` in the `build.properties` of the module.
+The value may point to a directory or directly to a POM file, and it is resolved relative to the module directory:
+
+```properties
+bin.includes = META-INF/,.
+source.. = src/
+tycho.pomless.parent = ../../releng/parent
+```
+
+The same property can be set as a system property to change the default for every module of the build, for example in `.mvn/maven.config`:
+
+```properties
+-Dtycho.pomless.parent=../../releng/parent
+```
+
+Absolute values are used as they are, which is useful if the modules are not all at the same depth, for example because they come from Git submodules or from a flat layout.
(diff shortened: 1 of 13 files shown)


laeubi:
> The `tycho.pomless.parent` property is currently only discoverable by reading `AbstractTychoMapping` or a few integration test projects

It is documented here

https://github.com/eclipse-tycho/tycho/wiki/Tycho-Pomless#specify-parent-path

Please also be aware of the new AI disclosure policy see https://github.com/eclipse-tycho/tycho?tab=contributing-ov-file#using-ai--genai-tools-to-contribute when AI acts on your behalf and the PR looks like it was entirely generated and created by an AI agent.

vogella:
Funny, looks like the "Using AI / GenAI Tools to Contribute" is written by AI. :-)

laeubi:
> Funny, looks like the "Using AI / GenAI Tools to Contribute" is written by AI. :-)

It is and follows what is proposed there as an example see:

- https://github.com/eclipse-tycho/tycho/pull/6227

src/site/markdown/StructuredBuild.md
```diff
@@ -88,3 +88,33 @@ Tycho however can derive most of the information from other already existing fil
...
+If the parent is somewhere else, set the property `tycho.pomless.parent` in the `build.properties` of the module.
+The value may point to a directory or directly to a POM file, and it is resolved relative to the module directory:
+
+```properties
+bin.includes = META-INF/,.
+source.. = src/
+tycho.pomless.parent = ../../releng/parent
+```
+
+The same property can be set as a system property to change the default for every module of the build, for example in `.mvn/maven.config`:
```
laeubi, line 106: Maven refers to them as User Properties see https://maven.apache.org/ref/4-LATEST/configuring.html

laeubi (commented):
We have a similar documentation already at https://github.com/eclipse-tycho/tycho/wiki/Tycho-Pomless#define-properties so it should at least reference that - also likley if we describe the parent thing, we should also possibly extend the docs about group/artifact id and how they are derived and can be overriden.

Issue #3057, opened 2023-11-19
Tycho doesn't use credentials while downloading artifacts from secured p2


masofcon:
We used 2.x Tycho to build our project.
Some of our repos in target platform are secured.
We faced with auth problems while migrating to 4.x Tycho.

Quick debug shows that MavenAuthenticator uses different instances of DefaultRepositoryIdManager during build.

I've attached some stacks during our build.

DefaultRepositoryIdManager first initialization
[first-init-stack.txt](https://github.com/eclipse-tycho/tycho/files/13403410/first-init-stack.txt)

Population with repos ids
[adds-mappings-to-created-repidmanager-stack.txt](https://github.com/eclipse-tycho/tycho/files/13403409/adds-mappings-to-created-repidmanager-stack.txt)

Access to populated instance
[getKnownMavenRepositoryLocations-from-first-repoidmanager-stack.txt](https://github.com/eclipse-tycho/tycho/files/13403411/getKnownMavenRepositoryLocations-from-first-repoidmanager-stack.txt)

During validate-classpath stage of tycho-compiler plugin Tycho uses newly created instance of DefaultRepositoryIdManager which is not populated with information about repo to id mappings
[second-init-stack.txt](https://github.com/eclipse-tycho/tycho/files/13403413/second-init-stack.txt)
[getKnownMavenRepositoryLocations-from-second-repoidmanager-stack.txt](https://github.com/eclipse-tycho/tycho/files/13403412/getKnownMavenRepositoryLocations-from-second-repoidmanager-stack.txt)

I've also tried to create example in PasswordProtectedP2RepositoryTest but have no luck.

tretyakevich:
This behavior is reproduced in Tycho 4.0.2 at least. The reason is in HttpTransportProtocolHandler.transportFactoryMap.
During the validation phase contains two pairs of an URLHttpTransportFactory'es (two for JavaUrl and two for LazyPlexusBean).
One of JavaUrl's is not properly initialized.

tretyakevich:
[sisu.log](https://github.com/eclipse-tycho/tycho/files/13407368/sisu.log)

Added SISU tracing which isn't clear for me. I can see a normal provider mapping of HttpTransportProtocolHandler (mapped by the class itself) on start. With singleton scope an so on.
But at the end (shutdown) I can see two entries for HttpTransportProtocolHandler - provider (exepcted) and some implicit constructor contribution (and that is strange, I can't see any constructor mappings in the code)

[...]

PR #6287, opened 2026-08-31
Match Maven settings server id additionaly with target location URL

HannesWell:
## What does this PR do?

Avoid the need to specify an `id` attribute for a target location repository, to supply e.g. credentials for that server, by matching also the repository's location URL against all server IDs specified in the 'settings.xml'.

If the p2-repository's location is `https://my.company.com/repo/foo` and the 'settings.xml' contains an entry like
```
<servers>
	<server>
		<id>https://my.company.com/repo/foo</id>
		<username>...
```
the credentials defined for that server are used to authenticate against that server.

The repository location is not only matched literally, but also all it's parent paths plus all variants without URL scheme. For a target definition like above, all the following server IDs match:
- `https://my.company.com/repo/foo`
- `https://my.company.com/repo`
- `https://my.company.com`
- `my.company.com/repo/foo`
- `my.company.com/repo`
- `my.company.com`

More specific server IDs are preferred.

## How was this tested?

A new test case was added for an existing test class to cover the new behavior.

## Checklist

- [x] I have searched existing PRs/issues and this is not a duplicate.
- [x] `mvn clean install -DskipTests` / relevant integration tests pass locally.
- [x] Commits are small and self-contained (see [CONTRIBUTING.md](../CONTRIBUTING.md#granularity)).

### DIFF
--- a/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/DefaultMavenRepositorySettings.java
+++ b/p2-maven-plugin/src/main/java/org/eclipse/tycho/p2maven/repository/DefaultMavenRepositorySettings.java
@@ -14,11 +14,15 @@
 package org.eclipse.tycho.p2maven.repository;
 
 import java.net.URI;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
 import org.apache.maven.RepositoryUtils;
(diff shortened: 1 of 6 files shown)


HannesWell:
This feature was already described in https://wiki.eclipse.org/Tycho/Target_Platform/Authentication_and_Mirrors:
```
When accessing a p2 repository, Tycho will check if there is a <server> entry with matching ID and use these credentials.
The ID of a p2 repository can be specified explicitly, e.g. when the repository is [added in the POM](https://wiki.eclipse.org/Tycho/Target_Platform#Layout_p2).
If no ID is specified, e.g. in a target file, the repository URL (without trailing slash) is used as ID (since Tycho 0.16.0).
```
But it seem to be lost in the meantime. Maybe due to
- https://github.com/eclipse-tycho/tycho/issues/3057#issuecomment-1853444479

This approach is IMO different and save, since it matches the most specific ID. At the same time it is still flexible since it also considers parent paths on no specific matches. So if suitable, one can e.g. cover multiple p2-repos at the same host (or sub-path).
The ability to skip the scheme part (e.g. `https://` or `http://`), is useful when using the new `mvn-server-credentials` property of the [setup-java@v6 github action](https://github.com/actions/setup-java/tree/dd06d9cba3e5552c54d9f8ea23572deb30010f7c#inputs), which lists servers in the format `server-id:USERNAME_ENV:PASSWORD_ENV`.



I didn't found any reference on defining server credentials in the current Tycho doc, but I believe it would be a good fit to add it to
- https://tycho.eclipseprojects.io/doc/5.0.4/TargetPlatform.html

@laeubi WDYT?

merks:
It sounds very flexible and useful.

HannesWell:
> It sounds very flexible and useful.

Thanks.

After adding a section about `Authentication for p2 Repositories` in the `TargetPlatform.md` documentation, I consider this ready.

laeubi (approved):
Looks good

laeubi:
Aggregator Check fails because it requires Java 25 now but this is a different fix.

- https://github.com/eclipse-tycho/tycho/pull/6294

PR #6309, opened 2026-09-08
Make DefaultTimestampFinder thread safe

HeikoKlare:
## What does this PR do?

`DefaultTimestampFinder` is a `@Named @Singleton` component, so Sisu creates one instance per build. It holds a map of `SimpleDateFormat` instances and reuses them for every parse. `SimpleDateFormat` is not thread safe.

`BuildQualifierAggregatorMojo` is declared `threadSafe = true`, so in a parallel build (`-T 1C`) several `eclipse-repository` projects run `build-qualifier-aggregator` concurrently and all call into that one shared finder. The concurrent parses corrupt each other's state in `java.text.DigitList`:

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-packaging-plugin:5.0.3:build-qualifier-aggregator
  (default-build-qualifier-aggregator) on project <an eclipse-repository project>:
  Execution default-build-qualifier-aggregator of goal ...:build-qualifier-aggregator failed: multiple points
Caused by: java.lang.NumberFormatException: multiple points
    at jdk.internal.math.FloatingDecimal.checkMultiplePoints (FloatingDecimal.java:2308)
    at java.text.DigitList.getDouble (DigitList.java:173)
    at java.text.DecimalFormat.parse (DecimalFormat.java:2303)
    at java.text.SimpleDateFormat.subParse (SimpleDateFormat.java:1976)
    at java.text.SimpleDateFormat.parse (SimpleDateFormat.java:1579)
    at org.eclipse.tycho.buildversion.TimestampFinder.parseTimestamp (TimestampFinder.java:109)
    at org.eclipse.tycho.buildversion.TimestampFinder.findInString (TimestampFinder.java:99)
    at org.eclipse.tycho.buildversion.TimestampFinder.discoverTimestamp (TimestampFinder.java:91)
    at org.eclipse.tycho.buildversion.TimestampFinder.parseQualifier (TimestampFinder.java:87)
    at org.eclipse.tycho.buildversion.TimestampFinder.findByDescriptor (TimestampFinder.java:69)
    at org.eclipse.tycho.buildversion.BuildQualifierAggregatorMojo$1.visitArtifact (BuildQualifierAggregatorMojo.java:107)
    at org.eclipse.tycho.buildversion.BuildQualifierMojo.execute (BuildQualifierMojo.java:154)
    at io.takari.maven.builder.smart.SmartBuilderImpl.buildProject (SmartBuilderImpl.java:206)
```

Observed sporadically on Tycho 5.0.3 / Maven 3.9.11 / JDK 25 / Windows, in a reactor with four `eclipse-repository` projects and `<timestampProvider>jgit</timestampProvider>`, built with `-T 1C`. The aggregation walk only runs for a non-`default` `timestampProvider`, which is why this needs a custom provider to show up.

Note the race does not always throw. It can also silently return a **wrong** timestamp, and therefore a wrong build qualifier, which is harder to notice than the build failure.

Fix: key the pattern map by the date pattern `String` rather than by a shared `SimpleDateFormat`, and construct the formatter per parse. A formatter is only constructed after its regex matched, so a successful lookup normally creates one.

Alternatives considered: `synchronized` would serialize concurrent repository builds; a `java.time.DateTimeFormatter` migration is a larger behavioural risk here (year-of-era vs. proleptic year, missing time fields in `yyyyMMdd`, resolver strictness vs. today's lenient parsing).

## How was this tested?

New unit test `TimestampFinderTest.testFindInStringIsThreadSafe`: 8 threads release from a common latch and each parse their own qualifier through a single shared `DefaultTimestampFinder`, failing on either a thrown exception or an incorrect result.

It fails reliably on the unfixed code and passes with the fix, and takes ~0.5 s.

No integration test: reproducing this through `tycho-its` would need a multi-project parallel build racing on wall-clock timing, which would be slow and flaky. The unit test targets the shared state directly.

## AI/GenAI disclosure

- Tool/agent + model/version + mode: GitHub Copilot CLI, Claude Opus 5, interactive
- [x] I have personally reviewed all code, tests, and text in this PR and understand and stand behind every change, per the [Eclipse GenAI contribution guidelines](https://www.eclipse.org/projects/handbook/#genai).
- [x] I will personally review and confirm any AI-assisted follow-up changes made in response to review comments before pushing them (see [AGENTS.md](../AGENTS.md) rule on reading full thread history).

## Checklist

- [x] I have searched existing PRs/issues and this is not a duplicate.
- [x] `mvn clean install -DskipTests` / relevant integration tests pass locally.
- [x] Commits are small and self-contained (see [CONTRIBUTING.md](../CONTRIBUTING.md#granularity)).

### DIFF
--- a/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/DefaultTimestampFinder.java
+++ b/tycho-packaging-plugin/src/main/java/org/eclipse/tycho/buildversion/DefaultTimestampFinder.java
@@ -14,11 +14,18 @@ package org.eclipse.tycho.buildversion;
 
 import java.text.ParsePosition;
 import java.text.SimpleDateFormat;
+import java.time.DateTimeException;
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.format.ResolverStyle;
+import java.time.temporal.ChronoField;
 import java.util.Date;
 import java.util.LinkedHashMap;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.TimeZone;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -42,21 +49,30 @@ public class DefaultTimestampFinder implements TimestampFinder {
 	Logger logger;
 
 
-    private static Map<SimpleDateFormat, Pattern> defaultPatterns() {
-        Map<SimpleDateFormat, Pattern> result = new LinkedHashMap<>();
-        result.put(utcFormat("yyyyMMddHHmm"), Pattern.compile("([0-9]{12})"));
-        result.put(utcFormat("yyyyMMdd-HHmm"), Pattern.compile("([0-9]{8})-([0-9]{4})"));
-        result.put(utcFormat("yyyyMMdd"), Pattern.compile("([0-9]{8})"));
+    private static Map<DateTimeFormatter, Pattern> defaultPatterns() {
+        Map<DateTimeFormatter, Pattern> result = new LinkedHashMap<>();
+        result.put(utcFormat("uuuuMMddHHmm"), Pattern.compile("([0-9]{12})"));
+        result.put(utcFormat("uuuuMMdd-HHmm"), Pattern.compile("([0-9]{8})-([0-9]{4})"));
+        result.put(utcFormat("uuuuMMdd"), Pattern.compile("([0-9]{8})"));
         return result;
     }
 
-    private static SimpleDateFormat utcFormat(String pattern) {
-        SimpleDateFormat format = new SimpleDateFormat(pattern);
-        format.setTimeZone(TimeZone.getTimeZone("UTC"));
-        return format;
+    /**
+     * Creates an immutable, and therefore thread safe, formatter for the given pattern. The pattern
+     * uses {@code uuuu} rather than {@code yyyy} to parse a proleptic year, so that no era is
+     * required. Missing time fields default to midnight, and parsing is lenient to keep the
+     * behavior of the previously used {@link SimpleDateFormat}.
+     */
+    private static DateTimeFormatter utcFormat(String pattern) {
+        return new DateTimeFormatterBuilder().appendPattern(pattern)
+                .parseDefaulting(ChronoField.HOUR_OF_DAY, 0)
+                .parseDefaulting(ChronoField.MINUTE_OF_HOUR, 0)
+                .toFormatter(Locale.ROOT)
+                .withResolverStyle(ResolverStyle.LENIENT)
+                .withZone(ZoneOffset.UTC);
     }
 
-    private final Map<SimpleDateFormat, Pattern> datePatternsByRegularExpressions;
+    private final Map<DateTimeFormatter, Pattern> datePatternsByRegularExpressions;
 
     public DefaultTimestampFinder() {
         datePatternsByRegularExpressions = defaultPatterns();
@@ -97,7 +113,7 @@ public class DefaultTimestampFinder implements TimestampFinder {
 
 	@Override
     public Date findInString(String string) {
-        for (Entry<SimpleDateFormat, Pattern> e : datePatternsByRegularExpressions.entrySet()) {
+        for (Entry<DateTimeFormatter, Pattern> e : datePatternsByRegularExpressions.entrySet()) {
             Matcher matcher = e.getValue().matcher(string);
             if (matcher.find()) {
                 String group = matcher.group();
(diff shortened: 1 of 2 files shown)


laeubi (commented):
Good catch, in general looks fine, I just wonder if one should probably use a `ThreadLocal` here to avoid constructing new formatter from strings over and over again.

As an alternative one might synchronize on the dateaformat itself what will decrease performance a little bit maybe.

merks:
It brings back memories 

https://github.com/eclipse-emf/org.eclipse.emf/blob/5504314ab7dcdfe8e4b57f0d589af331f895b0d1/plugins/org.eclipse.emf.ecore/src/org/eclipse/emf/ecore/impl/EFactoryImpl.java#L771

HeikoKlare:
Thank you for your feedback!

> I just wonder if one should probably use a `ThreadLocal` here to avoid constructing new formatter from strings over and over again.
>
> As an alternative one might synchronize on the dateaformat itself what will decrease performance a little bit maybe.

Both are valid options. I had a discussion about exactly these options with the agent when analyzing our build failure and chose to follow the simple, immutable-String solution just. This was because I found it to be the one with lowest cognitive complexity (in comparison to `ThreadLocal`) and the lowest performance impact (compared to `synchronized`) while I considered the overhead due  to reconstruction low. But I have to admit that the argument against `synchronized` is rather weak as there are not so many calls, and execution and synchronization time will be small compared to the business logic executed around it.

I post below the statement of the agent on `ThreadLocal`, based on which I made the decision. If you think either `ThreadLocal` or `synchronized` is the better option here, I am happy to adapt.

> - **It leaks in a Plexus singleton.** `TimestampFinder` lives for the entire build, and Maven/Takari worker threads outlive any individual mojo execution. A `ThreadLocal` on a long-lived component pins one formatter per worker thread indefinitely, with no natural cleanup point — a classic slow leak in container-managed code.
> - **It preserves mutable state for no gain.** The formatters are used on a cold path: a handful of qualifier strings per module. Allocating a formatter per call costs nothing measurable, so caching thread-confined mutable objects buys performance nobody needs while leaving a footgun for the next maintainer.
> - **Immutable keys are simpler to prove correct.** With `String` keys and a locally-scoped formatter created in `findInString`, there is no shared mutable state at all — correctness is obvious by inspection. `ThreadLocal` is correct only for as long as nobody later hands that formatter across a thread boundary.
> - **`DateTimeFormatter` is the modern answer.** It is immutable and thread-safe by design; `ThreadLocal<SimpleDateFormat>` is the pre-Java-8 workaround for the same problem.

laeubi:
`DateTimeFormatter` might be really better then as we require Java 8+ anyways.

HeikoKlare:
> `DateTimeFormatter` might be really better then as we require Java 8+ anyways.

Sounds reasonable. That's a bit more invasive and will have a slightly higher regression risk (I know that we had some issues in our product when moving from `SimpleDateFormat` to `DateTimeFormatter` but IIRC they were just related to very specific dates centuries ago), but it's probably better to just move to new API.
I will update the PR accordingly.

HeikoKlare:
I've updated to `DateTimeFormatter` as suggested. Since `DateTimeFormatter` is immutable, the singleton can keep sharing the formatter instances. So this version also avoids the repeated construction of formatters raised on the initial proposal, without needing a `ThreadLocal` or any locking.

Three details were required to keep the previous parsing behaviour:

- `uuuu` instead of `yyyy`, so that a proleptic year is parsed and no era has to be supplied.
- `parseDefaulting(HOUR_OF_DAY, 0)` and `parseDefaulting(MINUTE_OF_HOUR, 0)`, because the date-only `uuuuMMdd` pattern otherwise cannot be resolved to an `Instant`. These only apply when the field is absent, so the patterns that do contain a time are unaffected.
- `ResolverStyle.LENIENT`, which reproduces the roll-over behaviour of a lenient `SimpleDateFormat` for out-of-range fields.

`DateTimeFormatter.parse(CharSequence)` rejects trailing input, which replaces the previous `ParsePosition` full-consumption check.

Regarding the regression risk I mentioned: The agent compared the old and the new parsing over 1.8 million generated qualifier strings.

| Input | Compared | Differences |
| --- | ---: | ---: |
| Valid dates, years 1970–2099 | 900 000 | 0 |
| Arbitrary digit strings, year >= 1583 | 757 766 | 0 |
| Arbitrary digit strings, year < 1583 | 142 234 | 133 254 |

So `ResolverStyle.LENIENT` matches `SimpleDateFormat`'s roll-over exactly, including inputs such as month `99` or day `89`. The only divergence is for years before 1583, where `GregorianCalendar` switches to the Julian calendar at the 1582 cutover while `java.time` is proleptic Gregorian. That is probably the effect for which I mentioned earlier that we have experienced it in other such changes, but it cannot occur for a real build qualifier.

The regression test is unchanged and still fails reliably without the fix.

laeubi:
Looks like Jenkins is havin problems - I restarted the build.

HeikoKlare:
> Looks like Jenkins is havin problems - I restarted the build.

Last run was successful again.

akurtakov:
Thanks for this one. It's good to go in.

akurtakov:
@HeikoKlare backport failed so please provide backport manually.

HeikoKlare:
> backport failed so please provide backport manually.

Backport is created: https://github.com/eclipse-tycho/tycho/pull/6364

PR #6314, opened 2026-09-10
Fix useJDK=BREE toolchain selection to respect jre.compilation.profile

laeubi:
When a bundle's manifest has no `Bundle-RequiredExecutionEnvironment`, the
`tycho-compiler-plugin`'s `useJDK=BREE` toolchain lookup (`configureJavaHome()`)
fell back directly to the reactor-wide `target-platform-configuration`
`executionEnvironment`, ignoring `build.properties`' `jre.compilation.profile`.

This is inconsistent with `getSourceLevel()`/`getTargetLevel()` in the same
class, which already give `jre.compilation.profile` precedence over the
manifest BREE and the target-platform EE fallback.

The practical impact: a bundle can be compiled at bytecode level 1.8 (per
`jre.compilation.profile`) while the *toolchain/bootclasspath* used for that
compilation is silently selected from an unrelated, possibly much newer EE
(e.g. `JavaSE-17`, chosen only for test-execution purposes elsewhere in the
reactor). If that EE's toolchain ever resolves to a JDK21+ install, newer JDK
API can leak into bytecode declared for a much older target - this is a germ
of a currently open regression, see
eclipse-platform/eclipse.platform.releng.aggregator#4038 and
eclipse-jdt/eclipse.jdt.core#5389.

This PR makes `configureJavaHome()` mirror the same precedence:
`jre.compilation.profile` > manifest BREE > target-platform EE fallback.

Includes a regression test (`compiler.bree.jreCompilationProfile`) using two
distinguishable fake toolchains to prove the correct one is selected.

---
This PR was drafted with the assistance of GitHub Copilot CLI (agent mode,
Claude Sonnet 5) under human review and direction.

### DIFF
--- a/tycho-its/src/test/java/org/eclipse/tycho/test/compiler/CompilerBreeJreCompilationProfileTest.java
+++ b/tycho-its/src/test/java/org/eclipse/tycho/test/compiler/CompilerBreeJreCompilationProfileTest.java
@@ -12,25 +12,30 @@ package org.eclipse.tycho.test.compiler;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 
 import java.io.File;
+import java.nio.file.Path;
 
+import org.apache.commons.lang3.SystemUtils;
 import org.apache.maven.it.Verifier;
 import org.eclipse.tycho.test.AbstractTychoIntegrationTest;
 import org.junit.jupiter.api.Test;
 
 /**
- * When a bundle has no <code>Bundle-RequiredExecutionEnvironment</code> manifest header, the
- * compiler's <code>useJDK=BREE</code> toolchain lookup must prefer the effective compiler
- * level, i.e. build.properties' <code>jre.compilation.profile</code> (the same setting that
- * already determines the effective -source/-target level, see
- * AbstractOsgiCompilerMojo#getSourceLevel()/getTargetLevel()), over the reactor-wide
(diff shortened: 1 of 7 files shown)


laeubi:
`testDependencyCheck` is preexisting but windows is likely having a path issue in its test.

tycho-its/src/test/java/org/eclipse/tycho/test/compiler/CompilerBreeJreCompilationProfileTest.java
```diff
@@ -0,0 +1,61 @@
...
+ * second JDK.
+ */
+public class CompilerBreeJreCompilationProfileTest extends AbstractTychoIntegrationTest {
+
+	@Test
+	public void testJreCompilationProfileTakesPrecedenceOverTargetPlatformEE() throws Exception {
+		Verifier verifier = getVerifier("compiler.bree.jreCompilationProfile", false);
+		File toolchains = new File(verifier.getBasedir(), "toolchains.xml");
+		verifier.addCliOption("--toolchains " + toolchains.getCanonicalPath());
+		verifier.executeGoal("compile");
```
HannesWell, line 47: Changing this line on Windows to the following fixes the test on Windows:
```suggestion
		verifier.verifyTextInLog("fake-jdk-home-18\\bin\\java.exe not found");
```
I pushed a more general version in a separate commit.
Please squash it if you think it fits.

PR #6336, opened 2026-09-12
Raise jenkins timeout

akurtakov:
Trying to unbreak the build. If/when build consistently is under 3h, this one should be reverted.


akurtakov:
Merge it now as tests will fail due to https://github.com/eclipse-tycho/tycho/pull/6332

PR #6344, opened 2026-09-12
Remove mockito-core from modules whose tests do not use it

akurtakov:
tycho-eclipse-plugin, tycho-source-plugin and tycho-surefire-plugin declare a test dependency on mockito-core, but none of their test sources import org.mockito.

Assisted-by: Anthropic Claude Code (claude-fable-5-1)


akurtakov:
License check is irrelevant for dependency removal PR. Merging.

PR #6363, opened 2026-09-16
CI friendly Set-version does not work with mixed reactor setup

phermsdorf:
## What does this PR do?

Currently it demonstrates a problem where the MANIFEST.MF version is not changed when the versions ${revision} property is derived from the parent pom.
If it's a valid requirement then the fix should happen here as well.
(sorry for the whitespace and line break changes, happens automatically on save)

## How was this tested?

see TychoVersionsPluginTest#testCiFriendlyVersionMixedReactor

## Checklist

- [ ] I have searched existing PRs/issues and this is not a duplicate.
- [ ] `mvn clean install -DskipTests` / relevant integration tests pass locally.
- [ ] Commits are small and self-contained (see [CONTRIBUTING.md](../CONTRIBUTING.md#granularity)).


phermsdorf:
I tried to debug a bit in DefaultPomManipulator, but it would be very helpful if someone could point me in the right direction ...

phermsdorf:
I think the problem is that since the pom's of the sub modules are not changed (since they derive the changed Revision property simply from the parent) they are not treated as "version changed project" and then the logic to update the osgi artefacts is not executed ...

phermsdorf:
the problem seems to be that here
https://github.com/eclipse-tycho/tycho/blob/4870bc1f1aeeeb410b1225ea2b543e4e881b0ec7/tycho-versions-plugin/src/main/java/org/eclipse/tycho/versions/pom/PomFile.java#L154-L162
`getProperties()` only returns the properties of the current pom.xml and not the derived properties from parent's.

Any suggestions on how to get all properties here?

